PR #7, opened 2016-02-24
deprecate scala_test suites

johnynek:
In using the scala rules for a short time, we have already seen several instances of people forgetting to update the suites attribute.

This changes to always run all tests in the target, print the timing for each test in the target (in the standard out) as well as full stack traces of errors.

what do you think @dinowernli

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,6 +277,7 @@ scala_test(name, srcs, suites, deps, data, main_class, resources, scalacopts, jv
 using the `scalatest` library. It may depend on `scala_library`,
 `scala_macro_library` and `java_library` rules.
 
-A `scala_test` requires a `suites` attribute, specifying the fully qualified
-(canonical) names of the test suites to run. In a future version, we might
-investigate lifting this requirement.
+A `scala_test` by default runs all tests in a given target.
+For backwards compatiblity it accepts a `suites` attribute which
+is ignored due to the ease with which that field is not correctly
+populated and tests are not run.
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -145,12 +145,6 @@ cd $0.runfiles
       output=ctx.outputs.executable,
       content=content)
 
-def _args_for_suites(suites):
-  args = ["-o"]
-  for suite in suites:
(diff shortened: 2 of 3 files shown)


dinowernli:
@johnynek 

I believe I originally tried using the -R flag of the scalatest runner and ran into problems where the tests were not being picked up. However, it is possible that I hadn't found the magic combination of flags.

So, bottom line: if this works: LGTM.

scala/scala.bzl
```diff
@@ -152,16 +152,18 @@ def _args_for_suites(suites):
...
+
   content = """#!/bin/bash
 cd $0.runfiles
 {java} -cp {cp} {name} {args} "$@"
 """
   content = content.format(
       java=ctx.file._java.path,
+      cp=":".join([j.short_path for j in jars]),
       name=ctx.attr.main_class,
-      args=' '.join(_args_for_suites(ctx.attr.suites)),
```
dinowernli: This probably means we can remove _args_for_suites.

README.md
```diff
@@ -277,6 +277,7 @@ scala_test(name, srcs, suites, deps, data, main_class, resources, scalacopts, jv
 using the `scalatest` library. It may depend on `scala_library`,
 `scala_macro_library` and `java_library` rules.
 
-A `scala_test` requires a `suites` attribute, specifying the fully qualified
-(canonical) names of the test suites to run. In a future version, we might
-investigate lifting this requirement.
+A `scala_test` by default runs all tests in a given target.
```
dinowernli: "In a given target" is a bit ambiguous. How about: "runs alls tests contained in the files specified in srcs"?

johnynek:
@dinowernli hmm. It is scary for tests not to be run. And indeed, without the `-o` flag, it didn't find the tests, but I did verify they failed if I broke the test intentionally.

I'm inclined to merge and if we find an example of it not working then add that to the repo as a test case, and fix then.

What do you think?

dinowernli:
@johnynek 

Absolutely, I'm really happy you found a way to make it work :) LGTM.

Issue #14, opened 2016-02-29
Support different versions of scala


johnynek:
users should easily be able to use a particular major version of scala.

Two approaches:

1) each tag/commit only supports one and users pick.
2) have `scala_2_10.bzl`, `scala_2_11.bzl` etc... Hopefully they can share code and the external repositories work just fine.

liucijus:
addressed by #1113

PR #15, opened 2016-03-01
Add scala_version function

johnynek:
this is related to #14 

We give a function `scala_version` and a `scala_mvn_artifact` that can be used in name mangling for scala jars. This can help deal with making WORKSPACES less tied to a particular version of scala.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -327,6 +327,17 @@ scala_test = rule(
   test=True,
 )
 
+def scala_version():
+  """return the scala version for use in maven coordinates"""
+  return "2.11"
+
+def scala_mvn_artifact(artifact):
+  gav = artifact.split(":")
+  groupid = gav[0]
+  artifactid = gav[1]
+  version = gav[2]
+  return "%s:%s_%s:%s" % (groupid, artifactid, scala_version(), version)
+
 SCALA_BUILD_FILE = """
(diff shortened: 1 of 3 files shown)


johnynek:
@dinowernli what do you think of this?

scala/scala.bzl
```diff
@@ -327,6 +327,17 @@ scala_test = rule(
   test=True,
 )
 
+def scala_version():
+  """return the scala version for use in maven coordinates"""
+  return "2.11"
+
+def scala_mvn_artifact(artifact):
```
dinowernli: I'm wondering, should we really be parsing the artifact and stitching it back together? It seems like it might be more natural to just explicitly pass in the parts, but maybe I'm missing something.

johnynek: I agree that such stringly typed code is not great. That said, see: 

https://github.com/bazelbuild/rules_scala/pull/15/files#diff-09498dbadf45966909850dc8a47ebb13R9

bazel's `maven_jar` expects the artifact in one big string.

We saw massive pains at Twitter with pants when we had `_2.9.2` baked into artifact names rather than just one function we could update. Want to make sure the same does not happen here.

dinowernli: Ok, fair enough ;)

dinowernli:
LGTM

PR #24, opened 2016-03-25
Add a bazel xml test formatter

### LINKED ISSUE #23: we should create a single xml output
We should create a new reporter:
https://github.com/scalatest/scalatest/blob/release-2.2.6-for-scala-2.11-and-2.10/src/main/scala/org/scalatest/Reporter.scala

which is basically a clone of this:
https://github.com/scalatest/scalatest/blob/master/scalatest/src/main/scala/org/scalatest/tools/JUnitXmlReporter.scala

except that it creates a single file given by the environmental variable `$XML_OUTPUT_FILE`

See this tool here:

https://gist.githubusercontent.com/cgoldberg/4320815/raw/75ba2b7cc8b477bc784631e28ffd586f79efa3f7/merge_junit_results.py

johnynek:
closes #23 

@dinowernli @smparkes 

review?


smparkes:
LGTM

johnynek:
Note this is not usable (that I can see) until bazel 0.2.1 is out (I tested using it with the RC and it worked) due the reference to the scala reporter not being found with the current (I believe incorrect) namespacing.

johnynek:
I'll wait to merge until 0.2.1 is out.

dinowernli:
LGTM

xynny:
0.2.1 is released now it seems like?

johnynek:
yes. I'll update the travis to use 0.2.1 soon.

PR #27, opened 2016-04-06
Add disappearing class test

ianoc:
We came across a case where deleting a class would let the Downstreams using it still compile. This should highlight that bad case

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -73,6 +73,7 @@ def _compile(ctx, jars, buildijar):
       out=ctx.outputs.jar.path,
       ijar_out=ctx.outputs.ijar.path)
   cmd = """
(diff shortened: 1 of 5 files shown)


ianoc:
I signed it!

johnynek:
@damienmg We have signed a corporate CLA as @stripe but it seems it still thinks @ianoc is not legit.

Any suggestions?

damienmg:
Yes googlebot doesn't seems to recognize corporate emails :(

damienmg:
(manually verified that stripe has indeed signed the corp CLA, feel free to manually change the CLA tag for next stripe contributions)

johnynek:
@damienmg thank you, sir!

Issue #6, opened 2016-02-24
add support for scrooge thrift generator


johnynek:
See: http://twitter.github.io/scrooge/CommandLine.html

something basic can be done probably with a genrule + a macro once we solve #2.

An problem (to me) is dealing with dependencies between thrift targets. One approach is to pull the transitive thrift files, compile all of them using `--gen-file-map`, then only put the output classes from the current target into the current jar. This may be too slow for deep dependencies however.

johnynek:
see: https://github.com/wt/bazel_thrift

Issue #2, opened 2016-02-24
add support for src jars


johnynek:
for code gen, such as scrooge, we probably want to create src jars, then reuse the existing scala rules. Right now we require .scala files as source.

I guess this would use `find` after we unjar to make sure all the files can be handled.

PR #1, opened 2016-02-24
Change the scala compiler download to https

johnynek:
this started failing today prompting:

https://github.com/bazelbuild/bazel/issues/969

quickfix here.

Review @dinowernli

PR #29, opened 2016-04-13
Add scrooge support

jcoveney:
Ok, there is a LOT to be desired here... there are a ton of TODOS

BUT IT WORKS

I currently have a project which has two thrift files which depend on each other, BUILDS for them, and then a scala_library which depends on that and the thrifts...it builds. My guess is right now transitive stuff will not work.

So my immediate first goal is to make a test directory with some thrifts and some stuff that depends on them, to make sure that transitive stuff works properly.

After that is just...making it less crazy. I really, really need the guidance of people who have done this... I will put a microscope to everything I did, but I'm finding the "throw structs around with attributes! oh some attributes are missing!" to really resist abstraction and promote hacks...

But it works :) so I will do the above, hopefully refactor with guidance and time, and then will try it on one of our internal repositories.

One thing that I would like to put off, though, is moving it into different files etc. Right now I want to just make sure that we get the flow of everything down properly.


johnynek:
note stripe has signed the CLA.

scala/scala.bzl
```diff
@@ -72,11 +73,30 @@ def _compile(ctx, jars, buildijar):
...
+  #TODO probably can be in a helper function
+  srcjar_cmd = "\n_SRCJAR_ARG=\" \""
+  srcjars = _srcjar_filetype.filter(ctx.files.srcjars + list(_collect_srcjars(ctx.attr.deps)))
+  if len(srcjars) > 0:
+    # TODO Somehow, the manifest is sneaking in here?! so we filter... but really
+    #      should try and understand why the manifest is being put here!
+    # TODO the FileType.filter also seemed to get rid of it...??
+    for srcjar in srcjars:
+      srcjar_cmd += """
+mkdir -p {{out}}_tmp/expand_srcjars/{srcjar}
```
johnynek: we should delete this first to make sure we don't keep old ones around

scala/scala.bzl
```diff
@@ -72,11 +73,30 @@ def _compile(ctx, jars, buildijar):
...
+  #TODO here is the issue...when scala library depends on this, it's not picking up anything
+
+  #TODO I need a command that will explode out the source jars
+  #TODO probably can be in a helper function
+  srcjar_cmd = "\n_SRCJAR_ARG=\" \""
+  srcjars = _srcjar_filetype.filter(ctx.files.srcjars + list(_collect_srcjars(ctx.attr.deps)))
+  if len(srcjars) > 0:
+    # TODO Somehow, the manifest is sneaking in here?! so we filter... but really
+    #      should try and understand why the manifest is being put here!
+    # TODO the FileType.filter also seemed to get rid of it...??
```
johnynek: I'm confused by the comment. The filter should remove the name of the manifest. What does your TODO mean?

scala/scala.bzl
```diff
@@ -109,7 +132,8 @@ rm -rf {out}_tmp
       arguments=[f.path for f in ctx.files.srcs])
 
 def _compile_or_empty(ctx, jars, buildijar):
-  if len(ctx.files.srcs) == 0:
+  # TODO We assume that if a srcjar is present, that it is not empty... is there a way to check if it has files?
```
johnynek: no, we can't read the file before we do an action. I think we should not support empty srcjar.

scala/scala.bzl
```diff
@@ -384,3 +424,190 @@ def scala_repositories():
     url = "https://oss.sonatype.org/content/groups/public/org/scalatest/scalatest_2.11/2.2.6/scalatest_2.11-2.2.6.jar",
     sha256 = "f198967436a5e7a69cfd182902adcfbcb9f2e41b349e1a5c8881a2407f615962",
   )
+
+
+# TODO is it possible to share scala-library with the above?
+def scrooge():
```
johnynek: can we move the rest of this into a second file (and probably second PR)? Can we tackle just srcjar first?

scala/scala.bzl
```diff
@@ -282,6 +319,9 @@ scala_library = rule(
   implementation=_scala_library_impl,
   attrs={
       "main_class": attr.string(),
+      #TODO the implementation of java_library leads me to believe
+      # we may be able to eliminate this
+      "srcjars": attr.label_list(),
```
johnynek: ideally, this is just in srcs like the java rule:

http://bazel.io/docs/be/java.html

scala/scala.bzl
```diff
@@ -164,6 +187,17 @@ def _write_test_launcher(ctx, jars):
       output=ctx.outputs.executable,
       content=content)
 
+def _collect_srcjars(targets):
+  srcjars = set()
+  for target in targets:
+    print(target) #TODO remove
+    #TODO perhaps we want to unify the naming?
+    if hasattr(target, "srcjar"):
```
johnynek: I think `srcjar` is only used in `java_import` and we don't really have an analog of that, so I think we can skip it.

scala/scala.bzl
```diff
@@ -72,11 +73,30 @@ def _compile(ctx, jars, buildijar):
...
+    # TODO Somehow, the manifest is sneaking in here?! so we filter... but really
+    #      should try and understand why the manifest is being put here!
+    # TODO the FileType.filter also seemed to get rid of it...??
+    for srcjar in srcjars:
+      srcjar_cmd += """
+mkdir -p {{out}}_tmp/expand_srcjars/{srcjar}
+unzip {srcjar} -d {{out}}_tmp/expand_srcjars/{srcjar}
+""".format(srcjar = srcjar.path)
+    srcjar_cmd += """
+for f in `find {out}_tmp/expand_srcjars -type f -name "*.scala"`; do _SRCJAR_ARG=$_SRCJAR_ARG$f" "; done
```
johnynek: this is just building up an environmental variable that has a space separated list of all files right?

scala/scala.bzl
```diff
@@ -72,11 +73,30 @@ def _compile(ctx, jars, buildijar):
...
+  srcjar_cmd = "\n_SRCJAR_ARG=\" \""
+  srcjars = _srcjar_filetype.filter(ctx.files.srcjars + list(_collect_srcjars(ctx.attr.deps)))
+  if len(srcjars) > 0:
+    # TODO Somehow, the manifest is sneaking in here?! so we filter... but really
+    #      should try and understand why the manifest is being put here!
+    # TODO the FileType.filter also seemed to get rid of it...??
+    for srcjar in srcjars:
+      srcjar_cmd += """
+mkdir -p {{out}}_tmp/expand_srcjars/{srcjar}
+unzip {srcjar} -d {{out}}_tmp/expand_srcjars/{srcjar}
```
johnynek: can you comment about the double escaping here? It is pretty tricky at first glance.

scala/scala.bzl
```diff
@@ -72,11 +73,30 @@ def _compile(ctx, jars, buildijar):
       ijar=ctx.file._ijar.path,
       out=ctx.outputs.jar.path,
       ijar_out=ctx.outputs.ijar.path)
+  #TODO here is the issue...when scala library depends on this, it's not picking up anything
+
+  #TODO I need a command that will explode out the source jars
+  #TODO probably can be in a helper function
```
johnynek: agree we need to factor this somehow.

scala/scala.bzl
```diff
@@ -92,13 +112,16 @@ rm -rf {out}_tmp
       manifest=ctx.outputs.manifest.path,
       jar=ctx.file._jar.path,
       ijar=ctx.file._ijar.path,
-      jars=":".join([j.path for j in jars]),)
+      jars=":".join([j.path for j in jars]),
+      srcjar_cmd = srcjar_cmd
```
johnynek: seems like you are not using this anymore.

scala/scala.bzl
```diff
@@ -72,11 +73,30 @@ def _compile(ctx, jars, buildijar):
...
+    # TODO the FileType.filter also seemed to get rid of it...??
+    for srcjar in srcjars:
+      srcjar_cmd += """
+mkdir -p {{out}}_tmp/expand_srcjars/{srcjar}
+unzip {srcjar} -d {{out}}_tmp/expand_srcjars/{srcjar}
+""".format(srcjar = srcjar.path)
+    srcjar_cmd += """
+for f in `find {out}_tmp/expand_srcjars -type f -name "*.scala"`; do _SRCJAR_ARG=$_SRCJAR_ARG$f" "; done
+"""
+  #TODO need to make sure that the ijar and any other artifacts contain source stuff
```
johnynek: should get this for free since ijar runs after you make the jar.

scala/scala.bzl
```diff
@@ -16,6 +16,7 @@
 
 
 _scala_filetype = FileType([".scala"])
+_srcjar_filetype = FileType([".srcjar"])
```
johnynek: I think we want `_scala_filetype = FileType([".scala", ".srcjar"])` and put the srcjar and scala files in srcs (to follow the java convention).

scala/scala.bzl
```diff
@@ -384,3 +424,190 @@ def scala_repositories():
...
+          ctx.files._scopt + \
+          ctx.files._twitter_util + \
+          ctx.files._util_logging + \
+          ctx.files._scala_parser_combinators + \
+          transitive_cjars
+  _write_manifest(ctx) #TODO do we need something like this?
+
+  #TODO we need to use the gen-file-map for caching
+  #TODO what imports do we need?
+  #TODO do we want an ijar for this as well?
```
johnynek: no ijar, because you are making a src jar. Then when we depend on this, we will depend on the scala_library target that has compiled the src.

scala/scala.bzl
```diff
@@ -384,3 +424,190 @@ def scala_repositories():
...
+  cmd = """
+rm -rf {out}_tmp
+set -e
+{java} -classpath "{jars}" com.twitter.scrooge.Main -d {out}_tmp -s -l scala $@
```
johnynek: I think you need the `env JAVACMD` thing here

scala/scala.bzl
```diff
@@ -384,3 +424,190 @@ def scala_repositories():
...
+  return struct(scala = scalaattr, srcjar = ctx.outputs.srcjar)
+
+  #TODO the struct that we return from this is where we can set the scala attributes it wants, like the transitive dependencies
+
+
+scrooge_library = rule(
+  implementation = _scrooge_library_impl,
+  attrs = {
+      "srcs": attr.label_list(allow_files=_thrift_filetype),
+      "deps": attr.label_list(allow_files=_scroogelib_tar_filetype),
```
johnynek: I think we only want other scrooge libraries here, not the tars.

scala/scala.bzl
jcoveney: I thought that was just for the scala compiler? In this case, we are actually invoking the java commandline, not scalac

scala/scala.bzl
```diff
@@ -384,3 +424,190 @@ def scala_repositories():
...
+            default=Label("@scala_parser_combinators//jar"),
+            allow_files=True),
+    } + _implicit_deps,
+    outputs={
+      "srcjar": "lib%{name}.srcjar",
+      "manifest": "lib%{name}_MANIFEST.MF",
+    },
+)
+
+def scrooge_scala_library(name, scrooge_library, deps=[], visibility=None):
```
johnynek: I think we want to take srcs and deps here, right?

scala/scala.bzl
```diff
@@ -384,3 +424,190 @@ def scala_repositories():
...
+    visibility = visibility,
+  )
+  scala_library(
+    name = name,
+    srcjars = [name + '_srcjar'],
+    #TODO do these need to be here, given the transitive compile piece above?
+    deps = deps + [
+      "@libthrift//jar",
+      "@scrooge_core//jar",
+      "@finagle_thrift//jar",
```
johnynek: not sure if we want finagle on everything. Probably not. Probably an option to the rule.

scala/scala.bzl
jcoveney: yes...wasn't sure how else to do it. In a sense, bash is kind of like a monad you can't get out of... we can put information into it, but we can't get it out. I'm not sure outside of bash how we can say "we know that this is an archive with files in it, what are all the files?" Of course, an option would be to wrap scrooge as you suggested (this is not a bad option!), but for now this is all I could come up with.

scala/scala.bzl
jcoveney: yeah, I think we can get rid of it...I added this before I realized how structs were used to pass around information via their attributes. will tackle this tomorrow

scala/scala.bzl
johnynek: okay, actually you are right.

scala/scala.bzl
jcoveney: I mean, I can. I guess I can create a fake source jar and test off of that. The nice thing about having the two together right now is that it forces me to deal with the compile dependencies properly, etc. So I'm happy doing this, but I'm afraid that decoupling it will make it harder to tackle the issues around properly passing through the compile time and run time dependencies.

What do you think?

WORKSPACE
```diff
@@ -1,10 +1,11 @@
 workspace(name = "io_bazel_rules_scala")
 
-load("//scala:scala.bzl", "scala_repositories", "scala_mvn_artifact")
+load("//scala:scala.bzl", "scala_repositories", "scala_mvn_artifact", "scrooge")
 scala_repositories()
+scrooge()
```
johnynek: should probably call this `scala_scrooge` or `twitter_scrooge` or something a bit more namespaced.

scala/scala.bzl
jcoveney: Ah, I should get rid of this. I had a separate bug which I had not diagnosed yet, but the manifestation is that the filter call seemed like it was filtering everything. Will remove comment.

scala/scala.bzl
jcoveney: yeah, we rm -rf {out}_tmp so it should be removed as part of that

scala/scala.bzl
```diff
@@ -384,3 +424,193 @@ def scala_repositories():
     url = "https://oss.sonatype.org/content/groups/public/org/scalatest/scalatest_2.11/2.2.6/scalatest_2.11-2.2.6.jar",
     sha256 = "f198967436a5e7a69cfd182902adcfbcb9f2e41b349e1a5c8881a2407f615962",
   )
+
+
+# TODO is it possible to share scala-library with the above?
+def scrooge():
+  native.maven_jar(
+    name = "libthrift",
+    artifact = "org.apache.thrift:libthrift:0.8.0"
```
johnynek: we should add shas to these for better security for the consumers.

scala/scala.bzl
jcoveney: sure, though we will still need a _srcjar_filetype to properly filter the files given to _compile which are source jars, no?

scala/scala.bzl
jcoveney: for sure

scala/scala.bzl
```diff
@@ -384,3 +424,193 @@ def scala_repositories():
...
+                     transitive_runtime_deps = [], #TODO what are scrooge's runtime deps?
+                     transitive_compile_exports = transitive_cjars, #TODO are all necessary?
+                     transitive_runtime_exports = []
+                     )
+  return struct(scala = scalaattr, srcjar = ctx.outputs.srcjar)
+
+  #TODO the struct that we return from this is where we can set the scala attributes it wants, like the transitive dependencies
+
+
+scrooge_library = rule(
```
johnynek: what does this one do? I don't follow the need.

scala/scala.bzl
jcoveney: well, it's also listed in outputs for scrooge_scala_srcjar, but I don't think that that is ever made visible (it seems like outputs are not available to the targets graph, only attributes are)

scala/scala.bzl
```diff
@@ -384,3 +424,193 @@ def scala_repositories():
...
+
+  cjars = ctx.files._scrooge_generator + \
+          ctx.files._scopt + \
+          ctx.files._twitter_util + \
+          ctx.files._util_logging + \
+          ctx.files._scala_parser_combinators + \
+          transitive_cjars
+  _write_manifest(ctx) #TODO do we need something like this?
+
+  #TODO we need to use the gen-file-map for caching
```
johnynek: I think we will need the gen-file-map to delete scala files that did not come from the thrifts in this target. Some of the thrifts will be from our dependencies, I think, but we will regen all the source with each run. We need to parse that file and only keep the ones from the `srcs` and not the `deps`.

scala/scala.bzl
jcoveney: we want a scrooge_library specifically, not srcs (if I'm understanding the comment properly). We will have to get the sources from the target graph.

scala/scala.bzl
jcoveney: I will see if it can compile without finagle. Scrooge docs say it's necessary but that might be only if you set the flag to make services

scala/scala.bzl
johnynek: I see, you use it below in the rule that generates the scala, so you will need to keep `srcjar` but I don't think we need `srcjars` below, do we?

scala/scala.bzl
jcoveney: So I'm following the pattern used here
https://github.com/wt/bazel_thrift/blob/master/tools/thrift.bzl
as well as in genproto, but I think the idea is you have a "library" rule which says "these thrifts form a library," and then specific language library rules which do the code gen. Scrooge library makes the thrifts available as thrifts (and doesn't force the user to muck with issues around dependencies, scala versions, etc), and then it is up to the specific code gen targets to use that information to generate code for a given platform.

scala/scala.bzl
```diff
@@ -384,3 +424,193 @@ def scala_repositories():
...
+  transitive_archive_files = []
+  return struct(
+    srcs=ctx.files.srcs,
+    _transitive_archive_files=transitive_archive_files,
+  )
+
+def _gen_scrooge_srcjar_impl(ctx):
+  #TODO would there be a case in which we want/need to allow more jars here?
+  transitive_cjars = [ctx.files._scrooge_core, ctx.files._libthrift, ctx.files._finagle_thrift, ctx.files._scalalib]
+  transitive_cjars = [file for files in transitive_cjars for file in files]
```
johnynek: weird to reuse name like this, not sure I like that.

Do you need the transitive srcjars? or is that always handled by the scala_library's existing deps mechanism? I feel like we should be able to use that.

jcoveney: I will try to get it to work now that I understand how it works a bit better. I was failing to get it to work but maybe it works now. If you can scan it and given pointers on how to restructure it, this is definitely an area where I "yolo'd" a bit

johnynek:
This is super great! I would like two things:

1) let's break into two PRs (we can serialize, since one depends on the next): srcjar #2 and then #6 
2) let's add tests for srcjar support in a few cases: srcjar only, srcjar + src, multiple srcjars (with and without src).

scala/scala.bzl
jcoveney: I don't think I have a good understanding of the caching, its invalidation, and how a given target/rule can influence that... because it seems like if a source changes, wouldn't it just blow everything away and rebuild? Any pointers on better understanding how caching works?

scala/scala.bzl
jcoveney: most definitely

WORKSPACE
jcoveney: ah, good call

scala/scala.bzl
jcoveney: yeah, you are right. this is 100% cargo culted from: https://github.com/wt/bazel_thrift/blob/master/tools/thrift.bzl#L95

Though that raises the question: what are they trying to express?

scala/scala.bzl
jcoveney: definitely need a pass of garbage collection, especially around the deps stuff I think I am not fully leveraging the work you did as well. that's gonna be the next major pass at this, I think... getting rid of srcjars and properly leveraging srcs and passing through deps.

jcoveney:
Ok, CI passes, though I need to add real tests.

Definitely agree with #2, #1 I agree with but I want to do after I have a couple of passes to answer some of the issues. Without the scrooge stuff, there is no use case that will properly ensure that the srcjars are being used as intended, and that transitive information is passed along properly. Though I have a feeling that if I answer some of your comments above, the decoupling of the 2 PRs will be a lot cleaner.

scala/scala.bzl
johnynek: gen-file-map is not about caching. It is about isolation. Consider two targets: A, B. My B target depends on A. So, when I create the scala srcjar, scooge will need both A.thrift, and B.thrift. But if I put A.scala in my srcjar, then if someone depends on both A and B that src will show up twice in the compiler's path.

We may sidestep the issue by copying the files and somehow making sure they only appear once that way (assuming the file names would be the same in both the run to create A, and B (which also needs A)).

Does my concern make sense? The dependency shows up twice: 1) to render thrift -> scala, 2) to compile scala -> class.

scala/scala.bzl
johnynek: Okay. I'm not sure we need that, I might punt on that (it is basically just taring up thrifts and has nothing to do with thrift, I think, this is just a tar rule, right?)

jcoveney:
Ok, so something I want to discuss a little is testing...  it doesn't seem like, as is, there is a way to test if certain targets create (or don't create) certain artifacts. I assume it is ok if I add this to run_test? Nothing crazy, but for the transitive thrift stuff, want to make sure we do/don't make artifacts at the right time.

jcoveney:
Ok, this is maturing a lot. There are two things I'd like to do now:
- make sure it works with scala binary (which will also ensure that the runtime dependencies work...my guess is right now they aren't)
- figure out how we want to deal with the compile time dependencies
- split it out into multiple files etc (I'm still unsure that it's worth splitting into two PRs, but am happy to do that eventually...much more focused on getting the implementation working and having something actually use source jars has been invaluable)

On the second point above, I have a comment in there, but the question is really how we want to deal with the compile time dependencies of a source jar...there's no guarantee that the immediate next target will compile it (you can have a string of targets and only the last one will actually do the compiling, this is true of normal sources as well), but the compiler dependencies will only go one level.

thinking on it, what I think we can do is carry along the compile dependencies...UNTIL a compile happens. So "transitive_compile_export" would have the same spirit as today, but it would carry it forward and be removed once a compile actually happens (instead of right now where it is removed by the dependent node). The rest would be the same.

Now that I've written it out, I think that is the right approach to take.

jcoveney:
Ah, hm. On further thought, not sure if that will work. Gonna mull on it and explore code...

jcoveney:
The fix is being discussed https://github.com/bazelbuild/rules_scala/pull/30, will work on the rest in the meantime

johnynek:
can we split this into three PRs:

1) support for srcjar
2) thrift_library (which we might want to move out of this repo one day)
3) scala_scrooge_library (depends on 1 and 2).

jcoveney:
I'm moving in that direction... I'm about to push a change which splits them out into files, but I still want to get it all working, because this is a case where the consumer is going to help me figure out the design of the parent

jcoveney:
Hm, something in the rebase broke it. That's what I get >_< will figure it out

scala/scala.bzl
```diff
@@ -123,7 +149,7 @@ def _compile_or_empty(ctx, jars, buildijar):
       ijar = ctx.outputs.jar
     return struct(ijar=ijar, class_jar=ctx.outputs.jar)
 
-def _write_manifest(ctx):
+def write_manifest(ctx):
```
johnynek: why did you drop the `_` here?

thrift/thrift.bzl
```diff
@@ -0,0 +1,62 @@
...
+# the scrooge_scala_library is what handles the specifics of code generation...
+# this is useful because it means that if there are different code generation
+# targets, we don't need to have a whole separate tree of targets organizing
+# the thrifts.
+thrift_library = rule(
+  implementation = _thrift_library_impl,
+  attrs = {
+      "srcs": attr.label_list(allow_files=_thrift_filetype),
+      "deps": attr.label_list(),
+  } + implicit_deps(),
```
johnynek: does this really need this? Seems like you would not.

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -0,0 +1,155 @@
...
+    jars=":".join([f.path for f in cjars]),
+  )
+  ctx.action(
+    #TODO we should curate this a bit more deliberately...make sure it is accurate when it all is done
+    inputs=list(ctx.attr.thrift_library.transitive_srcs) +
+        cjars +
+        ctx.files._jdk +
+        ctx.files._scalasdk +
+        [ctx.outputs.manifest, ctx.file._jar],
+    outputs=[ctx.outputs.srcjar], #TODO ijar?
```
johnynek: I don't think so, since this code mechanically depends on the thrift, and if the thrift changes, the ijar is quite likely to change. We don't change logic without changing interface, is my point.

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -0,0 +1,155 @@
...
+    artifact = scala_mvn_artifact("com.twitter:util-core:6.33.0"),
+    sha1 = "bb49fa66a3ca9b7db8cd764d0b26ce498bbccc83",
+  )
+  native.maven_jar(
+    name = "util_logging",
+    artifact = scala_mvn_artifact("com.twitter:util-logging:6.33.0"),
+    sha1 = "3d28e46f8ee3b7ad1b98a51b98089fc01c9755dd",
+  )
+  native.maven_jar(
+    name = "scala_parser_combinators",
```
johnynek: this is in the scala .tar.gz. Let's reuse that:

https://github.com/bazelbuild/rules_scala/blob/master/scala/scala.bzl#L366

test/src/main/scala/scala/test/twitter_scrooge/BUILD
```diff
@@ -0,0 +1,100 @@
+load("//scala:scala.bzl",   "scala_binary", "scala_library")
+load("//twitter_scrooge:twitter_scrooge.bzl", "scrooge_scala_library")
+
+scrooge_scala_library(
+  name = "thrift1",
```
johnynek: can we call this `scrooge1` or something, it is a bit confusing with thrift libraries and scrooge libraries.

thrift/thrift.bzl
```diff
@@ -0,0 +1,62 @@
...
+    command = cmd,
+    progress_message = "making thrift archive %s" % ctx.label,
+    arguments = [f.path for f in ctx.files.srcs],
+  )
+  transitive_archives = _collect_thrift_tars(ctx.attr.deps)
+  transitive_archives += [ctx.outputs.libarchive]
+
+  transitive_srcs = _collect_thrift_srcs(ctx.attr.deps)
+  transitive_srcs += ctx.files.srcs
+  return struct(
```
johnynek: can we do the nesting pattern here under `thrift` so we can see the "type" of this thing later?

`struct(thrift = struct(...)``

thrift/thrift.bzl
```diff
@@ -0,0 +1,62 @@
...
+_thrift_filetype = FileType([".thrift"])
+
+#TODO some unanswered questions...
+# - should the tar for a given target include the deps? I believe the answer is no (they should be transitive)
+# - should "building" this tar trigger the builds of the dependent tars? I believe the answer should be yes...not sure how to trigger that?
+#   hmm, on the other side, though... perhaps the answer is no, and it should only be triggered when it is actually used/needed? yes, that
+#   seems more in line with bazel philosophy of "only build what you need when you need it"
+def _thrift_library_impl(ctx):
+  #TODO how do we want to archive this? could just do another jar to be consistent with everything else...
+  cmd = "tar -c $@ > {out}".format(out=ctx.outputs.libarchive.path)
```
johnynek: can we set the times or otherwise normalize the attributes to be 100% sure this is a pure function of inputs.

thrift/thrift.bzl
jcoveney: indeed...must be a holdover from a previous rev. nice catch

twitter_scrooge/twitter_scrooge.bzl
jcoveney: this is what I wanted to do, but I've failed to find the syntax to reference those files. have any pointers? the exports_files docs don't say

test/src/main/scala/scala/test/twitter_scrooge/BUILD
jcoveney: yeah, good call. renamed everything.

thrift/thrift.bzl
jcoveney: do you mean the content of the inputs? Because the inputs should not change... I guess the question is what happens if someone updates the time on one of the source files, but doesn't do anything else. Is that what we want to protect against?

johnynek: yes. exactly. We want the output to be a function of the content of the inputs (and not when it was run). We do take steps to ensure this. That way, the hash function used to check if we need to rebuild works.

twitter_scrooge/twitter_scrooge.bzl
johnynek: we are using other examples `@scala//:lib/scala-parser-combinators_2.11-1.0.4.jar` should do it.

test/src/main/scala/scala/test/twitter_scrooge/BUILD
```diff
@@ -0,0 +1,100 @@
...
+  thrift_library = "//test/src/main/scala/scala/test/twitter_scrooge/thrift",
+  visibility = ["//visibility:public"],
+)
+
+scrooge_scala_library(
```
johnynek: I worry about this:
```
thriftA -> thriftB
scroogeA -> thriftA (will generate thriftB?)
scoogeB -> thriftB
scala_lib -> scroogeA, scroogeB (now we have the scala from thriftB on the path twice, right?)
```
what is the right way to tackle this?

My thinking was that scrooge_scala_srcjar would be non-transitive. It would only generate a srcjar containing files from the given target, but it would depend on a srcjar target for all the deps to the src jar. What do you think?

I think this requires walking backwards on the graph (seeing the scrooge_scala_srcjar that depends on the thrift, and there should be exactly 1) or using the assumption of a standard naming as you have done here.

The last challenge is how to remove the files we don't want. I think we use the mapfile to do that: we only keep scala files that are associated with the direct dependencies in this rule.

What do you think?

twitter_scrooge/twitter_scrooge.bzl
jcoveney: yeah, that makes sense

thrift/thrift.bzl
jcoveney: good idea

test/src/main/scala/scala/test/twitter_scrooge/BUILD
jcoveney: Do we care which pass it was compiled in? It sounds like we just need to keep track of which files were already compiled, and ignore them (as they'll already be on the classpath). Shouldn't keeping a set of already compiled thrift paths be sufficient, then?

So in this case, we don't care if it compiles scroogeA or scroogeB first... if it compiles scroogeA first, it'll compile/code gen thriftA/thriftB at once. If it does scroogeB first, it'll gen/compile scrooge B and then when it gets to scrooge A, it'll know to omit scrooge B.

johnynek: Since we only look up the graph, and each node is an independent operation, `scroogeA` and `scroogeB` need to know what to generate without looking at children, only parents.

The question is how to deal with people who only depend on `scroogeA` and those that depend on both `scroogeA` and `scroogeB`. I think the answer may be something like `exports`. It may be that  you would only compile `thriftA` in `scroogeA`, but you probably add `scroogeB` to the exports. That makes writing the BUILD pretty tough i think.

johnynek: we could make it a build error to not export `scroogeB` in `scala_scrooge_library`. So it would work like this. scrooge_scala_library would have `srcs` that could be thrift files, `deps` that can be thrift_libraries or other `scala_scrooge_library`s and `exports` which can only be other `scala_scrooge_libraries`. We generate code for all the `srcs` and direct `deps`, but we must cover all the transitive `thrift_library` deps with the direct deps of the `exports`, otherwise we have an error (some files would not be owned by a target).

Note, the above includes the ability to add thrifts directly to a `scala_scrooge_library` which would create an ad-hoc thrift-library target and add it to the deps (this makes it nicer to use in the common case of just adding a bunch of thrifts and getting something we can depend on with a scala_library).

What do you think of this UX for BUILD writers?

jcoveney: But as the targets are traversed, they can label every source file with the target it comes from. This way in the final code gen target, we can just make sure that there are no identical files that have different content. Shouldn't that be sufficient? I like the conceptual simplicity of it right now and would prefer to keep it if possible.

thrift/thrift.bzl
jcoveney: gotcha!

twitter_scrooge/twitter_scrooge.bzl
jcoveney: used the scala libs wherever possible now!

scala/scala.bzl
jcoveney: _ makes it private, and it is used in twitter_scrooge

jcoveney:
So the thrift_library issue is outstanding, but I am going to take this chance to try and split it out into three PRs

jcoveney:
Ok, I've split out the two leaf aspects of this
https://github.com/bazelbuild/rules_scala/pull/33
https://github.com/bazelbuild/rules_scala/pull/34

Then I will rebase this

jcoveney:
closes https://github.com/bazelbuild/rules_scala/issues/6

jcoveney:
I have rebased this with the other branches that have been merged... still need to finish the work potentially using the gen-file-map

jcoveney:
Ok… so it does what we want. I am going to spend some time explaining the semantics, because even after our conversation, oscar, parts were not properly thought through. And in explaining this, I’ll explain this “extra_information” thing in scala.bzl… basically, this is a way to let things which depend on scala_library inject information into the graph that they can retrieve later, without scala.bzl having to know.

So, let’s say we have two thrift files, thrift a and its dependency, thrift b

thrift_library_a
- src: thrift_a
- dep: thrift_library_b

thrift_library_b
- src: thrift_b

our goal is that we have some unit which owns the compilation of PART of the thrift graph, not necessarily all…this is desirable for speed (you can cache parts of the dag instead of having to compile everything) and for correctness (you can’t have file conflicts, since only one target will own compilation)

ok, so we have

scrooge_scala_library_a
- dep: thrift_library_a, ???

scrooge_scala_library_b
- dep: thrift_library_b

and we note that scrooge_scala_library is a macro which creates a scala library name a and a scrooge_scala_srcjar named a_srcjar…

Note the ??? above, that is where I hit a snag. Oscar, per our discussion, it seemed like the goal would be to have

scrooge_scala_library_a
- dep: thrift_library_a, scrooge_scala_library_b_srcjar

scrooge_scala_library_b
- dep: thrift_library_b

Basically, any scrooge_scala_library “owns” the generation of it’s immediate dependencies (so a for a, and b for b, above). But it also needs to make sure that the transitive dependency graph under a is covered, this is where the dep on scrooge_scala_library_a_srcjar comes in… it knows the transitive dependency graph, and we can check that thrift_library_a’s graph is covered.

But this has an issue! Which is that scrooge_scala_library_b_srcjar doesn’t actually know how to compile the code… only how to generate sources. Which means that in the above setup, scrooge_scala_library_a would compile the code for everything (which is slow), and it won’t really benefit from caching, as scrooge_scala_library_b_srcjar doesn’t have actual code to cache. So what we actually want is…

scrooge_scala_library_a
- dep: thrift_library_a, scala_library_b

scrooge_scala_library_b
- dep: thrift_library_b

scala_library is what knows how to take sources and turn them into class files, and what we want when we do a scrooge code gen is to generate code for the transitive graph (can’t really avoid that), but only compile the files that we “own”…the rest we can take advantage of caching, and the fact that they are owned by some set of other targets (this is what we held above).

However, this has another issue: scala_library_b has no notion of owning targets! That was a concern of _srcjar target that got created…

So this is the purpose of the extra_information. It’s basically a generic way for you to stuff information into the graph. When is this useful? Well, in any case like this one where you have a new target which depends on/uses scala library, like so: new_target -> scala_library -> new_target. Since we do not want scala_library to have to be modified for new targets, the only way I could come up to support this was to have a way to pass through information.

Open to other ways to resolve the above issue!

Note: I tried using jar to decompress, but the jar command doesn’t let you give it a directory to target. This is why I’m currently using unzip.

OH, also, I added a python file… but I’m not sure if projects depending on this will pick it up. I hope so?? we can reorganize accordingly.

jcoveney:
Ok… so it does what we want. I am going to spend some time explaining the semantics, because even after our conversation, oscar, parts were not properly thought through. And in explaining this, I’ll explain this “extra_information” thing in scala.bzl… basically, this is a way to let things which depend on scala_library inject information into the graph that they can retrieve later, without scala.bzl having to know.

So, let’s say we have two thrift files, thrift a and its dependency, thrift b

thrift_library_a
- src: thrift_a
- dep: thrift_library_b

thrift_library_b
- src: thrift_b

our goal is that we have some unit which owns the compilation of PART of the thrift graph, not necessarily all…this is desirable for speed (you can cache parts of the dag instead of having to compile everything) and for correctness (you can’t have file conflicts, since only one target will own compilation)

ok, so we have

scrooge_scala_library_a
- dep: thrift_library_a, ???

scrooge_scala_library_b
- dep: thrift_library_b

and we note that scrooge_scala_library is a macro which creates a scala library name a and a scrooge_scala_srcjar named a_srcjar…

Note the ??? above, that is where I hit a snag. Oscar, per our discussion, it seemed like the goal would be to have

scrooge_scala_library_a
- dep: thrift_library_a, scrooge_scala_library_b_srcjar

scrooge_scala_library_b
- dep: thrift_library_b

Basically, any scrooge_scala_library “owns” the generation of it’s immediate dependencies (so a for a, and b for b, above). But it also needs to make sure that the transitive dependency graph under a is covered, this is where the dep on scrooge_scala_library_a_srcjar comes in… it knows the transitive dependency graph, and we can check that thrift_library_a’s graph is covered.

But this has an issue! Which is that scrooge_scala_library_b_srcjar doesn’t actually know how to compile the code… only how to generate sources. Which means that in the above setup, scrooge_scala_library_a would compile the code for everything (which is slow), and it won’t really benefit from caching, as scrooge_scala_library_b_srcjar doesn’t have actual code to cache. So what we actually want is…

scrooge_scala_library_a
- dep: thrift_library_a, scala_library_b

scrooge_scala_library_b
- dep: thrift_library_b

scala_library is what knows how to take sources and turn them into class files, and what we want when we do a scrooge code gen is to generate code for the transitive graph (can’t really avoid that), but only compile the files that we “own”…the rest we can take advantage of caching, and the fact that they are owned by some set of other targets (this is what we held above).

However, this has another issue: scala_library_b has no notion of owning targets! That was a concern of _srcjar target that got created…

So this is the purpose of the extra_information. It’s basically a generic way for you to stuff information into the graph. When is this useful? Well, in any case like this one where you have a new target which depends on/uses scala library, like so: new_target -> scala_library -> new_target. Since we do not want scala_library to have to be modified for new targets, the only way I could come up to support this was to have a way to pass through information.

Open to other ways to resolve the above issue!

Note: I tried using jar to decompress, but the jar command doesn’t let you give it a directory to target. This is why I’m currently using unzip.

OH, also, I added a python file… but I’m not sure if projects depending on this will pick it up. I hope so?? we can reorganize accordingly.

test/src/main/scala/scala/test/twitter_scrooge/BUILD
```diff
@@ -0,0 +1,137 @@
...
+  deps = [
+    ":justscrooge1",
+    ":justscrooge2a",
+    ":justscrooge2b",
+    ":justscrooge3",
+  ],
+)
+
+scala_library(
+  name = "twodeep",
```
johnynek: I thought this was a twadoop joke.

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -0,0 +1,271 @@
...
+def _collect_owned_srcs(targets):
+  r = set()
+  for _target in targets:
+    if hasattr(_target, "extra_information"):
+      for target in _target.extra_information:
+        if hasattr(target, "scrooge_srcjar"):
+          r += target.scrooge_srcjar.transitive_owned_srcs
+  return r
+
+#TODO this is the same as in scala.bzl...just depend on that one?
```
johnynek: yes

test/src/main/scala/scala/test/twitter_scrooge/BUILD
jcoveney: now I wish it was

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -0,0 +1,271 @@
...
+rm -rf {out}_tmp
+set -e
+mkdir {out}_tmp_polluted
+mkdir {out}_tmp
+unzip -o {polluted} -d {out}_tmp_polluted 2>/dev/null
+touch -t 198001010000 {manifest}
+touch {out}_tmp_polluted/gen_pluck_cmd
+{pluck} {jar} {manifest} {out} {out}_tmp_polluted/gen_pluck_cmd {out}_tmp_polluted {genmap} {include_base} {pluck_base} $@
+bash {out}_tmp_polluted/gen_pluck_cmd
+echo {out}_tmp_polluted
```
johnynek: do we need the echo?

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -0,0 +1,271 @@
...
+  r = set()
+  for target in targets:
+    if hasattr(target, "thrift"):
+      r += target.thrift.srcs
+  return r
+
+def _assert_set_is_subset(left, right):
+  for l in left:
+    if l not in right:
+      fail('scrooge_srcjar target must depend on scrooge_srcjar targets sufficient to ' +
```
johnynek: can we print the missing items?

twitter_scrooge/twitter_scrooge.bzl
jcoveney: ah, I thought I had removed all of the vestiges of debugging... definitely not, and I'll make sure to remove the rest... this PR has had a lot of writing and rewriting!

twitter_scrooge/twitter_scrooge.bzl
jcoveney: good idea

twitter_scrooge/twitter_scrooge.bzl
jcoveney: actually, since I wrote that comment it's not the same anymore... I'll split it into the shared and not

test/src/main/scala/scala/test/twitter_scrooge/BUILD
```diff
@@ -0,0 +1,137 @@
...
+)
+
+scala_binary(
+    name = "justscrooges",
+    srcs = ["JustScrooge1.scala"],
+    deps = [":allscrooges"],
+    main_class = "scala.test.twitter_scrooge.JustScrooge1",
+)
+
+#scala_test(
```
johnynek: can we remove these commented items?

test/src/main/scala/scala/test/twitter_scrooge/BUILD
```diff
@@ -0,0 +1,137 @@
...
+    ":scrooge3_srcjar",
+  ],
+  visibility = ["//visibility:public"],
+)
+
+scrooge_scala_library(
+  name = "scrooge2_b",
+  deps = [
+    "//test/src/main/scala/scala/test/twitter_scrooge/thrift/thrift2:thrift2_b",
+    ":scrooge3_srcjar",
```
johnynek: why are you depending on `_srcjar` here? For a test or is this a common path?

jcoveney: fixed...this is from the initial runs of it. I did most of my testing with the thinner targets below

scala/scala.bzl
```diff
@@ -264,7 +255,16 @@ def _lib(ctx, non_macro_lib):
       collect_data = True)
   return struct(
       scala = scalaattr,
-      runfiles=runfiles)
+      runfiles=runfiles,
+      extra_information=_collect_extra_information(ctx.attr.deps),
```
johnynek: do we need to collect from exports as well? If not, can we comment why not?

src/python/scripts/twitter_scrooge_pluck_cmd.py
```diff
@@ -0,0 +1,48 @@
...
+      k = g.group(1)
+      v = m.get(k, set())
+      v.add(g.group(2))
+      m[k] = v
+  return m
+
+def main(jar_cmd, manifest, output_jar, cmd_file, srcjar_polluted, gen_file_map, include_base, pluck_base, files):
+
+  gen_map = parse_gen_file_map(gen_file_map)
+  #TODO we should just walk srcjar_polluted, and only include mentioned sources
```
johnynek: why is that better? Also, didn't you remove srcjar_polluted?

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -0,0 +1,274 @@
...
+  pluck_cmd="""
+rm -rf {out}_tmp_polluted
+rm -rf {out}_tmp
+set -e
+mkdir {out}_tmp_polluted
+mkdir {out}_tmp
+unzip -o {polluted} -d {out}_tmp_polluted 2>/dev/null
+touch -t 198001010000 {manifest}
+touch {out}_tmp_polluted/gen_pluck_cmd
+{pluck} {jar} {manifest} {out} {out}_tmp_polluted/gen_pluck_cmd {out}_tmp_polluted {genmap} {include_base} {pluck_base} $@
```
johnynek: I think it is fine to merge using python, but actually due to bazelbuild/bazel#890, we are hoping that there is a system python and it is the right version.

This is fine for now, but rewriting that code in scala should not be hard (you can even make the jar without shelling out then: https://docs.oracle.com/javase/7/docs/api/java/util/jar/package-summary.html

src/python/scripts/twitter_scrooge_pluck_cmd.py
johnynek: I see we still have srcjar_polluted, but I'm not sure why we need it. Given that creating those files is not free, I would say we should try to minimize the jars we create.

jcoveney:
I think this is ready, but there are some followups that would be good...
- more pie in the sky: generalize how we propagate information through the targets... extra_information is the beginning of this, but we generally are passing information through the graph a lot (sometimes transitive, sometimes transitive with rules). We should fully generalize this.
- ensure 100% that you can't have the same class file twice. First step is probably just to check and error out, but we want to probably also use extra_information here to make sure that we can't have diamonds (right now it might be possible with overlapping scala_scrooge_library)
- more practical: let's get rid of all of the bash scripts, and push it into an actual language (in this case, scala)

@johnynek: anything else? and of course, we should go over it once more

johnynek:
So, looking at this:

https://github.com/twitter/scrooge/blob/develop/scrooge-generator/src/main/scala/com/twitter/scrooge/Compiler.scala#L35

I think we would have `includes` in `scala_scrooge_library` and pass the thrift files there. Which is a bit weird, but I guess you can either do `deps` or `includes`. deps is relative to the current directory and `includes` makes the directory magically appear under any path.

johnynek:
@jcoveney see:

https://github.com/twitter/scrooge/blob/develop/scrooge-maven-plugin/src/main/scala/com/twitter/ScroogeRunner.scala#L40

looks like includes is exactly what we are seeing. With include support, we should be good to go.

jcoveney:
Ooook...so includes works! There are two new additions that we probably need to discuss... the first is absolute_prefix (I welcome new names!). The issue is this: our thrift usually exists in a hierarchy like src/main/thrift/com/stripe/blah. Absolute imports are with respect to com/stripe/blah, but there's no way (that I can think of?) to just "know" that you should be doing your imports off of src/main/thrift as the "base" for includes. I can't think of a way to do this automatically that wouldn't create the possibility for erroneous includes.

The second one is the so-called "remote_jar"...this one I think we may want to think through a little more. The issue is this: we have thrifts that exist in the world, but not necessarily under the control of bazel. remote_jar is how you tell the codegen that you have the missing thrifts... right now, it assumes that this jar has both thrifts+classpath files (this is how we published them at stripe). This can all be changed, but something we need to sort of decide on. This was instrumental in getting an important internal respository to work with this... and I don't think "convert it all to bazel" will be feasible for many thrift deployments going forward. I think we need something, and it should be correct, and doesn't have to be amazing.

scala/scala.bzl
```diff
@@ -95,35 +91,29 @@ def _compile(ctx, jars, dep_srcjars, buildijar):
...
 
   cmd = """
+rm -rf {out}_tmp_expand_srcjars
 rm -rf {out}_tmp
 set -e
 rm -rf {out}_args
 mkdir -p {out}_args
 touch {out}_args/files_from_jar
 mkdir -p {out}_tmp""" + srcjar_cmd + """
+#TODO THIS IS A HACK TO SEE IF IT WORKS WITH ZINC...MUST REMOVE!!
```
johnynek: can we remove this?

scala/scala.bzl
```diff
@@ -147,7 +137,7 @@ rm -rf {out}_tmp
     outs.extend([ctx.outputs.ijar])
   ctx.action(
       inputs=list(jars) +
-          list(dep_srcjars) +
+          list(all_srcjars) +
           ctx.files.srcs +
```
johnynek: shouldn't this be `sources` now? You partitioned that into `sources` and `srcjars`.

src/python/scripts/twitter_scrooge_pluck_cmd.py
```diff
@@ -0,0 +1,49 @@
...
+    m = {}
+    for line in file:
+      g = r.match(line)
+      k = g.group(1)
+      v = m.get(k, set())
+      v.add(g.group(2))
+      m[k] = v
+  return m
+
+def main(jar_cmd, manifest, output_jar, cmd_file, srcjar_polluted, gen_file_map, include_base, pluck_base, files):
```
johnynek: do we still use this with the scala one? If not, can we remove? python not being hermetic worries me. bazelbuild/bazel#890

src/scala/scripts/TwitterScroogeGenerator.scala
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  def extractJarTo(_jar: String, _dest: String): List[Path] = {
+    val files = Buffer[Path]()
+    //TODO add to it, error if duplicate!
+    val jar = new JarFile(_jar)
+    val enumEntries = jar.entries()
+    while (enumEntries.hasMoreElements) {
+      val file = enumEntries.nextElement().asInstanceOf[JarEntry]
+      val f = new File(_dest, file.getName) //TODO pathify this
```
johnynek: what does this TODO mean? Is it still needed?

src/scala/scripts/TwitterScroogeGenerator.scala
```diff
@@ -0,0 +1,269 @@
...
+    try {
+      Files.walkFileTree(path, DeleteRecursively)
+    } catch {
+      case e: Exception => //TODO LOG
+    }
+  }
+
+  def extractJarTo(_jar: String, _dest: String): List[Path] = {
+    val files = Buffer[Path]()
+    //TODO add to it, error if duplicate!
```
johnynek: looks like we are erroring below, aren't we (it says it throws). What does this mean?

src/scala/scripts/TwitterScroogeGenerator.scala
```diff
@@ -0,0 +1,269 @@
...
+        files += path
+      }
+    }
+    files.toList
+  }
+
+  //TODO add logging that can be turned on/off?
+  def main(args: Array[String]) {
+    if (args.length < 4) sys.error("Need to ensure enough arguments! " +
+      "Required 3 arguments: onlyTransitiveThriftSrcs immediateThriftSrcs " +
```
johnynek: `Required 4` right?

src/scala/scripts/TwitterScroogeGenerator.scala
```diff
@@ -0,0 +1,269 @@
...
+
+        files += path
+      }
+    }
+    files.toList
+  }
+
+  //TODO add logging that can be turned on/off?
+  def main(args: Array[String]) {
+    if (args.length < 4) sys.error("Need to ensure enough arguments! " +
```
johnynek: what about `args.length != 4`, why not be strict?

src/scala/scripts/TwitterScroogeGenerator.scala
```diff
@@ -0,0 +1,269 @@
...
+        onlyTransitiveThriftSrcsFile,
+        scroogeOutput,
+        remoteJarsFile,
+        _tmp.toString
+      ).map(Paths.get(_))
+
+    scrooge.destFolder = scroogeOutput
+    scrooge.fileMapPath = Some(genFileMap)
+    //TODO NOOOOOOOooooOOO
+    scrooge.strict = false
```
johnynek: what is the story here? Our stuff does not compile with strict? If so, the TODO should probably be to make that an option we pass in.

src/scala/scripts/TwitterScroogeGenerator.scala
```diff
@@ -0,0 +1,269 @@
...
+    scrooge.run()
+
+    FinalJarCreator(jarOutput, immediateThriftSrcs, genFileMap, scroogeOutput)
+
+    // Clean it out to be idempotent
+    dirsToDelete.foreach { deleteDir(_) }
+  }
+}
+
+/**
```
johnynek: can the rest of this be removed?

thrift/thrift.bzl
```diff
@@ -3,37 +3,41 @@
...
     command = cmd,
     progress_message = "making thrift archive %s" % ctx.label,
     arguments = [f.path for f in ctx.files.srcs],
   )
-  transitive_archives = _collect_thrift_tars(ctx.attr.deps)
-  transitive_archives += [ctx.outputs.libarchive]
 
   transitive_srcs = _collect_thrift_srcs(ctx.attr.deps)
-  transitive_srcs += ctx.files.srcs
+  transitive_srcs += [ctx.outputs.libarchive]
```
johnynek: are these src files or are they the archives? Why is this here?

thrift/thrift.bzl
```diff
@@ -54,19 +55,34 @@ def _valid_thrift_deps(targets):
...
+      # that matches that, but that could create correctness issues if there are more
+      # than one in different parts of the tree. Another option is to take an argument
+      # that references namespace, and base the tree off of that. The downside
+      # to that is that thrift_library then gets enmeshed in the details of code
+      # generation. This could also be something punted to scrooge_scala_library
+      # or whatever, but I think that we should make it such that the archive
+      # created by this is created in such a way that absolute imports work...
+      "absolute_prefix": attr.string(default='', mandatory=False),
       "_jar": attr.label(executable=True, default=Label("@bazel_tools//tools/jdk:jar"), single_file=True, allow_files=True),
       "_jdk": attr.label(default=Label("//tools/defaults:jdk"), allow_files=True),
```
johnynek: I don't think we actually need this. We are just using `jar` here, not `javac` or the libraries or anything, right? So, if `jar` needs `jdk`, I guess it declares that.

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -0,0 +1,269 @@
...
+def _assert_set_is_subset(left, right):
+  missing = set()
+  for l in left:
+    if l not in right:
+      missing += [l]
+  if len(missing) > 0:
+    fail('scrooge_srcjar target must depend on scrooge_srcjar targets sufficient to ' +
+         'cover the transitive graph of thrift files. Uncovered sources: ' + missing)
+
+def _content_newline(data):
```
johnynek: can we say `_path_newline` or something to reflect that we are calling `.path`?

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -0,0 +1,269 @@
...
+  ctx.file_action(output=remote_jars_file, content=_content_newline(remote_jars))
+
+  only_transitive_thrift_srcs_file = ctx.new_file(ctx.outputs.srcjar, ctx.outputs.srcjar.short_path + "_only_transitive_thrift_srcs")
+  ctx.file_action(output = only_transitive_thrift_srcs_file, content = _content_newline(only_transitive_thrift_srcs))
+
+  immediate_thrift_srcs_file = ctx.new_file(ctx.outputs.srcjar, ctx.outputs.srcjar.short_path + "_immediate_thrift_srcs")
+  ctx.file_action(output = immediate_thrift_srcs_file, content = _content_newline(immediate_thrift_srcs))
+
+  ctx.action(
+    #TODO do the dependencies of the executable need to be listed?
```
johnynek: I think we bazel handles that. I think we just say what we need.

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -0,0 +1,269 @@
...
+
+  # We want to ensure that the thrift sources which we do not own (but need
+  # in order to generate code) have targets which will compile them.
+  _assert_set_is_subset(only_transitive_thrift_srcs, transitive_owned_srcs)
+
+  #TODO would there be a case in which we want/need to allow more jars here?
+  transitive_cjars = [ctx.files._scrooge_core, ctx.files._libthrift, ctx.files._scalalib]
+  transitive_cjars = [file for files in transitive_cjars for file in files]
+
+  cjars = ctx.files._scrooge_generator + \
```
johnynek: I think these should not be here. Are they needed? I think the deps for the action should live there. You shouldn't need to know about them here.

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -0,0 +1,269 @@
...
+            allow_files=True),
+        "_scala_parser_combinators": attr.label(
+            default=Label("@scala//:lib/scala-parser-combinators_2.11-1.0.4.jar"),
+            allow_files=True),
+        "_pluck_scrooge_scala": attr.label(
+          executable=True,
+          default=Label("//src/scala/scripts:generator"),
+          #single_file=True,
+          allow_files=True),
+        "_pluck_scrooge": attr.label(
```
johnynek: we don't need this anymore, right?

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -0,0 +1,269 @@
...
+        #     "covered," as well as needing the thrifts to
+        #     do the code gen.
+        "remote_jars": attr.label_list(),
+        "_scrooge_core": attr.label(
+            default=Label("@scrooge_core//jar"),
+            allow_files=True),
+        "_libthrift": attr.label(
+            default=Label("@libthrift//jar"),
+            allow_files=True),
+        "_scrooge_generator": attr.label(
```
johnynek: this should not be a dep here, I think. This should only be a dep or the pluck_scala binary.

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -0,0 +1,269 @@
...
+    transitive_runtime_deps = transitive_runtime_deps,
+    transitive_compile_exports = transitive_compile_exports,
+    transitive_runtime_exports = transitive_runtime_exports,
+  )
+
+scrooge_scala_srcjar = rule(
+    _gen_scrooge_srcjar_impl,
+    attrs={
+        "deps": attr.label_list(mandatory=True),
+        #TODO we should think more about how we want to deal
```
johnynek: these are jars that have both thrift and scala code in them, right?

Can we exercise this in tests?

src/python/scripts/twitter_scrooge_pluck_cmd.py
jcoveney: whoops, yeah, this is removable now

src/scala/scripts/TwitterScroogeGenerator.scala
jcoveney: our stuff does not compile with strict, because somebody used a thrift reserved keyword as an identifier

making it an option is a great idea

src/scala/scripts/TwitterScroogeGenerator.scala
jcoveney: yes

thrift/thrift.bzl
jcoveney: so one of the changes that I made is that we actually use the archive produced by thrift_library. this made everything much more consistent, and also helped the actual code gen, because instead of wrangling sources, I just have archives that I could unzip wherever I wanted. also, it felt more in line with the bazel philosophy of depending on the outputs of previous targets, instead of just sort of grabbing the file tree.

twitter_scrooge/twitter_scrooge.bzl
jcoveney: it was necessary when I was calling scrooge manually, but it may not be necessary now that we have a binary with scrooge's deps there...I will clean up

twitter_scrooge/twitter_scrooge.bzl
jcoveney: sure, though may be moot if we go with the new_git_repo approach?

src/scala/scripts/TwitterScroogeGenerator.scala
```diff
@@ -0,0 +1,269 @@
...
+    //TODO WE NEED TO TEST THIS!!
+    val remoteSrcJars = Source.fromFile(remoteJarsFile).getLines.toSet
+    extract(remoteSrcJars)
+
+    val dirsToDelete =
+      Set(
+        immediateThriftSrcsFile,
+        onlyTransitiveThriftSrcsFile,
+        scroogeOutput,
+        remoteJarsFile,
```
johnynek: we should not be deleting the inputs. Only delete the tmp directory and the scroogeOutput.

src/scala/scripts/TwitterScroogeGenerator.scala
```diff
@@ -0,0 +1,269 @@
...
+    val genFileMap = s"$scroogeOutput/gen-file-map.txt"
+
+    val scrooge = new Compiler
+
+    // we need to extract into the same tree, as that is the only way to get relative imports between them working..
+    // AS SUCH, we are just going to try extracting EVERYTHING to the same tree, and we will just error if there
+    // are more than one.
+    val _tmp =  Files.createTempDirectory(tmp, "jar")
+    // This will only be meaningful if they have absolute_prefix set
+    scrooge.includePaths += _tmp.toString
```
johnynek: actually, includePaths accepts jars:

https://github.com/twitter/scrooge/blob/develop/scrooge-generator/src/main/scala/com/twitter/scrooge/frontend/Importer.scala#L56

https://github.com/twitter/scrooge/blob/develop/scrooge-generator/src/main/scala/com/twitter/scrooge/Compiler.scala#L64

so we can just point to a zip or a jar.

johnynek: I did a quick check and it didn't work for me. Let's punt on this and just fix the caching issue and merge. We can optimize performance later.

src/scala/scripts/TwitterScroogeGenerator.scala
jcoveney: ah... I forgot here that the file_action is an action in its own right, and as such, its outputs will be cached/preserved by bazel.

twitter_scrooge/twitter_scrooge.bzl
jcoveney: not that we're using the scala_binary we can clean up these deps a lot...doing so

scala/scala.bzl
jcoveney: it should be all 3. ctx.files.srcs goes to sources and srcjars, and dep_srcjars came from ctx.attr.deps

src/scala/scripts/TwitterScroogeGenerator.scala
jcoveney: this file is a bit inconsistent in the way it deals with files/paths...it mainly uses nio2 Path, but then here creeps in File (this is from a code snippet about jar extraction). I've made it consistent

src/scala/scripts/TwitterScroogeGenerator.scala
jcoveney: dated todo, removed in cleanup

src/scala/scripts/TwitterScroogeGenerator.scala
jcoveney: all fixed

thrift/thrift.bzl
jcoveney: at one point I guess I needed it, but now it is not necessary

twitter_scrooge/twitter_scrooge.bzl
jcoveney: #namingthingsishard

twitter_scrooge/twitter_scrooge.bzl
jcoveney: graphs are a beautiful thing

twitter_scrooge/twitter_scrooge.bzl
jcoveney: nope. cleaned up the old python stuff laying around as well

jcoveney:
Lots of cleanup just pushed

jcoveney:
So it sounds like you're ok with absolute_prefix, and we should at least get a POC that new_git_repository can eliminate the need for remote_jars

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -209,36 +191,9 @@ scrooge_scala_srcjar = rule(
...
           executable=True,
           default=Label("//src/scala/scripts:generator"),
-          #single_file=True,
-          allow_files=True),
-        "_pluck_scrooge": attr.label(
-          executable=True,
-          default=Label("//src/python/scripts:twitter_scrooge_pluck_cmd"),
-          #single_file=True,
           allow_files=True),
     } + implicit_deps(),
```
johnynek: I don't think you need the scala implicit deps, do you?

johnynek:
One minor comment, then ready to go. Thanks for pushing this through!

jcoveney:
So what do we want to do about the remote_jars? Leave them for now, but comment that they may be removed soon? Or get the new_git_repository thing working?

twitter_scrooge/twitter_scrooge.bzl
jcoveney: good catch...man, using the executable is a really nice way to simplify it!

jcoveney:
Oh, and re: _jdk, I'm gonna put it back... I totally forgot about https://github.com/bazelbuild/rules_scala/pull/37/ !

jcoveney:
so I guess _jdk is a dependency of _jar? hm, I guess we will see :)

johnynek:
crap, I merged your other one and now this won't merge cleanly

jcoveney:
I'll fix it now

jcoveney:
Fixing it was pretty easy actually... I kept the commits, though it was a rebase, so they'll get new sha's. But hopefully you still get the incremental view?

johnynek:
Awesome!

damienmg:
For your information this has broken ci.bazel.io build (missing dependency). See http://ci.bazel.io/job/rules_scala/105/

PR #30, opened 2016-04-13
Add test that shows failure of current exports handling

jcoveney:


jcoveney:
Would love to know how to tie this email to stripe's corporate one :) Or since this isn't my corporate email, should I sign one regardless?

johnynek:
did you not add stripe's email to this account, or do you have two accounts?

test/A.scala
```diff
@@ -0,0 +1,7 @@
+package scala.test
+
+object A {
+  def main(args: Array[String]) {
+    println("4 8 15 16 23 42")
```
johnynek: nice. :)

jcoveney:
I guess I need to add stripe's email to this account

jcoveney:
BTW, I pushed a fix to this. I tried a couple bigger refactors, but the issue is that... this is really what it seems we want. We don't do anything more fancy with this information, we just don't want to include the ijars -- everything else is there.

I think we should consider a more principled refactor of this, but this fixes the bug... anything bigger imo is trickier than it seems, and I just want to unblock the scrooge work for now

scala/scala.bzl
```diff
@@ -200,13 +202,17 @@ def _lib(ctx, non_macro_lib):
 
   if not non_macro_lib:
```
johnynek: I think the defensive thing is to move this above line 198. We don't have a test that fails now, but macros should need scala-reflect at compile time (maybe not runtime). It may be that somehow we are putting scala reflect on every classpath elsewhere.

PR #33, opened 2016-04-18
add support for source jars to scala_library

jcoveney:
This is pulled out from https://github.com/bazelbuild/rules_scala/pull/29

This adds support for source jars. This will be used and tested immediately by another PR


jcoveney:
closes https://github.com/bazelbuild/rules_scala/issues/2

scala/scala.bzl
```diff
@@ -64,23 +66,59 @@ touch -t 198001010000 {manifest}
...
+
   cmd = """
 rm -rf {out}_tmp
 set -e
-mkdir -p {out}_tmp
-env JAVACMD={java} {scalac} {scala_opts} {jvm_flags} -classpath "{jars}" $@ -d {out}_tmp
+mkdir -p {out}_tmp""" + srcjar_cmd + """
+env JAVACMD={java} {scalac} {scala_opts} {jvm_flags} -classpath "{jars}" $_SRCJAR_ARG $@ -d {out}_tmp
 # Make jar file deterministic by setting the timestamp of files
+rm -rf
```
johnynek: what are we removing here?

jcoveney: good catch. fixed. this is from when I was moving around how the srcjar stuff was expanded

johnynek:
👍 will merge when green.

PR #34, opened 2016-04-18
add support for thrift_library to organize .thrift files

jcoveney:
https://github.com/bazelbuild/rules_scala/pull/29

This adds the thrift_library rule. This will be used and more thoroughly tested in a followup PR with codegen that depends on this

### DIFF
--- a/test/src/main/scala/scala/test/twitter_scrooge/thrift/BUILD
+++ b/test/src/main/scala/scala/test/twitter_scrooge/thrift/BUILD
@@ -1,0 +1,11 @@
+load("//thrift:thrift.bzl", "thrift_library")
+
+thrift_library(
+  name = "thrift",
+  srcs = ["Thrift1.thrift"],
+  deps = [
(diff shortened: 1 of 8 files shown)


jcoveney:
step one to https://github.com/bazelbuild/rules_scala/issues/6

jcoveney:
Ok, I removed the rsync dependency and instead depended on jar, which seems fine. See no reason not to get this in?

damienmg:
This broke on ci.bazel.io, probably due to sandboxing, see  http://ci.bazel.io/job/rules_scala/75/

johnynek:
thanks @damienmg ! glad we have that wired in.

damienmg:
btw I am revamping ci.bazel.io so PR vetting is coming along. We should
have it in a couple of weeks normally :)

johnynek:
@damienmg that will be super great. Since travis uses docker, we don't see this (and we use OSX locally).

PR #36, opened 2016-04-19
Fix thrift target in sandbox

jcoveney:
I had forgotten to make the jar a dependency (via the action's input) of the thrift action, so the sandbox was hiding it.

### DIFF
--- a/thrift/thrift.bzl
+++ b/thrift/thrift.bzl
@@ -18,7 +18,7 @@ find {out}_tmp -exec touch -t 198001010000 {{}} \;
 rm -rf {out}_tmp""".format(out=ctx.outputs.libarchive.path,
                            jar=ctx.file._jar.path)
   ctx.action(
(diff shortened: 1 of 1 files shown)


damienmg:
Still broken with sandboxing :( -> http://ci.bazel.io/job/rules_scala/

thrift/thrift.bzl
```diff
@@ -18,7 +18,7 @@ find {out}_tmp -exec touch -t 198001010000 {{}} \;
 rm -rf {out}_tmp""".format(out=ctx.outputs.libarchive.path,
                            jar=ctx.file._jar.path)
   ctx.action(
-    inputs = ctx.files.srcs,
+    inputs = ctx.files.srcs + ctx.files._jar,
```
damienmg: you needs to add the files from @local_jdk//:jdk-default also

johnynek:
@jcoveney

PR #37, opened 2016-04-20
Make jdk dependency explicit

jcoveney:
This should fix the issue that came up after the fact here

https://github.com/bazelbuild/rules_scala/pull/36

@damienmg, thank you for pointing this out! A question to mull over with @johnynek: how can we avoid this?

Is there an issue if we set up the automatic CI to run on different architectures? Is it a cost concern?

Also, any word on sandboxing in OSX? I see this https://github.com/bazelbuild/bazel/issues/902, but there's not code or anything.

### DIFF
--- a/thrift/thrift.bzl
+++ b/thrift/thrift.bzl
@@ -18,7 +18,7 @@ find {out}_tmp -exec touch -t 198001010000 {{}} \;
 rm -rf {out}_tmp""".format(out=ctx.outputs.libarchive.path,
                            jar=ctx.file._jar.path)
   ctx.action(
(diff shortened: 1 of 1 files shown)


johnynek:
I think @damienmg may be able to set up bazel's CI that does not use Docker.

damienmg:
This has fixed the build \o/

Yes I am working on making our CI vet every PR but it is not going fast because we need to scale it (and I am on rotation this week, which mean I have to ping everybody for breakage, you included :))

damienmg:
@hermione521 is working on sandboxing for OS X currently, see https://bazel-review.googlesource.com/#/c/3383/

PR #38, opened 2016-04-21
Fix scala rules with bazel HEAD

damienmg:
Bazel HEAD has removed //tools/defaults:ijar, rules should
now depends on @bazel_tools//tools/jdk:ijar.


johnynek:
Thanks, sir!

PR #39, opened 2016-04-21
use an arg file for scalac to deal with large targets

johnynek:
we don't want large targets for build speed and incrementalism, but for thrift it might happen (and we'd rather be slow than fail).


johnynek:
@jcoveney what do you think?

smparkes:
LGTM

PR #47, opened 2016-04-30
Add support for scalac compiler plugins

johnynek:
This seems to work to add plugins.

Only tested with existing ones, but it *should* work for plugins in the repo as well.

It it not clear to me what the classpath for plugins is (do they need to be fat jars, or does it share a classpath with compilation?)

This seems to work.

review? @smparkes @dinowernli @jcoveney


dinowernli:
LGTM, looks legit.

smparkes:
I'm not using it right now but I may be. Anyway, WFM.

PR #51, opened 2016-05-05
Remove the baked-in akka library from the scalasdk filegroup.

dinowernli:
@unlikely-leo @johnynek @smparkes


dinowernli:
I have tested this on Linux, don't actually have a Mac at hand.

PR #54, opened 2016-05-18
change script to support 0.2.3

johnynek:
due to the path changes in bazel 0.2.3, the plugin needs to be updated.


smparkes:
LGTM

Issue #53, opened 2016-05-18
md5sum hashes do not match when you build the same code twice on macos


xynny:
Just discovered this recently since I was working on build caching, but it seems like the default scala rule here do not  build  jars with the same hash on two separate runs with  the same inputs.  This seems to break the caching code in bazel that gets a digest with md5. 
This is what I ran on this repo here:
```
bazel build test/... 
md5 bazel-bin/test/*.jar > hash
bazel clean
bazel build test/... 
md5 bazel-bin/test/*.jar > hash2
```

When you diff the two files it shows that all the jars have different hashes.  even the ijars. 

```
Xins-MacBook-Pro:rules_scala xinlu$ diff hash hash2
1,3c1,3
< MD5 (bazel-bin/test/ExportOnly_deploy.jar) = e513f71cb2024bf2a5c6ac0a0208fb5e
< MD5 (bazel-bin/test/ExportOnly_ijar.jar) = e513f71cb2024bf2a5c6ac0a0208fb5e
< MD5 (bazel-bin/test/Exported_deploy.jar) = 4e1b1b620a002219a2edda47dc9a9cd2
---
> MD5 (bazel-bin/test/ExportOnly_deploy.jar) = 764d68cad512ea85baf75aec623c465e
> MD5 (bazel-bin/test/ExportOnly_ijar.jar) = 764d68cad512ea85baf75aec623c465e
> MD5 (bazel-bin/test/Exported_deploy.jar) = 31ea9c77a3bbf21d2b0ffc6def3247e4
5,6c5,6
< MD5 (bazel-bin/test/HelloLibTest_deploy.jar) = 1faf15bba18d60a4d9fee7937f3d18d6
< MD5 (bazel-bin/test/HelloLib_deploy.jar) = d72085331bdb480df424be08c0b8b5bd
---
> MD5 (bazel-bin/test/HelloLibTest_deploy.jar) = 9822c828fc3a1cbb87dbcd3800ba9e62
> MD5 (bazel-bin/test/HelloLib_deploy.jar) = 398401f2f02554b2d60aa62288cb255f
10,11c10,11
< MD5 (bazel-bin/test/MacroTest_deploy.jar) = 8c4767eead2843d79b850e56c86ea42e
< MD5 (bazel-bin/test/OtherLib_deploy.jar) = 2fc708e492140ed78ff0a489df106dcd
---
[...]

PR #55, opened 2016-05-19
Check that jar output is byte-for-byte identical

johnynek:
This is to address #53 

This adds a jar creator that is a minor edit of:
https://github.com/bazelbuild/bazel/blob/master/src/java_tools/buildjar/java/com/google/devtools/build/buildjar/jarhelper/JarCreator.java

to remove the unneeded guava dependency (so it would build easier), which sets the timestamps to fixed values, which makes the jars pure functions of the contents.

### DIFF
--- a/src/java/io/bazel/rulesscala/jar/BUILD
+++ b/src/java/io/bazel/rulesscala/jar/BUILD
@@ -1,0 +1,5 @@
+java_binary(name = "jar",
+  srcs = ["JarCreator.java", "JarHelper.java"],
+  main_class = "io.bazel.rulesscala.jar.JarCreator",
+  visibility = ["//visibility:public"],
+)
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -29,31 +29,48 @@ def _adjust_resources_path(path):
     return dir_1 + dir_2, rel_path
   return "", path
 
-def _add_resources_cmd(ctx):
+def _add_resources_cmd(ctx, dest):
   res_cmd = ""
   for f in ctx.files.resources:
     c_dir, res_path = _adjust_resources_path(f.path)
-    change_dir = "-C " + c_dir if c_dir else ""
-    res_cmd += "\n{jar} uf {out} " + change_dir + " " + res_path
+    res_cmd += "\nmkdir -p $(dirname {out_dir}{res_path})\ncp {c_dir}{res_path} {out_dir}{res_path}".format(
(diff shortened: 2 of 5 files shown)


johnynek:
this seems to pass on linux, but it fails on my OSX machine.

damienmg:
test this please

bazel-io:

johnynek:
Here is the issue:

http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/sun/tools/jar/Main.java#445

jar seems to hardwire writing the current time. Not sure how linux escapes this.

johnynek:
http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/sun/tools/jar/Main.java?av=f#526

johnynek:
This passes on OSX.

@ianoc @non this is what I came up with.

This is minor edit of:
https://github.com/bazelbuild/bazel/blob/master/src/java_tools/buildjar/java/com/google/devtools/build/buildjar/jarhelper/JarCreator.java

to remove the unneeded guava dependency (so it would build easier).

johnynek:
this closes #53

johnynek:
@smparkes @dinowernli review?

This does not change the build time for an internal repo relative to sha: 7b891adb975b4e3e6569b763d39ab6e9234196c9

smparkes:
Nice. LGTM

src/java/io/bazel/rulesscala/jar/BUILD
```diff
@@ -0,0 +1,5 @@
+java_binary(name = "jar",
```
damienmg: FWIW, @bazel_tools//tools/zip:zipper is a zip implementation that strip out timestamp. Might be easier to refer to it that to fork SingleJar's Jar implementation. (Also faster)

johnynek: That is good to know about instead of using system `zip`, but what about making manifests? Seems nice to reuse that bit?

dinowernli:
Sorry, out of town ;)

LGTM

src/java/io/bazel/rulesscala/jar/BUILD
damienmg: Oh I see, then maybe you can use directly the SingleJar binary. Just ask if you need to get rid of the file fork and I can probably help you sort that out.

PR #56, opened 2016-05-19
scala-lang should be a runtime dependency of any scala_library

colinmarc:
That way, a java_binary pulling it in will have scala-lang on the classpath.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -280,6 +280,7 @@ def _lib(ctx, non_macro_lib):
   rjars += [ctx.outputs.jar]
   rjars += _collect_jars(ctx.attr.runtime_deps).runtime
 
+  rjars += [ctx.file._scalalib, ctx.file._scalareflect]
   if not non_macro_lib:
     #  macros need the scala reflect jar
     rjars += [ctx.file._scalareflect]


bazel-io:

johnynek:
@damienmg I don't know what this means: https://github.com/bazelbuild/rules_scala/pull/56#issuecomment-220401788

damienmg:
I have send a mail about that last week on bazel-dev@. Basically it is asking if the change is safe to run on ci.bazel.io. to tell it safe you must be either in the Google org or in the bazelbuild org and make it publicly visible. See the mail for more details

johnynek:
Thanks, sorry for the noise.

On Thu, May 19, 2016 at 8:01 AM, Damien Martin-Guillerez <
notifications@github.com> wrote:

> I have send a mail about that last week on bazel-dev@. Basically it is
> asking if the change is safe to run on ci.bazel.io. to tell it safe you
> must be either in the Google org or in the bazelbuild org and make it
> publicly visible. See the mail for more details
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/bazelbuild/rules_scala/pull/56#issuecomment-220404474>
>



-- 
P. Oscar Boykin, Ph.D. | http://twitter.com/posco | http://pobox.com/~boykin

damienmg:
(no need to excuse, it is perfectly fine and sorry for the tone, I was writing on my cellphone while commuting).

johnynek:
test this please

damienmg:
You have to accept the invitation to the bazel org and make it public so Jenkins knows that you are an "admin".

In the mean time, for Jenkins: test this please

johnynek:
@damienmg how do I accept that? Did you send it to @johnynek ? I don't see it in my email or any notifications on github (but the latter can easily be lost for me because I get so many github notifications)

johnynek:
@colinmarc can you think of any way we could have a test for this?

Like, you uncovered in another bazel rule, but we didn't see it here because scala-library is always used by scala-binary, which also adds the dependency.

johnynek:
one idea could be to improve this:

https://github.com/bazelbuild/rules_scala/blob/master/scala/scala.bzl#L490

so it is a skylark rule, not a macro, that consumes targets and looks at their runtime and compile-time jars, and just exports them (right now it is not saving much effort).

Then, that could be used to test this feature. I'll add an issue.

Issue #20, opened 2016-03-07
scala_binary _deploy.jar target is not flattened


smparkes:
The goal of the _deploy.jar targets is that they are self-contained, that they can be executed with
`java -jar X_deploy.jar`
The current `_deploy.jar` target looks to be identical to the non-`_deploy.jar` target.

johnynek:
@smparkes see #61 (which has been merged). This should fix the issue. Closing for now.

Issue #57, opened 2016-05-19
Improve scala to java exports


johnynek:
See: https://github.com/bazelbuild/rules_scala/pull/56#issuecomment-220453040

```
one idea could be to improve this:

https://github.com/bazelbuild/rules_scala/blob/master/scala/scala.bzl#L490

so it is a skylark rule, not a macro, that consumes targets and looks at their runtime and compile-time jars, and just exports them (right now it is not saving much effort).

Then, that could be used to test this feature. I'll add an issue.
```

the fix we added for #18 is still pretty mechanical: you pass output jar names. What we really want is to just list targets and get all their exports and runtime deps for use from java.

In this way, we can test that the runtime paths are correct without conflating with `scala_binary`.

colinmarc:
What about adding a `java` provider to `scala_library`? Will the java rules pick that up automatically?

colinmarc:
Also, if that doesn't work, we could generate the `java_import` rules preemptively as `foo_javaimport`.

johnynek:
There is a ticket for some API to make this work, but I think the java rules don't have a contract like this. I wouldn't want to add preemptively out of anxiety that it would slow down the build by doubling the number of targets, half of which would not be used in scala-only repos (even with mixed java-scala it is not that common to depend across languages).

johnynek:
I don't see how to do the `java_import`. We can only call native rules from macros, not other skylark rules. But we can't see in a macro all the dependencies we need to add to the java_import.

I think we may just have to wait until https://github.com/bazelbuild/bazel/issues/970 is solved. I can't see any non-manual way to work around this, but maybe I'm missing something.

colinmarc:
What if we benchmarked preemptively creating `java_import` (assuming that actually works)? It may be that perf works fine with double the targets.

johnynek:
Sure, that would be fine. I actually don't know how to do it though.

Can you see a way?

On Wednesday, May 25, 2016, Colin Marc <notifications@github.com> wrote:

> What if we benchmarked preemptively creating java_import (assuming that
> actually works)? It may be that perf works fine with double the targets.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/bazelbuild/rules_scala/issues/57#issuecomment-221806161>
>


-- 
P. Oscar Boykin, Ph.D. | http://twitter.com/posco | http://pobox.com/~boykin

colinmarc:
[...]

Issue #25, opened 2016-04-06
No support for building deploy jars in scala


damienmg:
See bazelbuild/bazel#468

jcoveney:
It's unclear to me from that ticket what work is remaining to be done on the rules_scala side...care to elaborate?

johnynek:
i think this is a duplicate of #20

PR #61, opened 2016-05-23
Fix deployable JARs created by Scala rules.

devnev:
"Deployable" Jars, as created by the java_binary rule, are Jars with all runtime dependencies included. These are usable as standalone Jars that can be deployed onto servers and such. The reason why we need these for Scala binaries and libraries is as a bandaid fix for Java/Scala interop in Bazel. The Java ecosystem of rules is closed: AFAICT, only known-good rules (cc_library, java_library, java_import, ...) can be used as dependencies of java libraries and binaries. To work around this, the scala_export_to_java macro injects scala into the java ecosystem via "precompiled" jars in the java_imports rule. However, this rule does not pick up transitive runtime dependencies of the imported Jars, so when using these Jars in java_tests, the tests fail because they do not have access to the runtime dependencies of the imported Scala Jar. To make these imports work, the imported Jar must therefore be one that includes the full runtime dependencies in the Jar itself, which is what correctly-built "deploy" Jars provide.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -192,6 +192,25 @@ def _compile_or_empty(ctx, jars, srcjars, buildijar):
       ijar = ctx.outputs.jar
     return struct(ijar=ijar, class_jar=ctx.outputs.jar)
 
+def _build_deployable(ctx, jars):
+  cmd = "rm -rf {out}_tmp\n"
+  cmd += "mkdir -p {out}_tmp\n"
+  for jar in jars:
+    cmd += "unzip -o {jar} -d {{out}}_tmp >/dev/null\n".format(jar=jar.path)
+  cmd += "{jar} cmf {manifest} {out} -C {out}_tmp .\n"
+  cmd += "rm -rf {out}_tmp\n"
+
+  cmd = cmd.format(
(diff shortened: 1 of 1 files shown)


bazel-io:

dinowernli:
This looks legit to me, but please wait for one of @johnynek to ack.

johnynek:
This looks good, and fixes #20 #25 . I do have anxiety that these jars may be quite big, and have a lot of duplication, so actually solving #57 is still worth doing if that is your goal.

Also note, due to #53 , on OSX, jar is not respecting timestamps, so each run invalidates the cache, which is a real bummer. A fix for that will have fix this code as well.

johnynek:
@devnev thanks for your contribution!

johnynek:
@devnev we are having some issues with this PR. It significantly increases the build time for all our targets.

We need to find a way to deal with this. Is it the case that you only want the `scala_to_java_export` to work correctly? If so, we can fix that to actually compute the real deps and not generate the deploy jar for all libraries. It seems that in bazel, deploy jars are only implicit outputs of `java_binary` so maybe we should follow that approach and not build them for `scala_library`.

johnynek:
also, see issue bazelbuild/bazel#970

PR #63, opened 2016-05-24
Add support for a repl target

### LINKED ISSUE #58: REPL support
I don't think that bazel has any notion a repl, but I can imagine adding macro like

```
scala_repl(name = "foo", deps = [":a", ":b"])
```
which will create a repl with the classpath given by the runtime dependencies of `:a, :b`.

Actually, remembering terminal pains, the easiest thing might be to do a real skylark rule with `scala` in the scala-sdk, or perhaps use ammonite: http://www.lihaoyi.com/Ammonite/#Features

johnynek:
closes #58

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,3 +285,19 @@ A `scala_test` by default runs all tests in a given target.
 For backwards compatiblity it accepts a `suites` attribute which
 is ignored due to the ease with which that field is not correctly
 populated and tests are not run.
+
+
+<a name="scala_repl"></a>
+## scala_repl
+```python
+scala_repl(name, deps, scalacopts, jvm_flags)
+```
+A scala repl allows you to add library dependendencies (not currently `scala_binary` targets)
+to generate a script to run which starts a REPL.
+Since `bazel run` closes stdin, it cannot be used to start the REPL. Instead,
+you use `bazel build` to build the script, then run that script as normal to start a REPL
+session. An example in this repo:
+```
+bazel build test:HelloLibRepl
+bazel-bin/test/HelloLibRepl
+```
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -366,6 +366,37 @@ def _scala_binary_impl(ctx):
   _write_launcher(ctx, rjars)
   return _scala_binary_common(ctx, cjars, rjars)
 
+def _scala_repl_impl(ctx):
+  jars = _collect_jars(ctx.attr.deps)
+  rjars = jars.runtime
+  rjars += [ctx.file._scalalib, ctx.file._scalareflect]
+  rjars += _collect_jars(ctx.attr.runtime_deps).runtime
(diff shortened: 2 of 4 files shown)


bazel-io:

johnynek:
@smparkes @dinowernli any comments on this approach?

ittaiz:
@johnynek I hope I'm not out of line but aren't you worried about future regression?
There are no tests for the output (script) and you don't know that it actually works (without manually running it afterwards).

johnynek:
I am certainly worried about it. I added a test target, but I'm not sure
the best way to test it.

That said, a repl is considerably less critical than targets to build
deployed code, so the risk is also smaller.

Do you have any suggestions on how to test this? I could probably redirect
stdin, assuming the scala repl supports that, and use the existing test
script.

On Monday, May 23, 2016, Ittai Zeidman <notifications@github.com> wrote:

> @johnynek <https://github.com/johnynek> I hope I'm not out of line but
> aren't you worried about future regression?
> There are no tests for the output (script) and you don't know that it
> actually works (without manually running it afterwards).
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/bazelbuild/rules_scala/pull/63#issuecomment-221176354>
>


-- 
P. Oscar Boykin, Ph.D. | http://twitter.com/posco | http://pobox.com/~boykin

ittaiz:
stdin redirect is what I had in mind

dinowernli:
@johnynek  @ittaiz 

I think the approach looks good, but I am also a bit worried about this breaking inadvertedly. If there is a convenient way to test this, I feel that would be helpful.

That said, generally LGTM :)

johnynek:
Thanks for the review folks. I added a test.

johnynek:
look good? @dinowernli @ittaiz ?

ittaiz:
I might be way off but isn't the assert stage missing? Or maybe if you mean for the assert to be on the fact that it's an executable script then assert is a bit narrow.
The feature, as far as I understand, is:
"Enable the creation of an executable script that when runs creates a Scala REPL with the test classpath and can be configured with various JVM opts".
We can decide the configure part won't be tested but to test the other parts maybe we just need to add an stdout redirect which asserts the REPL output (just "foo"?). This tells us the missing pieces about a Scala REPL with the right classpath.

johnynek:
yeah, I foolishly did not check that I could make that fail, and must have thought a failing import would give an error code, which of course it can't unless the REPL fails immediately on any error.

Now I am checking that I see "foo scala" which is printed by that command, which should not show up if we can't set up the classpath correctly. Checking for a bad classpath fails this test locally.

@ittaiz can you take another look?

ittaiz:
👍🏽

dinowernli:
Slightly late, but LGTM :)

PR #64, opened 2016-05-24
fixes #60 by clarifying the getting started path

### LINKED ISSUE #60: Possible confusion in the docs //scala:scala.bzl vs @io_bazel_rules_scala//scala:scala.bzl
First of all thanks a lot for all your hard work, it's really appreciated!

I'm not sure this is in the scope of this repo but the fact of the matter is that I went along the readme and since it mentions the `load` with respect to `prelude_bazel` I wasn't sure if there isn't a different syntax for that use-case.
I went to the tests since they are the executable specification and failed since they use the absolute(?) path `//scala:scala.bzl`
This is an on-boarding issue since if I was familiar with Bazel syntax well enough I wouldn't have made that mistake but I think we want to make on-boarding easier.
2 options I had in mind:
1. Add a note in the readme at the end of the overview explaining the difference
2. Add a scala app to the bazelbuild/examples repo
The added value of #2 is that it serves as a bit of a test application since it has to use the syntax like "regular" users.

I'd be happy to contribute both options but didn't want to barge in with a PR before we discuss it.

ittaiz:
@johnynek I took the liberty of adding another sentence about the prelude since that by itself was another gotcha for me when I started out

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ This rule is used for building [Scala][scala] projects with Bazel. There are
 currently four rules, `scala_library`, `scala_macro_library`, `scala_binary`
 and `scala_test`.
 
+## Getting started
+
 In order to use `scala_library`, `scala_macro_library`, and `scala_binary`,
 you must have bazel 0.2.3 and add the following to your WORKSPACE file:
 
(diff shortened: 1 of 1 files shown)


bazel-io:

johnynek:
@ittaiz can you sign the google CLA, and then I will merge your PR?

ittaiz:
Will do soon, waiting for our "people" to give me the green light

ittaiz:
I signed it!

ittaiz:
I signed it as Wix.com btw with my email as the contact person

PR #67, opened 2016-05-27
short paths may not always be valid file names, targets labels seems …

ianoc:
…like a better bet for this.

This fixes an issue with external git repository building with the master copy of Bazel too.


bazel-io:

johnynek:
Thanks @ianoc!

PR #68, opened 2016-05-31
Don't allow transitive scala_library dependencies onto the classpath

colinmarc:
Without this patch, scala and java "deps of deps" end up on the
classpath. This isn't the way java_library rules work:

> A rule X can access the code in Y if there exists a dependency
> path between them that begins with a deps edge followed by zero
> or more exports edges.

I verified my understanding of this behavior with a small test repo,
copying one of the tests from this repo and changing 'scala_' to
'java_':
https://gist.github.com/colinmarc/c683652ef9a7d817c5e7ce9ca1236a6d

This patch fixes that, and all the tests that were incorrectly
depending on the behavior. It also adds some negative tests for the
behavior, under `test_transitive_deps`.

Unfortunately, this is going to be majorly backwards incompatible
with existing scala repositories.


bazel-io:

test/Exported.scala
```diff
@@ -1,11 +1,5 @@
 package scala.test
 
 object Exported {
-  def message: String = {
-    // terrible, don't do this in real code:
-    val msg = Class.forName("scala.test.Runtime")
```
johnynek: please don't remove this. This was testing something spefic: the presence of `scala.test.Runtime` on the runtime classpath. You should add what you need to make this work, not remove the testing of runtime classpaths.

test/HelloLib.scala
```diff
@@ -23,7 +23,6 @@ object HelloLib {
     MacroTest.hello(arg == "yo")
     println(getOtherLibMessage(arg))
     println(getOtherJavaLibMessage(arg))
-    println(Exported.message)
```
johnynek: please do not remove this. Update the builds to make this work.

test/src/main/scala/scala/test/twitter_scrooge/BUILD
```diff
@@ -9,6 +9,12 @@ scrooge_scala_library(
     ":scrooge2_b",
     ":scrooge3",
   ],
+  exports = [
+    "//test/src/main/scala/scala/test/twitter_scrooge/thrift",
```
johnynek: this is a thrift target. Why are you exporting it? What does that mean?

test/src/main/scala/scala/test/twitter_scrooge/BUILD
```diff
@@ -18,6 +24,7 @@ scrooge_scala_library(
     "//test/src/main/scala/scala/test/twitter_scrooge/thrift/thrift2:thrift2_a",
     ":scrooge3",
   ],
+  exports = [":scrooge3"],
```
johnynek: why are all the scrooge deps also exported now? Can you add a comment next to each export explaining why it is needed?

johnynek:
@colinmarc will you be able to update this patch?

colinmarc:
Sorry for the delay - want to take another look?

As discussed offline, I gave scrooge_scala_library the ability to automatically export its deps, which is literally just `exports = deps + exports`. I still think this is the wrong move, especially since I still had to change some `scala_library` targets in the scrooge tests anyway.

I also found some tests that I had forgotten to commit before.

johnynek:
tests are red.

One thought. Can we move `test_transitive...` to `test_expect_failure` or something and have one directory where we expect failing targets.

colinmarc:
Passing now. I forgot to `exit 0` at the end of the should-fail test.

johnynek:
Thanks for the fix for consistency with java_library.

PR #69, opened 2016-05-31
Use short_path for runfiles paths

kchodorow:
_java.path was (incidentally) working, but is now broken. short_path is
for paths relative to the runfiles tree.

This fixes ci.bazel.io's rules_closure project. See https://github.com/bazelbuild/bazel/issues/1316 for more info.


johnynek:
Thanks for fixing this!

PR #70, opened 2016-06-01
Add repl support to binary and test targets

johnynek:
actually, starting repls for test code and binaries is really useful, so I added support for that as well. This is done by making the binary targets return their deps correctly.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -368,14 +368,23 @@ def _scala_macro_library_impl(ctx):
 # Common code shared by all scala binary implementations.
 def _scala_binary_common(ctx, cjars, rjars):
   write_manifest(ctx)
-  _compile_or_empty(ctx, cjars, [], False)  # no need to build an ijar for an executable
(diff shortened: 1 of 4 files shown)


bazel-io:

johnynek:
@dinowernli @smparkes @colinmarc @ianoc ?

scala/scala.bzl
```diff
@@ -368,14 +368,23 @@ def _scala_macro_library_impl(ctx):
 # Common code shared by all scala binary implementations.
 def _scala_binary_common(ctx, cjars, rjars):
   write_manifest(ctx)
-  _compile_or_empty(ctx, cjars, [], False)  # no need to build an ijar for an executable
+  outputs = _compile_or_empty(ctx, cjars, [], False)  # no need to build an ijar for an executable
```
ianoc: should we give this a different name so its separate from the outputs defined on 379?

ianoc:
looks fine to me

dinowernli:
LGTM

Issue #26, opened 2016-04-06
Support mixed scala/java targets


damienmg:
Scala rules version of bug bazelbuild/bazel#500. Not sure if there is still work to do or not.

johnynek:
for mixed single targets we need this:
http://www.codecommit.com/blog/scala/joint-compilation-of-scala-and-java-sources

ittaiz:
any plans for this?
I work with @barlevalon (who opened the issue in bazel) and I'll probably be trying to tackle it in the next few days following the link @johnynek added here.
If there are objections or tips I'd love to hear them

johnynek:
I'm happy to review and merge code here, but I really don't need joint scala/java targets.

(I find that scalac by itself is slow enough that I really try to keep targets very small, that said, old projects might be easier to build with this if we have this capability).

ittaiz:
Awesome, thanks!

For us it's really mandatory since our codebase has really evolved side by
side and many of our core projects still contain Java "pockets".

Any tips?
My current direction is to add a joint_scala_java method similar to
scala_library which will be similar to the existing scala_library but will
run scalac and javac like in the article above.
Does that sound good? If so I'll drill down.

johnynek:
I think ideally, we would just use `scala_library` and `scala_binary` and use a filetype filter to select the `.java` and `.scala` sources. Then the work flow is identical except that we also accept some java files. I would error if there is not at least one `.scala` file in the sources, however.

As a test, you might first make a totally separate rule as you suggest, but my guess is there will be a lot of duplication.

ittaiz:
👍🏽

ittaiz:
@johnynek I've started fiddling around and indeed got the scala compiler to compile with very small changes like you suggested.
I'm a bit stuck on how to invoke the java compiler so that the resulting jar will contain all the needed .class files.
I've asked in bazel-discuss but I'll use this platform also if that's ok to ask you: It seems to me that I need to run the javac in `_compile` after `scalac` runs but I don't know how to invoke multiple actions in the same rule. Any pointers? Thanks!

ittaiz:
update: brian silverman helped me in discuss by pointing me to macros.
Current direction (which I don't really like) is to have a macro which will call:
scala_library(java+scala) and then scala_export_to_java and then native.java_library(java)
[...]

PR #71, opened 2016-06-01
[WIP] initial support of mixing java scala sources

ittaiz:
relates to #26 
Current approach generates a macro which creates a scala_library target and then a scala_export_to_java target and then a java_library target.
I've added a Binary to test everything works.
Open issues:
1. Is the test ok? I don't really like the reliance on the run of the binary and I feel it's become too cluttered but I'm trying to learn the lay of the land.
2.  I still need to see how I support and pass through all the properties java_library and scala_library ryles support
3. I want to unite java sources and scala sources back to one and then filter the scala sources when calling the java library. Doesn't work for me both on types and the filter breaks. Would love pointers.
4. Additionally currently only array works, no globbing
5. run time deps, what to do with them? just another parameter to pass through?
6. Should I add a test which checks scala_binary?
seems redundant since the below rough draft of a test didn't break the build so I'm guessing the binary uses the compile as well, do you want it here as regression?

```python
#scala_binary(
#    name = "MixJavaScalaLibBinary",
#    srcs = ["src/main/scala/scala/test/MixJavaScalaBinary.scala"] + ["src/main/scala/scala/test/MixJavaScalaForBinary.java"],
#    main_class = "scala.test.MixJavaScalaLibBinary",
#    deps = ["MixJavaScalaLib"],
#) 
```
Thanks!


bazel-io:

ittaiz:
I signed the cla already on behalf of my employer. How do we connect that info in?

ittaiz:
I signed it!

ittaiz:
@johnynek would appreciate your feedback?
I think the macro choice might be a problem since I've read in one of the threads that it doesn't support globbing. Is that correct? If so, do you have a different direction I can take it to?
I'm not sure how relevant it will be for us without globbing

johnynek:
This will certainly work for a class of mixed projects, but what if the scala depends on the java, or there are circular dependencies? I *thought* the approach of the blog post I linked to allows more general kinds of dependencies than here, and would have the benefit that we would allow java sources mixed with the scala sources in any of our targets.

What do you think about the approach of having scalac parse the java, then calling javac to finish the compilation of the java sources?

damienmg:
For the CLA: the googlebot does not handle correctly corporate CLA :(

Verified that Wix have sign the CLA :)

ittaiz:
@johnynek the scala depends on the java in my scenario:
Bar(scala) extends Baz(java) extends Foo(scala)
Scalac indeed parses the java to creates stubs (that's the only way Bar can compile) and then javac is called (via the java_library rule).
I've of course saw the test fail with scalac not able to compile Bar because it doesn't know of Baz.
I also added the binary part to the test to make sure that the mixed rule actually compiles the java itself (hence the main method working).

A drawback I see to my implementation choice is that it uses a macro (due to bazelbuild/bazel#970) which has strict limitations on globbing AFAICT.
Do you think I should call javac myself in the scala_library rule?

johnynek:
I'm just concerned about this using a macro and further digging us in on bazelbuild/bazel#970 . The use of the macro uses deploy jars, which are slow to build and make the compile classpath super huge. We lose downstream ability to see when `deps` are not declared correctly because they will get everything in the full transitive runtime path.

If we call `javac` ourselves to compile the java portion rather than using `java_library` we can deal with that, although that has other issues (namely, java code is being build differently that it would elsewhere).

I also don't like how this has two `srcs` rather than just doing one like other bazel rules.

ittaiz:
+1 about the two srcs completely. Unless I'm mistaken the macro path
mandates it, right? I tried to take one and filter it but got errors.

Q- let's say bazelbuild/bazel#970 is solved, that will only save us the
export, right? AFAICT We'd still be stuck with the macro because I haven't
found a way for a rule to *use* another rule.

If there is consensus about duplicating the code of the java_library rule I
can do that.
Would love some more opinions

ittaiz:
btw,
this discussion led me towards using the macro and not copying the java_lib code:
https://groups.google.com/forum/#!searchin/bazel-discuss/multiple/bazel-discuss/4OLv-OJ3J-o/nw_nWlyhAgAJ

johnynek:
@ittaiz that post on the master pattern is interesting:

http://www.bazel.io/docs/skylark/cookbook.html#master-rule

could we avoid using the scala export (which forces deploy jars) by just making a target that exports both the scala and java target inside the macro?

ittaiz:
@johnynek I might be missing something but how would this help us in having java code depend on scala code?
The java_library target needs to have in its inputs the scala outputs. That's currently not an option due to bazelbuild/bazel#970

johnynek:
yeah, you are right. I keep forgetting that skylark rules can't call `native.java_library` so you either have macro power to do that, or skylark rules.

I guess this is the best we can do until bazelbuild/bazel#970 is solved. Do you agree?

ittaiz:
Agree about duplicating the javac or about using the export?
I talked with @cgrushko yesterday and *maybe* there's an intermediate step that won't solve 970 but will remove the current blocker.

ittaiz:
@johnynek I've updated the PR with unification of the srcs arguments to one (learned the limitations of macros and how to use them).
I think that we should hold off until @cgrushko fixes 970 or at least exposes an intermediate solution.
Once that's in I'll be able to remove the use of scala_export_to_java.
Given we've past that do you have any tips for me on how to handle all of the args? Should I declare the union of scala_library and java_library arguments and pass them in?

ittaiz:
@johnynek I got some interesting feedback in this [discussion] (https://groups.google.com/forum/#!topic/bazel-discuss/kp4zE4XBjyo) which basically encouraged me to use **kwargs as a means to not needing to play catchup with the rules I'm shading.
Open questions:
1. WDYT about this approach?
2. Does bazel support the python3 keyword only synatx (, *,)?
3. If it doesn't how do you want to handle mandatory arguments of both java and scala rules? One option is to just define them by name ("risking" they are called positional) and then pass those to the respective rules + the kwargs
4. How does a rule handle additional arguments from this dictionary (just haven't gotten around to checking it)?

ittaiz:
@johnynek I continued with the `**kwargs` approach and I'd love to hear your opinion about the code.
I know you don't like the `scala_export_to_java` usage and I hope to either drop it and inline all of the macro into `scala_library` (and copy `java_library` in) or get some news about bazelbuild/bazel#970. 
In any case I think this commit is valuable since whenever bazelbuild/bazel#970 is solved I think this should be our approach.
I'm exporting the scala jar from the java target since I think that's mandatory for users of the target to be able to use both java and scala code but please correct me if I'm mistaken. 
The only open issue AFAIU is the tests- I didn't add tests that parameterize exports/runtime_deps/deps or any other property of scala/java library rules. I didn't add them because creating a binary and running it feels very coarse grained and I wasn't sure if you want me to add all those permutations inside.
WDYT? How would you like to proceed?

ittaiz:
@johnynek @damienmg 
I just stumbled on to `java_rules_skylark.bzl` in bazel itself and I was wondering maybe the correct temporary hack is to use that instead of manually copying and calling javac.
The file does have a "not production ready" warning but maybe it's good enough until 970 is solved.
WDYT?
Link:
https://github.com/bazelbuild/bazel/blob/master/tools/build_rules/java_rules_skylark.bzl
Warning in the file:
"This is a quick and dirty rule to make Bazel compile itself. It's not production ready."

johnynek:
I don't know the status of that, but it could be a possibility indeed. We might need to refactor it somewhat, or perhaps all the functions we need to call are exposed.

The lack of production readiness worries me (but that said, the scala rules are still evolving quite a lot, so I think the same may be said of them if you are not willing to roll up your sleeves).

ittaiz:
Don't know if @damienmg wants to refactor that rule. I tried to use it and it doesn't support exports and I guess a lot of other flags of java_library.
What is our alternative? When you say call javac directly aren't you talking about basically copying all of the code in that bzl and adding to it what we need? Is there a simpler way I'm not seeing?
Maybe you mean to expose it without customization?

scala/scala.bzl
```diff
@@ -519,6 +520,27 @@ scala_repl = rule(
   executable=True,
 )
 
+def mixed_scala_java_library(name, srcs, **kwargs):
+    scala_library(
+        name = name + "_mix_scala",
+        srcs = srcs,
+        **kwargs
```
bsilver8192: I think it's nicer to leave visibility out of kwargs (default it to None) and pass it just to the java_library, and then make the other two `visibility = ['//visibility:private']`. That makes it harder for other code to depend on implementation details, and makes the intention a bit more obvious to anybody reading.

johnynek: +1

ittaiz:
@bsilver8192 thanks for the idea! Done.
@johnynek 
I need your input on how to proceed:
Simplest option I see is to PR bazel to add support to java_rules_skylark.bzl only for `exports` and hope that 970 is solved before people need proper customization (like javacopts, licenses etc).
Other option is to "copy all of the code in that bzl and add to it what we need".
WDYT?
This is somewhat of a blocker for me

johnynek:
what do you think of looking here:

https://github.com/ittaiz/rules_scala/blob/c145b2a53ddb47cde5ddaaa635b3faa1d9591c3e/scala/scala.bzl#L156

and adding a call to `javac` if the list of java files are not empty. Then we add the resulting classes to jar, and we have already set up the classpath.

Would that not work? If we did that then all scala rules would also support java without any extra work, which seems nicer to me.

ittaiz:
Sure.
I'll give it a try and ping back.

ittaiz:
@johnynek done.
Two small issues:
for scalac we have scala_opts but i don't pass the same for javac- defer until someone needs it?
the formatting of "ins" is in one line but I wasn't able to break it down due to indentation. Tell me what you want to do with it

scala/scala.bzl
```diff
@@ -128,6 +130,20 @@ def _compile(ctx, _jars, dep_srcjars, buildijar):
...
+    javac_args_file = ctx.new_file(ctx.outputs.jar, ctx.label.name + "_javac_args")
+    javac_args = """ -classpath "{jars}:{out}_tmp" -d {out}_tmp {files}""".format(
+      jars=":".join([j.path for j in jars]),
+      files=" ".join([f.path for f in java_srcs]),
+      out=ctx.outputs.jar.path
+      )
+    ctx.file_action(output = javac_args_file, content = javac_args)
+    javac_sources_cmd = """
+    cat {javac_args} {{out}}_args/files_from_jar > {{out}}_args/java_args
+    env JAVACMD={{java}} {javac} {{jvm_flags}} @{{out}}_args/java_args""".format(javac_args = javac_args_file.path,javac=ctx.file._javac.path)
```
johnynek: this environmental variable is used by scalac (JAVACMD) is the same used by javac? Did you test with a broken value to see if that fails?

scala/scala.bzl
```diff
@@ -174,22 +190,11 @@ rm -rf {out}_tmp_expand_srcjars
   outs = [ctx.outputs.jar]
   if buildijar:
     outs.extend([ctx.outputs.ijar])
+  ins = list(jars) + list(dep_srcjars) + list(srcjars) + list(sources) + ctx.files.srcs + ctx.files.plugins + ctx.files.resources + ctx.files._jdk + ctx.files._jar + ctx.files._scalasdk + [ctx.outputs.manifest, ctx.file._ijar, ctx.file._scalac, ctx.file._java, scalac_args_file]
```
johnynek: can we wrap this line again, it is really long.

scala/scala.bzl
ittaiz: What do you mean by broken value?

scala/scala.bzl
ittaiz: I wrote about it in the comments. Tried to but couldn't due to indentation.
Python noob and probably making a small mistake.

johnynek:
seems like following:
http://www.bazel.io/docs/be/java.html#java_library

and adding `javacopts` to scala_library rules seems easy to support.

This seems pretty good. What do you think? I think the current PR is the cleanest from a usability standpoint (clearly) not all java features of java_library are supported, but if you need more, factor the code and make a java_library (or submit a PR).

What do you think?

Any comments @smparkes @dinowernli ?

scala/scala.bzl
johnynek: So, try it with `env JAVACMD=brokenvalue` and I expect it to fail if `javac` is really using it.

We should try to set the correct java command for javac (which I *think* is a java program itself, but I may be wrong).

ittaiz: ok tried the following (a bit brute force since I'm still wrapping my head around some of the semantics):
```python
javac_sources_cmd = """
    cat {javac_args} {{out}}_args/files_from_jar > {{out}}_args/java_args
    env JAVACMD2={{java}} {javac} {{jvm_flags}} @{{out}}_args/java_args""".format(javac_args = javac_args_file.path,javac=ctx.file._javac.path)
``` - notice the "2" which passed the build
```python
javac_sources_cmd = """
    cat {javac_args} {{out}}_args/files_from_jar > {{out}}_args/java_args
    {{java}} {javac} {{jvm_flags}} @{{out}}_args/java_args""".format(javac_args = javac_args_file.path,javac=ctx.file._javac.path)
``` - without the "env JAVACMD=" which failed (Could not find or load main class external.local_jdk.bin.javac)
```python
javac_sources_cmd = """
    cat {javac_args} {{out}}_args/files_from_jar > {{out}}_args/java_args
    env JAVACMD=brokenValue""".format(javac_args = javac_args_file.path,javac=ctx.file._javac.path)
``` - with the broken value which failed (javac couldn't find the scala source it needed on the classpath)

Sadly enough I'm not sure what this *means* so would appreciate your narration

**Comment by ittaiz** [comment]
I can add javacopts if that's the only thing you think is missing.

File: scala/scala.bzl (hunk 2 above)
Comment (ittaiz) [inline scala/scala.bzl]: I also tried using just `javac` without `java` which is what I thought you meant but that doesn't work as well
```python
    javac_sources_cmd = """
    cat {javac_args} {{out}}_args/files_from_jar > {{out}}_args/java_args
    env JAVACMD={javac} {{jvm_flags}} @{{out}}_args/java_args""".format(javac_args = javac_args_file.path,javac=ctx.file._javac.path)
```
error:
```
env: @bazel-out/local-fastbuild/bin/test/MixJavaScalaLib.jar_args/java_args: No such file or directory

ittaiz:
@johnynek any more feedback?

ittaiz:
@johnynek how would you like to proceed?

scala/scala.bzl
johnynek: in `scalac`, which is a bash script that starts a jvm and runs scala code (the compiler), `JAVACMD` is the environmental variable that *may* be set to control which `java` command to use. Since we want to use the one that bazel knows about, we set it.

The more I think about it, it seems like since bazel is setting up javac for us, it should be correct. So I think it should be fine to remove `env JAVACMD={java}` and just do `{{javac}}`. If we see a problem later, we will fix it.

johnynek:
Okay, just remove the (I think) unneeded `JAVACMD` and we'll merge.

Thanks for doing this!

ittaiz:
@johnynek done!
Thanks for your mentorship on this issue.
Hopefully many more to come :)

ittaiz:
@johnynek can we merge?

johnynek:
Yes! This is a great improvement!

Thanks for your hard work!

Issue #76, opened 2016-06-13
test_run should be a sh_test


damienmg:
This would allow to test it with `bazel test`

To do so, shipping the source of the repository might be needed. See how it is done in bazel in src/test/shell/bazel/...

ittaiz:
@johnynek seeing as test_run does `bazel clean` this means that running `bazel test //...` to get feedback on changes will take longer right?
Is it just something to live with since the repo is small? Is there anything to be done with it?

damienmg:
You want to recreate a very simple repo inside the sandbox for the test. This would clean only that repo. You have to depends on the bazel binary though.

johnynek:
here is the link: https://github.com/bazelbuild/bazel/tree/master/src/test/shell/bazel

this seems like a good thing to do for a shell-master that would like to contribute. Maybe we can reuse some of the above. Honestly, my shell programming is so weak, this would take me a lot of time, and I am not sure I'll have time to do it soon.

or-shachar:
@damienmg @johnynek I'm gonna work on this task.

I'm evaluating different options for implementation and will be happy to hear your thoughts

Here are some of the challenges. Some of them were already mentioned above.

### (A) current to run `test_run.sh` we need the whole repo
like @damienmg , I had to allow copying the sources of each package and recreate the same workspace in the test run environment. I started with a very naive solution, will make it better in the future.

### (B) running with `bazel test` uses vanilla env and causes long build time
just like @ittaiz mentioned. It takes a lot more time to do it inside bazel test run environment. 
Running the bazel build alone took > 5 minutes on my machine (mac pro).
We have a theory that it's downloading scala binary on each run, which takes most of the time. I'm looking in the `src/test/shell/bazel...` for similar executions to find some optimization. If you can point me out to specific test target - would be awesome!

@damienmg  - can you elaborate about your idea to recreate a very simple repo inside the sandbox? wouldn't it create a whole different test?

### (C) splitting to different test targets
It seems like it would be clearer to run each `run_test` in a17 different `test_sh` targets. It would make the result more readable and allow parallel run in the future. Problem is that most tests rely on the bazel build step, and since bazel is creating isolated env for each test run - it would re-build all sources and re-import all dependencies for each step. What can we do about it? Should we leave it as one `enormous` test?

### (D) Running `bazel` inside bazel execution behaves differently
As mentioned - I managed to recreate the workspace in the test environment.
This is the output:

> INFO: Found 1 test target...
[...]

PR #83, opened 2016-07-30
Fix for thrift rule issue that manifests with remote repos

johnynek:
The problem is that the path is not the same when remote as local. The current fix:

1. makes sure that a prefix, if given, is an actual substring of the common prefix.
2. ignores anything *before* the prefix that matches. So it is like the regex `.*prefix(.*)$`.

The above two fixes make this rule usable in both local and remote contexts for me.

### DIFF
--- a/thrift/thrift.bzl
+++ b/thrift/thrift.bzl
@@ -2,11 +2,40 @@
 
 _thrift_filetype = FileType([".thrift"])
 
+def _common_prefix(strings):
+  pref = None
+  for s in strings:
+    if pref == None:
+      pref = s
+    elif s.startswith(pref):
+      pass
+    else:
+      tmp_pref = pref
+      for end in range(0, len(pref) + 1):
+        test = pref[0:end]
+        if s.startswith(test):
+          tmp_pref = test
+      pref = tmp_pref
+  return pref
+
 def _thrift_library_impl(ctx):
   prefix = ctx.attr.absolute_prefix
+  src_paths = [f.path for f in ctx.files.srcs]
+  if len(src_paths) <= 0:
+    fail("we require at least one thrift file in a target")
+
   jarcmd = "{jar} cMf {out} -C {out}_tmp ."
   if prefix != '':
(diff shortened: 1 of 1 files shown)


bazel-io:

johnynek:
@ianoc review?

johnynek:
@jcoveney ?

thrift/thrift.bzl
```diff
@@ -2,11 +2,40 @@
...
+def _common_prefix(strings):
+  pref = None
+  for s in strings:
+    if pref == None:
+      pref = s
+    elif s.startswith(pref):
+      pass
+    else:
+      tmp_pref = pref
+      for end in range(0, len(pref) + 1):
```
johnynek: no `while` in skylark. :(

johnynek: (skylark may be a total language... No recursion, only for loops on collections, I think everything terminates).

If you consider any method error to be a type error (which a better compiler would have caught, I think it may be total....)

ianoc:
looks reasonable to me. Is it possible to add a test for the issue so it doesn't re-occur?

johnynek:
@ianoc we can add a test. It is not super easy to do so. We need to set up some kind of remote repo (maybe a nested local repository will do it), then verify that the build of a target that depends on the remote thrift target has all the thrift files we expect.

ianoc:
I'd be inclined to merge it since its strictly better, but maybe we should just file an issue for future clean up?

johnynek:
@ianoc yeah, added: https://github.com/bazelbuild/rules_scala/issues/87

The testing story is not great. The bazel folks want us to drive the tests in bazel, but that requires a lot of shell hacking: #76

ittaiz:
Just as general inquiry:
What do you mean by drive the tests in Bazel?

johnynek:
@ittaiz the suggestion is for the tests to be run with `bazel test`. Then you inside a bazel `sh_test` you create a new temporary WORKSPACE and BUILD files, and inside that script again call `bazel`.

Did you see the movie Inception? Like that.

ittaiz:
:) Got you.
Thanks for the explanation.

PR #84, opened 2016-08-09
Use bind's for twitter scrooge so local repo's can override scrooge and thrift versions

ianoc:


bazel-io:

johnynek:
looking

PR #85, opened 2016-08-16
improve the deploy jar creation

johnynek:
This does not pass the zip contents through the file system and makes deploy jar creation part a executable action. It should be both faster, but also safer for creation of deploy jars on linux systems (such as Docker or with encrypted home directories) that often have limited file lengths that might be violated by jars.

review? @ianoc @ittaiz ?

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -233,9 +233,18 @@ def _compile_or_empty(ctx, jars, srcjars, buildijar):
     return struct(ijar=ijar, class_jar=ctx.outputs.jar)
 
 def _build_deployable(ctx, jars):
+  # the _jar_bin program we call below expects one optional argument:
   # -m is the argument to pass a manifest to our jar creation code
-  # the next argument is the manifest itself, then the target jars
-  # finally the list of jars to merge
+  # the next argument is the path manifest itself
+  # the manifest is set up by methods that call this function (see usages
+  # of _build_deployable and note that they always first call write_manifest).
+  # that is what creates the manifest content
+  #
+  # following the manifest argument and the manifest, the next argument is
+  # the output path for the target jar
+  #
+  # finally all the rest of the arguments are jars to be flattened into one
+  # fat jar
   args = ["-m", ctx.outputs.manifest.path, ctx.outputs.deploy_jar.path]
   args.extend([j.path for j in jars])
   ctx.action(
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -233,10 +233,13 @@ def _compile_or_empty(ctx, jars, srcjars, buildijar):
     return struct(ijar=ijar, class_jar=ctx.outputs.jar)
 
 def _build_deployable(ctx, jars):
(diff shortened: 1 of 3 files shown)


bazel-io:

scala/scala.bzl
```diff
@@ -233,25 +233,15 @@ def _compile_or_empty(ctx, jars, srcjars, buildijar):
...
-    cmd += "unzip -o {jar} -d {{out}}_tmp >/dev/null\n".format(jar=jar.path)
-  cmd += "{java} -jar {jar} -m {manifest} {out} {out}_tmp\n"
-  cmd += "rm -rf {out}_tmp\n"
-
-  cmd = cmd.format(
-      out=ctx.outputs.deploy_jar.path,
-      jar=_get_jar_path(ctx.files._jar),
-      java=ctx.file._java.path,
-      manifest=ctx.outputs.manifest.path)
+  args = ["-m", ctx.outputs.manifest.path, ctx.outputs.deploy_jar.path]
```
ittaiz: what does -m stand for? Obviously a noob question but maybe should be a constant? a bit of a magic symbol?

src/java/io/bazel/rulesscala/jar/JarHelper.java
```diff
@@ -189,24 +196,53 @@ protected void copyEntry(String name, File file) throws IOException {
...
-        outEntry.setSize(size);
-        if (size == 0L) {
-          outEntry.setMethod(JarEntry.STORED);
-          outEntry.setCrc(0);
-          out.putNextEntry(outEntry);
-        } else {
-          outEntry.setMethod(storageMethod);
-          if (storageMethod == JarEntry.STORED) {
-            outEntry.setCrc(hashFile(file));
+        if (JarHelper.isJar(name)) {
```
ittaiz: WDYT about extracting this to a private method? what this does is unpack the jar into the deployable?

ittaiz:
@johnynek are there any tests around this feature? If not I'd add one or two. If there are then this does look like refactoring so it's not needed

johnynek:
@ittaiz there are existing tests that exercise this (the java consumption of scala deploy jars) and on top of that, this fixes the issue seen on linux (the contents of some jars were too long, so we could not make the fat jar).

johnynek:
@ittaiz I am not sure I follow. That flag is how we signal to the jar program that we are passing a manifest file. I think your suggestion is confusing, I don't understand it. In the interest of you saying "ship it" I am tempted to take the suggestion, but it seems to mislead the reader if we say "USE_EXISTING_MANIFEST_FLAG". What is the existing one? The ones in the inner jars?

By the way, we generally have not documented flags in other scripts. So this is a bit of a break to concern ourselves so deeply with this one flag.

I'll try one more time and see if you like it.

johnynek:
@ittaiz I updated the code. If you have any suggestions on how to improve the wording, please make a suggestion. Otherwise, I can merge and you can make a PR that improves it if you like.

I appreciate you taking the time to look at the change.

ianoc:
LGTM

PR #86, opened 2016-08-16
Exclude non-ASCII file names

unlikely-leo:
This fixes the "Bad magic" error encountered before, at the cost of, well, excluding some files that might be necessary.


bazel-io:

PR #88, opened 2016-09-01
Add specs2 support

mvaitkus:
Adding two rules to run specs2 tests:
- scala_specs2_test for running standalone specs2 tests
- scala_specs2_junit_test uses junit runner for test execution

Junit xml output is not yet supported.

Rules tested in:
https://github.com/mvaitkus/bazel_scala_sample


bazel-io:

ittaiz:
@johnynek they're Wix employees, they should be covered by the Wix organizational CLA right? cc @damienmg

damienmg:
CLA bot got it wrong again.

mvaitkus:
There are some things we are not sure about. 

Currently in order to use the new rules, you need to add following lines to project WORKSPACE file:
`load("@io_bazel_rules_scala//specs2:specs2.bzl", "specs2")`
`specs2()`
Executing it will load specs2 dependencies to workspace using native.maven_jar(... so they can be accessed later in rules. We assume it is possible to somehow load them in scala_repositories method and would most likely not be downloaded if specs2 rules are not used, but we where not sure if it is good idea to pollute workspace with unused labels.

Another thing we'd appreciate feedback on is how we produce a list of test classes from the jar (we use `jar tf` command on the output jar to do this and then `grep | sed | xarg run`). Wondering if there's a better way to do this using bazel/skylark instead of using this bash.

Do you have any recommendations on xml output? 
What we could think of - write a custom runner(s) with junit/specs event listeners. The runner(s) could probably also handle the tests to run list instead of bash command.

ittaiz:
👍🏽 thanks!

johnynek:
Sorry for the delay. I've been on vacation. Will look tomorrow. Thanks for
the contribution.

specs2/specs2.bzl
```diff
@@ -0,0 +1,112 @@
+def specs2():
+
+  # org.specs2:specs2-common_2.11:jar:3.8.4
+  native.maven_jar(
```
johnynek: this is already in the `@scala` target. Can we not use that one?

specs2/specs2.bzl
```diff
@@ -0,0 +1,112 @@
...
+
+  native.maven_jar(
+      name = "org_specs2_specs2_core_2_11",
+      artifact = "org.specs2:specs2-core_2.11:3.8.4",
+  )
+
+  # org.specs2:specs2-common_2.11:jar:3.8.4
+  # org.specs2:specs2-core_2.11:jar:3.8.4
+  native.maven_jar(
+      name = "org_scala_lang_scala_reflect",
```
johnynek: we should use the one in the `@scala` target.

specs2/specs2.bzl
```diff
@@ -0,0 +1,112 @@
...
+  # org.specs2:specs2-matcher_2.11:jar:3.8.4
+  native.maven_jar(
+      name = "org_specs2_specs2_common_2_11",
+      artifact = "org.specs2:specs2-common_2.11:3.8.4",
+      sha1 = "f3b280746b585b6872ef5ada0d78420b54226dfe",
+  )
+
+  # org.specs2:specs2-common_2.11:jar:3.8.4
+  native.maven_jar(
+      name = "org_scala_lang_modules_scala_xml_2_11",
```
johnynek: please use the one in the `@scala` target.

specs2/specs2.bzl
```diff
@@ -0,0 +1,112 @@
+def specs2():
```
johnynek: this should be called `spec2_repositories` or something like that.

johnynek:
can we add tests here?

One thing I am concerned about is that the dependencies of specs2 are a bit heavier.  So, there is a decent chance a user is going to already have some of these dependencies in their tree. How can we deal with that?

I think that we need to move to using `bind` so that people can bind their own versions of things like scalaz. We can punt on that here, but at least we should not duplicate the scala artifacts that are in the scala-sdk tarball we fetch.

scala/scala.bzl
```diff
@@ -548,6 +581,36 @@ scala_test = rule(
...
+     },
+  executable=True,
+  test=True,
+  )
+
+scala_specs2_junit_test = rule(
+  implementation=_scala_specs2_test_impl,
+  attrs={
+     "main_class": attr.string(default="org.junit.runner.JUnitCore"),
+     "_specs2_all": attr.label(default=Label("//specs2:specs2_with_junit"), allow_files=True),
```
johnynek: maybe we can just have a `junit` target, and then in a macro, etc... have junit + specs. Some might just want to run a junit test without any specs.

johnynek: and/or the target can be a bind variable, and the users can set up what to point to.

johnynek:
@ittaiz great to meet you and chat with you today!

ittaiz:
decided to implement it a bit differently in the end. I implemented JUnit support and added specs2 on top of that (requires usage of `SpecWithJUnit`).
Additionally I made JUnit support imitate the maven behavior of discovery of test classes by using regexes on class names. see #163 for more info.

mvaitkus:
Given that @ittaiz found a way to run specs2 on top of junit, I am closing this PR as no longer relevant.

PR #91, opened 2016-09-13
Persistent/worker scala compiler

ianoc:
Adds support for the persistant/worker Scala compiler. All the code for doing the protobufs is just lifted from the bazel example/tests for this. Thanks to @johnynek we've added in all the features via this route to get the build passing.


bazel-io:

ianoc:
This points at the bazel repo itself at a sha to get the protobuf code for the worker. Took this route following the changes over at https://github.com/bazelbuild/dash/pull/13 . Should we be trying to use bazel tools here or have the WORKSPACE's have a bazel checkout?

ittaiz:
Super exciting! @ianoc do you have any rough numbers on how this impacts build times?

ianoc:
Numbers are as you might imagine a bit all over the place depending on the nature/shape of your build targets (how many files and so on). But in a reasonable collection of targets it was 2.5mins -> 1min. On one target thats got some macros in it (its just 2 scala files...) 15sec -> 9sec -> ... -> 2.8sec. [Several timings i did over changing some code to re-run it, so the jit would get fully warmed and it went down over a few builds to the 2.8 sec range]

scala/scala.bzl
```diff
@@ -452,28 +467,36 @@ env JAVACMD=$0.runfiles/{repo}/{java} $0.runfiles/{repo}/{scala} {jvm_flags} -cl
...
+              ]
+    rjars += _collect_jars(ctx.attr.runtime_deps).runtime
+    _write_test_launcher(ctx, rjars)
+    return _scala_binary_common(ctx, cjars, rjars)
 
 _implicit_deps = {
   "_ijar": attr.label(executable=True, default=Label("@bazel_tools//tools/jdk:ijar"), single_file=True, allow_files=True),
   "_scala": attr.label(executable=True, default=Label("@scala//:bin/scala"), single_file=True, allow_files=True),
-  "_scalac": attr.label(executable=True, default=Label("@scala//:bin/scalac"), single_file=True, allow_files=True),
+  "_scalac": attr.label(cfg=HOST_CFG, executable=True, default=Label("//src/java/io/bazel/rulesscala/scalac"), allow_files=True),
```
johnynek: I don't know what HOST_CFG does in this context. I know that it means use the host configuration but what does that mean in this context? Can you add a comment?

src/java/io/bazel/rulesscala/jar/JarCreator.java
```diff
@@ -173,6 +173,36 @@ public void execute() throws IOException {
     }
   }
 
+  public static void buildJar(String[] args) throws IOException {
```
johnynek: Looks like duplication with main. Can we not call this method from main?

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
```diff
@@ -0,0 +1,366 @@
...
+  static int workUnitCounter = 1;
+
+  // If true, returns corrupt responses instead of correct protobufs.
+  static boolean poisoned = false;
+
+  // Keep state across multiple builds.
+  static final LinkedHashMap<String, String> inputs = new LinkedHashMap<>();
+
+  static class WorkerOptions {
+    public int exitAfter = 30;
```
johnynek: Why 30?

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
```diff
@@ -0,0 +1,366 @@
...
+
+  private static List<File> extractJar(String jarPath,
+      String outputFolder) throws IOException, FileNotFoundException {
+
+    List<File> outputPaths = new ArrayList<File>();
+    java.util.jar.JarFile jar = new java.util.jar.JarFile(jarPath);
+    java.util.Enumeration e = jar.entries();
+    while (e.hasMoreElements()) {
+      java.util.jar.JarEntry file = (java.util.jar.JarEntry) e.nextElement();
+      File f = new File(outputFolder + java.io.File.separator + file.getName());
```
johnynek: We already have File imported.

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
```diff
@@ -0,0 +1,366 @@
...
+    }
+    catch (Exception ex) {
+      throw new RuntimeException("nope", ex);
+    }
+  }
+
+  private static void processRequest(List<String> args) throws Exception {
+    Path tmpPath = null;
+    try {
+      if (args.size() == 1 && args.get(0).startsWith("@")) {
```
johnynek: Aren't you passing more than one arg now with the change to pass jvm_opts?

johnynek:
Can we break this into two PRs? One to restyle the python/bzl and then a second to add the worker? It's a bit hard to review with such an important change along with the restyling.

ianoc:
Does using https://github.com/bazelbuild/rules_scala/pull/91/files?w=1 clean it up enough or still rather the 2 prs?

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
ianoc: java doesn't pass args it consumes onto the runtime, so they never make it here if they are jvm_opts

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
ianoc: uhh good question. I lifted most of these things from the example, but not sure if i lifted this value.. I'm just going to kill the poison/exitAfter code

scala/scala.bzl
ianoc: I don't either ¯\_(ツ)_/¯ ! Lifted from example when i was trying to make this work. Happy to drop if it all still runs. I'll check

johnynek: Well, I don't know if you noticed, but we have been compiling targets
twice. Bazel does this for the host and target compilations. I don't see
why we need to do that with JVM code like this. I wonder if this flag (in
the right spot) could change that behavior.

ianoc: Do you mean we should want to add or remove? I've seen us compiling targets on different actions again, but haven't noticed us just doing the same thing twice over. But i'd believe it..

scala/scala.bzl
```diff
@@ -21,360 +21,375 @@ _srcjar_filetype = FileType([".srcjar"])
...
+        ijar_output_path = ctx.outputs.ijar.path
+        ijar_cmd_path = ctx.file._ijar.path
+
+    java_srcs = _java_filetype.filter(ctx.files.srcs)
+    sources = _scala_filetype.filter(ctx.files.srcs) + java_srcs
+    srcjars = _srcjar_filetype.filter(ctx.files.srcs)
+    all_srcjars = set(srcjars + list(dep_srcjars))
+    # look for any plugins:
+    plugins = _collect_plugin_paths(ctx.attr.plugins)
+    plugin_arg = ",".join(list(plugins))
```
johnynek: I'm a little worried about this. Are we sure plugins can't have `,` in them? This whole array encoding with `,` is worrisome. I wish there was a standard method to escape a set of characters and then unescape them in another language.

scala/scala.bzl
```diff
@@ -21,360 +21,375 @@ _srcjar_filetype = FileType([".srcjar"])
...
+Files: {files}
+EnableIjar: {enableijar}
+ijarOutput: {ijar_out}
+ijarCmdPath: {ijar_cmd_path}
+SourceJars: {srcjars}
+JavacPath: {javac_path}
+JavacOpts: {javac_opts}
+JavaFiles: {java_files}
+ResourceSrcs: {resource_src}
+ResourceDests: {resource_dest}
```
johnynek: we still need to pass the `jvm_opts` otherwise we can't use them to pass to javac targets, right?

scala/scala.bzl
```diff
@@ -21,360 +21,375 @@ _srcjar_filetype = FileType([".srcjar"])
...
+""".format(
+        out=ctx.outputs.jar.path,  # 0
```
johnynek: do we need the `#0, #1....` here?

scala/scala.bzl
```diff
@@ -21,360 +21,375 @@ _srcjar_filetype = FileType([".srcjar"])
...
-      return path
-  return None
+    for p in paths:
+        path = p.path
+        if path.endswith("/binary_deploy.jar"):
+            return path
+    return None
+
+
+def _get_scalac_jar_path(paths):
```
johnynek: I don't think this code is used.

scala/scala.bzl
ianoc: Ah, yes your right

scala/scala.bzl
```diff
@@ -21,360 +21,375 @@ _srcjar_filetype = FileType([".srcjar"])
...
-    ijar = None
+    scalac_args = """
+JarOutput: {out}
+Manifest: {manifest}
+ScalacOpts: {scala_opts}
+Plugins: {plugin_arg}
+Classpath: {cp}
+Files: {files}
+EnableIjar: {enableijar}
+ijarOutput: {ijar_out}
```
johnynek: seems like maybe we should make this `IjarOutput` to be consistent, right? Same below.

Also, can we sort this list since order does not matter?

scala/scala.bzl
ianoc: comma's are relatively common, so wouldn't surprise me if its in some one of their args :/. one option would be to json encode the list itself i suppose?

scala/scala.bzl
```diff
@@ -21,360 +21,375 @@ _srcjar_filetype = FileType([".srcjar"])
...
+            ctx.file._java,
+            argfile])
+    ctx.action(
+        inputs=ins,
+        outputs=outs,
+        executable=ctx.executable._scalac,
+        mnemonic="Scalac",
+        progress_message="scala %s" % ctx.label,
+        execution_requirements={"supports-workers": "1"},
+        arguments=list(ctx.attr.jvm_flags) + ["@" + argfile.path],
```
johnynek: can we add a comment here like:
```
when we run with a worker, the `@argfile.path` is removed and passed line by line as arguments in the protobuf. In that case, the rest of the arguments are passed to the process that starts up and stays resident.

In either case (worker or not), they will be jvm flags which will be correctly handled since the executable is a jvm app that will consume the flags on startup.
```
This confused me a bit, and I may forget why it works.

scala/scala.bzl
```diff
@@ -21,360 +21,375 @@ _srcjar_filetype = FileType([".srcjar"])
...
+        ijar = None
+        if buildijar:
+            ijar = ctx.outputs.ijar
+        else:
+            #  macro code needs to be available at compile-time,
+            #  so set ijar == jar
+            ijar = ctx.outputs.jar
+        return struct(ijar=ijar, class_jar=ctx.outputs.jar)
 
 def _build_deployable(ctx, jars):
```
johnynek: tempted to rename this `build_deplorable` to reflect recent news events. ;)

scala/scala.bzl
```diff
@@ -21,360 +21,375 @@ _srcjar_filetype = FileType([".srcjar"])
...
-  srcjars = collect_srcjars(ctx.attr.deps)
-  jars = _collect_jars(ctx.attr.deps)
-  (cjars, rjars) = (jars.compiletime, jars.runtime)
-  write_manifest(ctx)
-  outputs = _compile_or_empty(ctx, cjars, srcjars, non_macro_lib)
 
-  rjars += [ctx.outputs.jar]
-  rjars += _collect_jars(ctx.attr.runtime_deps).runtime
+def _collect_jars(targets):
+    """Compute the runtime and compile-time dependencies from the given targets"""  # noqa
```
johnynek: what's with the `noqa`?

ianoc: Stops linters caring about lines

scala/scala.bzl
ianoc: Haha, might just be hard to re-read the code in 2 years when we've forgotten the signifigance! ;) But it would be fun to name all methods for political stuff of the day.

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
```diff
@@ -0,0 +1,325 @@
...
+import java.util.Map.Entry;
+import java.util.Map;
+import java.util.TreeMap;
+import scala.Console$;
+import scala.tools.nsc.*;
+import scala.tools.nsc.reporters.ConsoleReporter;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+/**
+ * A class for creating Jar files. Allows normalization of Jar entries by setting their timestamp to
```
johnynek: I don't think this comment is up to date.

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
```diff
@@ -0,0 +1,325 @@
...
+
+      if (reporter.hasErrors()) {
+          reporter.printSummary();
+          reporter.flush();
+          throw new RuntimeException("Build failed");
+      } else {
+        /**
+         * See if there are java sources to compile
+         */
+        if (ops.javaFiles.length > 0) {
```
johnynek: should we move this to `compileJavaSources()` since it is kind of a distraction for an unusual case in the code.

johnynek:
okay, looks great. I'm going to run a quick test on a couple of repos and if they are green, I say we ship it. This is exciting.

johnynek:
Only thing missing, in my opinion, is to update the README to explain:

1. you need to add bazel to your WORKSPACE as well
2. how to run with persistent worker (`--strategy=Scalac=worker`), and how to add that to you `.bazelrc` to get it automatically for `build` and `test`.

This is really great.

ianoc:
updated the readme

README.md
```diff
@@ -40,6 +46,12 @@ load("@io_bazel_rules_scala//scala:scala.bzl", "scala_library", "scala_binary",
 ```
 You may wish to have these rules loaded by default using bazel's prelude. You can add the above to the file `tools/build_rules/prelude_bazel` in your repo (don't forget to have a, possibly empty, BUILD file there) and then it will be automatically prepended to every BUILD file in the workspace.
 
+To run with a persistant worker (much faster), you need to add
+```python
+--strategy=Scalac=worker
+```
+to your command line, or to enable by default for building/testing add it to your .bazelrc.
```
johnynek: to do this do you need to make lines like:
```
build --strategy=Scalac=worker
test --strategy=Scalac=worker
```
or can you just add `--strategy=Scalac=worker`?

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
```diff
@@ -0,0 +1,329 @@
...
+      try {
+        WorkRequest request = WorkRequest.parseDelimitedFrom(System.in);
+        if (request == null) {
+          break;
+        }
+
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        int exitCode = 0;
+
+        try (PrintStream ps = new PrintStream(baos)) {
```
johnynek: I got nervous about character sets here and found this API:
https://docs.oracle.com/javase/7/docs/api/java/io/PrintStream.html#PrintStream(java.io.OutputStream,%20boolean,%20java.lang.String)

I think we want `new PrintStream(baos, true, "UTF-8")`.

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
```diff
@@ -0,0 +1,329 @@
...
+            e.printStackTrace();
+            exitCode = 1;
+          }
+        } finally {
+          System.setOut(originalStdOut);
+          System.setErr(originalStdErr);
+        }
+
+        WorkResponse.newBuilder()
+            .setOutput(baos.toString())
```
johnynek: I think we want `new String(baos.toByteArray(), "UTF-8")` here, just to be sure about character sets.

johnynek: then again, I don't know, maybe it is fine to use the default here since we are using it twice (both to go to bytes and then back from bytes?)

ianoc: Yeah seems like it should be fine to me

README.md
ianoc: uhhh, i could be totally wrong, so lets just put that in

johnynek:
@philwo you might want to take a look. I'm going to merge this. It seems 2-6x faster depending on the target size (big wins for really small targets). Thanks for the work on this!

I would say it is a real bummer to have to check out all of bazel to get an easy to access java jar to use here (the generated from protobuf). Would be nice to have a smaller download (maybe a stand-alone protobuf + java library that lives outside of bazel and that bazel can depend on).

johnynek:
@philwo one note: it would be nice if we could consume a json stream, not a protobuf, which could have a lighter weight coupling between workers and bazel. Then we would not need to depend on the protobufs in the bazel repo and have this giant checkout.  I guess we could copy the file, but that also seems hacky.

philwo:
@ianoc @johnynek Thank you so much for this contribution & merging it! This is a really big win for Bazel.

I agree that the protobuf being hidden in the Bazel repository is annoying in a case like this. I'll figure something out.

PR #93, opened 2016-09-13
Fix jvm_flag support

johnynek:
bazel java binary targets have certain options:

https://github.com/bazelbuild/bazel/blob/c1b417549cf4ba719662f7562e99e807c606eae7/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt

use:
```bash
# --jvm_flag=<flag>     Pass <flag> to the "java" command itself.
#                       <flag> may contain spaces. Can be used multiple times.
```
to set the jvm flags.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -174,7 +174,7 @@ SourceJars: {srcjars}
         javac_opts=" ".join(ctx.attr.javacopts),
         javac_path=ctx.file._javac.path,
         java_files=",".join([f.path for f in java_srcs]),
-        jvm_flags=" ".join(["-J" + flag for flag in ctx.attr.jvm_flags]),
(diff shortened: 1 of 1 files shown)


bazel-io:

scala/scala.bzl
```diff
@@ -174,7 +174,7 @@ SourceJars: {srcjars}
         javac_opts=" ".join(ctx.attr.javacopts),
         javac_path=ctx.file._javac.path,
         java_files=",".join([f.path for f in java_srcs]),
-        jvm_flags=" ".join(["-J" + flag for flag in ctx.attr.jvm_flags]),
+        jvm_flags=" ".join(ctx.attr.jvm_flags),
```
ianoc: I think you need the -J still here, these are for java

ianoc:
LGTM

PR #95, opened 2016-09-14
Catch throwables from type errors

ianoc:
I ran into errors that were killing the standalone workers. Seems the Exception catching we have wasn't picking up these types. Not sure how safe it is to pickup all throwables so i'm just going to focus on the one we know should be captured..

### DIFF
--- a/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
+++ b/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
@@ -207,7 +207,17 @@ public class ScalaCInvoker {
 
       MainClass comp = new MainClass();
       long start = System.currentTimeMillis();
-      comp.process(compilerArgs);
+
+      try {
+        comp.process(compilerArgs);
+      } catch (Throwable ex) {
+        if(ex.toString().contains("scala.reflect.internal.Types$TypeError")){
+          throw new RuntimeException("Build failure with type error", ex);
+        } else {
+          throw ex;
+        }
+      }
+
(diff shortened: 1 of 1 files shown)


bazel-io:

johnynek:
So is this a case of a compiler crash? Don't we want to fail in that case?

Also, I don't see how this helps. You are just converting an exception to a RuntimException. Was it not an Exception? Maybe it is an Error (throwable that is not Exception). If so, maybe we can catch Error and send an Error response back? I'm not crazy about the String parsing here.

ianoc:
Catching error didn't catch it ¯\_(ツ)_/¯ ... throwable was the only thing i've seen catch it.

Also i don't see we would want to fail the worker, the JVM process is still fine, its just a code error/typer error?

johnynek:
Can you print the getClass of this thing? Maybe we can find it in the
Scalia code base?

ianoc:
@johnynek Its a subtype or path dependent type, after playing around in a debugger and trying a few variants I haven't found any type spec the java compiler would accept for catching that would catch this. 
Catch's in the scala code base refer to it directly like:
`} catch { case _: Exception | _: TypeError => errorMessage }`

its defined @ 
https://github.com/scala/scala/blob/2.11.x/src/reflect/scala/reflect/internal/Types.scala#L4473

johnynek:
Okay. I guess there is no chance we will get this wrong with that giant string. This is probably some subtype that is a pain to describe in Java?

ianoc:
its a path dependant type from what i could see. I don't how to express that/if you can

PR #97, opened 2016-10-08
Cleanup 0.3.2 warnings

johnynek:
This adds cfg settings to the executable labels (which 0.3.2 now warns about).

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -491,19 +491,19 @@ def _scala_test_impl(ctx):
     return _scala_binary_common(ctx, cjars, rjars)
 
 _implicit_deps = {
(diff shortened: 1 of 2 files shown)


johnynek:
@ianoc @ittaiz any comments?

johnynek:
note that `DATA_CFG` and `HOST_CFG` also warn now, and that `"host"` and `"data"` is the preferred form.

I put `"host"` on anything that is not derived from an external tarball. The docs are not super clear on what contexts are right for executable data, but I assume something cross platform (such as the bash script that is "/bin/scala" qualifies).

ianoc:
LGTM

ittaiz:
Are we specifying execution_requirements "supports-workers"?

johnynek:
Yes. That part is already set correctly. And we've been using at Stripe.
MUCH nicer!

ittaiz:
👍🏽

PR #98, opened 2016-10-20
Correctly set testonly to test targets

damienmg:
testonly was not enforced in previous version of Bazel but this will
be fixed in the next version of Bazel. Correct it for rules_scala.

Tested with `bazel build ...`

Tracking bug: bazelbuild/bazel#1967

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -1,3 +1,5 @@
+package(default_testonly = 1)
+
 load("//scala:scala.bzl",
     "scala_binary",
     "scala_library",


petemounce:
@damienmg I hit this (probably should have written it up here, but instead I did at https://github.com/bazelbuild/bazel/issues/1967#issuecomment-255741396). Is there a workaround I can apply...?

damienmg:
Adding @dinowernli to the party...

johnynek:
@petemounce can you make an issue here? I have not seen that issue before and we run on OS X mostly with 0.3.2. In the issue can you provide any other details (like does it come up in the latest head of this repo?)

dinowernli:
LGTM

PR #100, opened 2016-10-25
I think this should be tag, not commit, since the value is a tag

petemounce:


bazel-io:

petemounce:
`I signed it!`

This time, I've changed my github.com account to use my improbable.io account as my primary - we've signed it as a company. So, it would be great if the bot could be improved to account for when a contributor has more than one email and matching against at least one having signed the CLA.

johnynek:
Thanks!

bazel-io:

Issue #99, opened 2016-10-25
Problem building with bazel 0.3.2


petemounce:
From https://github.com/bazelbuild/bazel/issues/1967#issuecomment-255741396 and continued here:

@damienmg - yesterday, @dinowernli and I tried it on his ubuntu workstation, and got

```
ERROR: /home/dino/Private/code/everything/ArchUtil/src/main/java/improbable/archutil/testing/BUILD:7:1: scala //ArchUtil/src/main/java/improbable/archutil/testing:testing_scala failed: linux-sandbox failed: error executing command /home/dino/.cache/bazel/_bazel_dino/8d44bf8f062af3067a03c8a57a9b5ad0/execroot/everything/_bin/linux-sandbox ... (remaining 4 argument(s) skipped).
Exception in thread "main" java.lang.RuntimeException: nope
    at io.bazel.rulesscala.scalac.ScalaCInvoker.main(ScalaCInvoker.java:338)
Caused by: java.io.IOException: Cannot run program "external/local_jdk/bin/javac ": error=2, No such file or directory
    at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
    at io.bazel.rulesscala.scalac.ScalaCInvoker.compileJavaSources(ScalaCInvoker.java:293)
    at io.bazel.rulesscala.scalac.ScalaCInvoker.processRequest(ScalaCInvoker.java:237)
    at io.bazel.rulesscala.scalac.ScalaCInvoker.main(ScalaCInvoker.java:334)
Caused by: java.io.IOException: error=2, No such file or directory
    at java.lang.UNIXProcess.forkAndExec(Native Method)
    at java.lang.UNIXProcess.<init>(UNIXProcess.java:248)
    at java.lang.ProcessImpl.start(ProcessImpl.java:134)
    at java.lang.ProcessBuilder.start(ProcessBuilder.java:1029)
    ... 3 more
```

The BUILD file is:

```
package(
    default_visibility = ["//visibility:public"],
)

load("@io_bazel_rules_scala//scala:scala.bzl", "scala_export_to_java")

scala_library(
    name = "testing_scala",
    testonly = 1,
    srcs = glob(["*.java"]),
    visibility = ["//visibility:private"],
    deps = [
        "//third_party/scalatest",
    ],
)

scala_export_to_java(
    name = "testing",
    exports = [":testing_scala"],
    runtime_deps = [],
)
```

There is a single java source file in the package:

```
package improbable.archutil.testing;

[...]

PR #101, opened 2016-10-25
Add jvm_flags attribute test cases to expose issue building javac command line

petemounce:
This is mentioned in #99, and this is a reproduction for that (but I don't know where to fix it).

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -97,6 +97,13 @@ scala_library(
     deps = ["ExportOnly"],
 )
 
+scala_library(
+    name = "OtherLibButWithJvmFlags",
+    srcs = ["OtherLib.scala"],
+    deps = ["ExportOnly"],
+    jvm_flags = ["-Xms1G", "-Xmx4G"],
+)
+
 # Test of library without src
 scala_library(
     name = "ExportOnly",
@@ -153,6 +160,13 @@ scala_library(
 #Mix java scala
 scala_library(
     name = "MixJavaScalaLib",
+    srcs = glob(["src/main/scala/scala/test/mix_java_scala/*.scala"]) +
+           glob(["src/main/scala/scala/test/mix_java_scala/*.java"]),
+    jvm_flags = ["-Xms1G", "-Xmx4G"],
+)
+#Mix java scala
+scala_library(
+    name = "MixJavaScalaLibWithJvmFlags",
     srcs = glob(["src/main/scala/scala/test/mix_java_scala/*.scala"]) +
            glob(["src/main/scala/scala/test/mix_java_scala/*.java"]),
     jvm_flags = ["-Xms1G", "-Xmx4G"],
--- a/.bazel-installer-linux-x86_64.sh.sha256
+++ b/.bazel-installer-linux-x86_64.sh.sha256
@@ -1,1 +1,1 @@
-2744447c9999ceea3dc0b90013dcdfbc3683ebb416eb45287c98049bb7a1c6a8  bazel-0.3.2-installer-linux-x86_64.sh
+97dd53414e12da1c9a8a23911ebe732b4b278295ed6b226a5ddee4cd6775a01b  bazel-0.3.1-installer-linux-x86_64.sh
--- a/.bazel-installer-linux-x86_64.sh.sha256
+++ b/.bazel-installer-linux-x86_64.sh.sha256
@@ -1,1 +1,1 @@
-97dd53414e12da1c9a8a23911ebe732b4b278295ed6b226a5ddee4cd6775a01b  bazel-0.3.1-installer-linux-x86_64.sh
+2744447c9999ceea3dc0b90013dcdfbc3683ebb416eb45287c98049bb7a1c6a8  bazel-0.3.2-installer-linux-x86_64.sh


bazel-io:

petemounce:
With this diff, I now get what looks like the correct command line for javac, but the following error:

```diff
diff --git i/scala/scala.bzl w/scala/scala.bzl
index debac4a..aacef5f 100644
--- i/scala/scala.bzl
+++ w/scala/scala.bzl
@@ -202,6 +202,7 @@ SourceJars: {srcjars}
            [ctx.outputs.manifest,
             ctx.file._ijar,
             ctx.file._java,
+            ctx.file._javac,
             argfile])
     ctx.action(
         inputs=ins,
diff --git i/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java w/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
index 628b569..1298fda 100644
--- i/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
+++ w/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
@@ -274,6 +274,7 @@ public class ScalaCInvoker {
   private static void compileJavaSources(CompileOptions ops, Path tmpPath) throws IOException, InterruptedException {
     StringBuilder cmd = new StringBuilder();
     cmd.append(ops.javacPath);
+    cmd.append(" ");
     if (ops.jvmFlags != "") cmd.append(ops.jvmFlags);
     if (ops.javacOpts != "") cmd.append(ops.javacOpts);
```

```
$ bazel test --verbose_failures //...

...

INFO: Found 69 targets and 1 test target...
ERROR: /Users/peter/src/oss/rules_scala/test/BUILD:167:1: scala //test:MixJavaScalaLibWithJvmFlags failed: scalac failed: error executing command
  (cd /private/var/tmp/_bazel_peter/cb0f8a6254f19f3f9fe9f1ab970ecbf6/execroot/rules_scala && \
  exec env - \
  bazel-out/host/bin/src/java/io/bazel/rulesscala/scalac/scalac '--jvm_flag=-Xms1G' '--jvm_flag=-Xmx4G' @bazel-out/local-fastbuild/bin/test/MixJavaScalaLibWithJvmFlags_worker_input): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 1.
Exception in thread "main" java.lang.RuntimeException: nope
	at io.bazel.rulesscala.scalac.ScalaCInvoker.main(ScalaCInvoker.java:339)
Caused by: java.io.IOException: Cannot run program "external/local_jdk/bin/javac -J-Xms1G -J-Xmx4G": error=2, No such file or directory
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
	at io.bazel.rulesscala.scalac.ScalaCInvoker.compileJavaSources(ScalaCInvoker.java:294)
	at io.bazel.rulesscala.scalac.ScalaCInvoker.processRequest(ScalaCInvoker.java:237)
	at io.bazel.rulesscala.scalac.ScalaCInvoker.main(ScalaCInvoker.java:335)
Caused by: java.io.IOException: error=2, No such file or directory
	at java.lang.UNIXProcess.forkAndExec(Native Method)
	at java.lang.UNIXProcess.<init>(UNIXProcess.java:247)
	at java.lang.ProcessImpl.start(ProcessImpl.java:134)
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1029)
	... 3 more
INFO: Elapsed time: 10.626s, Critical Path: 7.67s
//test:HelloLibTest                                                   NO STATUS

Executed 0 out of 1 test: 1 was skipped.
```

johnynek:
I guess the issue is here:

https://github.com/bazelbuild/rules_scala/blob/master/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java#L288

this approach of gluing args onto the command is probably broken.

Probably better to use this method instead:
https://docs.oracle.com/javase/7/docs/api/java/lang/ProcessBuilder.html#command(java.util.List)

then we can add the args we need rather than string manipulation.

petemounce:
I expect #104 to fix this.

johnynek:
@petemounce can you merge master and see if the test goes green? I hope so, and then I'll merge.

petemounce:
@johnynek I rebased just now.

johnynek:
Huh. A test failed:

`Test test_build_is_identical failed`

Building twice does not produce the same output, but only with this change, it seems. :/

petemounce:
Locally (OSX 10.12), all the tests pass.

johnynek:
just to check, you did: https://github.com/bazelbuild/rules_scala/blob/master/test_run.sh `./test_run.sh` to test this?

clearly this is a test-only change, so if it fails, it is not a problem with the PR, but an actual bug.

If the code is non-identical on multiple runs, this usually means the caching is going to be broken and cause unneeded recomputation when there are non-semantic changes to the code (whitespace, for instance).

petemounce:
Yes; here's the output, and a `git status`.

```
peter at petermounce-mbp-imp in ~/src/oss/rules_scala (expose_jvm_flags_issue●)
$ ./test_run.sh
running test bazel build test/...
 Test bazel build test/... successful
running test bazel test test/...
 Test bazel test test/... successful
running test bazel run test/src/main/scala/scala/test/twitter_scrooge:justscrooges
 Test bazel run test/src/main/scala/scala/test/twitter_scrooge:justscrooges successful
running test bazel run test:JavaBinary
 Test bazel run test:JavaBinary successful
running test bazel run test:JavaBinary2
 Test bazel run test:JavaBinary2 successful
running test bazel run test:MixJavaScalaLibBinary
 Test bazel run test:MixJavaScalaLibBinary successful
running test bazel run test:ScalaBinary
 Test bazel run test:ScalaBinary successful
running test bazel run test:ScalaLibBinary
 Test bazel run test:ScalaLibBinary successful
running test test_disappearing_class
 Test test_disappearing_class successful
running test find -L ./bazel-testlogs -iname *.xml
 Test find -L ./bazel-testlogs -iname *.xml successful
running test xmllint_test
 Test xmllint_test successful
running test test_disappearing_class
 Test test_disappearing_class successful
running test test_build_is_identical
 Test test_build_is_identical successful
running test test_transitive_deps
 Test test_transitive_deps successful
running test test_repl
 Test test_repl successful

peter at petermounce-mbp-imp in ~/src/oss/rules_scala (expose_jvm_flags_issue●)
$ gs
On branch expose_jvm_flags_issue
Your branch is up-to-date with 'origin/expose_jvm_flags_issue'.
Untracked files:
	diff.patch
	hash1
	hash2

nothing added to commit but untracked files present
```

test/BUILD
```diff
@@ -157,6 +164,13 @@ scala_library(
            glob(["src/main/scala/scala/test/mix_java_scala/*.java"]),
     jvm_flags = ["-Xms1G", "-Xmx4G"],
 )
+#Mix java scala
+scala_library(
+    name = "MixJavaScalaLibWithJvmFlags",
```
ittaiz: @petemounce why is this target needed?
Asking for two reasons:
1. Nothing depends on it or runs it in the test_run.sh
2. The "existing" `MixJavaScalaLib` target also uses the `jvm_flags` attribute so other than the name I don't see any difference.

Did this push anything or conversely trying to protect from a regression? If so how is it different from the existing target?

ittaiz: oh sorry, I think the "existing" one got it from #104 which you rebased on top off.
Any objection to removing this target?

petemounce: No, no objection, I'll do that now.

petemounce: Ah - this one has a slightly more explicit name than the other scala+java mix target, which would make a test failure more explicit. In #104, the existing scala+java mix target had the jvm_flags attribute added by itself, so now could fail if there's an issue with mixed compilation _or_ jvm flags.

What do you think?

ittaiz: @petemounce sorry for the delay. fair enough. but wdyt about removing the jvm_flags from this target? It's purpose is to check JavaScala mix, the jvm_flags can be tested solely in the new target

petemounce: I think it was the addition of jvm_flags on a mixed scala+java library that caused the error.

ittaiz: ok, why do we need `OtherLibButWithJvmFlags`?
Can we also remove the jvm_flags from `MixJavaScalaLib` and only retain it in the new target?
Last question what is protected by `MixJavaScalaLibWithJvmFlags` (assuming `MixJavaScalaLib` doesn't have jvm_flags)? How is it protected? What test will break?

johnynek:
could you possibly rebase on master and see if the tests are green now? I think they should be.

ittaiz:
I think this might have been handled by #104 
@petemounce WDYT?

johnynek:
I think #234 also should put the nail in this. Reopen if there is still an issue.

PR #102, opened 2016-10-26
Add runtime_jvm_flags option to specify JVM flags for a scala program at runtime.

sixolet:
Prevously, we only had the `jvm_flags` option, which specified JVM flags for
scalac. This diff adds `jvm_runtime_flags`, which specifies JVM flags for the
JVM that will be running the scala program, in the shell-script shim that
invokes it.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -279,12 +279,17 @@ def _write_launcher(ctx, jars):
       ["$0.runfiles/%s/%s" % (ctx.workspace_name, f.short_path) for f in jars]
       )
 
+    runtime_jvm_flags = ""
+    for flag in ctx.attr.runtime_jvm_flags:
+        runtime_jvm_flags += "'{flag}' ".format(flag=flag)
+
     content = """#!/bin/bash
   export CLASSPATH={classpath}
-  $0.runfiles/{repo}/{java} {name} "$@"
+  $0.runfiles/{repo}/{java} {runtime_jvm_flags} {name} "$@"
   """.format(
       repo=ctx.workspace_name,
       java=ctx.file._java.short_path,
+      runtime_jvm_flags=runtime_jvm_flags,
       name=ctx.attr.main_class,
       deploy_jar=ctx.outputs.jar.path,
       classpath=classpath,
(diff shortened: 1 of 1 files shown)


bazel-io:

johnynek:
Thanks for the patch. Generally we try to keep the rules as close to the java rules as possible. I now realize that the java rules:

https://www.bazel.io/versions/master/docs/be/java.html#java_binary.jvm_flags

use `jvm_flags` for the application you have in mind. I'm a bit inclined to remove `jvm_flags` on `scala_library` and add `scalac_jvm_flags` or something, and make `jvm_flags` only available on `scala_binary` and `scala_test`.

What do you think of that?

sixolet:
That seems the appropriate end state, and matches my initial understanding of the `jvm_flags` option in the first place.  Are you not worried about backcompat, though?

ittaiz:
👍🏽

ittaiz:
@johnynek I'm now bitten by this. I can fix this PR or start a new one but first we should agree on the solution. Can we agree on your suggestion from before on making jvm_flags only available on the binary rules (binary, scalatest, junit, repl) and adding scalac_jvm_flags to the common_attr?
If so I think I'll open a new PR due to the conflicts this one has (unless anyone has strong objections)

johnynek:
Sorry we let this slip through.

I still think we need to use the same naming as the java rules so that means adding a new flag for the JVM flags for scalac. So I agree with your take.

If the original author wants to take it, great, otherwise we can start from this diff and add more to acknowledge the work of @sixolet

sixolet:
Go for it; I have left the job this was relevant for, and so don't have that tool chain easily handy anymore.

ittaiz:
Given the lack of CLA, file conflict and change on implementation direction I'll open a new PR

ittaiz:
so apparently as part of #171 (and without a test) this was added.
Unfortunately this was added by overloading the term of jvm_flags to be:
1. Scalac jvm flags
2. Javac jvm flags
3. Executable jvm flags
I'll open a separate ticket to consider splitting this

PR #104, opened 2016-10-30
Fix the behavior of the jvm_flags param to support multiple params 

dinowernli:
... by making the param a list rather than a string.

Drive-by:
* Add a few bits and pieces to .gitignore.
* Add a repro-case to test/BUILD which validates the fix.
* Use a list of arguments rather than a string to invoke javac.

Addresses: https://github.com/bazelbuild/rules_scala/issues/99

@petemounce @johnynek @damienmg

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 bazel-*
+*.swp
+*.idea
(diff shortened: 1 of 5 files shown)


bazel-io:

johnynek:
@dinowernli this is great, thanks.

Note the discussion in #102. I think we should probably rename this parameter, which we are using for compilation, not runtime. Something like `compile_jvm_flags` or something, since bazel's java rules use `jvm_flags` to signify flags the runtime system should get in the run script.

I think we should just merge this as is now, and as we address #102 make the change here.

PR #106, opened 2016-11-04
Java only sources

ittaiz:
Added an already passing test-target which shows support of targets with java only sources.
I think it has value even if it's already passing since at least currently this is a feature we want to support (until #970 in bazel itself is resolved) and I'd be calmer if there was any regression check for it.
Do you think an actual test is needed?

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -187,3 +187,9 @@ scala_binary(
     main_class = "scala.test.MixJavaScalaLibBinary",
     deps = ["MixJavaScalaLib"],
 )
+
+scala_library(
+    name = "SupportJavaOnlySourcesToEaseJavaScalaInterop",
+    srcs = glob(["src/main/scala/scala/test/only_java/*.java"]),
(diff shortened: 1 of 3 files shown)


bazel-io:

ittaiz:
I actually added the test since I thought about it and saw that I did it manually (added a compilation error in the class) and saw the build break so I decided to automate it.

johnynek:
For some reason, this test fails the identical output test. We need to figure out why. Somehow, building the second time is not producing exactly the same output on linux.

sentinelleader:
This looks good against the rebased master. Tests are passing on my local mac and on TravisCI too https://travis-ci.org/sentinelleader/rules_scala/builds/178413682, 😲

johnynek:
nice @ittaiz want to rebase and see if we can merge?

ittaiz:
@johnynek done. Let me know if anything else is needed.

johnynek:
Thanks!

PR #110, opened 2016-11-08
Removing redundant second call of disappearing class test

### LINKED ISSUE #107: Why does test_disappearing_class appear twice in test_run.sh
Hi,
`run_test test_disappearing_class` appears twice in the `test_run.sh` script, is this on purpose? If so why? Additionally do you think a comment might be helpful because for me it was confusing since I thought this might be a bug.

Additionally,
Is there an open issue to track converting test_run.sh to a "proper" bazel test run? I know it's not high on your priority but maybe if it's open someone will grab it.

ittaiz:
Fixed #107


bazel-io:

johnynek:
:+1:

PR #111, opened 2016-11-08
md5 bash script handling

ittaiz:
running md5 if osx, md5sum otherwise, also ignoring hash1/hash2 in git.
decided not to actually wait for #76  and just do the quick win.
WDYT?


bazel-io:

Issue #112, opened 2016-11-17
Issue with compiling multiple Java Source files


sentinelleader:
Hey Guys,

I've been trying to use `scala_library` to compile a mixture of both scala and java files. It seems the java file compilation fails when there are multiple java files. Up on looking at the error, it seems that the `javac` command that we invoke is taking the list of files as a single string block instead of taking it as a separate files. Below is the error that im getting while invoking javac

```
javac: file not found: infra/BetterOutput.java infra/Changeset.java infra/CompilationException.java
Usage: javac <options> <source files>
use -help for a list of possible options
Exception in thread "main" java.lang.RuntimeException: nope
	at io.bazel.rulesscala.scalac.ScalaCInvoker.main(ScalaCInvoker.java:345)
Caused by: java.lang.RuntimeException: javac process failed!
	at io.bazel.rulesscala.scalac.ScalaCInvoker.compileJavaSources(ScalaCInvoker.java:303)
	at io.bazel.rulesscala.scalac.ScalaCInvoker.processRequest(ScalaCInvoker.java:238)
	at io.bazel.rulesscala.scalac.ScalaCInvoker.main(ScalaCInvoker.java:341)
Target //infrastructure:infra-scala failed to build
```

Simplest fix was to append the file names directly to the command and that seems to be working fine

```
diff --git a/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java b/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
index 4c64c2c..4208123 100644
--- a/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
+++ b/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
@@ -281,19 +281,13 @@ public class ScalaCInvoker {
       commandParts.add(ops.javacOpts);
     }

-    StringBuilder files = new StringBuilder();
-    int cnt = 0;
-    for(String javaFile : ops.javaFiles) {
-      if (cnt > 0) files.append(" ");
-      files.append(javaFile);
-      cnt += 1;
-    }
-
     commandParts.add("-classpath");
     commandParts.add(ops.classpath + ":" + tmpPath.toString());
     commandParts.add("-d");
     commandParts.add(tmpPath.toString());
-    commandParts.add(files.toString());
+    for(String javaFile : ops.javaFiles) {
+      commandParts.add(javaFile.toString());
+    }

     Process iostat = new ProcessBuilder(commandParts)
[...]

PR #113, opened 2016-11-19
Fix Multiple Java source file compilation

sentinelleader:
This will fix javac compilation issue #112 when multiple java files are present in the src


sentinelleader:
I signed it!

PR #114, opened 2016-11-21
Adding a simple resource_strip_prefix option

sentinelleader:
This will add a `resource_strip_prefix` option like the one in Java. If the `resource_strip_prefix` is defined, we will strip out the prefix from the resource file path and forms the new destination path for the resource files. We will also throw a runtime exception if any of the resource file mentioned in the rule doesn't belongs to the specified path (which is the default behavior in Java rule too)

### DIFF
--- a/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
+++ b/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
@@ -350,7 +350,10 @@ public class ScalaCInvoker {
     // check if the Resource file is under the specified prefix to strip
     if (!sourcePath.startsWith(resourceStripPrefix)) {
       // Resource File is not under the specified prefix to strip
-      throw new RuntimeException("Resource File is not under the specified prefix to strip");
+      throw new RuntimeException("Resource File "
+        + sourcePath
+        + " is not under the specified strip prefix "
+        + resourceStripPrefix);
     }
     String newResPath = sourcePath.substring(resourceStripPrefix.length());
     return newResPath;
--- a/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
+++ b/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
@@ -320,12 +320,23 @@ public class ScalaCInvoker {
       Path dest) throws IOException {
     for(Entry<String, String> e : resources.entrySet()) {
       Path source = Paths.get(e.getKey());
-      String dstr = e.getValue();
-      if (dstr.charAt(0) == '/') dstr = dstr.substring(1);
+      String dstr;
       // Check if we need to modify resource destination path
       if (!"".equals(resourceStripPrefix)) {
+	/**
+	 * NOTE: We are not using the Resource Hash Value as the destination path
+	 * when `resource_strip_prefix` present. The path in the hash value is computed
+	 * by the `_adjust_resources_path` in `scala.bzl`. These are the default paths,
+	 * ie, path that are automatically computed when there is no `resource_strip_prefix`
+	 * present. But when `resource_strip_prefix` is present, we need to strip the prefix
+	 * from the Source Path and use that as the new destination path
+	 * Refer Bazel -> BazelJavaRuleClasses.java#L227 for details
+	 */
         dstr = getResourcePath(source, resourceStripPrefix);
+      } else {
+        dstr = e.getValue();
       }
+      if (dstr.charAt(0) == '/') dstr = dstr.substring(1);
       Path target = dest.resolve(dstr);
       File tfile = target.getParent().toFile();
       tfile.mkdirs();
@@ -341,13 +352,7 @@ public class ScalaCInvoker {
       // Resource File is not under the specified prefix to strip
       throw new RuntimeException("Resource File is not under the specified prefix to strip");
     }
-    /**
-     * Using a simple String replaceFirst for now,
-     * Once we get merge into Bazel, we can use the PathFragments from
-     * google's devtools vfs library (which the java rules uses)
-     */
-    String newResPath = sourcePath.replaceFirst(resourceStripPrefix, "");
-    if (newResPath.charAt(0) == '/') newResPath = newResPath.substring(1);
+    String newResPath = sourcePath.substring(resourceStripPrefix.length());
     return newResPath;
   }
 
--- a/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
+++ b/src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
@@ -240,7 +240,7 @@ public class ScalaCInvoker {
         /**
          * Copy the resources
          */
-        copyResources(ops.resourceFiles, tmpPath);
+        copyResources(ops.resourceFiles, ops.resourceStripPrefix, tmpPath);
         /**
          * Now build the output jar
          */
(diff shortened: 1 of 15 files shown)


sentinelleader:
Strange that the tests are passing locally. Let me check travis logs

bazel-io:

test/src/main/scala/scala/test/MoreScalaLibResources.scala
```diff
@@ -0,0 +1,29 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package scala.test
+
+object MoreScalaLibResources {
+  def getGreetings() = get("more hellos")
```
johnynek: how does this actually run? I thought you use `more-hellos` above?

Also, can we add a test that actually runs the code, and not just tests compilation?

test/src/main/resources/scala/test/more-hellos
```diff
@@ -0,0 +1,2 @@
+More Hello
```
johnynek: this file is not in the `exports_files` is that intentional? Can you add a comment in the `BUILD` as to why?

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
```diff
@@ -314,17 +314,42 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOEx
...
     }
   }
+  private static String getResourcePath(
+      Path source,
+      String resourceStripPrefix) throws RuntimeException {
+    String sourcePath = source.toString();
+    // check if the Resource file is under the specified prefix to strip
+    if (!sourcePath.startsWith(resourceStripPrefix)) {
+      // Resource File is not under the specified prefix to strip
+      throw new RuntimeException("Resource File is not under the specified prefix to strip");
```
johnynek: can you include `source` and `resourceStripPrefix` in the message to aid debugging?

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
```diff
@@ -314,17 +314,42 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOEx
...
+    }
+    /**
+     * Using a simple String replaceFirst for now,
+     * Once we get merge into Bazel, we can use the PathFragments from
```
johnynek: what is this comment about? There is no plan, that I know of, to put this back in bazel. This rule (as many rules) are intentionally developed outside so that it can evolve more quickly. As a note, it used to be in bazel mainline source.

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
```diff
@@ -314,17 +314,42 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOEx
...
+      String resourceStripPrefix,
+      Path dest) throws IOException {
     for(Entry<String, String> e : resources.entrySet()) {
       Path source = Paths.get(e.getKey());
       String dstr = e.getValue();
       if (dstr.charAt(0) == '/') dstr = dstr.substring(1);
+      // Check if we need to modify resource destination path
+      if (!"".equals(resourceStripPrefix)) {
+        dstr = getResourcePath(source, resourceStripPrefix);
+      }
```
johnynek: in the else branch is the only case where `e.getValue()` is used. Can we move it there to be clear:
```java
String dstr = null;
if (!"".equals(resourceStripPrefix)) {
  dstr = getResourcePath(source, resourceStripPrefix);
}
else {
  dstr = e.getValue();
}
// we can remove this line from getResourcePath now, right?
if (dstr.charAt(0) == '/') dstr = dstr.substring(1);
```

johnynek: I'm concerned that the value of the hashmap is not used at all if you strip a prefix. Can you add a comment as to why that is correct?

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
```diff
@@ -314,17 +314,42 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOEx
...
+    if (!sourcePath.startsWith(resourceStripPrefix)) {
+      // Resource File is not under the specified prefix to strip
+      throw new RuntimeException("Resource File is not under the specified prefix to strip");
+    }
+    /**
+     * Using a simple String replaceFirst for now,
+     * Once we get merge into Bazel, we can use the PathFragments from
+     * google's devtools vfs library (which the java rules uses)
+     */
+    String newResPath = sourcePath.replaceFirst(resourceStripPrefix, "");
```
johnynek: if it startsWith, which it does since you throw otherwise, why not just: `sourcePath.substring(resourceStripPrefix.length())`

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
```diff
@@ -314,17 +314,42 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOEx
...
+    if (newResPath.charAt(0) == '/') newResPath = newResPath.substring(1);
```
johnynek: can't we share this logic in the caller rather than duplicating?

test/src/main/scala/scala/test/MoreScalaLibResources.scala
sentinelleader: ah, i actually missed that. Will add that too :)

test/src/main/resources/scala/test/more-hellos
sentinelleader: It seems none of the files in `test/src/main/resources/scala/test/` are exported via `exports_files`, they were referenced directly (https://github.com/bazelbuild/rules_scala/blob/master/test/BUILD#L132-L133). Ill add a new BUILD file to export them too

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
sentinelleader: cool, ill tweak this and will also add the comment on the reason for ignoring the hash values

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
sentinelleader: oh ok, the devtools lib files are not visible to outside, so we cant import them directly from it. So i thought when it get merged into the core rules, we can actually use that as it's visible to the sub packages

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
sentinelleader: this makes sense :)

src/java/io/bazel/rulesscala/scalac/ScalaCInvoker.java
sentinelleader: 👍

sentinelleader:
Found out the reason why test's are exiting with 137 on travis. Just need to control the local_resources on the .bazelrc, so that we wont go OOM. Tested with a separate branch and seems to be working (https://travis-ci.org/sentinelleader/rules_scala/builds/177990436)

johnynek:
👍 

Nice work tracking this all down!

PR #120, opened 2017-01-09
Add in some meta targets

ianoc-stripe:
Both test and library meta targets

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -657,16 +657,25 @@ def scala_export_to_java(name, exports, runtime_deps):
     runtime_deps = ["@scala//:lib/scala-library.jar"] + runtime_deps
   )
 
+def _sanitize_string_for_usage(s):
+    res_array = []
+    for c in s:
+        if c.isalnum() or c == ".":
+            res_array.append(c)
+        else:
+            res_array.append("_")
+    return "".join(res_array)
+
 # This auto-generates a test suite based on the passed set of targets
 # we will add a root test_suite with the name of the passed name
 def scala_test_suite(name, srcs = [], deps = [], runtime_deps = [], data = [], resources = [],
                      scalacopts = [], jvm_flags = [], visibility = None, size = None):
     ts = []
     for test_file in srcs:
-        n = "suite_test_%s_%s_test" % (name, test_file)
-        scala_test(n, srcs = [test_file], deps = deps, runtime_deps = runtime_deps, resources=resources, scalacopts=scalacopts, jvm_flags=jvm_flags, visibility=visibility, size=size)
+        n = "%s_test_suite_%s" % (name, _sanitize_string_for_usage(test_file))
+        scala_test(name = n, srcs = [test_file], deps = deps, runtime_deps = runtime_deps, resources=resources, scalacopts=scalacopts, jvm_flags=jvm_flags, visibility=visibility, size=size)
         ts.append(n)
-    native.test_suite(name = name, tests = ts)
+    native.test_suite(name = name, tests = ts, visibility = visibility)
 
 # Scala library suite generates a series of scala libraries
 # then it depends on them with a meta one which exports all the sub targets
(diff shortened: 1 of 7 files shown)


bazel-io:

scala/scala.bzl
```diff
@@ -656,3 +656,52 @@ def scala_export_to_java(name, exports, runtime_deps):
...
     runtime_deps = ["@scala//:lib/scala-library.jar"] + runtime_deps
   )
+
+# This auto-generates a test suite based on the passed set of targets
+# we will add a root test_suite with the name of the passed name
+def scala_test_suite(name, srcs = [], deps = [], runtime_deps = [], data = [], resources = [],
+                     scalacopts = [], jvm_flags = [], visibility = None, size = None):
+    ts = []
+    for test_file in srcs:
+        n = "suite_test_%s_%s_test" % (name, test_file)
```
johnynek: do we need to normalize the test_file names at all? Remove spaces, lower case?

scala/scala.bzl
```diff
@@ -656,3 +656,52 @@ def scala_export_to_java(name, exports, runtime_deps):
...
+                        resource_strip_prefix = "",
+                        scalacopts = [],
+                        javacopts = [],
+                        jvm_flags = [],
+                        print_compile_time = False,
+                        visibility = None
+                        ):
+    ts = []
+    for src_file in srcs:
+        n = "%s_lib_%s" % (name, src_file)
```
johnynek: same comment/question about `src_file`.

scala/scala.bzl
```diff
@@ -656,3 +656,52 @@ def scala_export_to_java(name, exports, runtime_deps):
...
+                      plugins = plugins,
+                      runtime_deps = runtime_deps,
+                      data = data,
+                      resources=resources,
+                      resource_strip_prefix = resource_strip_prefix,
+                      scalacopts = scalacopts,
+                      javacopts = javacopts,
+                      jvm_flags = jvm_flags,
+                      print_compile_time = print_compile_time,
+                      visibility=visibility,
```
johnynek: should we expose the inner targets, or should we make them private, and only use the visibility on the item below?

scala/scala.bzl
```diff
@@ -656,3 +656,52 @@ def scala_export_to_java(name, exports, runtime_deps):
...
+                      exports=exports
+                      )
+        ts.append(n)
+    scala_library(name = name, exports = exports + ts)
```
johnynek: don't we need to set visibility here?

test/BUILD
```diff
@@ -6,7 +6,8 @@ load("//scala:scala.bzl",
     "scala_test",
     "scala_macro_library",
     "scala_export_to_java",
-    "scala_repl")
+    "scala_repl",
+    "scala_library_suite")
```
johnynek: we don't yet have a test for `scala_test_suite` do we?

scala/scala.bzl
ianoc-stripe: I'll add a sanitize, but cases it doesn't care about upper/lower case. And spaces aren't valid seemingly in input file names to scala library at all.

scala/scala.bzl
ianoc-stripe: Thats a good question, i'd have to go with i don't know. You could depend on the inner targets this way, and i know Avi expressed some interest in being able to do that. But then the generated target names need to be stable for referring in build files.

test/BUILD
ianoc-stripe: I had added some of the code below, forgot to wire it in the final bit.

ianoc-stripe:
Updated the readme, only outstanding thing is the visibility. I'm pretty good with either answer there.

scala/scala.bzl
johnynek: 👍 I guess we can just stay as is now and see later. Too public is probably better than too private for now.

johnynek:
@ittaiz it will not work if there are *ANY* dependencies between the files. They must be *independent* (which it does mention in the readme).

ittaiz:
@johnynek indeed it states that it will split the independent files. It wasn't clear to me if this can be of value if it can handle dependencies. I now understand this is more of a naive parallelization.
Is this really valuable in the library level? I can understand that tests are likely to be decoupled by nature but 10 files in the same source files usually have some relationship between them, don't they?

johnynek:
I think seldom will it help for libraries but frequently help with tests.

ittaiz:
Got you, that's what I thought. Thanks!

PR #123, opened 2017-01-12
Use a local copy of the generated java from the protobuf worker inter…

ianoc-stripe:
…face


bazel-io:

scala/scala.bzl
```diff
@@ -645,6 +645,25 @@ def scala_repositories():
     sha256 = "f198967436a5e7a69cfd182902adcfbcb9f2e41b349e1a5c8881a2407f615962",
   )
 
+  native.maven_server(
+    name = "scalac_deps_maven_server",
+    url = "http://bazel-mirror.storage.googleapis.com/repo1.maven.org/maven2/",
+  )
+
+  native.maven_jar(
+    name = "guava",
```
johnynek: have anxiety about this since I'm worried about the name collision. I guess if you collide, it is almost certainly the right jar? But still maybe something like: `rules_scala_guava` might be safer?

scala/scala.bzl
```diff
@@ -645,6 +645,25 @@ def scala_repositories():
...
+    artifact = "com.google.guava:guava:20.0",
+    sha1 = "89507701249388e1ed5ddcf8c41f4ce1be7831ef",
+    server = "scalac_deps_maven_server",
+  )
+
+  native.maven_jar(
+    name = "protobuf_java",
```
johnynek: same fear.

johnynek:
I'm +1, but maybe we can mangle the names a bit to avoid a chance at collision.

ittaiz:
+1 for name mangling

ianoc-stripe:
@johnynek updated

johnynek:
👍

PR #126, opened 2017-01-23
Fix edge case, you can export some targets but not depend on those. M…

ianoc-stripe:
…ostly not an issue, but building the library target should also build the children

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -732,6 +732,6 @@ def scala_library_suite(name,
                       exports=exports
                       )
         ts.append(n)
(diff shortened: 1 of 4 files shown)


bazel-io:

johnynek:
Can we add a test that exposes the issue? But if you're swamped this fix seems like the right thing.

ianoc-stripe:
I'll try come up with a test, i think it needs to be a failing build one. (If something else depends on this target i think it winds up getting the exports built).

ianoc-stripe:
Added a first commit, to get us a failing test, then will re-apply the fix

johnynek:
Thanks a ton for adding the test.

PR #129, opened 2017-01-25
Add ci.bazel.io build status to the README

damienmg:
This will currently show as broken because rules scala is failing due to sandboxing since https://github.com/bazelbuild/rules_scala/commit/55156d5fe501c112983deff6e8065eec2b2b366e. Travis CI does not catch those errors.


johnynek:
This is great! Will try to get it green soon.

johnynek:
@damienmg I think the issue is your CI is trying to build everything and we have a directory called `test_expect_failure` that we expect to fail.

Honestly, I'm not that pumped about having to build a giant pile of shell scripts to run expect failure tests with bazel. I see how bazel does it (it looks like you generate new projects as a `sh_test` and then call bazel recursively), but this seems pretty complex and heavy weight vs what we have now.

Is there any way you can just change the targets being built from `//...` to something like `//src/...` and `//test/...`? or even just `//test/...`?

johnynek:
nevermind. I worked around by changing our script with #135

damienmg:
@johnynek nice that you found a workaround but if that ever needs to be changed, changing the list of targets tested on ci.bazel.io is pretty easy so do not hesitate to ask if you need us to update it.

PR #135, opened 2017-02-07
Remove some unused deps from the rules

johnynek:
I can't see how we are actually using scalasdk, so I removed the target. If somehow this is being used we should try to add tests for that and add it back.


johnynek:
@ianoc @ittaiz take a look?

johnynek:
@ianoc what do you think of this hack to get the bazel ci green? I renamed the BUILD files in the expected fail to BUILD.fail, then our test runner moves them to BUILD before running the tests.

Seems to work.

ianoc:
Ah the bazel CI requires it to be able to build //... does it? if thats the case then yeah i don't think you've much of a choice other than rename stuffs.

ittaiz:
Don't you need to move them back to BUILD.fail afterwards to clean up after yourself?

damienmg:
For reference: https://bazel-review.googlesource.com/8690 is the change that replace the build pattern on the ci to only //test/..., I hold it until you are online (so until tomorrow probably, time zone are hard).

johnynek:
@damienmg the change to only build and test `//test/...` looks good to me. I'll merge this in the mean time and when the CI is updated I may revert the change. That said, it is kind of nice to be able to build `//...` so maybe not having broken targets sitting around is a good idea.

damienmg:
Don't merge it then, I'll just push the change to CI now

johnynek:
@damienmg okay. Let me know when the CI is ready with the new settings.

damienmg:
It is done :)

johnynek:
Thanks @damienmg for the super fast turn-around!

as a note, I also tested this PR on a linux VMWare instance to run all the tests in the `test_run.sh` script and also manually tried the REPL to make super sure I didn't break anything. I *think* this should be safe.

comment: sandboxing does not seem that valuable to me to run every time, but it does seem really valuable that the sandboxed mode builds. If we could run with sandboxing in CI (which we can't because we run in docker), it might be fine (since I understand sandboxing is slower, in particularly on macos).

damienmg:
ci.bazel.io runs with snadbox and as a bazelbuild member you should be able to run it (by saying "test this please")

johnynek:
@damienmg I meant at Stripe. Our CI is based on docker containers run by jenkins. We could imagine one that is just bazel on bare jenkins, which would be nice, but so far we just piggy-backed on the existing system.

Issue #41, opened 2016-04-27
Remove the need for remote_jars in twitter_scrooge


jcoveney:
https://github.com/bazelbuild/rules_scala/blob/master/twitter_scrooge/twitter_scrooge.bzl#L192

This feels unnecessary, we just need a POC/documentation on using new_git_repository or something like that to get rid of the need for it

PR #137, opened 2017-02-08
Avoid decompressing jars/zips in scrooge support

johnynek:
Use scrooge's native support for zip/jar to avoid all the decompression we were doing.

@ianoc please take a look.

We don't actually have tests in the repo that exercise "absolute_path" or the remote_jar which are a features we are using internally. I will try to add tests for that (I'm actually a bit confused about what absolute_path does, but I think it basically chops off a prefix of the path).

Before I merge I want to try to build an internal repo with this change and see if it improves performance. It should, since it avoids a lot of file system operations, and we possibly only decompress files that are actually needed, since it seems scrooge is reasonable smart about that.

### DIFF
--- a/src/scala/scripts/TwitterScroogeGenerator.scala
+++ b/src/scala/scripts/TwitterScroogeGenerator.scala
@@ -13,14 +13,14 @@ import java.util.jar.{ JarEntry, JarFile, JarOutputStream }
 object FinalJarCreator {
   val gm = """(\S+) -> (\S+)""".r
 
-  def apply(dest: Path, owned: Set[Path], genFileMap: Path, scroogeDir: Path) {
+  def apply(dest: Path, owned: Set[String], genFileMap: Path, scroogeDir: Path) {
     val genmap = Source.fromFile(genFileMap.toString)
       .getLines
       .foldLeft(Map.empty[String, Set[String]]) { case (m, gm(thrift, gen)) =>
         m.+((thrift, m.getOrElse(thrift, Set.empty[String]) + gen))
       }
     val shouldMove =
-      owned.map(_.toString).foldLeft(Set.empty[String]) { (s, n) =>
+      owned.foldLeft(Set.empty[String]) { (s, n) =>
         genmap.get(n).fold(s) { s ++ _ }
       }.map { Paths.get(_).normalize }
     val jar = new JarOutputStream(new FileOutputStream(dest.toFile))
@@ -79,24 +79,17 @@ object ScroogeGenerator {
     }
   }
 
-  def extractJarTo(_jar: Path, _dest: Path): List[Path] = {
-    val files = Buffer[Path]()
+  def listJar(_jar: Path): List[String] = {
+    val files = List.newBuilder[String]
     val jar = new JarFile(_jar.toFile)
     val enumEntries = jar.entries()
     while (enumEntries.hasMoreElements) {
       val file = enumEntries.nextElement().asInstanceOf[JarEntry]
-      val path = _dest.resolve(file.getName)
-      if (file.isDirectory) Files.createDirectories(path)
-      else {
-        val is = jar.getInputStream(file)
-
-        try Files.copy(is, path) // Will error out if path already exists
-        finally is.close()
-
-        files += path
+      if (!file.isDirectory) {
+        files += file.getName
       }
     }
-    files.toList
+    files.result()
   }
 
   def readLinesAsPaths(path: Path): Set[Path] =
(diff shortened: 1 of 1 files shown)


johnynek:
so, it looks like this change uncovers a hygiene issue. It passes the test locally, and on travis, but not on bazel CI, where there is sandboxing.

src/scala/scripts/TwitterScroogeGenerator.scala
```diff
@@ -126,37 +119,27 @@ object ScroogeGenerator {
...
-
-    def extract(jars: Set[Path]): Set[Path] =
-      jars.flatMap { jar =>
-        val files = extractJarTo(jar, _tmp)
-        files.foreach { scrooge.includePaths += _.toString }
-        files
-      }
-
-    val immediateThriftSrcs = extract(immediateThriftSrcJars)
+    // these are the ones we are actually generating
```
ianoc: this seems like a generic utility, probably doesn't need this comment?

src/scala/scripts/TwitterScroogeGenerator.scala
```diff
@@ -126,37 +119,27 @@ object ScroogeGenerator {
...
+    def allFilesInZips(i: Iterable[Path]): Set[String] =
+      i.flatMap(listJar(_)).toSet
 
-    immediateThriftSrcs.foreach { scrooge.thriftFiles += _.toString }
-
-    val onlyTransitiveThriftSrcs = extract(onlyTransitiveThriftSrcJars)
-
-    val intersect = onlyTransitiveThriftSrcs.intersect(immediateThriftSrcs)
+    val immediateThriftSrcs = allFilesInZips(immediateThriftSrcJars)
```
ianoc: comment to go here maybe?

src/scala/scripts/TwitterScroogeGenerator.scala
```diff
@@ -126,37 +119,27 @@ object ScroogeGenerator {
...
+    immediateThriftSrcs.foreach { scrooge.thriftFiles += _ }
+    immediateThriftSrcJars.foreach { path => scrooge.includePaths += path.toString }
```
ianoc: you can have an immediate thrift src jar, and an immediate thrift, TIL (do we need both?)

ianoc:
Seems fine to me, obviously need to fix the bug triggering the build failure.

Even if its the ~same speed, it seems much cleaner to me so worth merging

src/scala/scripts/TwitterScroogeGenerator.scala
johnynek: well, the srcs are just the files in the zip that we need to build. I'll comment.

ianoc: is that presuming we only need to build some subset thats in the zip ? (or is it a case of you have to put files always on the include path regardless if they are on the thriftFiles or not?

ianoc:
Could you add some flavor on what the issue was with the build hygiene and what the gist of all the new code is? (its a big diff again)

johnynek:
Yeah, let me know how much we should add to the comments:

1. The build hygiene was that the import happened to be the same relative path to the file on the disk after bazel does it's stuff in bazel-bin. Since scrooge always adds a `Importer(".")` it would find it from that path by happenstance of how bazel works, but not if there is sandboxing preventing it from seeing it.

2. The solution is make a new Compiler class, don't reuse thrifts, and have better control of how importing works. Doing this I learned that scrooge's zip/jar support is pretty broken and really only works if all imports are considered relative to some fictional root directory (a style done at Twitter, but not really standard for thrift). With my new Compiler, I could only add the importers I want. I made a new importer which is a zip importer that also knows where it is in the zip file. In this way, it could use a thrift jar that itself has internal relative imports (such as our monster thrift jar at Stripe).

I still have not tested it at Stripe, I am going to try that next.

oscar-stripe:
On an internal aggregate target with a lot of thrifts, this now takes:
(build 1 time, did a clean, timed the second build)
```
INFO: Elapsed time: 86.657s, Critical Path: 83.56s

real    1m26.787s
user    0m0.073s
sys     0m0.054s
```

with the old code which does decompression we have:
```
INFO: Elapsed time: 132.874s, Critical Path: 130.12s

real    2m12.979s
user    0m0.066s
sys     0m0.047s
```

which is a 65% reduction in elapsed time, so it looks like a fairly good win.

johnynek:
reusing the ZipFile didn't seem to speed anything up, but I guess is a bit friendlier on the GC, so we can keep it.

My guess is that the slow scala parser combinators are the next thing to look at, but at that point, you are really forking or doing a lot of work to upstream code to Twitter.

src/scala/io/bazel/rules_scala/scrooge_support/FocusedZipImporter.scala
```diff
@@ -18,9 +18,18 @@ case class FocusedZipImporter(focus: Option[File], zips: List[File]) extends Imp
   private def toZipEntryPath(n: String): String = focus match {
     case None => n
     case Some(f) =>
-      val str = (new File(f, n)).toString
-      if (str(0) == File.pathSeparatorChar) str.substring(1)
-      else str
+      def loop(leftPart: File, p: List[String]): File = p match {
```
ianoc: `@annotation.tailrec` ?

ianoc:
LGTM

At the top of the scrooge compiler copy pasta could you just outline the weaknesses from your comment ^^, and generally what we do different/why (so allowing paths used in the src jars)

johnynek:
The bazel ci seems to be stuck, but is reporting green if you click on it.

I'm going to merge based on @ianoc 's support.

johnynek:
closes #41

PR #141, opened 2017-02-10
Fix uses of bind for thrift rules

johnynek:
We use some binds, but not all. This uses only bind so users can fix the versions they need.


johnynek:
@ianoc how does this look (seems to work on zoolander)

ianoc:
LGTM

PR #142, opened 2017-02-14
Refactor worker support

johnynek:
If we want to make it easier to add more worker implementations, I wanted to share our existing worker code (particularly for thrift support for workers).

This just does the refactoring. It also removes the guava dependency which was really minor (but Worker still uses it, I think, so pretty weak win).

@ianoc ?


src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
```diff
@@ -0,0 +1,268 @@
...
+   * This is the reporter field for scalac, which we want to access
+   */
+  private static Field reporterField;
+  static {
+    try {
+      reporterField = Driver.class.getDeclaredField("reporter"); //NoSuchFieldException
+      reporterField.setAccessible(true);
+    }
+    catch (Exception ex) {
+      throw new RuntimeException("nope", ex);
```
softprops: I know this didn't change much before but I wonder if there isn't a better error message that can be reported

ianoc:
LGTM

johnynek:
@softprops what do you think of the latest?

softprops:
Much better :) lgtm

PR #146, opened 2017-02-21
added instructions on building and testing from source

ittaiz:
Since test_run.sh fails on OS X I tried running `bazel test //...` and it took me some time until I remembered the problem with the failing folder. Added a small section in the readme about it.


bazel-io:

PR #152, opened 2017-02-25
Make sure scala_library_suite has at least once source

oscar-stripe:
without this, it seems things do not do what is expected (which may be hiding other bugs).


bazel-io:

oscar-stripe:
probably we want to handle the case of exactly 1 src differently too: just emit a single scala_library, and not do the extra redirection.

We may want to do that for 0 src as well, so scala_library_suite is exactly the same as scala_library.

@ianoc thoughts?

oscar-stripe:
(this burned us internally when some runtime deps were lost in the case of no srcs on a library_suite).

ianoc-stripe:
So while it would be strange to use a suite maybe for 0 sources for obvious reasons, i think as you say for `<= 1` we should just drop the scala_library in there. Makes it work more broadly and isn't really surprising. ?

johnynek:
Yeah, the more I thought about it, I came to the same conclusion.

bazel-io:

johnynek:
@ianoc what do you think now?

@ittaiz ?

ianoc:
This lgtm

johnynek:
test this please

ittaiz:
1) AFAICT you and @ianoc talked about the "1" case but you ended up not handling it. Is that ok for you?
2) Maybe add a Binary that uses "Exported"? Better yet I'd take the actual bug and make a test out of it "(this burned us internally when some runtime deps were lost in the case of no srcs on a library_suite)."

johnynek:
test this please

johnynek:
what about now @ittaiz ?

johnynek:
I did add a test for the actual bug case. you could not depend on a `scala_test` or `scala_binary` after the provider change. This fixes the issue.

Exports are orthogonal to the observed bug. I don't want to totally duplicate all the testing at this point for scala_binary that we have for scala_library, although that might be useful in the future (in general it would be nice to have better test coverage).

ittaiz:
@johnynek I think something with the comments got mixed up.
I suggested to add a test here since AFAIU the above target passed during compilation but failed on runtime, no? The reply I got from you talked about the "provider" so I think you meant the other PR (which did have tests).

johnynek:
Yeah. I deleted the comment. Some web page back button operator error on my
part.

ittaiz:
Ok. So do you want to add a test here? Am I not seeing the test?

johnynek:
It's 21:00 here. I'll get to it tomorrow.

ittaiz:
Sure thing. Good night!

PR #153, opened 2017-02-26
Adds the ability to set a list of absolute prefixes

ianoc-stripe:
Adds the ability to do a list of absolute prefixes. Adds a test for same. Also added some documentation to describe how to use thrift_library.


bazel-io:

README.md
```diff
@@ -350,6 +350,69 @@ scala tests for. The outer target defines a native test suite to run all the inn
...
+        </p>
+      </td>
+    </tr>
+    <tr>
+      <td><code>absolute_prefixes</code></td>
+      <td>
+        <p><code>List of strings; optional</code></p>
+        <p>Each of these strings acts as a wildcard expression of the form <code>*string_value</code> that is removed from the start of the path.
+        Example: thrift is at <code>a/b/c/d/e/A.thrift</code> , prefix of <code>b/c/d</code>. Will mean other thrift targets can refer to this thrift
+        at <code>e/A.thrift</code>. Exactly one of these must match all thrift paths within the target, more than one or zero will fail the build.
```
johnynek: can you comment what the use case is here? Why give a list if exactly one can match? I'm guessing this is for macros so you can have some defaults but then also add more to the list (but if that is the case, maybe more than one would match). For that use case, maybe we want to have a list and stop at the first that matches, and the macro could add the default to the end of the list.

I'm not sure I know the use case.

test/src/main/scala/scala/test/twitter_scrooge/prefix_test/e/f/b/c/d/BUILD
```diff
@@ -0,0 +1,8 @@
+load("//thrift:thrift.bzl", "thrift_library")
+
+thrift_library(
+  name = "d",
```
johnynek: should this be `b`?

test/src/main/scala/scala/test/twitter_scrooge/prefix_test/a/b/c/d/BUILD
```diff
@@ -0,0 +1,8 @@
+load("//thrift:thrift.bzl", "thrift_library")
+
+thrift_library(
+  name = "d",
```
johnynek: should this be `a`?

test/src/main/scala/scala/test/twitter_scrooge/prefix_test/e/f/b/c/d/BUILD
```diff
@@ -0,0 +1,8 @@
...
+  srcs = ["B.thrift"],
+  absolute_prefixes = ["prefix_test/a", "prefix_test/e/f"],
```
johnynek: can you comment why the list here?

Also, shouldn't this have `a` as a dep?

README.md
ianoc-stripe: So the issue here is that when you do a source dependency you pick up the local version of the rule from its prelude. So when our repo with a 'monorepo' layout, depends on a repo with a 'maven' style layout you wind up wanting different prefixes. But you should only match one for any given target/path.

test/src/main/scala/scala/test/twitter_scrooge/prefix_test/e/f/b/c/d/BUILD
ianoc-stripe: We have the list because I wanted to ensure the list works to test the functionality.

Maybe, but deps on thrift don't do anything? So yes, but doesn't matter too much i guess?

test/src/main/scala/scala/test/twitter_scrooge/prefix_test/a/b/c/d/BUILD
ianoc-stripe: normal rules would say just name it after the folder its in?

test/src/main/scala/scala/test/twitter_scrooge/prefix_test/e/f/b/c/d/BUILD
ianoc-stripe: Folder name is d, usually name the only target after it?

johnynek:
👍 

Will squash merge when green.

Issue #151, opened 2017-02-24
too many java sources can cause build to fail


softprops:
For those running mixed scala/java builds you can currently run into a limit javac places on the length of its command line.

```
java.io.IOException: Cannot run program "external/local_jdk/bin/javac": error=7, Argument list too long
    at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
    at io.bazel.rulesscala.scalac.ScalacProcessor.compileJavaSources(ScalacProcessor.java:201)
    at io.bazel.rulesscala.scalac.ScalacProcessor.processRequest(ScalacProcessor.java:96)
    at io.bazel.rulesscala.worker.GenericWorker.run(GenericWorker.java:125)
    at io.bazel.rulesscala.scalac.ScalaCInvoker.main(ScalaCInvoker.java:42)
Caused by: java.io.IOException: error=7, Argument list too long
    at java.lang.UNIXProcess.forkAndExec(Native Method)
    at java.lang.UNIXProcess.<init>(UNIXProcess.java:248)
    at java.lang.ProcessImpl.start(ProcessImpl.java:134)
    at java.lang.ProcessBuilder.start(ProcessBuilder.java:1029)
    ... 4 more
```

sbt works around this by using a feature of javac that lets you provide arguments in an [arg file](https://github.com/sbt/sbt/blob/0.13/compile/src/main/scala/sbt/compiler/javac/ForkedJava.scala#L17)

It should be pretty straight foward to add that, though it may also be worth while seeing if there is something that can be borrowed from bazels built-in javac support. I did a little of splunking myself and found [this](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/rules/java/JavaCompileAction.java#L1114-L1121) Im not sure how reusable (or accessible) that is here but its worth taking note of the behavior java_xxx rules exhibit.

softprops:
Making a special note this limit seems variable based on platform. It was observed on osx but appears the same list of files was not hit on linux. It may just be a generally safe practice to just default to the arg file like sbt

johnynek:
closing this until we see evidence that #155 didn't fix it.

softprops:
Fixed it for us ;)

PR #155, opened 2017-02-27
Capture javac args in a @argfile

softprops:
This should fix issue #151

### DIFF
--- a/src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
+++ b/src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
@@ -227,7 +227,7 @@ class ScalacProcessor implements Processor {
     String contents = String.join("\n", args);
     BufferedWriter writer = Files.newBufferedWriter(argsFile);
     writer.write(contents);
-    writer.flush();
+    writer.close();
     return argsFile;
   }
 
--- a/src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
+++ b/src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
@@ -27,7 +27,6 @@ import java.util.Map.Entry;
 import java.util.Map;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
-import java.util.stream.Collectors;
 import scala.tools.nsc.*;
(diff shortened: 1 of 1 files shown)


bazel-io:

softprops:
I think the cla I had signed was for one email address but these commits are associated with another. I've added a cla agreement for my the email address associated with the commits in this pull

johnynek:
can you say "I signed it" or something, and see if the bot picks it up?

softprops:
@johnynek having my co-pilot  ( @jrodbeta ) verify his cla. I think its a combination of both of our agreements. I'll check in after

softprops:
I signed it

jrodbeta:
I signed it

johnynek:
"CLAs are signed, but unable to verify author consent"

So maybe all the authors need to have signed since there are multiple ones?

softprops:
I think we're just trying to figure out how to trip the pr bot now

johnynek:
cool, okay since it looks like I am just supposed to verify these multi-person commits, it looks like it is legit, and it looks like the build passes, so I'll merge.

Thanks for the contribution, @softprops and @jrodbeta

softprops:
thanks man

jrodbeta:
Thanks @johnynek  I'm wondering if it's because of company email.  I'll get the company CLA done for next time.

ittaiz:
as an FYI from my experiments with the bot it seems like the primary email needs to sign (Be it private or corporate)

ittaiz:
no test here :( @johnynek I'd suggest that at least for bug fixes targeted like this we mandate a test. WDYT?

softprops:
I'm happy to follow up with some unit tests.

ittaiz:
I would appreciate that

johnynek:
@ittaiz It would be nice to have a test that failed before and now works. We are testing this code in the existing java-scala mixed targets, which is why I accepted it without a stronger test.

I kind of bias towards being pretty certain it is not worse, especially for newer contributors. But, yes, ideally we could be stronger and be sure that it is even better than it was before.

ittaiz:
I understand. The problem I'm worried about is the fear of change or its
reckless friend the regression. We will eventually refactor and/or add
features there and we want that agility.

PR #156, opened 2017-03-01
WIP: issue-76

or-shachar:
working on #76 
Instead of running tests using bash, it will use sh_tests. 
To run this you can run
`bazel test :all`

Known issues that I need to find solution for:
1.  All runs must share the same repository cache, and there must be one rule that will run before tests and download the libraries once (otherwise all tests would try to download the libraries in parallel to the same path). **Currently you get consistent behavior and test time only from second run**
1. The targets that run java fail because they cannot find java executable - I need your advise about how to make java available in sh_test env
1. I added `.bashrc` that define some rules for persistent worker and shared repository cache. Not sure how this will behave in CI, or if it's allowed.

@damienmg @johnynek  - Of course this PR is mainly for your review and advise, and not for merge yet.


bazel-io:

BUILD
```diff
@@ -0,0 +1,131 @@
...
+         "//test_expect_failure/disappearing_class:sources",
+         "//test_expect_failure/scala_library_suite:sources",
+         "//test_expect_failure/transitive/java_to_scala:sources",
+         "//test_expect_failure/transitive/scala_to_java:sources",
+         "//test_expect_failure/transitive/scala_to_scala:sources",
+         "//thrift:sources",
+         "//tut_rule:sources",
+         "//twitter_scrooge:sources"]
+
+additions = []
```
damienmg: What is additions for?

BUILD
```diff
@@ -0,0 +1,131 @@
...
+         "//thrift:sources",
+         "//tut_rule:sources",
+         "//twitter_scrooge:sources"]
+
+additions = [] 
+
+sh_test(
+    name = "build_targets_under_tests_package",
+    size = "large",
+    data = repo,
```
damienmg: If you want the toolchain to be shiped in the test you need to pass  "@local_jdk//:jdk", in the data dependency. You will also need bazel itself as a dependency and specify it to the --javabase option of bazel.

See https://github.com/bazelbuild/bazel/blob/master/src/test/shell/testenv.sh#L246
and https://github.com/bazelbuild/bazel/blob/master/src/test/shell/testenv.sh#L87

BUILD
or-shachar: Bad naming... and currently redundant. At first I used this mechanism to only run tests that run bazel, the "additions" allowed me to define bazel arguments / flags that will be added to each execution. Later on I saw that shipping .bazelrc is much more effective because some other tests call bash function with inner calls to bazel.

BUILD
or-shachar: Adding the `local_jdk://jdk` was easy and we made some progress (It fails on another thing right now which is - it claims that it cannot load the main class JavaBinary... ([look in travis](https://travis-ci.org/bazelbuild/rules_scala/builds/207059839)).

As for "bazel itself as dependency with --javabase" - I tried to adjust Bazel's testenv.sh to our repository but it seems to have some dependencies on other packages that I couldn't find. I tried to simply add the `--javabase <path>` to the `.bashrc` but it did not make any different in the result...

damienmg: So let me rephrase, you need:
  1. to set --javabase and --host_javabase like it is done in testenv in Bazel
  2. You will need to have "bazel" as a dependency to run inside the sandbox. Might not be important if it's on the PATH and installed outside of the workspace but I think it will fails on ci.bazel.io.

damienmg:
Trigger bazel ci: test this please

or-shachar:
@damienmg  - I don't think that you run the test_sh script in Jenkins...

damienmg:
Right, we do bazel test //test/..., can you add a test_suite there that execute the sh_test?

or-shachar:
Sure! But I think that once it's working it should be moved to a different location... like "intergration-tests" or something of a kind.

damienmg:
sure it just implies another change to ci.bazel.io, so it's easier to do it that way right now.

or-shachar:
Ok... I added another folder under `test` called `integration-tests` and put all tests there. You can trigger bazel ci: test again. Thanks

damienmg:
test this please

.travis.yml
```diff
@@ -16,11 +16,11 @@ jdk:
...
-  - chmod +x bazel-0.3.2-installer-linux-x86_64.sh
-  - ./bazel-0.3.2-installer-linux-x86_64.sh --user
-  - mv .bazelrc.travis .bazelrc
+  - chmod +x bazel-0.4.4-installer-linux-x86_64.sh
+  - ./bazel-0.4.4-installer-linux-x86_64.sh --user
+  - cat .bazelrc.travis >> .bazelrc
 
 script:
-  - bash test_run.sh
+  - bazel test test/integration-tests/...
```
ittaiz: why not `bazel test test/...`?

or-shachar: In previous solution it would only run the test_sh.sh which is the integration tests.
Inside the integration tests in runs the `bazel test test/...`

or-shachar:
@damienmg  you were right. Under platform `darwin-x86_64` all tests fail because it cannot find bazel.

or-shachar:
So we are left with several problems:
1. For some reason the run_java rules cannot load the main class. (with error message - `Could not find or load main class *`). Note that adding the `--host_javabase... --javabase...` to the inner bazel run did not make any difference on my local.
2. bazel is missing in some platforms and must be imported.
3. `run_scalaBinary` is possibly flaky (failed on timeout on Linux platform only on Jenkins).
4. `test_disappearing_classes` is failing because it tries to change the content of files in the workspace and when running inside bazel test - the file is in read only mode.

ittaiz:
just had a talk with @dslomov which proposed using `sh_binary` on the downloaded `http_archive` and having the `sh_test` depend on the `sh_binary` target to get it into the sandbox.
WDYT?

bazel-io:

ittaiz:
obviously we won't resurrect this. If we'll want concepts we'll take a look at the branch.

Issue #157, opened 2017-03-03
the scala rules ignore java source files inside .srcjar


johnynek:
If you look at where we call the java compiler, we assume the only java files are passed explicitly:

https://github.com/bazelbuild/rules_scala/blob/master/src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java#L96

We should instead look at the extracted source jars and pass the scala and java files accordingly:
https://github.com/bazelbuild/rules_scala/blob/master/src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java#L134

johnynek:
This was fixed by #159 I believe.

Issue #150, opened 2017-02-22
add jmh benchmarking rule


johnynek:
it would be nice to have something like:

```
scala_benchmark_jmh(
  name = "my_benchmarks",
  options = ["-i 3",  "-wi 3",  "-f1", "-t1"] # "3 iterations" "3 warmup iterations" "1 fork" "1 thread"
  srcs = ["Foo.scala", "Bar.scala"],
  deps = [":barlib", ":foolib"])
```
and when I build this, I get the output of jmh test.

if you look here:
https://github.com/ktoso/sbt-jmh/blob/master/plugin/src/main/scala/pl/project13/scala/sbt/JmhPlugin.scala#L108

it seems we basically just need to compile the code, then pass it to:
`JmhBytecodeGenerator.main(Array(bytecodeDir, sourceDir, resourceDir, generatorType).map(_.toString))`

if you look at that, you can see what it does here:
http://hg.openjdk.java.net/code-tools/jmh/file/86072ed4c056/jmh-generator-bytecode/src/main/java/org/openjdk/jmh/generators/bytecode/JmhBytecodeGenerator.java

cc @jeremydhoon

PR #158, opened 2017-03-03
Add support for JMH benchmarks

jeremydhoon:
### Summary

This PR adds crude support for JMH benchmarks with a `scala_benchmark_jmh` target. To do this, it also allows compilation of Java sources in source JARs.

The meat of the logic to generate JMH benchmark code was cargo-culted from @tixxit's excellent SBT plug-in for JMH benchmarks:

https://github.com/tixxit/sbt-benchmark/blob/master/src/main/scala/net/tixxit/sbt/benchmark/BenchmarkPlugin.scala

### Motivation
- https://github.com/bazelbuild/rules_scala/issues/157
- https://github.com/bazelbuild/rules_scala/issues/150

### Testing
Added a very simple JMH benchmark under `tests/jmh`.

### TODO

- [x] Generate JARs without updating timestamps.
- [x] (likely) want to split this PR into two, one per issue.

r? @johnynek

### DIFF
--- a/jmh/jmh.bzl
+++ b/jmh/jmh.bzl
@@ -1,0 +1,123 @@
+load("//scala:scala.bzl", "scala_binary", "scala_library")
+
+def jmh_repositories():
+  native.maven_jar(
+      name = "io_bazel_rules_scala_org_openjdk_jmh_jmh_core",
+      artifact = "org.openjdk.jmh:jmh-core:1.17.4",
+      sha1 = "126d989b196070a8b3653b5389e602a48fe6bb2f",
+  )
+  native.bind(
+      name = 'io_bazel_rules_scala/dependency/jmh/jmh_core',
+      actual = '@io_bazel_rules_scala_org_openjdk_jmh_jmh_core//jar',
+  )
+  native.maven_jar(
+      name = "io_bazel_rules_scala_org_openjdk_jmh_jmh_generator_asm",
+      artifact = "org.openjdk.jmh:jmh-generator-asm:1.17.4",
+      sha1 = "c85c3d8cfa194872b260e89770d41e2084ce2cb6",
+  )
+  native.bind(
+      name = 'io_bazel_rules_scala/dependency/jmh/jmh_generator_asm',
+      actual = '@io_bazel_rules_scala_org_openjdk_jmh_jmh_generator_asm//jar',
+  )
+  native.maven_jar(
+       name = "io_bazel_rules_scala_org_openjdk_jmh_jmh_generator_reflection",
+       artifact = "org.openjdk.jmh:jmh-generator-reflection:1.17.4",
+       sha1 = "f75a7823c9fcf03feed6d74aa44ea61fc70a8439",
+  )
+  native.bind(
+       name = 'io_bazel_rules_scala/dependency/jmh/jmh_generator_reflection',
+       actual = '@io_bazel_rules_scala_org_openjdk_jmh_jmh_generator_reflection//jar',
+  )
+  native.maven_jar(
+      name = "io_bazel_rules_scala_org_ows2_asm_asm",
+      artifact = "org.ow2.asm:asm:5.0.4",
+      sha1 = "0da08b8cce7bbf903602a25a3a163ae252435795",
+  )
+  native.bind(
+      name = "io_bazel_rules_scala/dependency/jmh/org_ows2_asm_asm",
+      actual = '@io_bazel_rules_scala_org_ows2_asm_asm//jar',
+  )
+  native.maven_jar(
+      name = "io_bazel_rules_scala_net_sf_jopt_simple_jopt_simple",
+      artifact = "net.sf.jopt-simple:jopt-simple:5.0.3",
+      sha1 = "cdd846cfc4e0f7eefafc02c0f5dce32b9303aa2a",
+  )
+  native.bind(
+      name = "io_bazel_rules_scala/dependency/jmh/net_sf_jopt_simple_jopt_simple",
+      actual = '@io_bazel_rules_scala_net_sf_jopt_simple_jopt_simple//jar',
+  )
+  native.maven_jar(
+      name = "io_bazel_rules_scala_org_apache_commons_commons_math3",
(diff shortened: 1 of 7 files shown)


bazel-io:

johnynek:
This is great!

Could we actually split into 2 PRs:
1. fixing the java + srcjar and any scalac cleanups? Also, can we add a test for this directly?
2. leveraging 1 adding the new jmh support?

Note stripe has signed the CLA, so if you make the PR from your stripe account you should be fine, or you can personally also sign the CLA.

jhoon-stripe:
I signed it!

(The Google CLA, that is)

jeremydhoon:
I also signed the CLA personally, for this Github username (@jeremydhoon).

johnynek:
Want to rebase on master now?

jeremydhoon:
Thanks for merging #158!

I still need to generate proper JARs (without updating timestamps) before this will be ready to go.

jmh/jmh.bzl
```diff
@@ -0,0 +1,149 @@
...
+
+  codegen = name + "_codegen"
+  src_jar = "%s_jmh.srcjar" % name
+  benchmark_list = "resources/META-INF/BenchmarkList"
+  compiler_hints = "resources/META-INF/CompilerHints"
+
+  binary_deploy_jar = "//src/java/io/bazel/rulesscala/jar:binary_deploy.jar"
+
+  java_tool = "@local_jdk//:java"
+  jar_tool = "@local_jdk//:jar"
```
oscar-stripe: we have a tool in this repo that makes jars reproducibly:

https://github.com/bazelbuild/rules_scala/blob/master/src/java/io/bazel/rulesscala/jar/BUILD#L7

we should try to use that, I think since we are using it elsewhere and it tackles the issues involved.

jmh/jmh.bzl
```diff
@@ -0,0 +1,149 @@
...
+      name = compiled_lib,
+      srcs = [src_jar],
+      deps = deps + [
+          "//external:io_bazel_rules_scala/dependency/jmh/jmh_core",
+          lib,
+          codegen,
+      ],
+      resources = [benchmark_list, compiler_hints],
+  )
+  scala_binary(
```
oscar-stripe: so, to run the benchmark you do `bazel run //path/to/target` and you get the output?

Another approach might be to actually run the benchmark in the `build` phase. So we have one more genrule that calls this binary to produce an output which is the benchmark result. If we did this, we could use bazel caching to not rerun the benchmark if things change. With this, the options we pass to `jmh` become arguments to the build target (if you change those, bazel would automatically rerun I think).

jmh/jmh.bzl
```diff
@@ -0,0 +1,149 @@
...
+      srcs = srcs,
+      deps = deps + [
+          "//external:io_bazel_rules_scala/dependency/jmh/jmh_core",
+      ],
+      scalacopts = scalacopts,
+      visibility = ["//visibility:public"],
+  )
+  jmh_benchmark_generator_tool = "__jmh_benchmark_generator"
+  scala_binary(
+    name = jmh_benchmark_generator_tool,
```
jeremydhoon: This should be the exact same for all benchmarks. Should it just be a standalone target called `__jmh_bench_gen` or something?

jmh/jmh.bzl
jeremydhoon: After discussing, it sounds like it makes the most sense to leave as-is, with the build product being a runnable benchmark.

jmh/jmh.bzl
```diff
@@ -0,0 +1,138 @@
...
+      visibility = ["//visibility:public"],
+  )
+  jmh_benchmark_generator_tool = "//src/scala/io/bazel/rules_scala/jmh_support:benchmark_generator"
+
+  codegen = name + "_codegen"
+  src_jar = "%s_jmh.srcjar" % name
+  benchmark_list = "resources/META-INF/BenchmarkList"
+  compiler_hints = "resources/META-INF/CompilerHints"
+
+  jdk_tool = "//tools/defaults:jdk"
```
johnynek: do we still need this?

jmh/jmh.bzl
```diff
@@ -0,0 +1,138 @@
...
+  jar_tool = "@local_jdk//:jar"
+  jar_creator_tool = "//src/java/io/bazel/rulesscala/jar:binary"
+
+  native.genrule(
+      name = codegen,
+      srcs = [lib],
+      outs = [src_jar, benchmark_list, compiler_hints],
+      tools = [
+          jmh_benchmark_generator_tool,
+          jdk_tool,
```
johnynek: looks unused.

jmh/jmh.bzl
johnynek: 👍

test/jmh/BUILD
```diff
@@ -0,0 +1,6 @@
+load("//jmh:jmh.bzl", "scala_benchmark_jmh")
+
+scala_benchmark_jmh(
```
johnynek: can we add to the `run_test.sh` script to actually run a benchmark to make sure things run, not just build?

jmh/jmh.bzl
jeremydhoon: Without including `jdk_tool`, the JAR creator encounters the following error:

```
Error: could not find libjava.dylib
Error: Could not find Java SE Runtime Environment.
```

jmh/jmh.bzl
jeremydhoon: I can leave a comment to explain why this is necessary.

test/jmh/BUILD
jeremydhoon: Good thinking! Will specify some uselessly-low command line settings to make sure it doesn't take too long to run.

jmh/jmh.bzl
johnynek: huh, that shouldn't be needed, but oh well. I mean, it is transitively needed for sure, but i thought it would have been pulled by the jar creator.

johnynek:
looks good. Will merge when green.

PR #161, opened 2017-03-05
travis to work on newer versions and on multiplatforms

or-shachar:
Previous travis ran only with bazel **0.3.2** and on **linux**.
This update makes sure travis will have 4 permutations:
- versions: **latest** (0.4.4) and **head**.
- platforms: **linux** and **osx**

#Notes:
- The source of this file is from `go_rules` adjusted to this repo
 - I skipped all apt requirements but `libxml2-utils` that failed without it
- Those changes still don't solve issue #76 but will help test solutions to it
- I also added prints on duration of each test that is run by the test_run.sh

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,41 @@
-addons:
-  apt:
-    sources:
-      - ubuntu-toolchain-r-test
-    packages:
-      - gcc-4.8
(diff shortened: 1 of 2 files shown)


bazel-io:

Issue #72, opened 2016-06-07
specs2 support


ittaiz:
Hi,
Most of our tests use specs2 (a small portion uses JUnit and a handful use scalatest).
Do you think specs2 support should be part of this library?
If so do you think it should be via a flag to scala_test? via an additional rule?
Not sure I like the fact that scalatest is always on the classpath even if I don't need it but might be negligible.

I can add this support in an external repo but I think it's more natural for it to reside here.

ittaiz:
I'm actually closing this one since we actually use the JUnit runner to run our specs2 tests (`SpecificationWithJUnit`) and so I'll just solve it locally with compiling using the mix_java_scala and then running a JUnit runner ( in sh_test?)

johnynek:
Sorry I didn't reply sooner. I think a scala_specs_test could be fine to
add. I don't see a good way to do it in one rule.

I don't think you need the scalatest dependency if you don't use the rule.

On Monday, June 13, 2016, Ittai Zeidman <notifications@github.com> wrote:

> Closed #72 <https://github.com/bazelbuild/rules_scala/issues/72>.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/bazelbuild/rules_scala/issues/72#event-691332080>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAEJdkLvZPJ3jc7sV1eQSTlNYnl2y3zCks5qLjKXgaJpZM4Ivzaq>
> .
>


-- 
P. Oscar Boykin, Ph.D. | http://twitter.com/posco | http://pobox.com/~boykin

ittaiz:
No worries at all, hope you enjoyed the getaway and congratulations :)
Do you want me to reopen the issue?
Like I said I don't need this currently but others might. I might even get around to it later on but now I'm trying to face my blockers.

johnynek:
Well, we can leave it closed if you have a work around (at least give us
the impression the issue count is not *too* high. ;)

On Mon, Jun 13, 2016 at 7:14 PM, Ittai Zeidman <notifications@github.com>
wrote:

> No worries at all, hope you enjoyed the getaway and congratulations :)
> Do you want me to reopen the issue?
> Like I said I don't need this currently but others might. I might even get
> around to it later on but now I'm trying to face my blockers.
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
[...]

PR #163, opened 2017-03-09
Junit & Specs2 support

ittaiz:
Hi,
As discussed long ago ( #72  and https://github.com/bazelbuild/bazel/issues/503) this is junit and specs2 support.
It actually brings with it a powerful test discovery mechanism which imitates the maven behavior. One can define multiple patterns and all *classes* whose name fits the pattern will be run. In Scala the classes (vs file-name) emphasis is substantial due to many classes in a file and mismatch between file-names and class names.

Features:
JUnit and Specs2 (using specs2 spec) Tests can:
1. Have jvm_flags (we use this internally to increase xmx for some suites)
2. Run test classes based on multiple patterns (we use IT*,*IT,*E2E for example)
3. Use supplied deps and runtime_deps

Notes:
1. I used bind even though it looks like Bazel want to deprecate it since that's what's used here (and Oscar advocates it) 
2. Specs2 version- I extracted it to a method since I spent a few hours debugging a mismatch between the version of specs2 and specs2_junit :(
3. Scala build file changes- I've added filegroups since I using "@scala//:lib/scala-reflect.jar" as a label didn't work for me. If you know how I can skip that I'd love to know.

Dilemas I'd love feedback on:
1. Grep usage- I'm currently using grep to filter the relevant classes. I considered the implications of switching from grep to Scala code (which already does some manipulation on the output). The pro of grep is performance implications since AFAIR Ulf warned me about these traversals being expensive on the JVM when we're talking about big targets. The pro of moving to Scala is to have the logic more cohesive (currently the Scala code assumes stuff from the grep like the fact that everything ends with .class) and to have a more expected regex syntax.

2. Better handle where a target has no test classes- currently if there are no matches grep returns an exit code of 1 and you'll get a weird message 
```
ERROR: /Users/ittaiz/workspace/rules_scala/test/BUILD:260:1: error executing shell command: 'unzip -l bazel-out/local-fastbuild/bin/test/JunitWithNoTests.jar | grep  -e Test\.class > bazel-out/local-fastbuild/bin/test/JunitWithNoTests_discovered_classes.txt' failed: bash failed: error executing command /bin/bash -c 'unzip -l bazel-out/local-fastbuild/bin/test/JunitWithNoTests.jar | grep  -e Test\.class > bazel-out/local-fastbuild/bin/test/JunitCustomPattern2_discovered_classes.txt': com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 1.
Target //test:JunitWithNoTests failed to build
```
This might actually be another point towards moving this filtering to the Scala code since we can fail with a clearer message. I don't think this is a crucial point btw and we can live with it.

3. WDYT about adding default macros for unit tests and "bigger" tests where the first has default suffixes of (Test* and *Test) and the latter has the default suffixes of (IT*, *IT, *E2E)? We will definitely build this internally since that's what we use. I think this has value in the OSS but didn't want to push too much.

4. WDYT about removing the "suffix" and "multiple suffixes" tests? They helped me evolve the code but I think that the multiple suffixes test covers them.

5. Can I merge two structs? Currently I do it manually but didn't kno if there was an easier syntax for this.


bazel-io:

ittaiz:
cc @ulfjack and @damienmg for their thoughts about performance implications of using grep vs jvm filtering (more info in the above message)

ittaiz:
@johnynek wdyt?

scala/scala.bzl
```diff
@@ -494,6 +511,55 @@ def _scala_test_impl(ctx):
...
+def _discover_classes(ctx, patterns, archive):
+    discovered_classes = ctx.new_file(ctx.label.name + "_discovered_classes.txt")
+    ctx.action(
+      inputs=[archive],
+      outputs=[discovered_classes],
+      progress_message="Discovering classes with patterns of %s" % patterns,
+      #TODO consider with Damien/Ulf/Oscar the implications of switching from grep to scala code
+      #Pro-> logic will be cohesive (currently the scala code assumes stuff from the grep)
+      #Con-> IIRC Ulf warned me about performance implications of these traversals
+      command="unzip -l {archive} | grep {combined_patterns} > {out}".format(archive=archive.path, combined_patterns=_prep_grep_pattern_from(patterns), out=discovered_classes.path))
```
johnynek: can we use `jar tf` here and use the bazel_tools jar so we don't depend on unzip?

scala/scala.bzl
```diff
@@ -623,6 +689,30 @@ exports_files([
   "lib/scala-xml_2.11-1.0.4.jar",
   "lib/scalap-2.11.8.jar",
 ])
+
+filegroup(
```
johnynek: we already have these here:

https://github.com/bazelbuild/rules_scala/blob/master/src/scala/BUILD#L1

Can we consolidate on those targets? Why were these needed?

I'm concerned about supporting 2.12 and 2.11 with all these direct references in different places.

specs2/specs2.bzl
```diff
@@ -0,0 +1,51 @@
+def specs2_version():
+  return "3.8.8"
```
johnynek: do we need this special function?

specs2/specs2.bzl
```diff
@@ -0,0 +1,51 @@
...
+  native.bind(name = 'io_bazel_rules_scala/dependency/scalaz/scalaz_core', actual = '@io_bazel_rules_scala_org_scalaz_scalaz_core//jar')
+
+def specs2_dependencies():
+  return [
+  "//external:io_bazel_rules_scala/dependency/specs2/specs2_core",
+  "//external:io_bazel_rules_scala/dependency/specs2/specs2_common",
+  "//external:io_bazel_rules_scala/dependency/specs2/specs2_matcher",
+  "//external:io_bazel_rules_scala/dependency/scalaz/scalaz_effect",
+  "//external:io_bazel_rules_scala/dependency/scalaz/scalaz_core",
+  "@scala//:scala-xml",
```
johnynek: does this really need xml and parser combinators? If so, that's a bummer.

specs2/specs2_junit.bzl
```diff
@@ -0,0 +1,16 @@
+load("//specs2:specs2.bzl", "specs2_repositories", "specs2_dependencies", "specs2_version")
+load("//junit:junit.bzl", "junit_repositories")
+  
+def specs2_junit_repositories():
+  specs2_repositories()  
+  junit_repositories()
+  # Aditional dependencies for specs2 junit runner
+  native.maven_jar(
+      name = "io_bazel_rules_scala_org_specs2_specs2_junit_2_11",
+      artifact = "org.specs2:specs2-junit_2.11:" + specs2_version(),
```
johnynek: I don't see how that function helps (specs2_version) since only one version will have the right sha1 below, no (unless someone generates a sha1 collision for specs2! ;)

src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
```diff
@@ -0,0 +1,45 @@
+package io.bazel.rulesscala.test_discovery
+
+import org.junit.runner.RunWith
+import org.junit.runner.Runner
+import org.junit.runners.Suite
+import org.junit.runners.model.RunnerBuilder
+
+@RunWith(classOf[TextFileSuite])
+class DiscoveredTestSuite
```
johnynek: can you add a comment to the approach here? It looks like this is always registered at the test to run, and then by environmental variable all classes are passed in. Is this right?

src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
```diff
@@ -0,0 +1,45 @@
...
+	extends Suite(builder, testClass, TextFileSuite.discoveredClasses)
+
+object TextFileSuite {
+
+  private val discoveredClasses = readDiscoveredClasses(classesRegistry)
+
+  private def readDiscoveredClasses(classesRegistry: String): Array[Class[_]] =
+  	entries(classesRegistry)
+  		.map(filterWhitespace)
+      .map(filterMetadata)
```
johnynek: strange spacing here.

src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
```diff
@@ -0,0 +1,45 @@
...
+  		.map(Class.forName)
+
+  private def filterMetadata(zipEntryParts: Array[String]): String = 
+  	zipEntryParts.last
+
+  private def filterWhitespace(zipEntry: String): Array[String] = 
+    zipEntry.split("\\s+")
+
+  private def dropFileSuffix(classEntry: String): String = 
+  	classEntry.split("\\.").head
```
johnynek: inconsistent indentation in this file.

test/src/main/scala/scala/test/junit/specs2/Specs2Tests.scala
```diff
@@ -0,0 +1,16 @@
...
+import org.specs2.mutable.SpecWithJUnit
+import scala.test.junit.support.JUnitCompileTimeDep
+
+class JunitSpecs2Test extends SpecWithJUnit {
+
+  "specs2 tests" should {
+    "run smoothly in bazel" >> {
+      JUnitCompileTimeDep.hello
+      success
+    }
```
johnynek: can we add a negative test in the failing tests path? We should show that something that should fail does.

test_run.sh
```diff
@@ -112,6 +112,16 @@ xmllint_test() {
...
 
+multiple_junit_suffixes() {
+  bazel test //test:JunitMultipleSuffixes
+  matches=$(grep -c -e 'Running E2E' -e 'Running IT' ./bazel-testlogs/test/JunitMultipleSuffixes/test.log)
+  if [ $matches -eq 2 ]; then
+    return 0
+  else
+    return 1
+  fi
+}
```
johnynek: does this create .xml output? I had to do extra work for scalatest with that. we test it above, note. If this does generate xml output, we should make sure we are testing it.

scala/scala.bzl
ittaiz: we can, I actually thought about it but decided (for some reason I don't remember) on unzip. I'll change it.

scala/scala.bzl
ittaiz: we can, they were needed since using "@scala//:lib/scala-reflect.jar" as a label didn't work for me. If you know how I can skip that I'd love to know. I know my PR notes were long but this is one of the things I noted there.

specs2/specs2.bzl
ittaiz: it does. why is that a bummer?

specs2/specs2_junit.bzl
ittaiz: As I mentioned in the PR notes the reason I added this was because I (mistakenly) used one version in the specs2.bzl file and a different one in the specs2_junit.bzl file. Finding that bug took me a few hours since the runtime error messages are really unclear. Once I found it out I realized my problem was code duplication so I solved it with a method.

src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
ittaiz: true, sublime editing. Will fix.

src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
ittaiz: true, sublime editing. Will fix.

src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
ittaiz: that's right. I'll add a comment though I'd rather try and change the name to make it more explicit. If it was called `TextFileSuiteEntryPoint` would that be clearer that this is just a "vessel"? an entry point? and that the interesting bit is in the `TextFileSuite`?

test/src/main/scala/scala/test/junit/specs2/Specs2Tests.scala
ittaiz: I thought in the beginning I'll need to add such a test but didn't need to since the tests I added pushed very small parts of the code. I'm really confident about the regression here. In any case I'll add one per your request.

test_run.sh
ittaiz: JUnitCore creates an xml output. I actually don't think we should test this since I didn't add any code relating to it. Additionally when we'll (soon) move to java_test and remove the launcher this test will be even more redundant.
WDYT?

specs2/specs2.bzl
ittaiz: answered in the other comment about it

ittaiz:
@johnynek thanks for your review! Answered the comments but I'm not sure in the right github way since I started my own review (whatever this means). Hope all is clear.

scala/scala.bzl
ittaiz: I moved to jar but when I tried to use the one from bazel_tools I got the following error:
```
____Loading package: test ____Loading package: @local_config_cc// ____Loading package: @local_config_xcode// ____Loading complete. Analyzing... ____Loading package: @io_bazel_rules_scala_org_specs2_specs2_matcher//jar ____Loading package: @io_bazel_rules_scala_org_openjdk_jmh_jmh_generator_asm//jar ____Loading package: @io_bazel_rules_scala_org_openjdk_jmh_jmh_generator_reflection//jar ____Found 89 targets... ____Building... ____[342 / 391] Discovering classes with patterns of ["E2E", "IT"] ERROR: /Users/ittaiz/workspace/rules_scala/test/BUILD:245:1: error executing shell command: 'external/local_jdk/bin/jar -tf bazel-out/local-fastbuild/bin/test/JunitMultipleSuffixes.jar | grep -e E2E\.class -e IT\.class > bazel-out/local-fastbuild/bin/test/JunitMultipleSuffixes_discovered_...' failed: Process exited with status 1 [sandboxed]. Error: could not find libjava.dylib Error: Could not find Java SE Runtime Environment. Use --strategy=Generating=standalone to disable sandboxing for the failing actions. ____Building complete. ____Elapsed time: 3.011s, Critical Path: 0.66s
```
My attempt was via adding an internal attribute (like _ijar), add it to the inputs of the action and then use it instead of the vanilla jar command but that doesn't seem like it's enough. It seems more stuff are needed as inputs. Any idea?

scala/scala.bzl
ittaiz: :( Out of luck.
```
____Loading complete. Analyzing... ERROR: /Users/ittaiz/workspace/rules_scala/test/BUILD:258:1: no such target '//src/scala:parser_xml': target 'parser_xml' not declared in package 'src/scala' defined by /Users/ittaiz/workspace/rules_scala/src/scala/BUILD and referenced by '//test:Specs2Tests'. ERROR: /Users/ittaiz/workspace/rules_scala/test/BUILD:258:1: no such target '//src/scala:parser_library': target 'parser_library' not declared in package 'src/scala' defined by /Users/ittaiz/workspace/rules_scala/src/scala/BUILD and referenced by '//test:Specs2Tests'. ERROR: /Users/ittaiz/workspace/rules_scala/test/BUILD:258:1: no such target '//src/scala:parser_reflect': target 'parser_reflect' not declared in package 'src/scala' defined by /Users/ittaiz/workspace/rules_scala/src/scala/BUILD and referenced by '//test:Specs2Tests'. ERROR: Analysis of target '//test:Specs2Tests' failed; build aborted. ____Elapsed time: 0.073s
```
Any idea why I'm getting this?

test/src/main/scala/scala/test/junit/specs2/Specs2Tests.scala
ittaiz: added

ittaiz:
@johnynek I've pushed the changes I understand and I was able to do.
Would appreciate your feedback on the outstanding things.

specs2/specs2.bzl
johnynek: xml literals are deprecated, and the library is being removed from the standard library, I think.

src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
johnynek: I think documenting the architecture of how you are doing this is a good idea. To be honest, I'm only probably 80% aware how, and since I won't be using this code, in 6 months I'll probably be at 10% and may have to review patches at that time. I'd really appreciate a paragraph here explaining the approach rather than someone having to go read about junit architecture, then put together the approach from integrating the .bzl and scala code.

scala/scala.bzl
```diff
@@ -494,6 +511,55 @@ def _scala_test_impl(ctx):
     _write_test_launcher(ctx, rjars)
     return _scala_binary_common(ctx, cjars, rjars)
 
+def _prep_grep_pattern_from(patterns):
```
johnynek: can you add a comment on the motivation of patterns?

test/src/main/scala/scala/test/junit/JunitMultipleSuffixes.scala
```diff
@@ -0,0 +1,25 @@
...
+
+}
+class JunitSuffixIT {
+
+  @Test
+  def someTest: Unit = {
+  	println("Running IT")
+  	assertEquals(1, 1)
+  }
+
```
johnynek: can we add a class that does not match the pattern with a failing test and see that it isn't called?

I think the pattern thing is super dangerous, by the way, since it is very easy to not notice that your tests are not being called because they don't match the pattern. I'm not really sure we should have this feature and instead discover all the tests. We have, unfortunately, had test refactors that disabled huge groups of tests and we didn't notice which ones had been disabled for a while.

test_run.sh
johnynek: I just mean that bazel expects test xml on a certain path:
https://github.com/bazelbuild/rules_scala/blob/master/scala/support/JUnitXmlReporter.scala#L82

so, I wonder if we can see that we are generating it on the right path.

scala/scala.bzl
johnynek: any comment about this?

ittaiz:
By Sunday noon I'll either comment or fix everything

ulfjack:
I don't think that Scala vs. Grep is really contentious. What's expensive is introspecting the classpath for tests with very large classpaths, because it involves loading every single class in the transitive closure, while many (or even most) of those will not be needed.

ittaiz:
Thanks @ulfjack!
@johnynek haven't forgotten. Got bogged down with some stuff here but will get to this as soon as I can...

test_run.sh
ittaiz: added

scala/scala.bzl
ittaiz: Ok, moved to `//src/scala`.
Gave `@scala//:lib/scala-library.jar` another try but got the following message so am staying with the filegroup.
```
ERROR: /Users/ittaiz/workspace/rules_scala/test/BUILD:258:1: in deps attribute of scala_junit_test rule //test:Specs2Tests: file '@scala//:lib/scala-library.jar' is misplaced here (expected no files). Since this rule was created by the macro 'scala_specs2_junit_test', the error might have been caused by the macro implementation in /Users/ittaiz/workspace/rules_scala/scala/scala.bzl:817:11.
```

scala/scala.bzl
```diff
@@ -728,4 +818,24 @@ def scala_library_suite(name,
...
+  implementation=_scala_junit_test_impl,
+  attrs={
+      "patterns": attr.string_list(default=["Test"]),
+      "_junit": attr.label(default=Label("//external:io_bazel_rules_scala/dependency/junit/junit"), single_file=True),
+      "_hamcrest": attr.label(default=Label("//external:io_bazel_rules_scala/dependency/hamcrest/hamcrest_core"), single_file=True),
+      "_suite": attr.label(default=Label("//src/java/io/bazel/rulesscala/test_discovery:test_discovery")),
+      } + _implicit_deps + _common_attrs,
+  outputs={
+      "jar": "%{name}.jar",
+      "deploy_jar": "%{name}_deploy.jar",
```
ittaiz: Because of the code reuse of `scala_binary_common`. `scala_test` does the same.
We can change the common one to optionally generate it and then change both test rules to not generate them. WDYT?
In any case feels like a different PR.

test/src/main/scala/scala/test/junit/JunitMultipleSuffixes.scala
ittaiz: We have `ClassCoveringRegressionFromTakingAllClassesInArchive`.
If the pattern is ignored then JUnit fails because there are no tests there.
Good enough?

scala/scala.bzl
```diff
@@ -494,6 +511,55 @@ def _scala_test_impl(ctx):
...
+def _discover_classes(ctx, patterns, archive):
+    discovered_classes = ctx.new_file(ctx.label.name + "_discovered_classes.txt")
+    ctx.action(
+      inputs=[archive],
+      outputs=[discovered_classes],
+      progress_message="Discovering classes with patterns of %s" % patterns,
+      #TODO consider with Damien/Ulf/Oscar the implications of switching from grep to scala code
+      #Pro-> logic will be cohesive (currently the scala code assumes stuff from the grep)
+      #Con-> IIRC Ulf warned me about performance implications of these traversals
+      command="jar -tf {archive} | grep {combined_patterns} > {out}".format(archive=archive.path, combined_patterns=_prep_grep_pattern_from(patterns), out=discovered_classes.path))
```
ittaiz: Thanks for the reference! I indeed tried it before using the shell `jar` but what I missed was the need for `_jdk` which the [comment](https://github.com/bazelbuild/rules_scala/blob/master/thrift/thrift.bzl#L67) on line 67 gave me.
It's fixed but while fixing it I noticed that `_implicit_deps` contain an entry for `_jar` but it points to `"//src/java/io/bazel/rulesscala/jar:binary_deploy.jar"`. This sounds to me like it should be `jar_creator_deploy_jar` or something of the sort. WDYT?

ittaiz:
@johnynek Handled all comments. Would appreciate your feedback.
Additionally if you can give your opinion about dilemmas #3,#4,#5 I'd appreciate it.
Thanks

test/src/main/scala/scala/test/junit/JunitMultipleSuffixes.scala
ittaiz: Oh and about the danger of the patterns- This is how maven works and is a big need for us to migrate substantial codebases since people won't manage the targets manually. On the other hand they have code supporting the tests side by side with the tests relying on the pattern.
Would you like the default pattern to be "*.class"? I don't like it but can live with it since we'll use our own defaults.

scala/scala.bzl
ittaiz: I added in the general comment about how the design works. Do you want another comment here?
This is another indication that maybe I should move the grepping to scala code

scala/scala.bzl
ittaiz: @johnynek `//src/scala` doesn't work when I try to use it in an external repository.
I think it's because it's a macro.
If I prefix it with the name of the rules_scala repo in my workspace then it works but that doesn't sound like the right solution since people might name it otherwise, right?
Any thoughts?

scala/scala.bzl
```diff
@@ -494,6 +511,57 @@ def _scala_test_impl(ctx):
...
+    combined_pattern = ""
+    for pattern in patterns:
+      combined_pattern += " -e {}\.class".format(pattern)
+    return combined_pattern
+
+def _discover_classes(ctx, patterns, archive):
+    discovered_classes = ctx.new_file(ctx.label.name + "_discovered_classes.txt")
+    ctx.action(
+      #We need _jdk to even run _jar. Depending on _jar is not enough with sandbox
+      inputs= [archive] + ctx.files._jar + ctx.files._jdk,
```
johnynek: I guess this is supposed to be `ctx.executable._jar` as shown here: https://github.com/bazelbuild/rules_scala/pull/171/files#diff-9957b181001dad28e385437a86e99126R106

then I think we don't need `files._jdk`.

scala/scala.bzl
```diff
@@ -494,6 +511,57 @@ def _scala_test_impl(ctx):
...
+      outputs=[discovered_classes],
+      progress_message="Discovering classes with patterns of %s" % patterns,
+      #TODO consider with Damien/Ulf/Oscar the implications of switching from grep to scala code
+      #Pro-> logic will be cohesive (currently the scala code assumes stuff from the grep)
+      #Con-> IIRC Ulf warned me about performance implications of these traversals
+      command="{jar} -tf {archive} | grep {combined_patterns} > {out}".format(archive=archive.path, combined_patterns=_prep_grep_pattern_from(patterns), out=discovered_classes.path, 
+      jar=ctx.file._jar.path))
```
johnynek: `ctx.executable._jar.path` maybe?

scala/scala.bzl
```diff
@@ -494,6 +511,57 @@ def _scala_test_impl(ctx):
...
+    return discovered_classes
+
+def _gen_test_suite_based_on_prefix(ctx, archive):
+    discovered_classes = _discover_classes(ctx, ctx.attr.patterns, archive)
+    return struct(suite_class = "io.bazel.rulesscala.test_discovery.DiscoveredTestSuite", classesFlag = "-Dbazel.discovered.classes.file.path=%s" % discovered_classes.short_path, discovered_classes = discovered_classes)
```
johnynek: can we wrap this line on `,`?
```python
return struct(
    suite_class = "foo",
    classesFlag = "bar",
    ...
```

scala/scala.bzl
```diff
@@ -494,6 +511,57 @@ def _scala_test_impl(ctx):
...
+
+def _scala_junit_test_impl(ctx):
+    deps = ctx.attr.deps + [ctx.attr._suite]
+    jars = _collect_jars(deps)
+    (cjars, rjars) = (jars.compiletime, jars.runtime)
+    junit_deps = [ctx.file._junit,ctx.file._hamcrest]
+    cjars += junit_deps
+    rjars += [
```
johnynek: can we move the next line up to start here: `[ctx.outputs.jar,...`

specs2/specs2.bzl
```diff
@@ -0,0 +1,51 @@
...
+  native.bind(name = 'io_bazel_rules_scala/dependency/scalaz/scalaz_core', actual = '@io_bazel_rules_scala_org_scalaz_scalaz_core//jar')
+
+def specs2_dependencies():
+  return [
+  "//external:io_bazel_rules_scala/dependency/specs2/specs2_core",
+  "//external:io_bazel_rules_scala/dependency/specs2/specs2_common",
+  "//external:io_bazel_rules_scala/dependency/specs2/specs2_matcher",
+  "//external:io_bazel_rules_scala/dependency/scalaz/scalaz_effect",
+  "//external:io_bazel_rules_scala/dependency/scalaz/scalaz_core",
+  "//src/scala:scala_xml",
```
johnynek: since these are just strings, in other repos, this won't work, I guess. It only works when the rule is run in this repo....

test_run.sh
```diff
@@ -112,6 +124,21 @@ xmllint_test() {
...
+
+junit_generates_xml_logs() {
+  bazel test //test:JunitTestWithDeps
+  test -e ./bazel-testlogs/test/JunitTestWithDeps/test.xml
```
johnynek: should this be non-empty?

ittaiz: Do you mean to check it's not empty? Isn't this the same check you do for scala_test?

specs2/specs2.bzl
ittaiz: Indeed. I think I commented somewhere and said it doesn't work for me without prefixing it with the repo name I gave rules_scala.
Should I revert it to @scala like my previous solution?

test_run.sh
johnynek: yeah, just double checked `xmllint` fails on an empty file. Disregard this comment.

specs2/specs2.bzl
ittaiz: I'll push a change with bind and ping you to give your opinion which of the two you prefer. In any case even if it will be a scala_toolchain it will probably be a humble beginning. Sounds good?

johnynek:
sounds good.

ittaiz:
@johnynek moved from explicitly depending on `@scala` to using `external` and using `bind` in `scala_repositories`. Problem is that I couldn't avoid keeping the filegroup public which doesn't look that well to me. Any suggestions?
Handled formatting.
Trying to fix the _jar problem.

johnynek:
look like good changes to me. Can you answer each of my comments and let me know when you think it is ready to merge.

This is going to conflict somewhat with #171 if we don't synchronize a bit (the use of executable vs. file seems meaningful).

scala/scala.bzl
sdtwigg: TL;DR: Definitely prefer ctx.executable._jar here, especially since you are using it as an executable.

So, I did some major experimentation with this. _jar being marked as executable should be sufficient to automatically pull in its runfiles regardless of whether you access via executable or files. (ctx.action WILL NOT find runfiles on all inputs, just those marked with executable=True. Note that you can still find runfiles on any label, executable or not, via ctx.attr.....data_runfiles|default_runfiles.files. 

That said, I think it is safest to access via ctx.executable._jar so it is clear that it is being used as an executable and that there is an error if you accidentally remove executable=True.

scala/scala.bzl
sdtwigg: +1, since merely calling it to execute it here and I am targeting removing single_file=True in other PR.

scala/scala.bzl
```diff
@@ -728,4 +828,25 @@ def scala_library_suite(name,
...
     scala_library(name = name, deps = ts, exports = exports + ts, visibility = visibility)
 
+scala_junit_test = rule(
+  implementation=_scala_junit_test_impl,
+  attrs= _implicit_deps + _common_attrs + {
+      "patterns": attr.string_list(default=["Test"]),
+      "_junit": attr.label(default=Label("//external:io_bazel_rules_scala/dependency/junit/junit"), single_file=True),
+      "_hamcrest": attr.label(default=Label("//external:io_bazel_rules_scala/dependency/hamcrest/hamcrest_core"), single_file=True),
+      "_suite": attr.label(default=Label("//src/java/io/bazel/rulesscala/test_discovery:test_discovery")),
+      "_jar": attr.label(executable=True, cfg="host", default=Label("@bazel_tools//tools/jdk:jar"), single_file=True, allow_files=True),
```
sdtwigg: single_file=True should not be necessary here. Also, this should already be in _implicit_deps.

ittaiz:
@johnynek I think that I've answered all of the issues.
Big main change is that I've stopped using patterns in favor of prefixes/suffixes.
I think this will simplify usage greatly while still keeping flexibility high.
I'm also throwing a clear error if no tests are found so that if tests are suddenly lost then the person will (hopefully) get an error. This won't work if only one test is renamed but will catch some cases at least. 
This also allowed me to stop using grep and jar in favor of JarInputStream which gives me easier and more cohesive implementation and I won't get in the way of #171 .
On my side we're good to merge

ittaiz:
@sdtwigg I just dropped the whole jar running all together for other
reasons

scala/scala.bzl
```diff
@@ -298,6 +299,25 @@ def _write_launcher(ctx, jars):
...
+      archiveFlag = test_suite.archiveFlag,
+      prefixesFlag = test_suite.prefixesFlag,
+      suffixesFlag = test_suite.suffixesFlag,
+      printFlag = test_suite.printFlag,
+      #allows setting xmx for example for tests which use a lot of memory
+      jvm_flags = " ".join(ctx.attr.jvm_flags)
+    )
+    ctx.file_action(
+        output=ctx.outputs.executable,
+        content=content)
```
sdtwigg: (Depending on PR submission order one of us will have to remember this) this will be the call to the unified _write_launcher in #171:

mandatoryJvmFlags = ["-ea", test_suit.archiveFlag, test_suite.prefixesFlag, test_suite.suffixesFlag, test_suite.printFlag]
 _write_launcher(
      ctx = ctx,
      rjars = jars,
      main_class = "org.junit.runner.JUnitCore",
      jvm_flags = mandatoryJvmFlags + ctx.attr.jvm_flags,
      args = test_suite.suite_class,
      run_before_binary = "",
      run_after_binary = "",
  )

ittaiz:
Sure thing. I'd call it "launcherJvmFlags" but yeah.

damienmg:
Just testing ci:

retest this please

damienmg:
So seems like it work for PR

scala/scala.bzl
```diff
@@ -831,11 +807,12 @@ def scala_library_suite(name,
 scala_junit_test = rule(
   implementation=_scala_junit_test_impl,
   attrs= _implicit_deps + _common_attrs + {
-      "patterns": attr.string_list(default=["Test"]),
+      "prefixes": attr.string_list(default=[]),
+      "suffixes": attr.string_list(default=[]),
```
johnynek: I wonder if a default here like `["Test", "Tests", "Spec", "Specs"]` makes sense or something.

johnynek:
Okay, @ittaiz if you merge with master and make the last changes I think we are ready to merge. @sdtwigg can you do a review pass on the merge to make sure we combine with the unified launcher writing code?

I think we are closing in on what might almost be called a 1.0 release of these rules (scalatest + junit test support, jmh benchmarks support, and if we get java.providers in so java can depend on scala, I think we are in pretty decent shape actually).

scala/scala.bzl
ittaiz: Also thought about it (and noted in the PR comments) and thought about adding macros.
The reason I'd leave it as-is is so that the choices will be explicit. Even if we add the macros their names will suggest the prefixes/suffixes which are being used.
WDYT?

ittaiz:
@sdtwigg @johnynek I'm ready to go

sdtwigg:
LGTM

scala/scala.bzl
```diff
@@ -554,6 +554,40 @@ def _scala_test_impl(ctx):
...
+    jars = _collect_jars(deps)
+    (cjars, rjars) = (jars.compiletime, jars.runtime)
+    junit_deps = [ctx.file._junit,ctx.file._hamcrest]
+    cjars += junit_deps
+    rjars += [ctx.outputs.jar,
+              ctx.file._scalalib
+              ] + junit_deps
+    rjars += _collect_jars(ctx.attr.runtime_deps).runtime
+    test_suite = _gen_test_suite_flags_based_on_prefixes_and_suffixes(ctx, ctx.outputs.jar)
+    launcherJvmFlags = ["-ea", test_suit.archiveFlag, test_suite.prefixesFlag, test_suite.suffixesFlag, test_suite.printFlag]
```
sdtwigg: test_suite

ittaiz: Thanks!
Used github resolve conflicts and just copy pasted your snippet :( I checked it but missed that. Resolved.

ittaiz: Thanks!
Used github resolve conflicts and just copy pasted your snippet :( I checked it but missed that. Resolved.

ittaiz: Thanks!
Used github resolve conflicts and just copy pasted your snippet :( I checked it but missed that. Resolved.

sdtwigg: Yeah, sorry about that. My weak defense is that I was in a coffeeshop and quickly writing it before I had to run off.

ittaiz:
No need to apologize. I should have run the build locally but GH's resolve
conflicts UI was too compelling.

johnynek:
Looks good to me, but the tests are red. As soon as they are green we should merge.

ittaiz:
Oscar, ping? Anything left here?

johnynek:
test this please

johnynek:
I think it looks good when the CI is green.

ittaiz:
this still fails locally but with a weird error:
trying to understand how these changes relate to `_build_nosrc_jar` for `twitter_scrooge:allscrooges`
```
____Loading complete.  Analyzing...
____Found 102 targets...
____[1 / 24] scala //test/src/main/scala/scala/test/twitter_scrooge:allscrooges
ERROR: /Users/ittaiz/workspace/rules_scala/src/scala/scripts/BUILD:19:1: error executing shell command: '
  rm -rf bazel-out/host/bin/src/scala/scripts/generator.jar_tmp
  set -e
  mkdir -p bazel-out/host/bin/src/scala/scripts/generator.jar_tmp
  # copy any resources

  external/local_jdk/bin/java -...' failed: Process exited with status 2 [sandboxed].
Error: could not find libjava.dylib
Error: Could not find Java SE Runtime Environment.
Use --strategy=Generating=standalone to disable sandboxing for the failing actions.
____Elapsed time: 0.737s, Critical Path: 0.61s
```

ittaiz:
I wonder if I can trigger as well-
test this please

johnynek:
test this please

johnynek:
This is weird. I wonder if master is actually broken and somehow we didn't see? But I thought everything was green?

ittaiz:
`test bazel build test/...` passes on travis and fails on my machine so I'm not really certain what's going on here :(

johnynek:
Travis can't use sandboxing, I don't think, so I think we don't see issues there. I would look most closely at the ci.bazel.io.

I wonder if we can really remove the dependencies on the jdk as was done. cc @sdtwigg

ittaiz:
@johnynek seems like mac os is broken on CI which makes sense since that's my os. getting kids ready for school. will take a look soon.

ittaiz:
@johnynek @sdtwigg checkout master and indeed it's broken.

sdtwigg:
I'm re-reviewing as well.... Is there a way to get the error log from the failing build?
I had tried the CL in a local hermetic build and it was fine with the JDK removed but it IS possible something strange happened. :/

sdtwigg:
No, we need to add the dependency on the jdk back. The default bazel jdk/java tooling setup, located at https://github.com/bazelbuild/bazel/blob/master/src/main/tools/jdk.BUILD, does not properly associate dependencies for the java and javac targets.

'Amusingly' (and I'll try to figure out some way to avert this for future tests), I HAD run this on a hermetic BUILD system; however, the system jdk/java tooling in that system is slightly different. Notably, it was establishing a dependency for java and javac on the jdk.

ittaiz:
I think the main way to avert this in the future is to mandate an automatic
ci.bazel.io green light and not settle for a Travis one or a manual run.
Should we merge this and then Stephen will fix on top of it? It was green
before merging master in

sdtwigg:
Since you use _scala_binary_common rather than create your own binary, should be fine with merging this before #172. (I want to wait for that to show green in CI so concerned with delaying your PR further.)

johnynek:
okay @ittaiz we should be good to go if you can merge master one more time. Sorry for the pain and thanks for bearing with it in this contribution.

The reason bazel ci is not automatic, I understand, is to limit security risk. A human looks first to make sure it looks sane.

ittaiz:
@johnynek merged. can you please trigger CI?

damienmg:
test this please

ittaiz:
@johnynek @damienmg 
```
Caused by: java.lang.NoClassDefFoundError: Could not initialize class hudson.util.ProcessTree$UnixReflection
``` 
:( build failed and I have no idea why this is happening.

damienmg:
A problem on our CI ? /cc @philwo that have been touching our VM today

And maybe that will do it: retest this please

johnynek:
retest this please

johnynek:
It *looks* like the CI is having an issue after all the tests have passed. Since 2/3 of the platforms on bazel ci are green and travis is green, I'm going to merge.

Thanks for carrying this one over the finish line @ittaiz

philwo:
@johnynek Yes, Bazel CI is currently having trouble, because I replaced the old Ubuntu 15.10 slaves with Ubuntu 16.04. That version is somehow insisting on making a pre-release version of Java 9 the default on the system, which breaks the Jenkins slave service, giving the NoClassDefFoundError exception.

I'll fix this today.

johnynek:
Thanks for the update.

ittaiz:
Sure thing, thank you for your review.
Happy to see this baby growing :)

PR #165, opened 2017-03-13
Transitive dependencies for JMH benchmarks

jeremydhoon:
### Summary

This PR properly gathers the dependencies for JMH benchmark generation. It uses a Skylark rule to assemble the runtime dependencies for the compiled library for which JMH will generate benchmark code. We need the runtime dependencies (rather than the `ijar` compile-only dependencies) since JMH's generators actually need to loaded classes depended upon.

### Motivation

We were failing to build benchmarks with dependencies on targets exported from the benchmarks' dependencies.

### Testing

Added some Java and Scala dependencies to test target for JMH benchmarks.

r? @johnynek

### DIFF
--- a/jmh/jmh.bzl
+++ b/jmh/jmh.bzl
@@ -76,10 +76,10 @@ def _scala_generate_benchmark(ctx):
   class_jar = ctx.attr.src.scala.outputs.class_jar
   classpath = _scala_construct_runtime_classpath([ctx.attr.src])
   ctx.action(
-      outputs = [ctx.outputs.benchmark_list, ctx.outputs.compiler_hints, ctx.outputs.src_jar],
+      outputs = [ctx.outputs.src_jar, ctx.outputs.resource_jar],
       inputs = [class_jar] + classpath,
       executable = ctx.executable._generator,
-      arguments = [f.path for f in [class_jar, ctx.outputs.src_jar] + classpath],
+      arguments = [f.path for f in [class_jar, ctx.outputs.src_jar, ctx.outputs.resource_jar] + classpath],
       progress_message = "Generating benchmark code for %s" % ctx.label,
   )
 
@@ -91,8 +91,7 @@ scala_generate_benchmark = rule(
     },
     outputs = {
         "src_jar": "%{name}.srcjar",
-        "benchmark_list": "resources/META-INF/BenchmarkList",
-        "compiler_hints": "resources/META-INF/CompilerHints",
+        "resource_jar": "%{name}_resources.jar",
     },
 )
 
@@ -115,11 +114,7 @@ def scala_benchmark_jmh(**kw):
   )
 
   codegen = name + "_codegen"
-  benchmark_list = "resources/META-INF/BenchmarkList"
-  compiler_hints = "resources/META-INF/CompilerHints"
-
   scala_generate_benchmark(name=codegen, src=lib)
-
   compiled_lib = name + "_compiled_benchmark_lib"
   scala_library(
       name = compiled_lib,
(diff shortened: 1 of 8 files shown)


bazel-io:

jmh/jmh.bzl
```diff
@@ -84,47 +122,15 @@ def scala_benchmark_jmh(**kw):
...
-      outs = [src_jar, benchmark_list, compiler_hints],
-      tools = [
-          jmh_benchmark_generator_tool,
-          # Without the JDK, local_jdk//:jar complains of failing to find
-          # a Java SE Runtime Environment.
-          jdk_tool,
-          jar_creator_tool,
-          jar_tool,
-      ],
-      cmd = """
```
jeremydhoon: As a nice benefit, we no longer need to run bash to build benchmarks.

jmh/jmh.bzl
```diff
@@ -61,6 +61,44 @@ jdk_tool = Label("//tools/defaults:jdk")
...
+  classpath = _scala_construct_runtime_classpath([ctx.attr.src])
+  ctx.action(
+      outputs = [ctx.outputs.benchmark_list, ctx.outputs.compiler_hints, ctx.outputs.src_jar],
+      inputs = [class_jar] + classpath,
+      executable = list(ctx.attr._generator.files)[0],
+      arguments = [f.path for f in [class_jar, ctx.outputs.src_jar] + classpath],
+      progress_message = "Generating benchmark code for %s" % ctx.label,
+  )
+  return struct(
+      files=set([ctx.outputs.src_jar])
```
johnynek: I'm not sure if we need this. Can you comment as to why?

jmh/jmh.bzl
```diff
@@ -61,6 +61,44 @@ jdk_tool = Label("//tools/defaults:jdk")
...
+
+scala_generate_benchmark = rule(
+    implementation = _scala_generate_benchmark,
+    attrs = {
+        "src": attr.label(allow_single_file=True, mandatory=True),
+        "_generator": attr.label(executable=True, cfg="host", default=Label("//src/scala/io/bazel/rules_scala/jmh_support:benchmark_generator"))
+    },
+    outputs = {
+        "src_jar": "%{name}.srcjar",
+        "benchmark_list": "resources/META-INF/BenchmarkList",
```
johnynek: I'm worried you will have a name collision if you have two benchmarks in the same file. Can we not put there into the .srcjar or in another jar?

Or at least make sure in a test (perhaps below) where you have two benchmarks in the same BUILD.

jmh/jmh.bzl
johnynek: nice!

jmh/jmh.bzl
```diff
@@ -84,47 +122,15 @@ def scala_benchmark_jmh(**kw):
   benchmark_list = "resources/META-INF/BenchmarkList"
```
johnynek: I don't think you are using `src_jar` right above this line.

test/jmh/BUILD
```diff
@@ -1,6 +1,27 @@
-load("//jmh:jmh.bzl", "scala_benchmark_jmh")
+load("//jmh:jmh.bzl", "scala_library", "scala_benchmark_jmh")
+
+native.java_library(
```
johnynek: I didn't even know you could write `native.java_library` but `java_library` should work.

johnynek:
such a champ! This looks great. Will hold off on the 👍 until we merge the resource_jar path and maybe we can rebase to clean up this diff?

Thanks so much for pushing this through. It will be really nice to have the ability to make benchmarks.

jhoon-stripe:
Confirming I am ok contributing these commits.

johnynek:
@jeremydhoon can you rebase on master? It's a little hard for me to see the diff, which I think is much smaller now.

PR #167, opened 2017-03-14
Implement resource_jars attribute for Scala targets

jeremydhoon:
### Summary

Implements the `resource_jars` attribute which is already available for Java targets:

https://bazel.build/versions/master/docs/be/java.html#java_library.resource_jars

This implementation follows the behavior described in the Java target documentation:

> Set of archives containing Java resources. If specified, the contents of these jars are merged into the output jar.

The Java behavior appears to be implemented here...
https://github.com/bazelbuild/bazel/blob/master/src/java_tools/buildjar/java/com/google/devtools/build/buildjar/SimpleJavaLibraryBuilder.java#L203-L226

...and it appears to generally agree with the behavior in this PR.

### Motivation

Easy collection of generated resources in https://github.com/bazelbuild/rules_scala/pull/165

### Testing

Added a test case; also got this logic working with #165

r? @johnynek


bazel-io:

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
```diff
@@ -87,6 +87,10 @@ public void processRequest(List<String> args) throws Exception {
        */
       copyResources(ops.resourceFiles, ops.resourceStripPrefix, tmpPath);
       /**
+       * Extract and copy resources from resource jars
+       */
+      copyResourceJars(ops.resourceJars, tmpPath);
```
jeremydhoon: Should we copy-in the contents of resource jars before copying in resources, in order to give individual resources precedence?

johnynek:
osx head seems to be getting stuck, which is a bit worrisome.

jeremydhoon:
Looks like it unstuck!

jhoon-stripe:
I can indeed confirm I am happy to see these commits contributed to `rules_scala`.

PR #168, opened 2017-03-15
printing the output of failed sh_test run with line breaks

or-shachar:
sh_test don't print the each test run output unless it fails.
If it fails - it prints the output without line breaks, which makes it really hard to understand whats going on.

This tiny addition will fix it and make sure the output will be printed with line breaks


bazel-io:

johnynek:
👍 

thanks for the fix!

PR #175, opened 2017-04-18
Some small fixes to the scalac infrastructure

sdtwigg:
Removing unused imports, variables.

Properly typing the Enumeration meant no longer needed type-cast, which
is nice. Avoids implicit 'inference' of Enumneration<?> by javac which leads to 'fun'.

constParams removal because was dead code.

== is reference equality and .equals is potentially value equality.

### DIFF
--- a/src/java/io/bazel/rulesscala/scalac/CompileOptions.java
+++ b/src/java/io/bazel/rulesscala/scalac/CompileOptions.java
@@ -1,6 +1,5 @@
 package io.bazel.rulesscala.scalac;
 
-import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


johnynek:
👍 

This is great. Sorry some of our java is a bit rusty...

Will merge when travis goes green.

sdtwigg:
To be honest, amusingly, this is mostly to satisfy our presubmit checks when doing imports. Currently have to either explicitly suppress them or manually make these changes.

I was about to adjust our import script to automatically rewrite these lines but doing a PR just seemed simpler and better overall 🤣.

ittaiz:
All checks passed

PR #177, opened 2017-04-19
suggested different way to import rules_scala

or-shachar:
Instead of using git_repository use http_archive which should be more efficient


bazel-io:

johnynek:
👍

PR #178, opened 2017-04-19
Use consistent way of expressing external deps

johnynek:
Make all the deps use the bind targets.

My motivation for this is that we use:

https://github.com/johnynek/bazel-deps

to specify our maven dependencies and deal with updating the transitive versions at stripe. We'd rather pull all the jars just from maven than from the sdk. With the recent changes, that is now possible since we don't use the scala repl script any longer.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -598,9 +598,9 @@ def _scala_junit_test_impl(ctx):
 _implicit_deps = {
   "_ijar": attr.label(executable=True, cfg="host", default=Label("@bazel_tools//tools/jdk:ijar"), allow_files=True),
   "_scalac": attr.label(executable=True, cfg="host", default=Label("//src/java/io/bazel/rulesscala/scalac"), allow_files=True),
-  "_scalalib": attr.label(default=Label("@scala//:scala-library"), single_file=True, allow_files=True),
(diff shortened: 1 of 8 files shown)


johnynek:
@ittaiz what do you think?

sdtwigg:
Is there any way we can have the bulk of the scala_repositories code in its own bzl files?

It might help cut down on how long scala.bzl is getting. My critique is that scala_repositories is nominally about setting up a WORKSPACE whereas the other functions here are called in a BUILD file.

johnynek:
@sdtwigg I think that is fine, but I might add a function to redirect to it so we don't break people who bump up their versions.

Would you like to see that in this PR?

sdtwigg:
Redirect function definitely makes sense.

Up to you on this PR or new PR.

scala/scala.bzl
```diff
@@ -755,29 +755,26 @@ def scala_repositories():
   )
 
   native.maven_jar(
-    name = "scalac_rules_guava",
```
ittaiz: was this just unused?

johnynek:
Yes.

ittaiz:
@johnynek  LGTM. I agree with @sdtwigg about breaking out `scala_repositories` and it's definitely up to you if it's here or in a new PR

Issue #180, opened 2017-04-19
scala_repl takes common_attrs but never does any compilation


sdtwigg:
scala_repl takes common_attrs, which includes a bunch of stuff like srcs, deps, resources, etc. but only looks at deps and runtime_deps. There is no invocation of _lib or _scala_binary_common to actually compile the sources.

I can think of two paths I can think of depending on what do you think scala_repl should be:
1. scala_repl only takes deps and runtime_deps (where both are treated as runtime deps) as things to load into the resulting REPL environment.
2. scala_repl uses _scala_binary_common to actually do a compile of those sources similar to how scala_binary and scala_test can compile sources.

johnynek:
it might be nice to have some scratch code to do some testing with, so option 2 might be nice. Although, you can also solve the problem with an additional library target.

sdtwigg:
I'll try unifying with scala_binary_common.

It also would avoid having to change common_attrs while still allowing scala_repl to use it.

PR #181, opened 2017-04-20
Unified use of collect_jar and binary_common, other cleanups

sdtwigg:
Many places were directly using file or files on various jars when
building up the classpaths. This was not resilient to later changing
those jars to being true java targets (and thus having runtime
dependencies).

Also, there was significant redundancy in the various scala rules each
analyzing their deps and runtime_deps to build up the compile and
runtime jar lists.

A new function _collect_jars_from_common_ctx has been created to reduce
this redundancy. This function looks at ctx.attr.deps and
ctx.attr.runtime_deps to build up compiletime and runtime jars. It also
assumes an automatic dependency on scala-reflect and scala-library.
(This allows eliminating weirdness where both those jars were being
separately added to the compiletime and runtime classpath.) All these
collections are via _collect_jars, rather than files, and thus should be
resilient to later changing the jar sources to true java targets that
have other runtime dependencies. Note that _collect_jars has a failsafe
to look at files when given a non-java target anyway.

Note that the function has parameters for extra_deps and
extra_runtime_deps. This is used by scala_test and scala_junit_test to
implicitly include testing library dependencies and scala_repl to
implicitly include the necessary scala-compiler dependency (since that
holds the main_class runner).

Finally, scala_repl now uses scala_binary_common rather than setting up
its own scala provider, etc. Beforehand, scala_repl was using
commot_attrs, which stipulate srcs and so forth to compile but the
implementation was never actually invoking the compile. Now, scala_repl
just uses _scala_binary_common like scala_binary, scala_test, etc. Added
a test that a REPL with sources actually compiles and can use them.

This did require 'outputs' and generation of a deploy_jar for scala_repl
(since scala_binary_common references them).

Other cleanups or important notes:
The ClassPath of the written manifest is now blank (previously had the
scala-library in it). The restructuring adds the scala-library to the
rjars before passing them to the deploy_jar builder. Thus, scala-library
should end up placed in the deploy jar anyway.

As example of need to consider transitive dependencies for these files,
scalatest_reporter actually requires scala_xml. However, when added as
just a file, would not end up adding scala_xml. Now, using collect_jar
that xml addition happens automatically.

IMPORTANT NOTE: Also for scala_test, CANNOT directly depend on
_scalatest using _collect_jar. _scalatest points to an http_jar, which
runs ijar and exposes that result as the compiletime jar. However,
scalatest defines and exposes macros and so this falls apart. Ideally,
the scalatest label would be retargeted to something like
scala_macro_library, which exposes the right compiletime jars, runtime
jars and providers. (Reverting back to http_file would work but cause
other issues so the stopgap was to mnaually inject the runtime jars as
the compiletime jars. Since no actual dependencies in current scalatest
import, this ends up looking identical as before.)

Further note for future reference that http_jar runs ijar because it is
roughly equivalent to http_file followed by java_import (and java_import
runs ijar).

Remove the various requirements that single_file = true since using
_collect_jar means don't need reliance on file.

Removed scalacompiler from the compiletime class path for compilation of
a scala target. It was not being added to the runtime classpath so this
could have led to code compiling but then failing to load the proper
classes at runtime.

In thrift.bzl, use executable to refer to jar instead of file.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -149,12 +149,7 @@ def _compile(ctx, _jars, dep_srcjars, buildijar):
     plugins = _collect_plugin_paths(ctx.attr.plugins)
     plugin_arg = ",".join(list(plugins))
 
-    compiler_classpath = '{scalalib}:{scalacompiler}:{scalareflect}:{jars}'.format(  # noqa
-        scalalib=ctx.file._scalalib.path,
-        scalacompiler=ctx.file._scalacompiler.path,
-        scalareflect=ctx.file._scalareflect.path,
-        jars=":".join([j.path for j in jars]),
-    )
+    compiler_classpath = ":".join([j.path for j in jars])
 
     scalac_args = """
 Classpath: {cp}
@@ -225,9 +220,6 @@ SourceJars: {srcjars}
             ctx.executable._ijar,
             ctx.executable._java,
             ctx.executable._javac,
-            ctx.file._scalacompiler,
-            ctx.file._scalareflect,
-            ctx.file._scalalib,
             argfile])
     ctx.action(
         inputs=ins,
@@ -291,7 +283,7 @@ def _build_deployable(ctx, jars):
 
 def write_manifest(ctx):
     # TODO(bazel-team): I don't think this classpath is what you want
-    manifest = "Class-Path: %s\n" % ctx.file._scalalib.path
+    manifest = "Class-Path: \n"
     if getattr(ctx.attr, "main_class", ""):
         manifest += "Main-Class: %s\n" % ctx.attr.main_class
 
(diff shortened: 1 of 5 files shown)


sdtwigg:
Please notice the warning regarding the (somewhat temporary) workaround of _scalatest using http_jar

sdtwigg:
Addresses #180 and #181

sdtwigg:
As a bonus demonstration of the change, scala_test no longer needs to directly depend on scala-xml since that dependency is carried via test_reporter.

Note also the movement of test_reporter to being a runtime dependency.

johnynek:
This is great!

sdtwigg:
I should warn (in case some random target somewhere breaks after this) that this does tighten up some of the dependencies a little (you no longer automatically get the compile-time dependency on scalacompiler all the time, scala_test no longer gets an automatic dependency on scala-xml and the test_reporter, etc.).

So, it is possible previous targets that were not properly declaring their dependencies will need cleanup. However, I assert that this is good in the long run anyway....

johnynek:
Bugs in overbroad deps are always better to fix, if you ask me, so I agree
with you.

johnynek:
by the way, this came in handy for me today. Bumping to scalatest 3.0.1 was as easy as using this with the right bind target to my repos local target for scalatest.

Thanks for cleaning this up!

sdtwigg:
Great to hear other users are also seeing benefits :D

PR #186, opened 2017-04-27
Don't write XML output if the envvar unset

### LINKED ISSUE #185: XML junit reporter crashes if the env var is not set
https://github.com/bazelbuild/rules_scala/blob/master/scala/support/JUnitXmlReporter.scala#L82

we could just skip writing if it is not set, though bazel always sets it. While testing by hand, you might not, and thus you will see an NPE.

johnynek:
closes #185

### DIFF
--- a/scala/support/JUnitXmlReporter.scala
+++ b/scala/support/JUnitXmlReporter.scala
@@ -78,13 +78,14 @@ class JUnitXmlReporter extends Reporter {
   // Writes the xml file for a single test suite.  Removes processed
   // events from the events Set as they are used.
   //
-  private def write(suites: Set[Testsuite]): Unit = {
-    val filespec  = System.getenv.get("XML_OUTPUT_FILE")
-    val out = new PrintWriter(filespec, "UTF-8")
-    val xml    = xmlify(suites.toList.sortBy(_.name))
-    out.print(prettyXml(xml))
-    out.close()
-  }
+  private def write(suites: Set[Testsuite]): Unit =
+    Option(System.getenv.get("XML_OUTPUT_FILE"))
+      .foreach { filespec =>
+        val out = new PrintWriter(filespec, "UTF-8")
+        val xml = xmlify(suites.toList.sortBy(_.name))
+        out.print(prettyXml(xml))
+        out.close()
+      }
 
   //
   // Constructs a Testsuite object corresponding to a specified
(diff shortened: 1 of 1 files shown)


scala/support/JUnitXmlReporter.scala
```diff
@@ -78,13 +78,14 @@ class JUnitXmlReporter extends Reporter {
...
   //
-  private def write(suites: Set[Testsuite]): Unit = {
-    val filespec  = System.getenv.get("XML_OUTPUT_FILE")
-    val out = new PrintWriter(filespec, "UTF-8")
-    val xml    = xmlify(suites.toList.sortBy(_.name))
-    out.print(prettyXml(xml))
-    out.close()
-  }
+  private def write(suites: Set[Testsuite]): Unit =
+    Option(System.getenv.get("XML_OUTPUT_FILE"))
```
smparkes: `scala.util.Properties.envOrNone("XML_OUTPUT_FILE")`?

johnynek: I didn't even know that existed! If you'd really prefer that, I will change it, but all things being equal I'd minimize my labor on it and just merge as is.

johnynek: PS: thanks for reviewing!

smparkes: +1 either way

scala/support/JUnitXmlReporter.scala
```diff
@@ -78,13 +78,14 @@ class JUnitXmlReporter extends Reporter {
...
+      .foreach { filespec =>
+        val out = new PrintWriter(filespec, "UTF-8")
+        val xml = xmlify(suites.toList.sortBy(_.name))
+        out.print(prettyXml(xml))
+        out.close()
```
sdtwigg: As a matter of best-practices pedantry, put the close in a finally in case out.print or prettyXml have an 'issue'?

ittaiz:
I agree with Stephen's remark. Other than that LGTM

johnynek:
looks like the testing of HEAD on travis is broken. I'm not sure why yet, but this is going to fail any new PRs.

cc @ittaiz if you have a chance to look at it.

ittaiz:
Haven't had a chance to look yet but is there a chance it's related to this
bazel issue? https://github.com/bazelbuild/bazel/issues/2819

ittaiz:
Just to update- I took a look and around a week ago Bazel [build](http://ci.bazel.io/job/Bazel/JAVA_VERSION=1.8,PLATFORM_NAME=darwin-x86_64/1458/) started to output another "without-jdk" installer which causes the travis script to fail. The script tries to do some parsing on the ci.bazel.io output to find the download url and that parsing needs to be changed to handle the new case. I'll try to work on it tomorrow but no can't make any promises.

PR #190, opened 2017-05-07
travis hardcoded to bazel--installer.sh for HEAD

ittaiz:
Around a week ago Bazel build started to output another "without-jdk" installer which causes the travis script to fail. The script tries to do some parsing on the ci.bazel.io output to find the download url and that parsing needs to be changed to handle the new case.
It looks like the parsing was done to be more resilient to output names but on the other hand the output name seems fixed so I've hardcoded that name (bazel--installer.sh).
Let's see if travis will pass for this one

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,7 @@ before_install:
       OS=linux
     fi
     if [[ "${V}" == "HEAD" ]]; then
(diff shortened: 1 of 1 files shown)


ittaiz:
@johnynek travis build passes. Any objections to merge?

johnynek:
Thanks for fixing this!

PR #191, opened 2017-05-12
Use new mirror.bazel.build hostname

jart:
CC: @damienmg


johnynek:
seems travis is failing again. We thought we fixed that with #190 but maybe the bazel installer path changed again on head.

jart:
Yes, I don't even bother testing Bazel at HEAD using Travis specifically for that reason. https://github.com/bazelbuild/rules_closure/blob/master/.travis.yml

johnynek:
Thank you!

Issue #192, opened 2017-05-12
add option for comparable to sbt's CompilationOrder


softprops:
I ran into an issue today with a module that contains both scala and java sources when using srcjar that contains java code that scala source files rely on. My gut feeling is it's related to [this code](https://github.com/bazelbuild/rules_scala/blob/aaf8e46942bc67c8485909408cc61b4d3d58d864/src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java#L66-L78). However if I depend on the java generated java sources directly mixed complication seems to work just fine. There's probably a bit more digging I need to do but wanted to report the issue here in case anyone has has ran into the same problem and found a work around

sdtwigg:
Oh, on first glance, it looks like this code isn't passing the java sources to scalac so that scalac can analyze them. Actually, it also isn't clear that the scala jar is available to the java compilation. (I think it might end up emitted to a directly on the classpath so this might not actually be an issue but would be nice if the classpath to javac was amended to clearly include this.)

For others, I found this documentation on sbt Compilation order: http://www.scala-sbt.org/0.13/docs/Java-Sources.html

PS: This situation might be more broadly solved by our other discussions of just using Zinc instead of raw scalac....

softprops:
Yea hot on your tail :) Yea it looks like java sources are added to the [sources](https://github.com/bazelbuild/rules_scala/blob/aaf8e46942bc67c8485909408cc61b4d3d58d864/scala/scala.bzl#L145) make their way into the compile options for `files` which get interpreted as [scalaSources](https://github.com/bazelbuild/rules_scala/blob/aaf8e46942bc67c8485909408cc61b4d3d58d864/src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java#L60) perhaps correctly or maybe not by design. That answers my question about why passing .java sources as is works

softprops:
the use of Zinc would sounds promising but Im going to be looking for a fix sooner than that happens. I can try to come up with a solution on my local fork but wanted to reach out and see if this was  a bug or a feature and if others have bumped into the problem before :)

sdtwigg:
Oh, sorry, I completely misread 'srcjar' in your initial description as 'src'. This explains how your situation is different than the test for mixing scala and java sources:
https://github.com/bazelbuild/rules_scala/blob/master/test/BUILD#L205

[...]

PR #193, opened 2017-05-15
Srcjar mixed compilation

softprops:
This change addresses #192. I've tested it locally against a repository build that was failing without the change and confirmed the change fixes the issue after the change. I've also provided some additions to the test harness to make sure this case has test coverage.


bazel-io:

softprops:
Test failure is an unrelated [test](https://travis-ci.org/bazelbuild/rules_scala/jobs/232243539#L626) I'll look into that. Pointers are welcome

johnynek:
This this please

(Msg to CI bot)

johnynek:
Test this please

softprops:
Fixed that last issue. This time it looks like bazel isn't on the path on one of the test envs? https://travis-ci.org/bazelbuild/rules_scala/jobs/232250662#L104

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
```diff
@@ -58,6 +58,9 @@ public void processRequest(List<String> args) throws Exception {
       List<File> javaJarFiles = filterFilesByExtension(jarFiles, ".java");
 
       String[] scalaSources = GenericWorker.appendToString(ops.files, scalaJarFiles);
+      // include java files from .srcjar's to support mixed compilation
+      scalaSources = GenericWorker.appendToString(scalaSources, javaJarFiles);
```
ittaiz: How about extracting the new line and the existing one to "appendSrcJarSources"? I think inside that context we can live without the comment. Wdyt?

ittaiz:
LGTM in general. I think the Travis failure might be due to corrupted installer download. If you'll make another push we'll see. If not I'll see how I can re trigger it (AFK)

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
softprops: I'm cool with that. I was thinking about introducing a new variable but hesitated to reduce the potential of error for future errors if someone references `scalaSources` when they meant to reference `appendSrcJarSources`.

softprops:
> LGTM in general. I think the Travis failure might be due to corrupted installer download. If you'll make another push we'll see. If not I'll see how I can re trigger it (AFK)

Cool. If its a cached corruption you can always delete the branch cache from travis settings

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
ittaiz: I think a small private method which handles `srcjar`s (both scala and java) would be somewhat better since it can remove the need for the comment.
You can decide to have the new variable inside that method to also strengthen the point about mixed sources. I'll not insist but I think it will be clearer.

softprops:
pushed a new change. looks like quite a bit of travis queue backup?

ittaiz:
@softprops build passed. Can you reply on the two comments?

softprops:
ok. all tests pass now @ittaiz @johnynek let me know what you think

ittaiz:
@softprops I already replied (few times). Can you reply to my two questions?
Thanks!

johnynek:
👍 

I think we can merge after you confirm with @ittaiz that the tests fail if we remove the fix you have (to prevent future regression).

softprops:
replied. I think [this](https://github.com/bazelbuild/rules_scala/pull/193#commitcomment-22142884) was the last open question

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
softprops: I'll add that

softprops: pushed up a few commits to address that

ittaiz:
Thanks!

softprops:
thanks @ittaiz / @johnynek I like the testing setup for this repo. I'm going to steal some of this for some rules my team is work on

softprops:
side question: we're okay with using commit's as checkpoints for now but is there any potential for a more formal release module for skylark rules repos. something to track changes with semantic versioning?

johnynek:
I'm a bit negative on this honestly, @softprops. I think we should keep compatibility as enforced by the tests, and people should choose the git-sha they want.

softprops:
git logs are also honest change logs :)

ittaiz:
Oscar,
Why do you think some kind of semver is a bad idea?

johnynek:
I think it can be valuable for marketing, and I should probably care more
about that. Aside from that, in my experience using MiMa for years, without
tooling humans are terrible at checking API compatibility for semver
communication. Elm has built in tooling for this in the packaging tool. I
think that's the right way to go. But we don't have anything like that here.

In lieu of that, I would suggest making marketing efforts around
significant new features, but trying our best to never break the build
rules we have tests for pretty much ever.

When we do break a rule, it should be to make it more correct and more
consistent with bazel.

I would suggest a new version number when we get support for java sandwich,
which I think is a considerable new feature. Also for scala_toolchain
allowing easily selecting which version of scala to use.

On the horizon, I can see marketing pushes around cross compilation support
(many JVM versions as well as scala native, scalajs) as well as tooling to
improve interop with sbt.

Semver sounds really nice, but without tooling is almost certainly a lie,
and with tooling I've found you have to bump versions much more rapidly
than you think, which destroys any value of semver. I should actually make
a blog post about my thoughts about it. I used to be very pro, now I think
we need much better tech and semver is not much help.

PR #194, opened 2017-05-15
added location expansion with data attribute

ittaiz:
We need $location expansion and that is supported by the java rules.
Not exactly sure why I had to add `ctx.attr.data` explicitly to the `expand_location` since, to me at least, once I add the label in the `data` attribute I think that should be a valid dependency but it's apparently not strict enough for bazel for some reason. I added this [SO](http://stackoverflow.com/questions/43982762/location-expansion-in-bazel) question to try and gather more details on why this works like it does.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -297,6 +297,10 @@ def _write_launcher(ctx, rjars, main_class, jvm_flags, args, run_before_binary, 
           repo = ctx.workspace_name, spath = f.short_path
       ) for f in rjars])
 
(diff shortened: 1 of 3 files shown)


johnynek:
I think the docs on it supports the way you did it:
https://bazel.build/versions/master/docs/skylark/lib/ctx.html#expand_location

👍

ittaiz:
Really? I read and reread that link but didn't find any corroborating evidence to "data" being special. I'll merge and update afterwards if we seem to be wrong.

Issue #138, opened 2017-02-08
Resources not included in jar if srcs is empty


seanmcl:
The following repo demonstrates a bug in `scala_library`

https://github.com/seanmcl/scala-bazel-bug

I expected the resources files to be included without srcs requiring a non-empty list.

johnynek:
Yeah, filegroup is not supported right now. See the test for how this works:

https://github.com/bazelbuild/rules_scala/blob/master/test/BUILD#L163

the resources should be individual files.

would welcome a PR to add support for filegroup resources.

or-shachar:
@johnynek  - It's not only filegroups. Also just simple:
```
scala_library(
    name = "resources",
    resources = ["a.txt"],
)
```
Does not copy the `a.txt` to the jar.

or-shachar:
fixed and merged

ittaiz:
@seanmcl please reopen if you think this wasn't resolved

PR #199, opened 2017-05-21
fixing #138 | missing resources on no_src scala_library

or-shachar:
Added tests, made sure it works both on direct files and filegroups

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -76,28 +76,30 @@ def _get_jar_path(paths):
 
 
 def _build_nosrc_jar(ctx, buildijar):
-    cp_resources = _add_resources_cmd(ctx, "{out}_tmp".format(
-      out=ctx.outputs.jar.path)
-    )
+    temp_resources_dir="{jar_output}_temp_resources_dir".format(jar_output=ctx.outputs.jar.path)
+    cp_resources = _add_resources_cmd(ctx, temp_resources_dir)
     ijar_cmd = ""
     if buildijar:
-        ijar_cmd = "\ncp {out} {ijar_out}".format(
-          out=ctx.outputs.jar.path,
-          ijar_out=ctx.outputs.ijar.path)
+        ijar_cmd = "\ncp {jar_output} {ijar_output}".format(
+          jar_output=ctx.outputs.jar.path,
+          ijar_output=ctx.outputs.ijar.path)
     cmd = """
-  rm -rf {out}_tmp
+  rm -rf {temp_resources_dir}
   set -e
-  mkdir -p {out}_tmp
+  mkdir -p {temp_resources_dir}
   # copy any resources
   {cp_resources}
-  {java} -jar {jar} -m {manifest} {out} {out}_tmp
+  # adding {temp_resources_dir} as last argument will copy its content into the jar output
+  {java} -jar {jarBuilder} -m {manifest} {jar_output} {temp_resources_dir}
   """ + ijar_cmd
     cmd = cmd.format(
+        temp_resources_dir=temp_resources_dir,
         cp_resources=cp_resources,
-        out=ctx.outputs.jar.path,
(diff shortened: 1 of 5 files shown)


bazel-io:

ittaiz:
test this please (for CI)

test/src/main/scala/scala/test/resources/BUILD
```diff
@@ -0,0 +1,34 @@
+
+load("//scala:scala.bzl", "scala_library")
+
+# Test of resources with srcs
```
ittaiz: please fix indentation in the whole file

test_run.sh
```diff
@@ -186,6 +186,43 @@ test_junit_test_errors_when_no_tests_found() {
...
 }
 
+test_resources() {
+  RESOURCE_NAME="resource.txt"
+  OUTPUT_JAR="bazel-bin/test/src/main/scala/scala/test/resources/$TARGET.jar"
+  FULL_TARGET="test/src/main/scala/scala/test/resources/$TARGET.jar"
+  bazel build $FULL_TARGET
+  jar tf $OUTPUT_JAR | grep $RESOURCE_NAME
+  RET=$?
+  if [ $RET -ne 0 ]; then
```
ittaiz: can't we reuse the existing failure mechanism?

test_run.sh
```diff
@@ -186,6 +186,43 @@ test_junit_test_errors_when_no_tests_found() {
...
+    echo "Missing ${RESOURCE_NAME} in jar based on $JAR_DESCRIPTION"
+    exit 1
+  fi
+}
+
+scala_library_jar_with_srcs_includes_direct_file_resources(){
```
ittaiz: do we need the "with srcs" cases?

scala/scala.bzl
```diff
@@ -90,7 +90,7 @@ def _build_nosrc_jar(ctx, buildijar):
   mkdir -p {out}_tmp
   # copy any resources
   {cp_resources}
-  {java} -jar {jar} -m {manifest} {out}
+  {java} -jar {jar} -m {manifest} {out} {out}_tmp
```
ittaiz: what does this do? I think the name isn't clear enough. Maybe also add a comment

test_run.sh
or-shachar: You are right, we don't it's already tested with `test:RunMoreScalaLibBinary`. Removing it.

scala/scala.bzl
```diff
@@ -76,28 +76,30 @@ def _get_jar_path(paths):
...
-  rm -rf {out}_tmp
+  rm -rf {temp_resources_dir}
   set -e
-  mkdir -p {out}_tmp
+  mkdir -p {temp_resources_dir}
   # copy any resources
   {cp_resources}
-  {java} -jar {jar} -m {manifest} {out}
+  # adding {temp_resources_dir} as last argument will copy its content into the jar output
+  {java} -jar {jarBuilder} -m {manifest} {jar_output} {temp_resources_dir}
```
ittaiz: looks much better! I'd really like to remove the comment above. @johnynek wdyt about introducing a `--additional-folders` flag (or something of the sort)? I think this could give very similar value to the comment but be more resilient. @or-shachar of course not in this PR.

ittaiz:
test this please (for ci)

scala/scala.bzl
or-shachar: I agree. It's a bit problematic that we have multiple unnamed arguments with different roles.

sdtwigg: What would be the plusses and minuses of moving the nosrc jar building logic into the scalac worker? The action already has all the infrastructure setup for passing and parsing the arguments.

johnynek: Yeah it might be nice to unify into the worker and have it sort out the no-source case separately.

PR #201, opened 2017-05-22
test discovery filters abstract classes and traits

ittaiz:
junit and specs2 fail with when trying to run abstract classes and traits.
We now filter those out.

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -328,3 +328,11 @@ scala_specs2_junit_test(
     data = ["//src/java/com/google/devtools/build/lib:worker"],
     jvm_flags = ["-Dlocation.expanded=$(location //src/java/com/google/devtools/build/lib:worker)"],
 )
+
+scala_junit_test(
+    name = "JunitFiltersAbstractClassesAndInterfaces",
+    srcs = ["src/main/scala/scala/test/junit/JunitAbstractClassAndInterface.scala"],
+    size = "small",
+    suffixes = ["Test"],
+    print_discovered_classes = True
+)
--- a/test/src/main/scala/scala/test/junit/JunitAbstractClassAndInterface.scala
+++ b/test/src/main/scala/scala/test/junit/JunitAbstractClassAndInterface.scala
@@ -1,0 +1,21 @@
+package scala.test.junit
+
+import org.junit.Test
+
+abstract class SomeAbstractTest {
+  @Test
+  def abstractTest: Unit =
+  	println("abstract")
+} 
+
+trait SomeTraitTest {
+  @Test
+  def traitTest: Unit =
+  	println("trait")
+} 
+
+class SingleTestSoTargetWillNotFailDueToNoTestsTest {
+  @Test
+  def someTest: Unit =
+  	println("passing")
+} 
--- a/src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
+++ b/src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
@@ -7,6 +7,7 @@ import java.io.File
 import java.io.FileInputStream
 import java.util.jar.JarInputStream
 import java.util.jar.JarEntry
+import java.lang.reflect.Modifier
 /*
   The test running and discovery mechanism works in the following manner:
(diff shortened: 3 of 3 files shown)


ittaiz:
@or-shachar can you review this?

test/BUILD
```diff
@@ -328,3 +328,11 @@ scala_specs2_junit_test(
...
     jvm_flags = ["-Dlocation.expanded=$(location //src/java/com/google/devtools/build/lib:worker)"],
 )
+
+scala_junit_test(
+    name = "JunitFiltersAbstractClassesAndInterfaces",
+    srcs = ["src/main/scala/scala/test/junit/JunitAbstractClassAndInterface.scala"],
+    size = "small",
+    suffixes = ["Test"],
+    print_discovered_classes = True
+)
```
or-shachar: @ittaiz just to understand better - without your addition to the `DiscoveredTestSuite.scala` that test would cause exception, right? 
It might not be clear that this is what we're testing here. How can we explain that the runner **must filter out abstract and interfaces**? Maybe a comment?

test/src/main/scala/scala/test/junit/JunitAbstractClassAndInterface.scala
```diff
@@ -0,0 +1,21 @@
...
+  	println("abstract")
+} 
+
+trait SomeTraitTest {
+  @Test
+  def traitTest: Unit =
+  	println("trait")
+} 
+
+class SingleTestSoTargetWillNotFailDueToNoTestsTest {
```
or-shachar: And what about class that implements the trait or extends the abstract? Does it work the same?

test/BUILD
ittaiz: The test indeed failed before my change. It failed since JUnit (and specs2) fail when trying to instantiate traits and abstract classes via reflection.
To me targets in this package are somewhere between a unit test and a full blown e2e (which resides in `run_test.sh`).
If I can add a test here and not in `run_test.sh` I'd rather do it since the feedback cycle is much faster.

test/src/main/scala/scala/test/junit/JunitAbstractClassAndInterface.scala
ittaiz: They are not abstract and so they will just work

test/BUILD
or-shachar: yes. makes sense. Thanks for the explanation!

ittaiz:
@johnynek @or-shachar I'm not sure this was the right call.
I'm using this and my new PR #206 but I'm facing a new issue of failing to run regular classes which match the affix but don't have @RunWith or @Test methods.
I combed and debugged through the maven surefire codebase and found [JUnit4TestChecker](https://github.com/apache/maven-surefire/blob/surefire-2.19.1/surefire-providers/common-junit4/src/main/java/org/apache/maven/surefire/common/junit4/JUnit4TestChecker.java) and [JUnit48TestChecker](https://github.com/apache/maven-surefire/blob/surefire-2.19.1/surefire-providers/common-junit48/src/main/java/org/apache/maven/surefire/common/junit48/JUnit48TestChecker.java)
which add additional logic like verifying the class has either an @RunWith annotation on the class, an @Test annotated method or @RunWith with @Enclosed.
I can add this logic by myself but I'm deliberating if the correct behavior wouldn't be to use this code itself.
WDYT?

johnynek:
honestly, I am trusting you on this and just looking for style and correctness, but not usability issues.

We use scalatest and have no plans to change. I think a consumer is best able to discuss the usability.

That said, scalatest does discovery without us having to do anything. I think that is the right way to go if you possibly can, but it is up to you (or anyone else) to send a PR since I am way too behind to have time to work on it myself.

ittaiz:
I understand.
I actually think the best course of action would be to separate the test
discovery part out of rules scala to its separate rule.
Leave the implementation in the rules_scala with just a basic JUnit Suite
and have the more elaborate discovery suite outside since the changes are
not bound to scala.
I'm waiting for a runtime_location function from @ulfjack and I'll have the
machinery I need to split this out.
Meanwhile I'll take a stab at depending on the surefire plugin as an
internal dependency.

PR #202, opened 2017-05-22
Use Java launcher script. Fixes #198.

### LINKED ISSUE #198: Launcher script doesn't allow passing JVM flags on command line
The [launcher script for scala_binary](https://github.com/bazelbuild/rules_scala/blob/2d49d201b1756cb06bf188ecc8d3e2275556f6ce/scala/scala.bzl#L304) is pretty simple. There is no way to pass JVM flags via invocation, so it's difficult for an IDE to launch the binary with a debug port and attach to it.

See the (pretty complicated) [Java launcher script](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt) for reference. It has flags to launch the binary for debugging and arbitrary JVM flags on the command line.

chaoren:


bazel-io:

chaoren:
@johnynek @sdtwigg It is done. Please take a look.

scala/scala.bzl
```diff
@@ -293,52 +293,47 @@ def write_manifest(ctx):
...
+    ])
+    javabin = "/".join([
+        "${JAVA_RUNFILES}", ctx.workspace_name, ctx.executable._java.short_path
+    ])
+    template = ctx.attr._java_stub_template.files.to_list()[0]
+
+    command = " ".join([
+        "sed",
+        "\n".join([
+            # Insert the preamble at the first blank line
```
sdtwigg: Can you add a comment that it has to be after #! /bin/bash ? (and that would implicitly explain why less exotic prepend solutions don't quite work)

It is a subtlety I missed upon first review.

scala/scala.bzl
```diff
@@ -579,12 +573,13 @@ def _scala_junit_test_impl(ctx):
...
         jvm_flags = launcherJvmFlags + ctx.attr.jvm_flags,
         args = test_suite.suite_class,
-        run_before_binary = "",
-        run_after_binary = "",
     )
 
     return _scala_binary_common(ctx, cjars, rjars)
 
+_launcher_template = {
+  "_java_stub_template": attr.label(default=Label("@java_stub_template//file")),
```
sdtwigg: This is more of a question to you/Oscar. Looking at commit: https://github.com/bazelbuild/rules_scala/commit/0e4ff2d4e4d1d84f7c42a188ae75ded45cdea545

I don't have a good sense of when to use //external versus @... since other parts of file use @bazel_tools (although I presume because those aren't actually setup in the WORKSPACE and assumed to be premapped).

Also, test/BUILD still references @scala//:scala-library and I wasn't sure if that was an oversight or intentional demonstration that rules_scala 'clients' should still be using @scala.

scala/scala.bzl
```diff
@@ -484,37 +479,38 @@ def _scala_binary_impl(ctx):
...
-  args = " ".join(ctx.attr.scalacopts)
   _write_launcher(
       ctx = ctx,
       rjars = rjars,
       main_class = "scala.tools.nsc.MainGenericRunner",
       jvm_flags = ["-Dscala.usejavacp=true"] + ctx.attr.jvm_flags,
-      args = args,
-      run_before_binary = """
+      args = " ".join(ctx.attr.scalacopts),
+      preamble = """
```
sdtwigg: As discussed, add reminder for future maintainers to 'watch out' for the java launcher using trap as well.

sdtwigg:
LGTM

scala/scala.bzl
chaoren: Done.

scala/scala.bzl
chaoren: Done.

ittaiz:
@chaoren the Travis build is broken. Skimmed over the log from my phone and it seems like the setup of scala library is not right.

sdtwigg:
First discrepancy that comes up is that the regular java rules will put double quotes around the CLASSPATH. I changed it to:
raw_classpath = [what you have now for classpath]
classpath = "\"%s\"" % raw_classpath

However, it still seems to have issues.

What is odd is that locally all the tests pass but the same tut rules return odd and different errors involving the SymbolTable:
Exception in thread "main" java.lang.NullPointerException
        at scala.reflect.internal.SymbolTable.exitingPhase(SymbolTable.scala:256)
        at scala.tools.nsc.interpreter.IMain$Request.x$20$lzycompute(IMain.scala:903)

sdtwigg:
I just noticed that the java_stub_template is using exec. This will replace the shell with java. Thus, the trap being setup to restore stty will also be blown away.

chaoren:
@ittaiz Thanks for the heads up, I'll look into it.
@sdtwigg Fixed. Used a wrapper forJAVABIN like we discussed.

johnynek:
test this please

chaoren:
The failure looks like we exposed a tut bug: https://github.com/tpolecat/tut/issues/85
I don't know if I want to invest any more time fixing this.

scala/scala.bzl
johnynek: I think `@java_stub...` is the right way to go here.

Also, I think it is okay that `test/BUILD` uses `@scala//:` since we aren't actually depending on tests targets remotely.

scala/scala.bzl
```diff
@@ -661,9 +649,9 @@ scala_test = rule(
   attrs={
       "main_class": attr.string(default="org.scalatest.tools.Runner"),
       "suites": attr.string_list(),
-      "_scalatest": attr.label(default=Label('//external:io_bazel_rules_scala/dependency/scalatest/scalatest'), allow_files=True),
+      "_scalatest": attr.label(default=Label("//external:io_bazel_rules_scala/dependency/scalatest/scalatest"), allow_files=True),
```
johnynek: why the change to `"` from `'`? Seems like a minor stylistic thing that is adding a lot of lines to the diff.

scala/scala.bzl
```diff
@@ -756,19 +744,30 @@ def scala_repositories():
...
+
+  native.bind(name = "io_bazel_rules_scala/dependency/com_google_protobuf/protobuf_java", actual = "@scalac_rules_protobuf_java//jar")
 
-  native.bind(name = 'io_bazel_rules_scala/dependency/scala/parser_combinators', actual = '@scala//:scala-parser-combinators')
+  native.bind(name = "io_bazel_rules_scala/dependency/scala/parser_combinators", actual = "@scala//:scala-parser-combinators")
 
-  native.bind(name = 'io_bazel_rules_scala/dependency/scala/scala_compiler', actual = '@scala//:scala-compiler')
+  native.bind(name = "io_bazel_rules_scala/dependency/scala/scala_compiler", actual = "@scala//:scala-compiler")
 
-  native.bind(name = 'io_bazel_rules_scala/dependency/scala/scala_library', actual = '@scala//:scala-library')
```
johnynek: don't see why we needed to make these changes.

johnynek:
The tut rule is really simple:
https://github.com/bazelbuild/rules_scala/blob/master/tut_rule/tut.bzl#L17

My fear is that actually this script has changed the classpath seen by targets and so perhaps the `scala-compiler` is no longer on the classpath? I'd like to get tut working at a minimum.

It may be that tut should have had an explicit dependency on the compiler and this exposes a real issue, or it may be that the script will break others as well and our tests are weak and only doing very naive compilations/runs.

scala/scala.bzl
chaoren: Yes, minor stylistic thing, but it also makes some internal processes easier within Google.

sdtwigg:
I was getting the same failures when I just changed the original script from export CLASSPATH={classpath} to -classpath {classpath}.

Very unclear on the why though. I'm tempted to submit a PR with JUST that change to see if travis and CI fail with the same error as well.

scala/scala.bzl
sdtwigg: It was only this one. The _scalatest was the only label that used double quotes so it originally exposed an issue where our repo conversion script wasn't finding this string. Had put it in here originally so could do some internal tests but can do that in a separate PR as well.

chaoren:
> My fear is that actually this script has changed the classpath seen by targets and so perhaps the scala-compiler is no longer on the classpath? I'd like to get tut working at a minimum.

If that were the case, other scala rules shouldn't compile either.

johnynek:
@chaoren tut needs the compiler at runtime, generally scala builds don't need that (only at compile time).

johnynek:
well, we are depending on the compiler: https://github.com/bazelbuild/rules_scala/blob/master/src/scala/io/bazel/rules_scala/tut_support/BUILD#L8

but maybe we need `scala-reflect` at runtime?

chaoren:
@johnynek This is the classpath the tut compiler (`//test/tut/some_example_compiler`) is running with:
```
${TEST_SRCDIR}/io_bazel_rules_scala/../scala/lib/scala-library.jar:
${TEST_SRCDIR}/io_bazel_rules_scala/../scala/lib/scala-reflect.jar:
${TEST_SRCDIR}/io_bazel_rules_scala/test/tut/test_dep.jar:
${TEST_SRCDIR}/io_bazel_rules_scala/../scala/lib/scala-compiler.jar:
${TEST_SRCDIR}/io_bazel_rules_scala/../io_bazel_rules_scala_org_tpolecat_tut_core/jar/tut-core_2.11-0.4.8.jar:
${TEST_SRCDIR}/io_bazel_rules_scala/src/java/io/bazel/rulesscala/io_utils/libio_utils.jar:
${TEST_SRCDIR}/io_bazel_rules_scala/src/scala/io/bazel/rules_scala/tut_support/tut_compiler_lib.jar:
${TEST_SRCDIR}/io_bazel_rules_scala/test/tut/some_example_compiler.jar
```
It does include reflect and the compiler.

sdtwigg:
I pushed a PR at #203 that only changes from export CLASSPATH to -classpath (which, theoretically, is a more thoroughly tested path since that is what the native java rules are doing).

At least locally, I was seeing some divergences in behavior with tut and was curious if anyone else had insight.

johnynek:
I wonder if just bumping the tut version to something more recent will fix things?

chaoren:
Also @sdtwigg discovered that the issue only happens when the classpath is passed with the `-classpath` argument instead of using the `CLASSPATH` environment variable. The people on the tut bug (https://github.com/tpolecat/tut/issues/85) discovered that forking triggers the issue as well. I'm inclined to think that the issue might be improper propagation of classpath.

chaoren:
I tried the latest version: 0.5.2. Doesn't work either.

johnynek:
I wonder if the issue is that setting `CLASSPATH` means it is set when
`tut` calls the compiler, but when we do `-classpath` tut has no way of
reading it back and somehow it's invocations of scalac don't have the
classpath? Totally stabbing in the dark here.

johnynek:
https://gitter.im/tpolecat/tut?at=5924a22cfcbbe1891c3a4450

where I asked a question.

sdtwigg:
That would make sense:
Is tut calling the compiler by instantiating a compiler object in the same JVM and running it? If so, then it shouldn't matter.

If tut calling the compiler by making a subprocess? If so, then can definitely see how they might forget to propagate the classpath correctly and we were just getting lucky before due to env parenting.

chaoren:
Alright, I added a hack to bypass the bug. This should pass now.

johnynek:
test this please

(dearest cibot).

chaoren:
ERROR: /home/travis/build/bazelbuild/rules_scala/tut_rule/tut.bzl:15:3: no such package '@io_bazel_rules_scala_org_tpolecat_tut_core//jar': Failed to fetch Maven dependency: Could not transfer artifact org.tpolecat:tut-core_2.11:jar:0.4.8 from/to io_bazel_rules_scala_org_tpolecat_tut_core (https://dl.bintray.com/tpolecat/maven/): NullPointerException and referenced by '//external:io_bazel_rules_scala/dependency/tut/tut_core'.
ERROR: Analysis of target '//test/tut:some_example__tut' failed; build aborted.

:rage:

tpolecat:
@johnynek it looks like you need to add scala-reflect to the `-classpath` argument you pass to `TutMain`. That should get you over the immediate hump. I'll try to fix this for tut 0.5.3


    tut-test$ tree
    .
    └── in
        └── test.md

    1 directory, 1 file

    tut-test$ coursier launch -r "https://dl.bintray.com/tpolecat/maven/" org.tpolecat:tut-core_2.11:0.5.2 -- in out '.*\.md$'
    [tut] compiling: in/test.md
    Exception in thread "main" java.lang.NullPointerException
    	at scala.reflect.internal.SymbolTable.exitingPhase(SymbolTable.scala:256)
    	at scala.tools.nsc.interpreter.IMain$Request.x$20$lzycompute(IMain.scala:896)
            ...

    tut-test$ coursier launch -r "https://dl.bintray.com/tpolecat/maven/" org.tpolecat:tut-core_2.11:0.5.2 -- in out '.*\.md$' -classpath $(coursier fetch -p org.scala-lang:scala-reflect:2.11.8)
    [tut] compiling: in/test.md

    tut-test$ cat out/test.md 
    ```scala
    scala> 1 + 1
    res0: Int = 2
    ```

chaoren:
@tpolecat That is not the problem. See [this comment](https://github.com/bazelbuild/rules_scala/pull/202#issuecomment-303527838) for the classpath given to the tut compiler. It does include scala-reflect. See [this hack](https://github.com/bazelbuild/rules_scala/pull/202/commits/b0622d6ae69e29d433ede3c43721d3a983aac95b) that we're using to bypass the issue. It just puts everything passed through the `-classpath` argument into the `CLASSPATH` environment variable.

johnynek:
@chaoren I think he is saying that TutMain also has a -classpath argument. Is that right @tpolecat?

chaoren:
> TutMain also has a -classpath argument.

Do you mean [this thing](https://github.com/tpolecat/tut/blob/master/core/src/main/scala/tut/TutMain.scala#L8) takes a classpath argument? That doesn't make any sense. The JVM already has the classpath.

johnynek:
@chaoren I'm guessing but it isn't that weird. Tut should have two
classpaths: the one it needs and the one the code it is compiling needs.

tpolecat:
The arguments to `TutMain` are

`<input-path> <output-dir> <pattern> [<scalac-arg> ...]`

where any provided args are used to construct the `Settings` used to construct the `IMain` that interprets your code blocks. The REPL has its own classloader and its own Scala runtime settings, independent of the caller's, which makes sense. You're running an embedded environment.

Application/library classes on the caller's classloader are available by delegation, so it is not necessary to pass the classpath again if you're running in-process, as when called from sbt. But if you're running out of process it makes more sense to pass the classpath as an argument to `TutMain`. For some reason the `scala-reflect` library seems to be twitchy about how it's specified. I don't know why. This is what I need to investigate. But I think you can solve your problem by passing everything as an argument.

chaoren:
Cool fixed.

johnynek:
test this please

(kind ci-bot).

Also thanks @tpolecat for looking into a solution for us, and thanks @chaoren for pushing this all the way through.

Let's merge when green.

chaoren:
test this please (does it work when I say it?)

johnynek:
test this please

johnynek:
Thanks again all!

PR #203, opened 2017-05-23
Shift from using CLASSPATH env variable to arg

sdtwigg:
This changes the launcher script from using export CLASSPATH to using
-classpath argument to the java executable. In theory, this should not
cause any differences but seems to break only tut for some reason.


johnynek:
https://gitter.im/tpolecat/tut?at=5924a22cfcbbe1891c3a4450

johnynek:
I think we can close due to #202

PR #207, opened 2017-05-28
added a test to demonstrate large classpath is broken

ittaiz:
It seems that #202 broke the ability to run large classpaths (450 deps on my machine).
I recreated the same with `java_binary` but the limit there was much higher (1755 deps).
@johnynek I'm thinking of reverting #202 for now, wdyt?

I'll wait to see this fails in travis but just as FYI when I reverted locally this test starts to pass
cc @chaoren @sdtwigg


sdtwigg:
Does it still break if you revert 202 but use 203?

I wonder if the issue is just -classpath not working quite right. Do you know if your example is triggering the 120k classpath length check and shifting to using posix substitution?

ittaiz:
passes with #203 manually applied.

chaoren:
> Do you know if your example is triggering the 120k classpath length check and shifting to using posix substitution?

Yes, it is. As far as I can tell the resulting command line is just:
`$java -classpath @<(echo $classpath) $main_class`

ittaiz:
Does this mean we're reading from a file?

chaoren:
It's from this line: https://github.com/bazelbuild/bazel/blob/8598824a26434167108a5a4925a971b498af109f/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt#L308

The [`<(...)`](http://tldp.org/LDP/abs/html/process-sub.html) syntax writes the output to a temporary file (e.g., /dev/fd/56).
And it *looks* like the `-classpath @file` syntax is supposed to tell the java binary to use the contents of the file as classpath, but it doesn't seem to be working when I try it.

test/src/main/scala/scala/test/large_classpath/BUILD
```diff
@@ -0,0 +1,910 @@
+load("//scala:scala.bzl", "scala_binary", "scala_library")
+scala_binary(
+    main_class = "scala.test.large_classpath.ObjectWithLargeClasspath",
+    name = "largeClasspath",
+    srcs = ["ObjectWithLargeClasspath.scala"],
+    deps = [
+        ":empty1",
```
chaoren: Please make a helper .bzl file to return both this list of targets, and create the list of libraries below.

test/src/main/scala/scala/test/large_classpath/BUILD
```diff
@@ -0,0 +1,910 @@
...
+        ":empty443",
+        ":empty444",
+        ":empty445",
+        ":empty446",
+        ":empty447",
+        ":empty448",
+        ":empty449",
+        ":empty450",
+    ],
+
```
chaoren: Extra white space.

chaoren:
Okay, so the difference between the java target and scala target is that their classpaths are all relative paths, while our classpaths are full paths, so they can take more targets before hitting the length limit.
However, the bigger problem is that it seems the solution for really large classpaths (the `@file` syntax) doesn't actually work, even for java targets.

I propose we "fix" this on our end by changing classpaths to use relative paths, then file a bug for the stub template.

chaoren:
Filed https://github.com/bazelbuild/bazel/issues/3069.

sdtwigg:
Thanks for looking into this @chaoren !

ittaiz:
Definitely! Sorry for only coming up with the repro, I'm just trying to put
off some other fires...

PR #208, opened 2017-05-28
Use shortened RUNPATH for classpaths. Temporary fix for #207.

chaoren:


johnynek:
Can we merge a test like #207 in and show that this fixes? I like to, if possible have some test that breaks without each improvement.

chaoren:
Sure the test @ittaiz is adding would be fixed, I'll just move that test here.

chaoren:
Done. This change now replaces https://github.com/bazelbuild/rules_scala/pull/207. @ittaiz please abandon that change.

chaoren:
Though, an even longer classpath would still fail because of https://github.com/bazelbuild/bazel/issues/3069. After that one is fixed, we can just add a target with like 10,000 dependencies to test.

johnynek:
Test this please

johnynek:
:+1:

chaoren:
Looks like some of the test runs timed out. I reduced the number of targets and made the target names longer to cut down on the number of things we need to build.

PR #209, opened 2017-05-29
test discovery enhancements

ittaiz:
test discovery filters classes which match affix without tests, includes classes with @RunWith and test methods in hierarchy

### DIFF
--- a/src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
+++ b/src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
@@ -159,6 +159,6 @@ object PrefixSuffixTestDiscoveringSuite {
       }
     }
 
-  private lazy val runWithAnnotation = classOf[RunWith]
-  private lazy val testAnnotation = classOf[Test]  
+  private val runWithAnnotation = classOf[RunWith]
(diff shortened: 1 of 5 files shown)


ittaiz:
@or-shachar I'd appreciate your review

src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
```diff
@@ -134,4 +138,27 @@ object PrefixSuffixTestDiscoveringSuite {
...
+    }
+
+  private def hasTestAnnotatedMethodsInCurrentClass(testClass: Class[_]): Boolean =
+    testClass.getDeclaredMethods.exists { method =>
+      method.getAnnotations.exists { annotation: Annotation =>
+        testAnnotation.isAssignableFrom(annotation.annotationType)
+      }
+    }
+
+  private lazy val runWithAnnotation = classOf[RunWith]
```
johnynek: why are these lazy? Seems like `val` is fine to me.

ittaiz: We only need it if we get to that phase. I had a different thought when I started but now it's probably just an over optimization. I'll drop it.

johnynek:
👍

Issue #188, opened 2017-05-04
unclear or missing support for `bazel test --test_filter`


softprops:
It doesn't outwardly appear as though the [test_filter](https://bazel.build/versions/master/docs/bazel-user-manual.html#flag--test_filter) option is supported in `scala_test` rules, though it would be very useful! I was looking for an analog to sbt's `testOnly` filter. If this is not implemented yet, I'd be happy to add support with some direction on where the idiomatic place to do this is. 

I dug a bit through bazel core and found it applies this filter option [here](https://github.com/bazelbuild/bazel/blob/25b952b8fec4a3e514b4f91fbbd5e5133fcab4b7/src/java_tools/junitrunner/java/com/google/testing/junit/runner/junit4/JUnit4Runner.java#L96-L97). Its threaded through a number of factories which seem at odds with how simple the skylark extensions are set up.

softprops:
Taking a closer peek it seems these could be supported by appending to these [args](https://github.com/bazelbuild/rules_scala/blob/f44c5f48ebd1f44843ec39dfe1b3cf0cbd2b574f/scala/scala.bzl#L547-L552) the `-n` and `-l` flags noted in the scala test [RunnerDocs](http://doc.scalatest.org/1.7/org/scalatest/tools/Runner$.html). Im not sure if there's a more straightforward way to get access to the parsed command line other than referencing [ctx.arguments](https://bazel.build/versions/master/docs/skylark/lib/ctx.html#parameters). This looks totally doable, happy to submit a PR if this seems like the right path to go down

johnynek:
@softprops one issue with bazel is that skylark is new and bolted on to the old system. As far as I know almost none of the bazel CLI options are visible to skylark rules.

The link you gave to action's arguments I *think* are arguments to pass to the action which might be a bash script or some code you control. I would love to be wrong thought and be able to see the bazel arguments somehow.

For now, I just make small test targets. We have a scala_test_suite macro that makes this pretty easy to do (one target per file).

softprops:
I'll try prototyping something as POC soon. The test suite thing may be a nice alternative. I hadn't considered that. Regardless I just wanted to tug on a few shirts in case this was supported. I'm happy to try and make it happen :)

ittaiz:
@softprops I actually thought about the same issue today (though it's a bit farther down the line for me). I was thinking about it with respect to running tests from the IDE which at least now enable us to run a specific example of specs2 which is very convenient many times.

[...]

PR #216, opened 2017-06-05
make scala_test honor --test_filter flag.

softprops:
This should address #188, enabling what sbt users are used to with it's `testOnly` feature, with the `--test-filter` option of bazel with `scala_test` rules

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -517,12 +517,12 @@ def _scala_test_impl(ctx):
           "suites attribute is deprecated. All scalatest test suites are run"
         )
     jars = _collect_jars_from_common_ctx(ctx,
-        extra_runtime_deps = [ctx.attr._scalatest_reporter],
+        extra_runtime_deps = [ctx.attr._scalatest_reporter, ctx.attr._scalatest_runner],
     )
     (cjars, rjars) = (jars.compiletime, jars.runtime)
     # _scalatest is an http_jar, so its compile jar is run through ijar
     # however, contains macros, so need to handle separately
-    scalatest_jars = _collect_jars([ctx.attr._scalatest, ctx.attr._scalatest_runner]).runtime
+    scalatest_jars = _collect_jars([ctx.attr._scalatest]).runtime
     cjars += scalatest_jars
     rjars += scalatest_jars
 
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -522,7 +522,7 @@ def _scala_test_impl(ctx):
     (cjars, rjars) = (jars.compiletime, jars.runtime)
     # _scalatest is an http_jar, so its compile jar is run through ijar
     # however, contains macros, so need to handle separately
-    scalatest_jars = _collect_jars([ctx.attr._scalatest]).runtime
+    scalatest_jars = _collect_jars([ctx.attr._scalatest, ctx.attr._scalatest_runner]).runtime
     cjars += scalatest_jars
     rjars += scalatest_jars
 
@@ -649,9 +649,10 @@ scala_binary = rule(
 scala_test = rule(
   implementation=_scala_test_impl,
   attrs={
-      "main_class": attr.string(default="org.scalatest.tools.Runner"),
+      "main_class": attr.string(default="io.bazel.rulesscala.scala_test.Runner"),
       "suites": attr.string_list(),
       "_scalatest": attr.label(default=Label("//external:io_bazel_rules_scala/dependency/scalatest/scalatest"), allow_files=True),
+      "_scalatest_runner": attr.label(executable=True, cfg="host", default=Label("//src/java/io/bazel/rulesscala/scala_test:runner.jar"), allow_files=True),
       "_scalatest_reporter": attr.label(default=Label("//scala/support:test_reporter")),
       } + _launcher_template + _implicit_deps + _common_attrs,
   outputs={
--- a/src/java/io/bazel/rulesscala/scala_test/BUILD
+++ b/src/java/io/bazel/rulesscala/scala_test/BUILD
@@ -1,9 +1,8 @@
-
 java_binary(
-    name="runner",
+    name = "runner",
     srcs = ["Runner.java"],
+    visibility = ["//visibility:public"],
     deps = [
-         "//external:io_bazel_rules_scala/dependency/scalatest/scalatest"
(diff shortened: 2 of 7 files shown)


bazel-io:

softprops:
reviewer notes: there are actually quite a few filtering options that that the [scala test runner supports](http://www.scalatest.org/user_guide/using_the_runner) I picked the one that most closely aligned with the type of filtering scala test users are used to with sbt. This covers filtering on the package and test suite name, ie. `com.company.foo.*` and `com.company.foo.bar.BazTest` ect. It does not filter on the per method level. Method level filtering [can be accomplished](https://stackoverflow.com/a/22490924/547339) by providing additional `--test_args` with the -z flag of the scala test runner. I was able to accomplish that with something like the following

```bash
bazel test --test_arg=-z --test-arg=single-test-pattern //...
```

Special note here that `--test_arg=-z=single-test-pattern` didn't seem to work as advertised [here](https://bazel.build/versions/master/docs/bazel-user-manual.html#flag--test_arg) but I don't think that's an issue specific to rules_scala

scala/scala.bzl
```diff
@@ -522,7 +522,7 @@ def _scala_test_impl(ctx):
     (cjars, rjars) = (jars.compiletime, jars.runtime)
     # _scalatest is an http_jar, so its compile jar is run through ijar
     # however, contains macros, so need to handle separately
-    scalatest_jars = _collect_jars([ctx.attr._scalatest]).runtime
+    scalatest_jars = _collect_jars([ctx.attr._scalatest, ctx.attr._scalatest_runner]).runtime
```
sdtwigg: Can you see if this works when adding it to the extra_runtime_deps list on line 520? (with _scalatest_reporter)

sdtwigg: Yes, this way the added classes in the runner it won't be on the compile-time classpath for all scala_test compiles. 

The scalatest_jars is setup like that to work-around (hack) an issue with java_import forcing ijar creation not meshing well with jars that contain scala macros. Thus, just want to contain how much that hack exposes/gets used.

johnynek:
Test this please

(Msg to cibot)

scala/scala.bzl
softprops: 👍 pushed up a a change for that

softprops:
This [build error](https://travis-ci.org/bazelbuild/rules_scala/jobs/239455178#L528)  looks unrelated to my change.

ittaiz:
@softprops, I restarted the travis job

softprops:
thanks @ittaiz!

johnynek:
👍 

Thanks for tackling this @softprops ! this will be a usability win for us!

softprops:
us tool!

softprops:
any thoughts on this folks?

softprops:
asking because I think I saw some changes to the test_run.sh that are probably going to cause a merge conflict in this pull once merged

chaoren:
> Method level filtering can be accomplished by providing additional --test_args with the -z flag of the scala test runner.

Hi, do you think it would be okay to add method filtering via the `class#method` syntax?
Right now bazel java uses `--test_filter=foo.Bar#` to match the class `foo.Bar` exactly, while `--test_filter=foo.Bar` matches `foo.BarBaz` as well. I think we should follow that behavior. It would help make our IDE work more consistent.

johnynek:
I think it is good to be consistent. Does that syntax not work? I think
@softprops just used the default scala_test filtering functions.

softprops:
According to bazels official test_filter docs here's what users should expect

> The particular interpretation of filter-expression is up to the test framework responsible for running the test.

Does that syntax come from a feature of the java test framework or does java rules parse that out the deliver components of the parsers expression to the underlying test framework? 

I would say it may possible with scala test but that it's not actually a feature of scala test  to provide a filter for a test suite and method in one argument. It has separate arguments for both. I choose the one most closely aligned with scala tests sbt interface because it's what the majority of scala users are used to ( sbt testOnly suite name )

In order to match the java syntax, the bazel scala test runner would need to acquire some intelligence about parsing the filter acording to the java test tool conventions and adapt to two or more arguments to the scala test runners feature set. 

I say that it may be possible but not certainly possible because I noticed when doing some initial testing with --test_filter for the suite name and --test_arg -z scala test didn't seem to honor both. More testing should be done to confirm.

softprops:
Here's a link to the scala test runners cms line args

http://www.scalatest.org/user_guide/using_the_runner

exp0nge:
Hey, quick question, anyway to pass -t or -z? `test_args` seemed to be broken

softprops:
Neither. It's `-s`. An example would be  `bazel test //foo:bar --test_filter=com.mycompany.foo.BarTest`

I choose this by team was in transition from sbt and that was the closest direct translation I could find since sbt has its own built in test discovery filters. I experimented with a few of the other options and, at least at the time, -s felt most appropriate.

exp0nge:
So no way to run a specific test within a suite since it's hardcoded to `-s`? Eg: `#"some test"`. `t`/`z` would have at least allowed some glob but maybe slower to scan everything. Maybe I can submit a patch to get those in with a different bazel flag?

ittaiz:
@softprops the original description said one can use test_arg does that not work?

softprops:
Enough time has passed that it would be worth trying again. It was 2017 when I last looked at this!  I believe the policy of the repo is the same, if you want to see change you going to have to be the maker of that change.

ittaiz:
@softprops I agree, I didn't ask you to do the change but rather asked for a clarification.
Mainly because I thought (maybe mistakenly) that y'all use this mechanism in the d2d and you can comment on how people use it.
I don't use scalatest so I'm unsure (can fiddle around with it but I'd rather invest the time elswhere).
My bottom line- do you use the following approach and can say if it still works (if not that's ok):

> Method level filtering can be accomplished by providing additional --test_args with the -z flag of the scala test runner. I was able to accomplish that with something like the following `bazel test --test_arg=-z --test_arg=single-test-pattern //...`

@exp0nge can you say what exactly is broken?

virusdave:
Any update on this?  Or has been abandoned again? :(

exp0nge:
Sorry forgot to respond. `bazel test //your-module:tests --test_arg=-z --test_arg="a single test"` does work for now. I think it would be nice to have some equivalent of `#` but we need work done for that

ittaiz:
Maybe send a documentation PR for now?

PR #218, opened 2017-06-05
Implement --test_filter for scala_junit_test. Fixes #212.

### LINKED ISSUE #212: Filter tests with --test_filter
Bazel allows filtering tests with [`--test_filter=`](https://bazel.build/versions/master/docs/bazel-user-manual.html#flag--test_filter), which passes the filter expression via the `TESTBRIDGE_TEST_ONLY` environment variable. [`DiscoveredTestSuite`](https://github.com/bazelbuild/rules_scala/blob/master/src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala) should filter tests with the `TESTBRIDGE_TEST_ONLY` variable to support the `--test_filter` argument.

chaoren:


johnynek:
can we have a test similar to #216

chaoren:
Done.

src/java/io/bazel/rulesscala/test_discovery/FilteredRunnerBuilder.scala
```diff
@@ -0,0 +1,34 @@
...
+import org.junit.runner.Runner
+import org.junit.runners.BlockJUnit4ClassRunner
+import org.junit.runners.model.FrameworkMethod
+import org.junit.runners.model.RunnerBuilder
+import scala.collection.JavaConverters._
+
+object FilteredRunnerBuilder extends RunnerBuilder {
+  // Defined by --test_filter bazel flag.
+  private def pattern =
+    sys.env.get("TESTBRIDGE_TEST_ONLY").map(Pattern.compile(_))
```
chaoren: Yes, I'm a scala scrub. That's what I intended.

src/java/io/bazel/rulesscala/test_discovery/FilteredRunnerBuilder.scala
```diff
@@ -0,0 +1,34 @@
...
+  override def runnerForClass(testClass: Class[_]): Runner = {
+    new BlockJUnit4ClassRunner(testClass) {
```
chaoren: I think the compiler should be smart enough to handle this, but I'll pull it out just in case.

src/java/io/bazel/rulesscala/test_discovery/FilteredRunnerBuilder.scala
chaoren: Done.

src/java/io/bazel/rulesscala/test_discovery/FilteredRunnerBuilder.scala
chaoren: Done.

src/java/io/bazel/rulesscala/test_discovery/FilteredRunnerBuilder.scala
```diff
@@ -0,0 +1,34 @@
...
+      override def getChildren: java.util.List[FrameworkMethod] = {
+        def children = super.getChildren
+        pattern match {
+          case Some(pattern) => {
```
chaoren: Done.

ittaiz:
@chaoren thanks! I think the changes really improved the readability.
Unfortunately travis is upset so I've re-triggered it and let's wait for it's judgment...
Appreciate your help.
Looks to me that supporting specs2 should be really easy, right?
Just extract the filter method contents to a "filter" and have a junit and specs2 filters. WDYT?
Do you know what format you'll be sending? [This](https://github.com/JetBrains/intellij-scala/blob/26f8c5b659c906ac645463d67602dcb13f83a4a0/src/org/jetbrains/plugins/scala/testingSupport/test/specs2/Specs2ConfigurationProducer.scala#L66) might be a lead to the format...

johnynek:
👍  when green.

johnynek:
this fails the bazel ci:

http://ci.bazel.io/job/PR/job/rules_scala/26/BAZEL_VERSION=latest,PLATFORM_NAME=linux-x86_64/console

not a noisey test.

ittaiz:
it seems that specs2 integration was broken. All tests which use specs2 seem to be broken.

src/java/io/bazel/rulesscala/test_discovery/FilteredRunnerBuilder.scala
```diff
@@ -0,0 +1,33 @@
...
+import org.junit.runners.model.RunnerBuilder
+import scala.collection.JavaConverters._
+
+object FilteredRunnerBuilder extends RunnerBuilder {
+  // Defined by --test_filter bazel flag.
+  private val pattern = sys.env.get("TESTBRIDGE_TEST_ONLY").map(Pattern.compile(_))
+
+  override def runnerForClass(testClass: Class[_]): Runner =
+    pattern match {
+      case Some(pattern) => new FilteredClassRunner(testClass, pattern)
```
ittaiz: btw, you can live it as is but just as an FYI binding in the case to the same name of the matched variable is a bit of a scala anti practice since you now have shadowing.
What many people do (and is not amazing but works) is to call you `private val pattern...` `private val maybePattern`

chaoren: Done.

ittaiz:
Thanks!

chaoren:
Hmm, yeah. Specs2 might use a different runner.

ittaiz:
@chaoren I'm thinking of two options:
1. Only enable this filtering if the junit API allows to know if it's "regular" junit. Don't know if the junit API exposes it and what constitues as "regular".
2. Change specs2_junit to pass some kind of flag and then only enable this filter if no flag exists (the flag can be something like custom_frameworks_using_junit and to enable others piggybacking on it).

The pro of 1 is that it seems to be more generic.
The pro of 2 is that I want to add some kind of identification so I'll be able to support test filtering for specs2 (like the one intellij does today with specs2).

WDYT?

chaoren:
Ideally we want this feature for specs2 as well. But they don't have test "methods" per se. Right now, I think I can just identify the parent runner, if it's a method list parent runner, I can replace it with the filtered one. Otherwise, I'll leave it untouched. It looks like specs2 uses a nested Suite runner. We can probably do some other kind of filtering with it later on.

chaoren:
Done. It looks like specs2 uses nested suite runners. I've excluded that case for future implementation. Please take a look.

chaoren:
Looks like two builders timed out. Please restart the test?

ittaiz:
test this please (cibot)

chaoren:
Last builder timed out again.

chaoren:
test this please (cibot)

ittaiz:
finally got to a green build. thanks @chaoren for the contribution and your patience. I hope we will be able to stabilize travis soon.

Issue #217, opened 2017-06-05
Should we delete/disable the "is_identical" test


ittaiz:
The test take a very long time to complete (~10 minutes IINM on travis) which causes flaky builds on travis. Additionally this test sometimes fails on my own machine (mac os) for an unclear reason which we have an issue for.
I just feel that the long feedback loop + travis flakiness is too much.
@johnynek wdyt?

softprops:
I actually liked this an use it as the baseline for some other rule repos I've created. The reason being is that bazels claims on reproducibility is one of its core tenets.

johnynek:
I don't want to give the test up either if we can avoid it. We can break
reproducibility in the rules by accidentally embedding timestamped or nonce
data. If we do that, we will destroy caching, so I think it is pretty
important to test.

johnynek:
We don't run a test like this at Stripe, but since the scala rules are
creating a lot of important outputs, it seems like testing them makes a lot
of sense.

softprops:
thats been an issue for is, mainly for srcjars. less so with binary jars.

sdtwigg:
Perhaps we could move it to the end?

This should also speed things up a little since it wipes out the cache and
thus the targets after it are slower.

softprops:
that makes sense to me and seems to address the faster feedback issue

johnynek:
👍 

We should add a comment to keep it at the end. I think the problem is that it was at the end, but then we added tests.

ittaiz:
Can we have two build configurations on travis? If so WDYT about separating it to a different build configuration? This way if we see flakiness we can judge if this is real or not. The problem is worsened by the fact that we run tests for 4 different configurations. A lot of times one of them times out on travis.

ittaiz:
Have an idea-
Maybe we'll move this test to a different repo and have that defined as another travis check? I think this can help since it will minimize the scope of clean/rebuild

ittaiz:
closing for now since the flakiness was solved. this is less of an issue now

PR #219, opened 2017-06-05
Fixup some of the tests

sdtwigg:
In test_run.sh:
* Move test_build_identical to the end (as per #217), add comment
* Move all the binary runs together, sort them

In test/BUILD:
* Add missing test binary runs (and sort them)
  Now is same as in test_run.sh

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -298,11 +298,12 @@ scala_specs2_junit_test(
 ) for binary in [
     "JavaBinary",
     "JavaBinary2",
+    "JavaOnlySources",
+    "MixJavaScalaLibBinary",
+    "MixJavaScalaSrcjarLibBinary",
+    "MoreScalaLibBinary",
     "ScalaBinary",
     "ScalaLibBinary",
-    "MoreScalaLibBinary",
-    "MixJavaScalaLibBinary",
-    "JavaOnlySources",
 ]]
 
 # Make sure scala_binary works in genrule environment
--- a/test_run.sh
+++ b/test_run.sh
@@ -208,18 +208,18 @@ run_test bazel test test/...
 run_test bazel run test/src/main/scala/scala/test/twitter_scrooge:justscrooges
 run_test bazel run test:JavaBinary
 run_test bazel run test:JavaBinary2
+run_test bazel run test:JavaOnlySources
 run_test bazel run test:MixJavaScalaLibBinary
 run_test bazel run test:MixJavaScalaSrcjarLibBinary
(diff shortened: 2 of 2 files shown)


johnynek:
👍

johnynek:
I'm okay with my commits (the merge conflict resolve) being included cibot.

ittaiz:
Why do we need to add all the binaries in the test/BUILD? What confidence does this bring us?

sdtwigg:
They were originally added to ensure there was no discrepancy between blaze
test using the binary and blaze run due to runfiles weirdness.

At this point though (with runfiles setup being sorted by the launcher), I
put them in test/BUILD because I'm more likely to do a quick blaze test
test/... than run test_run.sh when in the middle of changes.

ittaiz:
I see. Then maybe we should cut down the number of happy cases we have in
test_run.sh? We obviously need a few for confidence and regression of
"bazel run" but I definitely agree the feedback cycle of "bazel test
//test/..." is much faster

sdtwigg:
Sounds reasonable.

johnynek:
Thinking more about this, I'm not sure it does matter where the build is identical test goes in the order. Yes it removes everything but it also rebuilds everything.

Why does it matter where it goes in order?

ittaiz:
I think you're right. We can also in the build logs that the tests which
follow take a short time so it shouldn't be an issue.

sdtwigg:
I re-reviewed. For some reason, I earlier thought that the identical test was only checking some of the jars under test/..., but it seems to be checking all of them.

This is far less useful now. It is maybe nice to do the identical test at the end because it is the longest so if you break a test after it in test_run, you aren't waiting as long.

johnynek:
Yeah, that's true. For that reason running last is a good idea.

ittaiz:
I suggest, rebase, green build and let's merge

ittaiz:
@sdtwigg do you want to resolve the conflicts or do you have too much going on? I can do it if you're too busy. Obviously #227 is order of magnitudes more important :)

sdtwigg:
Just got back from a small vacation. Will try to hit both over the weekend (as catching up on my directwork related tasks must take priority :) )

ittaiz:
:)
Awesome! This is starting to be a blocker for us since we need *runtime*
location expansion and that is only supported (partially) in java_test.
This means I need to switch test running to be a macro (scala library and
java_test). If there is anything I can do to help please let me know.

ittaiz:
closed in favor of #398 thanks @sdtwigg!

Issue #211, opened 2017-05-31
test.xml output from scala_junit_test is lacking in detail


chaoren:
JUnit test class with 6 test methods, in which 1 fails gives the following output:

```
$ cat bazel-testlogs/foo/bar/test/test.xml
<?xml version=1.0 encoding=UTF-8?>
<testsuites>
  <testsuite name=foo/bar/test tests=1 failures=0 errors=1>
    <testcase name=foo/bar/test status=run>
      <error message=exited with error code 1></error>
    </testcase>
  </testsuite>
</testsuites>
```

chaoren:
The IDE needs a list of which methods passed, which failed, and the actual failure message(s) for each.

johnynek:
@ittaiz would this be easy to fix for scala junit?

Also @chaoren have you looked at `scala_test` outputs? Are they okay?

chaoren:
`scala_test` looks fine.

chaoren:
The `testsuite name="foo/bar/test"` should probably be the test class name instead of the bazel target name (though with `:` replaced by `/`).

ittaiz:
I'm on vacation with really limited reception so expect delays in response. In general I can say that I don't know how easy/hard it will be as I just used whatever JUnitCore currently outputs. 
@chaoren I think the bigger problem is that this currently outputs one xml file per target even if we had 3 classes, right?
This should be solved when java interop will be solved since we'll just delegate to java_test. @sdtwigg do you have any estimate for it?

softprops:
Im actually coincidentally working on a ticket gather and stage test.xml reports to jenkins. If my team feels some visibility pain here, I'd be happy to try and jump in and take a look as well

softprops:
correction, we're only using `scala_test` targets so we shouldn't hit this issue, but still happy to help where help is needed :)

PR #220, opened 2017-06-06
allow scala tests to omit verbose test logs and xml files

natansil:
changed `scala_junit_test` runner to `bazel_test_runner` instead of `JUnitCore`. this allows for complete logs and xml output to be generated. the `bazel_test_runne`r source code was copied to `rules_scala` in order to depend on it in the `scala_junit_test` rule.
This is a temporary measure, where the duplicate code can be deleted once java_test will be able to depend on scala_library.

This PR will allow to resolve issue https://github.com/bazelbuild/rules_scala/issues/211 (test.xml output from scala_junit_test is lacking in detail)


bazel-io:

johnynek:
Thanks for the PR.

I'm concerned about copying 8k lines from bazel. Is there any way we can:

1. use bazel code as a remote repository?
2. could we minimize this somewhat? The scala_test xml output is only ~500 lines: https://github.com/bazelbuild/rules_scala/blob/master/scala/support/JUnitXmlReporter.scala

ittaiz:
@johnynek I'm fairly certain we can't use it since the targets are not public. I asked Ulf about it sometime ago and he said to copy the code (he might thought it's less code though).
Main reason I think this is ok is because AFAIU @sdtwigg will finish his work very soon and enable us to delete all of this code since we'll be using java_test which has access to this Runner.

WORKSPACE
```diff
@@ -29,3 +29,17 @@ maven_jar(
   sha1 = "e5b3e2753d0817b622c32aedcb888bcf39e275b4")
 
 
+# dependencies needed for BazelTestRunner
+maven_server(
+    name = "one_hippo",
```
ittaiz: why do we need this server? doesn't maven central have jsr305?

WORKSPACE
```diff
@@ -29,3 +29,17 @@ maven_jar(
...
+    url = "http://maven.onehippo.com/maven2/",
+)
+
+maven_jar(
```
ittaiz: @johnynek This should probably move to the *_repositories pattern and bind, right?

ittaiz:
test this please (cibot)

natansil:
@johnynek, I've looked into minimizing the code. Unfortunately not a lot can be removed...
An added bonus is getting the tests sharding feature of the bazel test runner... 
I think it's the best temporary solution until the switch to java_test

ittaiz:
test this please (ci)

chaoren:
@ittaiz 
> I'm fairly certain we can't use it since the targets are not public.

Can we just download the files and make our own targets? Like we did with https://github.com/bazelbuild/rules_scala/blob/a94b6d0690f48cd41d3649d8aab9a7fd1491b76e/scala/scala.bzl#L753.

ittaiz:
@chaoren we're not generating any targets in the stub example. We're just using a source file. In this case we need a binary. I've tried looking in the zip and I don't see the runner jar in it. Am I looking in the wrong place? If there is an easy archive we can depend on which exposes the jar that's definitely an option

chaoren:
@ittaiz 
> I've tried looking in the zip and I don't see the runner jar in it. 

But the sources to build the jar is available from https://github.com/bazelbuild/bazel/archive/0.5.1.tar.gz, no?

Instead of just copying the sources, why not refer to them directly from the package? Just create an new_http_archive, and our own BUILD file for it.

natansil:
@chaoren, from what I see in `new_http_archive` documentation, only the source files are exposed. Do you suggest to merge all of the _bazel_ runner related BUILD files into one _rules_scala_ BUILD file?

chaoren:
@natansil yes, create our own BUILD file for those sources.

natansil:
@chaoren , it will take me a few days to get to this. so we have 3 options:
1. someone else try to create own BUILD file
2. merge the current solution
3. wait for when I have more availability for this

chaoren:
@natansil 
> 2. merge the current solution

Please don't.

I'll do it.

chaoren:
Okay, so after playing around a bit, I couldn't find a way across the package boundary created by the existing BUILD files. (Tried some weird tricks with `strip_prefix`.)

But, I did find the the jar we need at `@bazel_tools//tools/jdk:TestRunner_deploy.jar`. However, it is not publicly visible, unlike the [other deploy jars next to it](https://github.com/bazelbuild/bazel/blob/master/tools/jdk/BUILD). I'll send a pull request upstream to expose test runner for our use. If they accept, then problem solved.

If they don't accept, then we can still just build the jar from source and check it in. Of course there's the security implication of checking in a jar, but since bazel builds should be reproducible, the reviewer (or whoever cares) can just verify by building the jar themselves and comparing.

I think either solution is superior and more maintainable than this PR.

ittaiz:
Thanks!
Let's wait and see the response on your PR.
I agree both are probably better.

natansil:
superseded by https://github.com/bazelbuild/rules_scala/pull/228

PR #223, opened 2017-06-07
trying to mitigate the "timeout no output" fail

ittaiz:
using travis_wait to send a console message every minute to avoid timeout


ittaiz:
I have bigger faith in #224

PR #224, opened 2017-06-08
stabilize travis run

### LINKED ISSUE #221: Travis build flakiness

or-shachar:
fixes #221 
travis doesn't like it that our build takes > 30m to finish, especially when we don't print anything to stdout while one of the tests is running.

Hence I created an alternative `run_test` that would run only in CI that would run each test as spawned processes and print to stdout every 1 minutes while the spawned process is running.

**why didn't I just replace the current `run_test`?**
Because locally we don't want our test to spawn new processes (we want to be able to terminate the process tree with Ctrl+C, for instance).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,4 +36,4 @@ before_install:
   - cat .bazelrc.travis >> .bazelrc
 
 script:
-  - bash test_run.sh
+  - bash test_run.sh ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,18 @@ before_install:
       OS=linux
     fi
     if [[ "${V}" == "HEAD" ]]; then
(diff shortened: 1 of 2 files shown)


bazel-io:

johnynek:
thank you so much! I'm optimistic.

PR #226, opened 2017-06-12
forcing utf-8 encoding with javac

or-shachar:
`java_library` compiles UTF-8 files without special configurations.
On systems where UTF-8 is not the default encoding, `scala_library` fails when compiling java code with UTF-8 characters, showing error: 
```
error: unmappable character for encoding ASCII
```
As part of the effort to be compliant with java_library - I suggest we do the same.


bazel-io:

ittaiz:
test this please (cibot)

ittaiz:
test this please (cibot)

ittaiz:
test this please (cibot)

PR #227, opened 2017-06-13
Restructure scala provider, add JavaProvider

sdtwigg:
The key change is that the scala provider has been completely
restructured to match the structure of the JavaProvider. It no longer
tracks exports separately but instead has the following fields:

* outputs = contains all the outputs generated by the current rule
(nothing or ijar and class_jar); however, no rules here actually use it.

* compile_jars = contains all the jars dependent rules should compile
with. Thus, ijar and exports.compile_jars

* transitive_runtime_jars = contains all the jars needed to handle this
target at runtime. Thus, class_jar plus (deps + exports +
runtime_deps).transitive_runtime_jars

The created java provider (used by dependent native java rules) uses
just compile_jars and transitive_runtime_jars.

In general, this change was seamless. The major exception is if you were
specifically relying on only exports being re-exported by specifically
gen_scrooge_srcjar. Beforehand, any jars produced by direct dependencies
were not being exported to dependents but now they are. Note that, if
you were using scrooge_scala_library instead, nothing should change.

Other changes:
* Use depset instead of set. (set docs already say set should be
considered deprecated and just points to depset anyway)

* Tests amended to exploit that JavaProvider is properly constructed.
Note that things are still a little strange until Bazel releases
incorporate some fixes to JavaProvider and native provider rules.
Generally, only deps for java_library and java_binary work at the
moment.


sdtwigg:
I am still potentially fiddling with this a little. In particular, unsure if the tests properly capture what is going on yet.

I had been delaying sending this out until some Bazel changes were released (as until then the 'feature' of interop with JavaProvider is slightly incomplete).

However, the re-architecture here makes thinking about asks like "neverlink" much easier so just sending it out....

ittaiz:
Thanks!
To the best of my understanding 0.5.2 (containing a lot of contributions from you which we've been waiting for) will be released end of this week.
How different will the PR be?
Do you think we should wait?
I can wait with the neverlink for another week.
Also,
Given your Bazel changes java_test won't be a problem, right?

sdtwigg:
If next week, makes sense to wait. Would avert the backwards incompatibility issue since my bazel PRs slightly amended the argument types to experimental java_common.create_provider.

The changes are minimal (an adjustment to collect_jars and the code that constructs the provider in the _lib rule). I would probably continue using both the scala provider and java provider for this PR. Then, we can consider using just the java provider at a later date.

I never explicitly tested java_test (somehow never occurred to me); however, it looks like it piggybacks on java_binary, which was thoroughly tested.

ittaiz:
Got you. I'll ping on the Bazel ticket and update here.
I seemed to remember you "loosened" the java_test restrictions in the Bazel side to allow that to work with JavaProvider as well, am I mistaken?
Also,
Do you want to do the move to java_common.compile in a speparate PR? Leave it to someone else?

johnynek:
so it looks like this fails on head on travis. Is that expected?

scala/scala.bzl
```diff
@@ -390,42 +381,68 @@ def _collect_jars_from_common_ctx(ctx, extra_deps = [], extra_runtime_deps = [])
...
+        deploy_jar = ctx.outputs.deploy_jar,
+    )
+    # Note that, internally, rules only care about compile_jars and transitive_runtime_jars
+    # in a similar manner as the java_library and JavaProvider
+    scalaattr = struct(
+        outputs = rule_outputs,
+        compile_jars = next_cjars,
+        transitive_runtime_jars = transitive_rjars,
+    )
+    # TODO(twigg): Linearization is concerning here.
```
johnynek: is this something that has a fix on some open PR to bazel? Can we add a link to that if so?

scala/scala.bzl
```diff
@@ -786,6 +803,8 @@ def scala_repositories():
 
   native.bind(name = "io_bazel_rules_scala/dependency/scalatest/scalatest", actual = "@scalatest//jar")
 
+# With providers changes, this is completely deprecatable once bazel release includes
+# fixes to java rules using JavaProvider
```
johnynek: can we change the implementation of this to just do the naive thing? We could make a `java_library` target that has the given exports and runtime_deps?

scala/scala.bzl
sdtwigg: Sorry, yes, this was already fixed and is the reason this CL doesn't work on head.

See https://github.com/bazelbuild/bazel/commit/05ea0281ee4820a7fd70d166e140a8e0355d275b

scala/scala.bzl
sdtwigg: Sadly, not yet, exports was what was specifically broken and fixed by:

https://github.com/bazelbuild/bazel/commit/7e91ad631a660ae12a2b8dcfb8ba42574a7a76d3

However, might delay submitting this PR until that commit is released, in which case will definitely do that.

ittaiz:
Stephen,
Wdyt about changing the PR to assume your changes from Bazel head are in and we'll review that? We'll ignore the failures against 0.5.1 for now.
The reply on the ticket was that 0.5.2 is near.
Two more related questions:
1. java_common.compile can be in a separate PR right?
2. java_test dependency should work given your changes on Bazel HEAD, right?

sdtwigg:
Sounds good to me. I already have a setup on my local machine for testing against Bazel@HEAD (from when making Bazel changes).

1. Yes, definitely.
2. Yes, I would want to test it specifically but I did a review of the associated Bazel sources and it doesn't look like anything is there that would prevent it from working. (There is a bit of magic with using Java reflection that makes me nervous about making a definitive statement without a test. This is why my responses have been vague on this matter.)

ittaiz:
Sounds good. Waiting for your ping when you want us to review against Bazel HEAD and thanks for the clarifications!
Don't worry re `java_test` I'll test it out when we'll get to it

scala/scala.bzl
johnynek: It's pretty hard to see what is going on with bazel development without also figuring out the gerrit thing. I get that you all use it internally, but I never have and honestly it feels like some monstrous UI from 1993 or something. I am dreaming of the day all PRs go to github.

ittaiz:
@sdtwigg I really want to try this PR with Bazel's HEAD.
Do you have any estimate on when you'll have a chance to get to it?
If it will take you a while (which is fine) I'll try to hack together something on my own fork.
Context is that we have a blocker with this [issue](https://github.com/bazelbuild/bazel/issues/2475) and if we can change `scala_junit_test` to be a macro which composes `scala_library` and `java_test` that would work around it for now (since `java_test` already has the needed behavior)

sdtwigg:
Take a look. Although, don't merge this CL as-is since I kept as two commits for now in case I need to jump back to the 0.5.1 version.

scala/scala.bzl
```diff
@@ -356,32 +355,23 @@ def collect_srcjars(targets):
...
-            # compile_jars += [output.ijar for output in target.java.outputs.jars]
-            compile_jars += target.java.transitive_deps
-            runtime_jars += target.java.transitive_runtime_deps
-            found = True
-        if not found:
+        if java_common.provider in target:
+            java_provider = target[java_common.provider]
+            compile_jars += java_provider.compile_jars
+            runtime_jars += java_provider.transitive_runtime_jars
+        elif hasattr(target, "scala"):
```
johnynek: When will we hit this branch now? Won't the scala code hit the first provider?

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -153,13 +153,13 @@ def _gen_scrooge_srcjar_impl(ctx):
...
-  scalaattr = struct(outputs = None,
-                     transitive_runtime_deps = jars.transitive_runtime_deps,
-                     transitive_compile_exports = jars.transitive_compile_exports,
-                     transitive_runtime_exports = jars.transitive_runtime_exports,
-                     )
+  scalaattr = struct(
+      outputs = None,
+      compile_jars = deps_jars.compile_jars,
+      transitive_runtime_jars = deps_jars.transitive_runtime_jars,
+  )
```
johnynek: Can we not use a java provider here?

scala/scala.bzl
sdtwigg: This code should not reach this branch.

I actually wanted to delete it, although was afraid of someone having weird custom rules appended that relied on the old provider being picked up. Then, they will fail-over to the raw file insertion. However, in retrospect, this was a silly thought and will remove.

Note, as future work, am tempted to create the custom no-ijar java_import variant so this raw file insertion can be phased out.

twitter_scrooge/twitter_scrooge.bzl
sdtwigg: Will fix, good catch.

sdtwigg:
Pushed some updates from feedback.

As a small note, if using test_run.sh, make sure you clean when switching bazel versions/rules_scala versions. Some combination of changing both at once causes the identical test to fail for the deploy jars. I couldn't really dig into it further since everything works within a single version....

ittaiz:
@sdtwigg do you want to change the travis file as part of your PR to minimal 0.5.2 rc2?
will get us to green on travis at least

ittaiz:
@sdtwigg If I'm running locally with 0.5.2 rc4 do you see any reason not to cherry-pick your changes into my fork? anything unstable. From the chatter on the [ticket](https://github.com/bazelbuild/bazel/issues/3086) 0.5.2 is coming out today.
Main reason I need this change is because it will enable me to have java_test depend on scala_library which in turn will allow me to use $location expansion in jvm_flags AND get the runtime location (instead of the source location). This is currently a blocker for me to use remote execution. ([This](https://github.com/bazelbuild/bazel/issues/2475) currently works for some native cases but not for skylark)

sdtwigg:
It should all work. I looked it over earlier and it is basically where I
want it excluding small, likely non-functional cleanups.

ittaiz:
Great! 0.5.2 was finally released, can I add a commit to this PR to update the travis.yml from 0.4.5 to 0.5.2? We'll have good visibility then and can decide if we want to merge as is or wait for the cleanup?
Small headsup- we (Wix) want to make two significant changes in this area:
1. Refactor JUnit test
2. Better support "strict deps" and unused_deps.
The work on 2 started (based on these commits). If you'll refactor soon then we'll rebase and resolve the conflicts but we it will be a problem to do this in a few days time since the drift will be big.
*we'll open separate issues about these

johnynek:
What do you mean better support for strict deps and unused deps? Very
interesting!

ittaiz:
:) I'll open a ticket in an hour or two. Just wanted to give the heads up.

ittaiz:
Yay! The build is green. @sdtwigg waiting for your OK (or small cleanup) so we can merge.

johnynek:
Test this please

sdtwigg:
Confirm.

ittaiz:
@sdtwigg thank you so much for your contributions! This has been a long outstanding debt :)

PR #228, opened 2017-06-13
Junit (and specs2) tests now output logs with content

ittaiz:
Moved to using Bazel's Junit runner which apparently was already open.
based on @natansil's previous work on scala.bzl but without the code from Bazel.
Also changed the existing junit logs test to actually check that the test-cases are actually printed to the xml log file. Test was red and with the change it passed! :)
cc @chaoren

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -558,8 +558,7 @@ def _scala_test_impl(ctx):
     return _scala_binary_common(ctx, cjars, rjars)
 
 def _gen_test_suite_flags_based_on_prefixes_and_suffixes(ctx, archive):
-    return struct(suite_class = "io.bazel.rulesscala.test_discovery.DiscoveredTestSuite",
-    archiveFlag = "-Dbazel.discover.classes.archive.file.path=%s" % archive.short_path,
+    return struct(testSuiteFlag = "-Dbazel.test_suite=io.bazel.rulesscala.test_discovery.DiscoveredTestSuite", archiveFlag = "-Dbazel.discover.classes.archive.file.path=%s" % archive.short_path,
     prefixesFlag = "-Dbazel.discover.classes.prefixes=%s" % ",".join(ctx.attr.prefixes),
     suffixesFlag = "-Dbazel.discover.classes.suffixes=%s" % ",".join(ctx.attr.suffixes),
     printFlag = "-Dbazel.discover.classes.print.discovered=%s" % ctx.attr.print_discovered_classes)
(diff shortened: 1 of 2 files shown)


ittaiz:
the test is failing on linux. not sure why.
@or-shachar @johnynek would appreciate any ideas if you have any.
Trying to run it now on a remote linux machine I have but `bazel build` is failing there on 
```
____[440 / 779] Writing file test/src/main/scala/scala/test/large_classpath/dependency_dependency_dependency_dependency_dependency_dependency_dependency_dependency_dependency_dependency_dependency_dependency_dependency_dependency_dependency_dependency_dependency_dependency_dependency_dependency_67_MANIFEST.MF
____[527 / 934] Writing file test/src/main/scala/scala/test/resource_jars/resource_jars_worker_input
____[562 / 994] Creating source manifest for //src/scala/scripts:generator [for host]
____[581 / 1,013] Writing file test/src/main/scala/scala/test/twitter_scrooge/scrooge3_MANIFEST.MF
____[589 / 1,024] Extracting interface @scrooge_core//jar:jar
ERROR: /root/.cache/bazel/_bazel_root/8ebc253a2f259764cbc03f2ee667a0ac/external/scala/BUILD.bazel:15:1: Extracting interface @scala//:scala-library failed: Process exited with status 1 [sandboxed].
src/main/tools/linux-sandbox-pid1.cc:193: "mount(/tmp, /tmp, NULL, MS_BIND, NULL)": Invalid argument
```

ittaiz:
got this now on the remote machine:
```
ERROR: /root/.cache/bazel/_bazel_root/8ebc253a2f259764cbc03f2ee667a0ac/external/io_bazel_rules_scala_org_openjdk_jmh_jmh_generator_reflection/jar/BUILD.bazel:2:1: Extracting interface @io_bazel_rules_scala_org_openjdk_jmh_jmh_generator_reflection//jar:jar failed: Process exited with status 1 [sandboxed].
src/main/tools/linux-sandbox-pid1.cc:193: "mount(/tmp, /tmp, NULL, MS_BIND, NULL)": Invalid argument
```

johnynek:
some kind of sandboxing error it seems like you are getting. But travis is seeing something else: that the xml is not being generated but no other error, right?

ittaiz:
"xml is not being generated"? Why do you think?

ittaiz:
@johnynek solved the problem. Apparently `grep` on linux doesn't like escaping of single quote with `\`. Removed it and it works on mac and linux. I'll merge when green :)

PR #229, opened 2017-06-14
don't put signatures in deploy jar

oscar-stripe:
We hit some error with a jar including a signature, then it causing the deploy jar to fail verification.


bazel-io:

ittaiz:
test this please (cibot)

src/java/io/bazel/rulesscala/jar/JarCreator.java
```diff
@@ -173,6 +173,16 @@ public void execute() throws IOException {
     }
   }
 
+  @Override
+  protected boolean ignoreFileName(String name) {
+    /*
+     * It does not make sense to copy signature files
```
ittaiz: Do you think it would be wise to rename `JarCreator` to `FatJarCreator` / `DeployJarCreator` / `UberJarCreator`? Current name made me wonder why the override is like this and the base is just "false". Until I read the comment and remembered

johnynek:
yeah, I will add a test. I just wanted to get unblocked so I pushed it here.

ittaiz:
Sure. We can merge and you can add a test later if you're blocked. WDYT
about the JarCreator class name (again not blocking the merge)?

johnynek:
okay @ittaiz, I have added a test which exposes the problem if I don't have this change and does not fail with the change.

I don't want to rename things at this stage since this code was originally copied from bazel and they set the names (also, we have code that actually depend on it elsewhere so it will be a migration pain for us that I am hoping to avoid).

johnynek:
I consent

ittaiz:
Test this please

ittaiz:
LGTM

PR #232, opened 2017-06-18
renamed MoreScala* targets to ResourcesStripScala* targets

ittaiz:
Pretty confident from context and git commit log that this is the actual name.
I tried to understand this a few weeks ago and thought it's best to document this knowledge back in the code.
@johnynek,
Would appreciate your thoughts as well to make sure I'm not adding confusion here.

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -165,8 +165,8 @@ scala_library(
 )
 
 scala_library(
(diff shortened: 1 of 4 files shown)


ittaiz:
@johnynek wdyt?

PR #234, opened 2017-06-25
Split jvm_flags overloading

### LINKED ISSUE #187: jvm_flags is used for several different meanings
Currently jvm_flags are used for:
1. Scalac jvm flags
2. Javac jvm flags
3. End executable jvm flags
We should probably introduce `compiler_jvm_flags` and pass that to javac and scalac and keep jvm_flags only to exeutable targets.
Additionally whoever handles this should consider adding a more explicit test about use-case 3 (currently one of the junit tests handles it but it might not be that clear).
Testing use-cases 1&2 can be tricky (maybe setting a really low Xmx and seeing the compilation fails with a negative test)

ittaiz:
This PR fixes #187. The main motivation behind it came when we found out that the overloading of jvm_flags can cause each target to create it's own worker instance and take down the machine.
The chosen solution is to expose two new attributes- `scalac_jvm_flags` and `javac_jvm_flags` which allow the user to configure the `jvm_flags` of the respective compiler. `jvm_flags` remains solely for the executables (repl, binary, tests).
This will also make us more aligned with the java rules (though they don't have the compiler jvm_flags).
I also did a small refactor in the test_run.sh to reduce duplication between tests which expect failure.
Note- this is a breaking change for people who up until now used `jvm_flags` to also configure the compiler flags.


ittaiz:
@johnynek @or-shachar would appreciate your review

ittaiz:
I think (not 100% sure) that we can remove two existing test references to `jvm_flags`

test_expect_failure/compilers_jvm_flags/BUILD
```diff
@@ -0,0 +1,11 @@
...
+scala_library(
+	name = "can_configure_jvm_flags_for_scalac",
+	srcs = ["WillNotCompileScalaSinceXmxTooLow.scala"],
+	scalac_jvm_flags = ["-Xmx1M"],
+)
+scala_library(
+	name = "can_configure_jvm_flags_for_javac",
+	srcs = ["WillNotCompileJavaSinceXmxTooLow.java"],
+	javac_jvm_flags = ["-Xmx1M"],
+)
```
or-shachar: cosmetic: newline at the end of the file

test_run.sh
```diff
@@ -175,6 +159,20 @@ run_test_local() {
   fi
 }
 
+action_should_fail() {
```
or-shachar: +1 👍

or-shachar:
@ittaiz travis passed most permutations but one failed due to connection issue with maven central. please rerun

scala/scala.bzl
```diff
@@ -616,6 +615,8 @@ _common_attrs = {
   "scalacopts":attr.string_list(),
   "javacopts":attr.string_list(),
   "jvm_flags": attr.string_list(),
+  "scalac_jvm_flags": attr.string_list(),
+  "javac_jvm_flags": attr.string_list(),
```
johnynek: can we add a documentation of this to README.md?

or-shachar:
> I think (not 100% sure) that we can remove two existing test references to jvm_flags
@ittaiz WDYM? which?

test_expect_failure/compilers_jvm_flags/BUILD
ittaiz: done

scala/scala.bzl
ittaiz: @johnynek please review

ittaiz:
BTW, our docs have in general gone stale and we need to revamp them.
We are working very hard to onboard our first team into bazel but I hope we'll have some capacity sometime in Q3 to try and tackle this

johnynek:
👍

ittaiz:
@johnynek  Just to put a smile on your face- this change allowed us to use persistent workers which dropped the compilation time from 35 minutes to 7 minutes! A lot of fine grain targets reusing JVMs...

johnynek:
That's great news! Very glad to hear.

PR #238, opened 2017-06-29
Add a provider for binary targets

johnynek:
this being missing caused us some pain.

1) some targets depend on binaries or test. Maybe not awesome, but not sure we really want to make this impossible either.

2) we want tooling we have that deals with hadoop deploys to just use java providers, but scala_binary does not have one, so it fails.

I added two tests that failed before this change.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -487,8 +487,15 @@ def _scala_binary_common(ctx, cjars, rjars):
       compile_jars = depset([outputs.class_jar]),
       transitive_runtime_jars = rjars,
   )
+
+  java_provider = java_common.create_provider(
+      compile_time_jars = scalaattr.compile_jars,
+      runtime_jars = scalaattr.transitive_runtime_jars,
+  )
+
   return struct(
       files=set([ctx.outputs.executable]),
+      providers = [java_provider],
       scala = scalaattr,
       runfiles=runfiles)
 
--- a/test/BUILD
+++ b/test/BUILD
@@ -76,6 +76,13 @@ scala_test(
     ],
 )
 
+# test that a library can depend on a test:
+scala_library(
+    name = "lib_to_test",
+    srcs = ["LibToTest.scala"],
+    deps = [":HelloLibTest"],
+    exports = [":HelloLibTest"])
+
 scala_test_suite(
     name = "HelloLibTestSuite",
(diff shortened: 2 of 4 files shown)


scala/scala.bzl
```diff
@@ -487,8 +487,15 @@ def _scala_binary_common(ctx, cjars, rjars):
       compile_jars = depset([outputs.class_jar]),
       transitive_runtime_jars = rjars,
   )
+
+  java_provider = java_common.create_provider(
+      compile_time_jars = scalaattr.compile_jars,
+      runtime_jars = scalaattr.transitive_runtime_jars,
```
ittaiz: Is this the way to do it? I'm still honing providers-fu. Shouldn't we use the java provider instead of the scala attr?

test/BUILD
```diff
@@ -76,6 +76,13 @@ scala_test(
     ],
 )
 
+# test that a library can depend on a test:
+scala_library(
+    name = "lib_to_test",
+    srcs = ["LibToTest.scala"],
+    deps = [":HelloLibTest"],
+    exports = [":HelloLibTest"])
```
ittaiz: Why do we need the exports?

test/BUILD
```diff
@@ -176,6 +183,13 @@ scala_binary(
     deps = ["ResourcesStripScalaLib"],
 )
 
+# test that a library can depend on a binary:
+scala_library(
+    name = "lib_to_bin",
+    srcs = ["LibToBin.scala"],
+    deps = [":ScalaLibBinary"],
+    exports = [":ScalaLibBinary"])
```
ittaiz: Why do we need the exports?

scala/scala.bzl
johnynek: Consult the lib function to compare. This is the same way we do above. `scalaattr` is just a struct that already has this info.

test/BUILD
johnynek: we don't. I'll remove.

scala/scala.bzl
ittaiz: Ok. I think @sdtwiggs is planning on removing it but since I'm not sure and since he can just remove this when removing the other one I'm good with this.

johnynek:
what about now @ittaiz ?

PR #239, opened 2017-06-29
misc: add line for test and build formatting

mduvall:
This is adds an extra line of padding so you can read the "Building from source" section (sorry).


bazel-io:

ittaiz:
Why are you sorry? Thank you for taking the time!

PR #240, opened 2017-06-30
src: add support for scrooging local JARs

mduvall:
This adds support for providing local JARs full of Thrifts, and generating the scrooge classes for this in `scrooge_scala_library`.


bazel-io:

src/scala/io/bazel/rules_scala/scrooge_support/Compiler.scala
```diff
@@ -113,7 +113,7 @@ class Compiler {
     } yield inputFullPath
 
     val rootImporter = FocusedZipImporter.forPaths(None, allJars)
-    allPaths.foreach { inputFullPath =>
+    allPaths.filter(_.endsWith(".thrift")).foreach { inputFullPath =>
```
ianoc-stripe: Can we add a comment here that we only should pass thrift files to the compiler. Jars which have manifest or other files in there will cause errors here.

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -206,6 +211,7 @@ scrooge_scala_srcjar = rule(
         #     "covered," as well as needing the thrifts to
         #     do the code gen.
         "remote_jars": attr.label_list(),
+        "remote_self_jars": attr.label_list(),
```
ianoc-stripe: we need to add a comment here saying what remote_self_jars means

src/scala/scripts/TwitterScroogeGenerator.scala
```diff
@@ -53,12 +53,16 @@ class ScroogeGenerator extends Processor {
     // remote jars are jars that come from another repo, really no different from transitive
     val remoteSrcJars = getIdx(3)
 
+    // Remote jars that we should build scrooge classes for
```
ianoc-stripe: maybe lets expand this

Remote jars where we should export the generated class files built from that thrift. This is the remote jar version of `immediateThriftSrcJars`

test/src/main/scala/scala/test/twitter_scrooge/bare_jar_test/BUILD
```diff
@@ -0,0 +1,21 @@
+load("//scala:scala.bzl", "scala_binary", "scala_library")
```
mduvall: erm don't need `scala_library` here

src/scala/io/bazel/rules_scala/scrooge_support/Compiler.scala
mduvall: 👍

twitter_scrooge/twitter_scrooge.bzl
mduvall: 👍

src/scala/scripts/TwitterScroogeGenerator.scala
mduvall: 👍

test/src/main/scala/scala/test/twitter_scrooge/bare_jar_test/BUILD
```diff
@@ -0,0 +1,21 @@
+load("//scala:scala.bzl", "scala_binary")
+load("//thrift:thrift.bzl", "thrift_library")
+load("//twitter_scrooge:twitter_scrooge.bzl", "scrooge_scala_library")
+
+java_import(
```
johnynek: this, in my opinion, should be `filegroup`: https://docs.bazel.build/versions/master/be/general.html#filegroup

this is not a java library (since it likely has no java bytecode inside it). If it does, we handle it with `remote_jar` in scrooge already.

test/src/main/scala/scala/test/twitter_scrooge/bare_jar_test/BUILD
```diff
@@ -0,0 +1,21 @@
...
+    name = "barejar",
+    jars = ["bare-thrift.jar"]
+)
+
+scrooge_scala_library(
+    name = "bare_thrifts",
+    visibility = ["//visibility:public"],
+    remote_self_jars = [":barejar"]
```
johnynek: can we put this on `thrift_library` instead? That makes it clearer the jar is only thrift and really has nothing to do with scrooge.

johnynek:
basically, I'd rather here:
https://github.com/bazelbuild/rules_scala/blob/master/thrift/thrift.bzl#L80

have something like an additional `external_jars` attribute that we aggregate up. Then in the scrooge rule, we can look at the external_jars and handle them similarly to how you have it.

test/src/main/scala/scala/test/twitter_scrooge/bare_jar_test/BUILD
mduvall: I see that makes sense, will change

test/src/main/scala/scala/test/twitter_scrooge/bare_jar_test/BUILD
mduvall: Yep happy to!

mduvall:
I put `external_jars` onto `thrift_library`, and do the JAR reading in `scrooge_scala_library` now, PTAL @johnynek?

thrift/thrift.bzl
```diff
@@ -78,6 +78,7 @@ rm -rf {out}_tmp"""
     thrift = struct(
       srcs = ctx.outputs.libarchive,
       transitive_srcs = transitive_srcs,
+      external_jars = ctx.attr.external_jars,
```
johnynek: I think you need to keep the transitive union of external_jars to dependencies work correctly. See what is happening with `_collect_thrift_srcs`.

also while at it, can we use `depset` vs `set`. `set` is a deprecated alias for `depset`:
https://docs.bazel.build/versions/master/skylark/lib/set.html

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -69,6 +69,16 @@ def _collect_owned_srcs(targets):
           r += target.scrooge_srcjar.transitive_owned_srcs
   return r
 
+def _collect_external_jars(targets):
+  r = set()
```
johnynek: should be a `depset` I think.

mduvall: Cool, didn't know about depset! will update other usages of `set()` as well

thrift/thrift.bzl
mduvall: Interesting, I'll add a failing test case and fix it, good catch

mduvall:
@johnynek I've done the `depset` change and added handling for transitive deps. I'm having some build shenanigans at the moment when running the build:

```
$ bazel test //test/...
ERROR: /Users/matt/workspace/rules_scala/test/src/main/scala/scala/test/twitter_scrooge/BUILD:66:1: in scrooge_scala_srcjar rule //test/src/main/scala/scala/test/twitter_scrooge:bare_thrift_scrooge_srcjar:
	File "/Users/matt/workspace/rules_scala/test/src/main/scala/scala/test/twitter_scrooge/BUILD", line 66
		scrooge_scala_srcjar(name = 'bare_thrift_scrooge_srcjar')
	File "/Users/matt/workspace/rules_scala/twitter_scrooge/twitter_scrooge.bzl", line 141, in _gen_scrooge_srcjar_impl
		_assert_set_is_subset(only_transitive_thrift_srcs, trans...)
	File "/Users/matt/workspace/rules_scala/twitter_scrooge/twitter_scrooge.bzl", line 105, in _assert_set_is_subset
		fail("scrooge_srcjar target must depe...))
scrooge_srcjar target must depend on scrooge_srcjar targets sufficient to cover the transitive graph of thrift files. Uncovered sources: depset([File:[[/private/var/tmp/_bazel_matt/fc140044b49349304f9eee84fd180cdd/execroot/io_bazel_rules_scala]bazel-out/local-fastbuild/bin]test/src/main/scala/scala/test/twitter_scrooge/thrift/bare_jar_thrifts/bare_jar_1/libbare_jar_1.jar, File:[[/private/var/tmp/_bazel_matt/fc140044b49349304f9eee84fd180cdd/execroot/io_bazel_rules_scala]bazel-out/local-fastbuild/bin]test/src/main/scala/scala/test/twitter_scrooge/thrift/bare_jar_thrifts/bare_jar_2/libbare_jar_2.jar]).
ERROR: Analysis of target '//test/src/main/scala/scala/test/twitter_scrooge:bare_thrift_scrooge_srcjar' failed; build aborted.
INFO: Elapsed time: 0.678s
ERROR: Couldn't start the build. Unable to run tests.
```

This seems like a regression I introduced in this PR...but I am stumped as to exactly what the assertion is trying to tell me here (the JARs do end up being available and the tests pass when commenting out https://github.com/mduvall/rules_scala/blob/9d6b797a91a22d5fca31601554bba3fee433b90b/twitter_scrooge/twitter_scrooge.bzl#L141). I have a minimal reproduction of the issue I'm having on this branch here: https://github.com/bazelbuild/rules_scala/compare/master...mduvall:md-canidep?expand=1. This is likely PEBKAC; I am guessing the code does not quite use the `thrift_library` deps correctly, any pointers/obvious wrong things here?

mduvall:
The error on that branch mentioned above is similar:

```
ERROR: /Users/matt/workspace/rules_scala/test/src/main/scala/scala/test/twitter_scrooge/canidep/BUILD:11:1: in scrooge_scala_srcjar rule //test/src/main/scala/scala/test/twitter_scrooge/canidep:canidep_srcjar:
	File "/Users/matt/workspace/rules_scala/test/src/main/scala/scala/test/twitter_scrooge/canidep/BUILD", line 11
		scrooge_scala_srcjar(name = 'canidep_srcjar')
	File "/Users/matt/workspace/rules_scala/twitter_scrooge/twitter_scrooge.bzl", line 127, in _gen_scrooge_srcjar_impl
		_assert_set_is_subset(only_transitive_thrift_srcs, trans...)
	File "/Users/matt/workspace/rules_scala/twitter_scrooge/twitter_scrooge.bzl", line 94, in _assert_set_is_subset
		fail("scrooge_srcjar target must depe...))
scrooge_srcjar target must depend on scrooge_srcjar targets sufficient to cover the transitive graph of thrift files. Uncovered sources: depset([File:[[/private/var/tmp/_bazel_matt/fc140044b49349304f9eee84fd180cdd/execroot/io_bazel_rules_scala]bazel-out/local-fastbuild/bin]test/src/main/scala/scala/test/twitter_scrooge/canidep/Goodbye/libgoodbye.jar]).
```

mduvall:
🤦‍♀️ ignore me I didn't catch that we had to build the accompanying scrooge classes as well! Will update the test.

mduvall:
Okay that was embarrassing, PTAL @johnynek?

johnynek:
Looks good.

mduvall:
\o/ thanks!

johnynek:
@mduvall thanks for pushing this through!

Excited to have one more person at Stripe that has some rules hacking experience.

PR #241, opened 2017-06-30
Organize implicit dependencies necessary for IDEs.

chaoren:
Add a separate attribute whose sole purpose is to propagate dependency
information via aspects.
This will allow the IntelliJ aspect to get every necessary implicit
dependency by just following _scala_toolchain.
This also allows us to avoid the issue of certain dependencies being
required to be used as files instead of java targets.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -620,6 +620,28 @@ _implicit_deps = {
   "_jdk": attr.label(default=Label("//tools/defaults:jdk"), allow_files=True),
 }
 
+# Single dep to allow IDEs to pickup all the implicit dependencies.
+_resolve_deps = {
+  "_scala_toolchain" : attr.label_list(default=[
+    Label("//external:io_bazel_rules_scala/dependency/scala/scala_library"),
+  ], allow_files=False),
+}
+
+_test_resolve_deps = {
+  "_scala_toolchain" : attr.label_list(default=[
+    Label("//external:io_bazel_rules_scala/dependency/scala/scala_library"),
+    Label("//external:io_bazel_rules_scala/dependency/scalatest/scalatest"),
+  ], allow_files=False),
+}
+
+_junit_resolve_deps = {
+  "_scala_toolchain" : attr.label_list(default=[
+    Label("//external:io_bazel_rules_scala/dependency/scala/scala_library"),
+    Label("//external:io_bazel_rules_scala/dependency/junit/junit"),
+    Label("//external:io_bazel_rules_scala/dependency/hamcrest/hamcrest_core"),
+  ], allow_files=False),
+}
+
 # Common attributes reused across multiple rules.
 _common_attrs = {
   "srcs": attr.label_list(
@@ -644,7 +666,7 @@ scala_library = rule(
   attrs={
       "main_class": attr.string(),
       "exports": attr.label_list(allow_files=False),
-      } + _implicit_deps + _common_attrs,
+      } + _implicit_deps + _common_attrs + _resolve_deps,
   outputs={
       "jar": "%{name}.jar",
       "deploy_jar": "%{name}_deploy.jar",
@@ -658,7 +680,7 @@ scala_macro_library = rule(
   attrs={
       "main_class": attr.string(),
       "exports": attr.label_list(allow_files=False),
-      } + _implicit_deps + _common_attrs,
+      } + _implicit_deps + _common_attrs + _resolve_deps,
   outputs={
       "jar": "%{name}.jar",
       "deploy_jar": "%{name}_deploy.jar",
@@ -670,7 +692,7 @@ scala_binary = rule(
   implementation=_scala_binary_impl,
   attrs={
       "main_class": attr.string(mandatory=True),
-      } + _launcher_template + _implicit_deps + _common_attrs,
+      } + _launcher_template + _implicit_deps + _common_attrs + _resolve_deps,
   outputs={
       "jar": "%{name}.jar",
       "deploy_jar": "%{name}_deploy.jar",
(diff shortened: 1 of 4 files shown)


scala/scala.bzl
```diff
@@ -620,6 +620,30 @@ _implicit_deps = {
   "_jdk": attr.label(default=Label("//tools/defaults:jdk"), allow_files=True),
 }
 
+# Single dep to allow IDEs to pickup all the implicit dependencies.
+_shared_resolve_deps = [
+  Label("//external:io_bazel_rules_scala/dependency/scala/scala_library"),
+  Label("//external:io_bazel_rules_scala/dependency/scala/scala_compiler"),
```
johnynek: these are not really implicit dependencies. I think only `scala_library` is. The other ones need to be added explicitly to classpaths generally.

ittaiz:
@chaoren If a user has a scala library and in its does it has a "some_external_scala_lib//jar:file" won't we have the same problem in the IDE?

chaoren:
> @chaoren If a user has a scala library and in its does it has a "some_external_scala_lib//jar:file" won't we have the same problem in the IDE?

Yes we will. This is only a bandage for the implicit deps which goes on every rule. The bigger problem of needing to avoid jars for macros still has to be fixed. Though the user might be able to workaround this by depending on the jar in addition to the file.

scala/scala.bzl
chaoren: Done.

test/aspect/aspect.bzl
```diff
@@ -1,49 +1,56 @@
...
+                 ctx.attr.scala_test,
+                 ctx.attr.scala_junit_test]:
+        expected = deps[rule.label.name]
+        if rule.visited != expected:
+            content += """
+            echo Expected these deps from {name}: 1>&2
+            echo {expected}, 1>&2
+            echo but got these instead: 1>&2
+            echo {visited} 1>&2
+            false
```
ittaiz: Does "false" make the test fail? if so I'd really like to learn why, if not what makes the test fail?

test/aspect/aspect.bzl
```diff
@@ -1,49 +1,56 @@
...
-        content = "true"
-    else:
-        content = """
-        echo Expected these rules to be visited by the aspect: 1>&2
-        echo %s, 1>&2
-        echo but got these instead: 1>&2
-        echo %s 1>&2
-        false
-        """ % (', '.join(expected), ', '.join(visited))
+    deps = {
```
ittaiz: is this `expected_deps`?

test/aspect/aspect.bzl
```diff
@@ -52,7 +59,11 @@ def _rule_impl(ctx):
 
 aspect_test = rule(
     implementation = _rule_impl,
-    attrs = { "scala_rule" : attr.label(aspects = [test_aspect]) },
+    attrs = {
+        "scala_test" : attr.label(aspects = [test_aspect]),
+        "scala_library" : attr.label(aspects = [test_aspect]),
```
ittaiz: It took me some time to understand the motivation for these attributes.
WDYT about renaming them to include some context? maybe `scala_library_test_target` etc

chaoren:
> WDYT we should do with the specs2 deps? They are the most volatile and aren't included here.

Can you please list them for me?

> Rule maintainers will need to maintain both the regular deps and these ones since they are duplicates, right?

Yes. Though these are just the ones you need to resolve the source code. So it's only a subset of the regular deps.

> This PR isn't really related to the whole "...//jar:file" problem since none of these dependencies is being used as a file target.

Well that problem (on implicit deps) is apparently unique to your private repo, where you do make that change. But yes, in this public repo, it just helps consolidate all the resolve deps into one, so we don't break the aspect every time something gets added or moves.

test/aspect/aspect.bzl
chaoren: It's used below as: `expected = deps[rule.label.name]`
Do you just not like the name?

test/aspect/aspect.bzl
chaoren: Yes. Pass/fail of the test is based on return value. False returns 1, and if it's the last line in the script, then the script will also return 1, and be treated as a failure. Compared to return/exit it allows us to append commands at the end.
I'll add a comment explaining it.

test/aspect/aspect.bzl
chaoren: I'll just add a comment to clarify.

chaoren:
> > WDYT we should do with the specs2 deps? They are the most volatile and aren't included here.

> Can you please list them for me?

Is it everything under `specs2_dependencies()` + `specs2_junit_dependencies()`?

ittaiz:
Yes it is.

test/aspect/aspect.bzl
ittaiz: I'd appreciate it if you change it to expected_deps. When I scanned the code it took some time until I noticed this usage

chaoren: Done.

test/aspect/aspect.bzl
chaoren: Done.

test/aspect/aspect.bzl
chaoren: Done.

chaoren:
> WDYT we should do with the specs2 deps? They are the most volatile and aren't included here. They are also important since we want to turn the junit one to a macro like they are so maybe we should solve how macros can pass that as well.

We can't modify _scala_toolchain from macros, so there's no way to add them there. It looks like specs2 just adds libraries to regular deps. Our aspect always follows regular deps, so if all the deps are jars, then it wouldn't be a problem. In the case of your private repo, it'll just look like a regular target with file deps, which I don't think we can handle.

chaoren:
@sdtwigg may be interested in this change too.

PR #242, opened 2017-07-01
Use workspace name in tut rule

johnynek:
when we load this rule in a remote repo, this label looks like a local label. We need to use the full name of the target.

### DIFF
--- a/tut_rule/tut.bzl
+++ b/tut_rule/tut.bzl
@@ -23,7 +23,7 @@ def scala_tut_doc(**kw):
     name = tool,
     main_class = "io.bazel.rules_scala.tut_support.TutCompiler",
     deps = deps + [
-      "//src/scala/io/bazel/rules_scala/tut_support:tut_compiler_lib"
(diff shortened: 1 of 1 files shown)


tut_rule/tut.bzl
```diff
@@ -23,7 +23,7 @@ def scala_tut_doc(**kw):
     name = tool,
     main_class = "io.bazel.rules_scala.tut_support.TutCompiler",
     deps = deps + [
-      "//src/scala/io/bazel/rules_scala/tut_support:tut_compiler_lib"
+      "@io_bazel_rules_scala//src/scala/io/bazel/rules_scala/tut_support:tut_compiler_lib"
```
ittaiz: I think we tried this approach and realized this means that now every user has to use this exact name for the rules scala repository, doesn't it? If so do we have an alternative?

johnynek:
The workspace declares the name and it is an error (or will be soon) for
the name to not match (import vs declaration). I think this is safe.

johnynek:
PS: it used to not work because you couldn't refer to the current name, but
that seems to be fixed.

ittaiz:
Yeah, remember that it used o not work :) fair enough.

PR #244, opened 2017-07-04
add recommended setup for intellij .bazelproject to new scripts folder

natansil:
to help contributers who work on Intellij


bazel-io:

johnynek:
This looks good to me. What is it about Scrooge that doesn't work? I wonder if there is something we can fix.

ittaiz:
We don't really know. On one of the IntelliJ issues we opened @chaoren said
it's not supported

chaoren:
> On one of the IntelliJ issues we opened @chaoren said it's not supported

By that I mean I didn't get to add any special support for it. It it's just a bunch of libraries (jars not files) attached to scala rules (library, binary, test, etc), then it should work out of the box.

johnynek:
I'm happy to merge this, but I don't use intellij so I can't really evaluate it very well. I'll leave to @ittaiz to merge.

PR #245, opened 2017-07-06
Fix issue with empty args and workers

johnynek:
when workers are enabled, it seems empty lines are removed from the arguments list. This PR makes sure to always have 1 line per arg (by padding with a leading `_` which is discarded).

This is actually currently a blocker for us. I would love to have better tests around this (that exercise everything with and without workers, but I don't want to try to shave that yak before we get this build issue unblocked).

### DIFF
--- a/src/scala/scripts/TwitterScroogeGenerator.scala
+++ b/src/scala/scripts/TwitterScroogeGenerator.scala
@@ -41,7 +41,7 @@ class ScroogeGenerator extends Processor {
 
   def processRequest(args: java.util.List[String]) {
     def getIdx(i: Int): List[String] =
-      if (args.size > i) args.get(i).split(':').toList.filter(_.nonEmpty)
+      if (args.size > i) args.get(i).drop(1).split(':').toList.filter(_.nonEmpty)
       else Nil
 
     val jarOutput = args.get(0)
--- a/src/scala/scripts/TwitterScroogeGenerator.scala
+++ b/src/scala/scripts/TwitterScroogeGenerator.scala
@@ -40,9 +40,20 @@ class ScroogeGenerator extends Processor {
     }
 
   def processRequest(args: java.util.List[String]) {
-    def getIdx(i: Int): List[String] =
-      if (args.size > i) args.get(i).drop(1).split(':').toList.filter(_.nonEmpty)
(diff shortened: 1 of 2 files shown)


johnynek:
added #246

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -140,7 +140,7 @@ def _gen_scrooge_srcjar_impl(ctx):
   # in order to generate code) have targets which will compile them.
   _assert_set_is_subset(only_transitive_thrift_srcs, transitive_owned_srcs)
 
-  path_content = "\n".join([_colon_paths(ps) for ps in [immediate_thrift_srcs, only_transitive_thrift_srcs, remote_jars, external_jars]])
+  path_content = "\n".join(["_" + _colon_paths(ps) for ps in [immediate_thrift_srcs, only_transitive_thrift_srcs, remote_jars, external_jars]])
```
ittaiz: Can you extract this into a constant with a clear name? Something like PaddingToPreserverAruent or PaddingToSeparateArgumentToOwnLine? Names are different once I'm really not sure I got the point well enough...

src/scala/scripts/TwitterScroogeGenerator.scala
```diff
@@ -41,7 +41,7 @@ class ScroogeGenerator extends Processor {
 
   def processRequest(args: java.util.List[String]) {
     def getIdx(i: Int): List[String] =
-      if (args.size > i) args.get(i).split(':').toList.filter(_.nonEmpty)
+      if (args.size > i) args.get(i).drop(1).split(':').toList.filter(_.nonEmpty)
```
ittaiz: Can we extract 1 into a constant as well to express why we're dropping it? Something like .drop(PaddingCharacterUsedForArgumentPreservation)
I know the names are long but honestly I fear this will cryptic code really fast

johnynek:
Sounds good. Will update and add links to the issue I filed on bazel.

ittaiz:
LGTM. I'd wait for green just to be on the safe side. Going to a company party so you can merge when you want :)

johnynek:
@ittaiz thanks for the review, sir!

PR #247, opened 2017-07-07
Improve tut rule

johnynek:
We saw two issues which I'm not sure how we didn't catch before:

1) the tmp directory is (sometimes?) outside of bazel's control, so sandboxing fails in one case we have (I thought the build on bazel ci would have caught that).

2) the name for the tut rule is highly unintuitive and not the name you write, so you probably can only build it with a `...` pattern, which is annoying.

### DIFF
--- a/src/scala/io/bazel/rules_scala/tut_support/TutCompiler.scala
+++ b/src/scala/io/bazel/rules_scala/tut_support/TutCompiler.scala
@@ -7,8 +7,7 @@ import tut.TutMain
 
 object TutCompiler {
   def main(args: Array[String]): Unit = {
-    val tmp = Paths.get(Option(System.getenv("TMPDIR")).getOrElse("/tmp"))
-    val mdOutput = Files.createTempDirectory(tmp, "tut")
+    val mdOutput = Files.createTempDirectory("tut")
     val outfile = args(1)
     val classpath = System.getProperty("java.class.path")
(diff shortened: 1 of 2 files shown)


ittaiz:
LGTM.
At the risk of sounding like a broken record I really think tests would help here.
1. Without a repro we might break it again.
2. I love using a test to see how comfortable (or not) my api is. I think a test *might* have shown you the clumsiness of the name.

You can merge as is if you want.

johnynek:
actually, we couldn't repro locally, but still I wanted to remove the defensive approach. Bazel always sets `java.tmp.dir`, it seems.

johnynek:
by the way, I *love* tests, but to be clear, testing a lot of these things is very laborious. The fact is, we don't actually have a build team working on this. I take a moment away from ML infrastructure to try to fix issues that are biting us.

It would be great to have negative and positive tests for each of the important features. I highly support it, at the same time, it may often be the choice of merging small changes that appear to improve undertested items (note we do have tests, which passed for this, but exercising the case that wound up with the TMPDIR not being writable, we still don't know how to repro).

The next thing is purely convention: we want the name of the resulting tut doc to be what you give, not `name__tut`. We could test that, but the risk seems pretty minor.

The problem with this rule is that prior to us using it lately, I'm pretty sure this rule had no users.

PR #254, opened 2017-07-26
Rename set to depset

vladmos:


ittaiz:
Thanks!

johnynek:
Wait, this was red in CI, right?

vladmos:
It looks like flakiness or misconfiguration: test_run.sh: line 210: bazel: command not found

I don't see how this change can break anything, `set` and `depset` are completely equivalent and `depset` was introduced months ago.

ittaiz:
This was red in CI because Bazel removed the ability to test against HEAD.
That's also why master is broken.
Or will send a PR later to susped HEAD testing from travis.
We're waiting for
https://github.com/bazelbuild/continuous-integration/issues/95

ittaiz:
@johnynek @or-shachar noticed that they returned the binaries to the pipeline so I manually triggered the build and master is green again :)

PR #257, opened 2017-08-03
Scala Strict Deps toggling via cmd line flag

ittaiz:
Scala strict deps can now be turned on (error/warn) via cmd line flag which makes it actually usable!
There are still some question marks about the performance which is why this is still turned off by default for now.

Details:
upgraded bazel to 0.5.3 since that exposes strict java deps cmd line flag to skylark
many BUILD files were removed since the toggling can be done in the test via the cmd line flag and the same BUILD files can be reused.
weird error mode test is no longer needed since validation is done by bazel on the cmd line flag
error mode test is not needed since it's tested by other tests
warn doesn't have a unit-test so is still needed
off/default drive skylark code so they're needed as e2es


johnynek:
It's a little weird to use `--strict-java-deps` rather than `scala`, but I guess skylark does not give you the ability to add new flags.

Seems fine to me.

👍

ittaiz:
Exactly. I opened an issue to add a "strict_jvm_deps" flag. No real hopes
on it getting in but I'm trying.

scala/scala.bzl
```diff
@@ -782,10 +782,9 @@ _common_attrs_for_plugin_bootstrapping = {
 }
 
 _common_attrs = _common_attrs_for_plugin_bootstrapping + {
-  # dependency_analyzer_mode_soon_to_be_removed will be replaced by using command line flag called 'strict_java_deps' (https://github.com/bazelbuild/bazel/issues/3295)
+  # using stricts scala deps is done by using command line flag called 'strict_java_deps'
```
natansil: > using stricts scala deps is done by using command line flag called 'strict_java_deps' 
switching mode to "on" means...

I don't think this is clear enough, I think language should reflect how to use the command line flag to turn the feature on

scala/scala.bzl
```diff
@@ -464,15 +464,15 @@ def _collect_jars(dep_targets, dependency_analyzer_is_off = True):
     else:
       return _collect_jars_when_dependency_analyzer_is_on(dep_targets)
 
-def is_dependency_analyzer_off(ctx):
-  if not hasattr(ctx.attr, 'dependency_analyzer_mode_soon_to_be_removed'):
+def is_dependency_analyzer_on(ctx):
+  if (hasattr(ctx.attr,"_dependency_analyzer_plugin")
+    #when the strict deps FT is removed this will be removed as well
```
natansil: I think this comment is a bit confusing.
The FT is this line (line 471)...
Unless I am missing something which other line will be deleted when FT will be removed?

test_run.sh
```diff
@@ -68,15 +68,16 @@ test_expect_failure_or_warning_on_missing_direct_deps_with_expected_message() {
 
   expected_message=$1
   test_target=$2
-  operator=${3:-"eq"}
+  strict_deps_mode=$3
+  operator=${4:-"eq"}
```
natansil: WDYT about adding an alias to eq, ne
something like expecting_failure, expecting_success

scala/scala.bzl
ittaiz: I agree but since this is not really documented (i.e. not in readme) I'm good with this ambiguity.

test_run.sh
ittaiz: I want to delete this in general but it's too much of a hassle :(

scala/scala.bzl
ittaiz: thanks, updated the phrashing

PR #258, opened 2017-08-08
moved plugin under third_party in ij.bazelproject

ittaiz:
plugin is no more under the root


ittaiz:
@natansil please review

natansil:
👍  Thanks for the fix

PR #259, opened 2017-08-08
printing stderr / stdout of javac

ittaiz:
@johnynek I just spent some time trying to debug a failing javac issue and decided that this small change is good even though it's not the actual solution (which is moving to `java_common.compile`).

WDYT?

### DIFF
--- a/src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
+++ b/src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
@@ -270,15 +270,24 @@ class ScalacProcessor implements Processor {
     commandParts.add("@" + normalizeSlash(argsFile.toFile().getAbsolutePath()));
     try {
       Process iostat = new ProcessBuilder(commandParts)
-        .inheritIO()
         .start();
+      String error = getInputAsString(iostat.getErrorStream());
+      String output = getInputAsString(iostat.getInputStream());
       int exitCode = iostat.waitFor();
       if(exitCode != 0) {
-        throw new RuntimeException("javac process failed!");
+        throw new RuntimeException("javac process failed! "+ error + " " + output);
+      } else {
+        System.out.println("Javac stdout: " + output);
       }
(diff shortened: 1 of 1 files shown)


src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
```diff
@@ -270,18 +270,27 @@ private static void compileJavaSources(CompileOptions ops, String[] javaSources,
...
-        .inheritIO()
         .start();
+      String error = getInputAsString(iostat.getErrorStream());
+      String output = getInputAsString(iostat.getInputStream());
       int exitCode = iostat.waitFor();
       if(exitCode != 0) {
-        throw new RuntimeException("javac process failed!");
+        throw new RuntimeException("javac process failed! "+ error + " " + output);
+      } else {
+        System.out.println("Javac stdout: " + output);
```
johnynek: do we need this in the general case? I'm afraid it would add a bunch of noise to many builds.

ittaiz:
General case where you have java sources right?
And I meant to catch the warnings. Do you think they're printed to stderr?

johnynek:
if there is no output, should we maybe output nothing? I think this bazel does not expect compilers to print things in the usual case.

softprops:
What would this output look like volume-wise? We have a **** ton of java mixed in with our scala  so also have a fear that this would add a lot of noise to our build logs. That said, I'd be happy for to opt into a setting for more verbose output. I think there's already a rule attr to print timing information. Perhaps a setting to print more compiler output would be in line with that.

ittaiz:
I think this PR is moot since I realized I'm pretty much blocked by us not
using java_common.compile (I need javac to use strict java deps) and that
change will solve the logging part.
I'll see how I make progress with it and update here.
Worst case I'll drop the "happy path" print and just keep the error one.

ittaiz:
superseded by moving to java_common.compile

PR #260, opened 2017-08-09
handling none ijar case

ittaiz:
If ijar is none that means dep target does not contain ijar.
This happens to us when working on intellij integration (which we'll contribute back in a few weeks).
Essentially the intellij plugin needs the `scala_import` rule to set ijar as None so that it will take the `class_jar`. This is important for the whole macros handling.
This is also why there isn't a test for this right now. 
This will be covered when we contribute back the `scala_import`
@natansil I'd appreciate your review


natansil:
👍  Looks good.
I think breaking it into two lines will be even better.

PR #262, opened 2017-08-15
scrooge generator: add an option to enable finagle compilation (fixes #117)

### LINKED ISSUE #117: Scrooge --finagle
Is it possible to generate finagle services with current scrooge support? (like passing --finagle in sbt plugin)

kevingessner:
First commit fixes #117. Second is for forward compatibility with scrooge, as I use a newer version here.

### DIFF
--- a/src/scala/scripts/TwitterScroogeGenerator.scala
+++ b/src/scala/scripts/TwitterScroogeGenerator.scala
@@ -1,6 +1,7 @@
 package scripts
 
 import io.bazel.rules_scala.scrooge_support.{ Compiler, CompilerDefaults }
+import com.twitter.scrooge.backend.WithFinagle
 import io.bazel.rulesscala.jar.JarCreator
 import io.bazel.rulesscala.io_utils.DeleteRecursively
 import java.io.{ File, PrintStream }
@@ -69,10 +70,17 @@ class ScroogeGenerator extends Processor {
     // immediateThriftSrcJars.
     val remoteSelfThriftSources = getIdx(4)
 
+    // Further configuration options for scrooge.
+    val additionalFlags = getIdx(5)
+
     val tmp = Paths.get(Option(System.getenv("TMPDIR")).getOrElse("/tmp"))
     val scroogeOutput = Files.createTempDirectory(tmp, "scrooge")
 
     val scrooge = new Compiler
+
+    if (additionalFlags.contains("--with-finagle")) {
+      scrooge.flags += WithFinagle
+    }
 
     scrooge.compileJars ++= immediateThriftSrcJars
     scrooge.compileJars ++= remoteSelfThriftSources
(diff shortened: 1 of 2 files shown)


bazel-io:

kevingessner:
I signed it!  Via my company's Corporate CLA group.

kevingessner:
I signed it!  Take 2.

src/scala/scripts/TwitterScroogeGenerator.scala
```diff
@@ -97,6 +98,7 @@ class ScroogeGenerator extends Processor {
     scrooge.destFolder = scroogeOutput.toString
     //TODO we should make this configurable
     scrooge.strict = false
+    scrooge.flags += WithFinagle
```
kevingessner: OK!  I'll make that change.

src/scala/io/bazel/rules_scala/scrooge_support/Compiler.scala
```diff
@@ -128,7 +128,7 @@ class Compiler {
           defaultOptional = isJava,
           skipIncludes = false,
           documentCache
-        )(com.twitter.logging.NullLogger) // scrooge warns on file names with "/"
```
kevingessner: I removed it because I'm running a newer version of scrooge than 4.6.0 (4.15.0, via my project's WORKSPACE).  Everything works fine with the newer version except this:  in 4.7, scrooge [removed its dependency on twitter.util](https://github.com/twitter/scrooge/commit/de7e1d1c826d1549f6992ce5d68c444bc0720563), so this is java Logger now.  Removing this was the simplest way to make Compiler work with both versions of scrooge.

I can split this into a separate PR, if you'd prefer -- I think I'll upgrade rules_scala's scrooge to 4.7.0 and change this to a null logging implementation in the new interface, so it will work with newer versions as well.  Does that sound good?

johnynek: ahh, I see.. :/ I guess so, I just hope it does spit a ton of warnings out on us.

kevingessner:
@johnynek New patch with finagle support as an option.  New PR incoming with the logging change separate.

src/scala/scripts/TwitterScroogeGenerator.scala
```diff
@@ -69,11 +70,18 @@ class ScroogeGenerator extends Processor {
...
 
+    // Further configuration options for scrooge.
+    val additionalFlags = getIdx(5)
+
     val tmp = Paths.get(Option(System.getenv("TMPDIR")).getOrElse("/tmp"))
     val scroogeOutput = Files.createTempDirectory(tmp, "scrooge")
 
     val scrooge = new Compiler
 
+    if (additionalFlags.contains("--with-finagle")) {
```
kevingessner: This should give us an easy spot to add additional flags, without bringing in a full-fledged argument parser.

johnynek:
Nice! Thanks!

PR #264, opened 2017-08-18
scrooge compiler: disable logging via a method, not a NullLogger instance

kevingessner:
Versions of scrooge newer than 4.6.0 use java.util.logging directly, instead of com.twitter.logging.  The NullLogger instance is therefore incompatible with newer versions.  By changing to use setLevel, the logging is disabled for any logger: that method is the same on both the twitter and java loggers, and compiles correctly with more versions of scrooge.

I'm using scrooge 4.15.0 instead of the built-in 4.6.0, and everything works great once I have this change.


bazel-io:

PR #268, opened 2017-08-27
add support for rule customization scala test color and stack trace o…

softprops:
…utput

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -660,6 +660,17 @@ trap finish EXIT
 
   return _scala_binary_common(ctx, cjars, transitive_rjars, jars.transitive_compile_jars, jars.jars2labels)
 
+def _scala_test_flags(ctx):
+    # output report test duration
+    flags = "-oD"
+    if ctx.attr.full_stacktraces:
+        flags += "F"
+    else:
+        flags += "S"
+    if not ctx.attr.colors:
+        flags += "W"
+    return flags
+
 def _scala_test_impl(ctx):
     if len(ctx.attr.suites) != 0:
         print(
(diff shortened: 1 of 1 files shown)


bazel-io:

softprops:
Looking for some input on what a good test would look like for this

scala/scala.bzl
```diff
@@ -852,6 +860,8 @@ scala_test = rule(
   attrs={
       "main_class": attr.string(default="io.bazel.rulesscala.scala_test.Runner"),
       "suites": attr.string_list(),
+      "colors": attr.bool(default=False),
+      "full_stacktraces": attr.bool(default=False),
```
johnynek: I'm happy making both of these default to `True` and let people opt out, personally.

scala/scala.bzl
```diff
@@ -987,11 +997,12 @@ def _sanitize_string_for_usage(s):
 # This auto-generates a test suite based on the passed set of targets
 # we will add a root test_suite with the name of the passed name
 def scala_test_suite(name, srcs = [], deps = [], runtime_deps = [], data = [], resources = [],
-                     scalacopts = [], jvm_flags = [], visibility = None, size = None):
+                     scalacopts = [], jvm_flags = [], visibility = None, size = None,
+                     colors=False, full_stacktraces=False):
```
johnynek: again, I'm happy for these to default to true.

scala/scala.bzl
softprops: Me too actually. We'd opt to enable these as our internal defaults either way, but I wasn't sure how much pushback I'd get from changing existing behavior if I did. Looks like not much :)

scala/scala.bzl
```diff
@@ -682,9 +682,17 @@ def _scala_test_impl(ctx):
 
     transitive_rjars += [ctx.outputs.jar]
 
+    # output report test duration
```
ittaiz: Any chance you can extract this block into a "flags" method? This will decrease the likelihood of us having a collision.

softprops: absolutely. I eta on that is probably within the next day or so

softprops:
Couldn't help noticing travis build was failing. I don't _think_ it's related to these changes though

bazel command not found https://travis-ci.org/bazelbuild/rules_scala/jobs/268985675#L574

ittaiz:
no, our HEAD build is broken again.
It's against a "pirate" url so it keeps being broken.
I'm good with merging if 0.5.3 is passing but Oscar might have some
reservations.

softprops:
I'd like to follow up on making the more dev friendly settings the default if that's okay. I can also follow up with factoring out a python method for the flags if that makes it less likely to conflict with the other branch

softprops:
updated pull with factored out flag func and more useful defaults for test colors and stacktraces

softprops:
Bump

johnynek:
Thanks @softprops !

merged side stepping the HEAD issue which is persistent.

softprops:
Thanks @johnynek

Issue #164, opened 2017-03-09
Move from manual javac to java_common.compile


ittaiz:
Hi,
I'm trying to tackle moving from a manual `javac` to `java_common.compile` as discussed in https://github.com/bazelbuild/bazel/issues/970#issuecomment-284335067.
Should we just return two jars? If so any preference to the names? Do you think anyone counts on it being one jar? We can unzip and jar both of them back together but it's messy.
Other than that it looks pretty straightforward to me by just obliterating all javac references, calling `java_common.compile` at the end of `_compile(` and propagate the providers out

johnynek:
I think we will have re-architect the rules or have one output. We at least have to combine to make the _deploy.jar. But if we emit two jars then targets that consume this one will have to add both jars to their inputs (since the whole idea is to use this when there are circular dependencies).

We already have jar creation code, we can just make it work for jar concatenation without unjarring first. It's slightly more IO but I bet not st all the slow part compared to scalac.

ittaiz:
the problem with joining the jars is that currently this happens in the worker and the java_common will happen at skylark level. 
AFAIU we have two options:
1. Do the jar concatenation in bash on skylark
2. Change the worker to be able to get two types of requests and add a concat jars request type
3. Call JarCreator (or a variant) plainly without a worker after java_common and start a new jvm.

I think #1 sounds least expensive performance wise since IINM this problem isn't jvm related but just archive manipulation.
WDYT?

johnynek:
The problem with bash is that you have to jump through all kinds of hoops
to make it a reproducible jar (same order contents, no timestamps).

It would be nice to have a C++ implementation or jar in bazel that did the
right thing.

ittaiz:
Right...
So wdyt about having the worker support two type of commands?

johnynek:
I think that could be fine. Since it is still related to compiling scala.

So, we could add some optional args to the scalac worker to just act as a jar concatenator.

ittaiz:
Ok, any tips on how I should do that? Never worked with protobuf and the
current worker protocol feels a bit implicit to me.

johnynek:
you can look at the current invocation code. We have abstracted all the
protobuf out. You get a list of arguments that we then decode to pass to
scalac.

You will just need to add a new argument to our key-value approach.

sdtwigg:
How far along are you with this approach?

[...]

PR #269, opened 2017-08-28
Moved to java_common.compile

ittaiz:
Was supposed to be a refactoring but found a few missing tests.
The good news are that we indeed have good coverage since this hits many code paths and the tests gave me good (though long) feedback about what I'm breaking.

Possible open issues:
1) I haven't changed the "files" being returned- No tests demanded it and I'm trying to understand what needs this. I think our internal aspect actually will be broken by this so I'll try to add a test like it today and see if it breaks.
2) Not sure Intellij integration continues to work now- will see how I triage this. @chaoren can the intellij aspect try to use the java_provider before falling back to scala attr?

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -149,7 +149,7 @@ def _collect_plugin_paths(plugins):
     return paths
 
 
-def _compile(ctx, cjars, dep_srcjars, buildijar, transitive_compile_jars=[], labels = {}):
+def _compile(ctx, cjars, dep_srcjars, buildijar, transitive_compile_jars, labels, implicit_junit_deps_needed_for_java_compilation):
     ijar_output_path = ""
     ijar_cmd_path = ""
     if buildijar:
@@ -289,15 +289,16 @@ DependencyAnalyzerMode: {dependency_analyzer_mode}
         arguments=["--jvm_flag=%s" % flag for flag in ctx.attr.scalac_jvm_flags] + ["@" + argfile.path],
       )
 
-    java_jar = try_to_compile_java_jar(ctx, all_srcjars, java_srcs)
+    java_jar = try_to_compile_java_jar(ctx, all_srcjars, java_srcs, implicit_junit_deps_needed_for_java_compilation)
     return java_jar
 
 
-def try_to_compile_java_jar(ctx, all_srcjars, java_srcs):
+def try_to_compile_java_jar(ctx, all_srcjars, java_srcs, implicit_junit_deps_needed_for_java_compilation):
     if not java_srcs and not all_srcjars:
       return False
 
     providers_of_dependencies = collect_java_providers_of(ctx.attr.deps)
+    providers_of_dependencies += collect_java_providers_of(implicit_junit_deps_needed_for_java_compilation)
     scala_sources_java_provider = java_common.create_provider(
         #the line below means we can't use the provider from java_common.compile
         #since it will include the full jar when I want to expose only the ijar
@@ -331,14 +332,14 @@ def collect_java_providers_of(deps):
           providers.append(dep[java_common.provider])
     return providers
 
-def _compile_or_empty(ctx, jars, srcjars, buildijar, transitive_compile_jars, jars2labels):
+def _compile_or_empty(ctx, jars, srcjars, buildijar, transitive_compile_jars, jars2labels, implicit_junit_deps_needed_for_java_compilation):
     # We assume that if a srcjar is present, it is not empty
     if len(ctx.files.srcs) + len(srcjars) == 0:
         _build_nosrc_jar(ctx, buildijar)
         #  no need to build ijar when empty
         return struct(ijar=ctx.outputs.jar, class_jar=ctx.outputs.jar, java_jar = False)
     else:
-        java_jar = _compile(ctx, jars, srcjars, buildijar, transitive_compile_jars, jars2labels)
+        java_jar = _compile(ctx, jars, srcjars, buildijar, transitive_compile_jars, jars2labels, implicit_junit_deps_needed_for_java_compilation)
         ijar = None
         if buildijar:
             ijar = ctx.outputs.ijar
@@ -541,7 +542,7 @@ def _lib(ctx, non_macro_lib):
     (cjars, transitive_rjars) = (jars.compile_jars, jars.transitive_runtime_jars)
 
     write_manifest(ctx)
-    outputs = _compile_or_empty(ctx, cjars, srcjars, non_macro_lib, jars.transitive_compile_jars, jars.jars2labels)
+    outputs = _compile_or_empty(ctx, cjars, srcjars, non_macro_lib, jars.transitive_compile_jars, jars.jars2labels, [])
 
     transitive_rjars += [ctx.outputs.jar]
     if outputs.java_jar:
@@ -621,9 +622,9 @@ def _scala_macro_library_impl(ctx):
   return _lib(ctx, False)  # don't build the ijar for macros
 
 # Common code shared by all scala binary implementations.
-def _scala_binary_common(ctx, cjars, rjars, transitive_compile_time_jars, jars2labels):
+def _scala_binary_common(ctx, cjars, rjars, transitive_compile_time_jars, jars2labels, implicit_junit_deps_needed_for_java_compilation = []):
   write_manifest(ctx)
-  outputs = _compile_or_empty(ctx, cjars, [], False, transitive_compile_time_jars, jars2labels)  # no need to build an ijar for an executable
+  outputs = _compile_or_empty(ctx, cjars, [], False, transitive_compile_time_jars, jars2labels, implicit_junit_deps_needed_for_java_compilation)  # no need to build an ijar for an executable
   rjars += [ctx.outputs.jar]
   if outputs.java_jar:
     rjars += [outputs.java_jar]
@@ -659,6 +660,7 @@ def _scala_binary_common(ctx, cjars, rjars, transitive_compile_time_jars, jars2l
       providers = [java_provider],
       scala = scalaattr,
       transitive_rjars = rjars, #calling rules need this for the classpath in the launcher
+      java_jar = outputs.java_jar,
       runfiles=runfiles)
 
 def _scala_binary_impl(ctx):
@@ -752,12 +754,19 @@ def _scala_test_impl(ctx):
     )
     return out
 
-def _gen_test_suite_flags_based_on_prefixes_and_suffixes(ctx, archive):
+def _gen_test_suite_flags_based_on_prefixes_and_suffixes(ctx, scala_archive, maybe_java_archive):
+    archives = _get_archives_short_path(scala_archive, maybe_java_archive)
     return struct(testSuiteFlag = "-Dbazel.test_suite=io.bazel.rulesscala.test_discovery.DiscoveredTestSuite",
-    archiveFlag = "-Dbazel.discover.classes.archive.file.path=%s" % archive.short_path,
+    archiveFlag = "-Dbazel.discover.classes.archives.file.paths=%s" % archives,
     prefixesFlag = "-Dbazel.discover.classes.prefixes=%s" % ",".join(ctx.attr.prefixes),
     suffixesFlag = "-Dbazel.discover.classes.suffixes=%s" % ",".join(ctx.attr.suffixes),
     printFlag = "-Dbazel.discover.classes.print.discovered=%s" % ctx.attr.print_discovered_classes)
+
+def _get_archives_short_path(scala_archive, maybe_java_archive):
+  java_archive_short_path = ""
+  if maybe_java_archive:
+    java_archive_short_path = "," + maybe_java_archive.short_path
+  return scala_archive.short_path + java_archive_short_path
 
 def _scala_junit_test_impl(ctx):
     if (not(ctx.attr.prefixes) and not(ctx.attr.suffixes)):
@@ -766,9 +775,10 @@ def _scala_junit_test_impl(ctx):
         extra_deps = [ctx.attr._junit, ctx.attr._hamcrest, ctx.attr._suite, ctx.attr._bazel_test_runner],
     )
     (cjars, transitive_rjars) = (jars.compile_jars, jars.transitive_runtime_jars)
+    implicit_junit_deps_needed_for_java_compilation = [ctx.attr._junit, ctx.attr._hamcrest]
 
-    out =  _scala_binary_common(ctx, cjars, transitive_rjars, jars.transitive_compile_jars, jars.jars2labels)
-    test_suite = _gen_test_suite_flags_based_on_prefixes_and_suffixes(ctx, ctx.outputs.jar)
+    out =  _scala_binary_common(ctx, cjars, transitive_rjars, jars.transitive_compile_jars, jars.jars2labels, implicit_junit_deps_needed_for_java_compilation)
+    test_suite = _gen_test_suite_flags_based_on_prefixes_and_suffixes(ctx, ctx.outputs.jar, out.java_jar)
     launcherJvmFlags = ["-ea", test_suite.archiveFlag, test_suite.prefixesFlag, test_suite.suffixesFlag, test_suite.printFlag, test_suite.testSuiteFlag]
     _write_launcher(
         ctx = ctx,
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -289,15 +289,56 @@ DependencyAnalyzerMode: {dependency_analyzer_mode}
         arguments=["--jvm_flag=%s" % flag for flag in ctx.attr.scalac_jvm_flags] + ["@" + argfile.path],
       )
 
+    java_jar = try_to_compile_java_jar(ctx, all_srcjars, java_srcs)
+    return java_jar
+
+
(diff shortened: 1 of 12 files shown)


ittaiz:
fixes #164

ittaiz:
@or-shachar @natansil I'd appreciate your review as well please

ittaiz:
Travis build is green for 0.5.3
HEAD builds are broken again because of travis CI

scala/scala.bzl
```diff
@@ -599,9 +650,14 @@ def _scala_binary_common(ctx, cjars, rjars, transitive_compile_time_jars, jars2l
...
       deploy_jar=ctx.outputs.deploy_jar,
   )
+
+  compile_outputs = [outputs.class_jar]
+  if outputs.java_jar:
+    compile_outputs.extend([outputs.java_jar.ijar])
+
   scalaattr = struct(
       outputs = rule_outputs,
-      compile_jars = depset([outputs.class_jar]),
```
ittaiz: @johnynek is there any reason you use here `outputs.class_jar` and in `_lib` you use `ctx.outputs.jar`?
I did a few prints and they seem identical.

scala/scala.bzl
```diff
@@ -614,31 +670,33 @@ def _scala_binary_common(ctx, cjars, rjars, transitive_compile_time_jars, jars2l
       files=depset([ctx.outputs.executable]),
```
ittaiz: @johnynek can we add here the actual jars? is there a reason to **only** return the executable?

ittaiz:
@chaoren if java_provider isn't an option (though I'd really love to understand why) then I think I can just add an attribute `scala.outputs.jars` which will have an array of both the scala code and the java code and also add an empty `scala.transitive_exports` since I don't know what it means and I don't think we need it.
From poking around the aspect those seem to be the only mandatory parts, right?
WDYT in general?

chaoren:
@ittaiz 

> if java_provider isn't an option (though I'd really love to understand why)

We can't easily identify the rule as scala if we're not using scalaattr. It doesn't look like we can refer to the java provider by name (it's just a list of providers), so if another (unsupported) rule provides similar looking providers we can't tell them apart.

>  I think I can just add an attribute scala.outputs.jars which will have an array of both the scala code and the java code and also add an empty scala.transitive_exports since I don't know what it means and I don't think we need it.

Yes that seems fine.

src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
```diff
@@ -45,24 +45,30 @@ object PrefixSuffixTestDiscoveringSuite {
...
+    val classes = archives.flatMap(discoverClassesIn)
+    if (classes.isEmpty)
+      throw new IllegalStateException("Was not able to discover any classes " +
+                                      s"for archive=$archives, " +
+                                      s"prefixes=$prefixes, " +
+                                      s"suffixes=$suffixes")
+    classes
+  }
+
+  private def discoverClassesIn(archivePath: String) = {
```
johnynek: can we add a return type here. I think it is `Array[Class[_]]`

src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
```diff
@@ -108,11 +114,11 @@ object PrefixSuffixTestDiscoveringSuite {
     .map(_.getName)
     .toList
 
-  private def archiveInputStream() =
+  private def archiveInputStream(archivePath: String) =
     new JarInputStream(new FileInputStream(archivePath))
 
-  private def archivePath: String =
-    System.getProperty("bazel.discover.classes.archive.file.path")
```
johnynek: is this going to break anyone, or is this only expected to be set by a rule? If a rule, is it our rule? should we use `io.bazel.rules_scala` or something to distinguish? Can we add a comment in any case to ease reading the code?

scala/scala.bzl
```diff
@@ -289,23 +289,65 @@ DependencyAnalyzerMode: {dependency_analyzer_mode}
...
+def try_to_compile_java_jar(ctx, all_srcjars, java_srcs, implicit_junit_deps_needed_for_java_compilation):
+    if not java_srcs and not all_srcjars:
+      return False
+
+    providers_of_dependencies = collect_java_providers_of(ctx.attr.deps)
+    providers_of_dependencies += collect_java_providers_of(implicit_junit_deps_needed_for_java_compilation)
+    scala_sources_java_provider = java_common.create_provider(
+        #the line below means we can't use the provider from java_common.compile
+        #since it will include the full jar when I want to expose only the ijar
+        #on the other hand we can't supply the java_common.compile the scala ijar since it needs the internals
```
johnynek: I don't know what `internals` mean here. I'm not sure why the ijar of the scalac compilation along with all the ijars we need for compilation is not sufficient. Can you add to the comment to possibly explain this?

scala/scala.bzl
```diff
@@ -289,23 +289,65 @@ DependencyAnalyzerMode: {dependency_analyzer_mode}
...
+                output = full_java_jar,
+                javac_opts = ctx.attr.javacopts + ctx.attr.javac_jvm_flags,
+                deps = providers_of_dependencies,
+                #exports can be empty since the manually created provider exposes exports
+                #needs to be empty since we want the provider.compile_jars to only contain the sources ijar
+                exports = [],
+                java_toolchain = ctx.attr._java_toolchain,
+                host_javabase = ctx.attr._host_javabase,
+    )
+    return struct(jar = full_java_jar, ijar = provider.compile_jars.to_list().pop())
```
johnynek: is there really a contract that the first item is the ijar for this item? Can we add a comment to that effect?

This jar only includes the java output, not the scala, right?

scala/scala.bzl
```diff
@@ -289,23 +289,65 @@ DependencyAnalyzerMode: {dependency_analyzer_mode}
...
+          providers.append(dep[java_common.provider])
+    return providers
+
+def _compile_or_empty(ctx, jars, srcjars, buildijar, transitive_compile_jars, jars2labels, implicit_junit_deps_needed_for_java_compilation):
     # We assume that if a srcjar is present, it is not empty
     if len(ctx.files.srcs) + len(srcjars) == 0:
         _build_nosrc_jar(ctx, buildijar)
         #  no need to build ijar when empty
-        return struct(ijar=ctx.outputs.jar, class_jar=ctx.outputs.jar)
+        return struct(ijar=ctx.outputs.jar, class_jar=ctx.outputs.jar, java_jar = False)
```
johnynek: wait, here this is a bool, but below it is a struct, I think. Could we use `java_jar = None` here?

src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
ittaiz: Will do

src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
ittaiz: Won't break anyone
 our JUnit rule
I'm actually thinking of factoring out the JUnit "discover test classes" mechanism (which this is a part of) since it's not scala related but JUnit related.
When we added this we couldn't do this refactoring due to how scala and java interoped. Due to this I'd rather not change the name.
Re the comment- If one will search the repo for this text he will find it in the bzl so I think it's ok for now but if you feel strongly about it I'll add the comment

scala/scala.bzl
ittaiz: Sure but maybe let's first see I'm not just mistaken :) : Won't ijar strip parts from the scalac jar the java sources need? They're the same compilation "group"

scala/scala.bzl
ittaiz: See the comment 5 lines above. Is it clear enough?
Yes.

scala/scala.bzl
ittaiz: Will do. In python it's the same though I don't know what's the idiomatic way.

scala/scala.bzl
ittaiz: @johnynek ping

scala/scala.bzl
ittaiz: @johnynek ping

ittaiz:
@chaoren I hope it's ok I'm asking but what does a jvm rule (kotlin for example) need for it to be supported? Other than exposing java provider.

scala/scala.bzl
```diff
@@ -747,23 +749,31 @@ def _scala_test_impl(ctx):
...
     return struct(testSuiteFlag = "-Dbazel.test_suite=io.bazel.rulesscala.test_discovery.DiscoveredTestSuite",
-    archiveFlag = "-Dbazel.discover.classes.archive.file.path=%s" % archive.short_path,
+    archiveFlag = "-Dbazel.discover.classes.archives.file.paths=%s" % archives,
     prefixesFlag = "-Dbazel.discover.classes.prefixes=%s" % ",".join(ctx.attr.prefixes),
     suffixesFlag = "-Dbazel.discover.classes.suffixes=%s" % ",".join(ctx.attr.suffixes),
     printFlag = "-Dbazel.discover.classes.print.discovered=%s" % ctx.attr.print_discovered_classes)
 
+def _get_archives_short_path(scala_archive, maybe_java_archive):
+  java_archive_short_path = ""
+  if maybe_java_archive:
```
natansil: Maybe instead of this if, you can pass this method a list of archives, and some how archives that are None will not get into the list...

scala/scala.bzl
natansil: Maybe a test can be devised for this...

johnynek: ijar strips the code and leaves the classes and the signatures. Imagine all methods were replaced with `throw new Exception("")`. I don't see how javac can use that code at compile time except as annotation processors (which are the closest things java has to macros as far as I know).

So, I think using the ijar is exactly what we want since bazel is now driving this compilation. There could be times where I think bazel could recompile scala, but see that the java didn't change and the compiled ijar didn't change.

ittaiz:
Sounds good. I'll use the ijar.
I wonder if I should add a no recompilation for same target when scala
changes and java doesn't recompile.
I'll probably add it.

scala/scala.bzl
ittaiz: Sounds good. I'll use the ijar.
I wonder if I should add a no recompilation for same target when scala
changes and java doesn't recompile.
I'll probably add it

scala/scala.bzl
ittaiz: I have a feeling it's indirection and not abstraction. There are just two instances and if I understand correctly you're suggestion is to still have the if just for a "generic" item, right?

scala/scala.bzl
ittaiz: Done

chaoren:
> @chaoren I hope it's ok I'm asking but what does a jvm rule (kotlin for example) need for it to be supported? Other than exposing java provider.

Technically, we just need to be able to recognize and handle it in the aspect. Do you mean if we decide to support the rule? Then it's probably based on popularity and if we decide it's worth the maintenance cost.

ittaiz:
I was more aiming at: "can we have a generic jvm interface for languages to
implement and then automatically be supported by the plugin (like expose
java provider)" it's far away from the original issue that I'll drop it.
I'll change the attribute to return both jars.
Thanks!

ittaiz:
@johnynek apart from the ijar change and the IntelliJ change, are we good?
I'd like to merge it tomorrow.
Next change is to add support for strict deps (we first need @iirina to allow passing warn/default from skylark)

scala/scala.bzl
johnynek: 👍 

Sounds good. Would be nice to see a test that we can change some private method contents and not recompile the java if we can test that.

scala/scala.bzl
johnynek: yes that's fine.

scala/scala.bzl
johnynek: I don't think there was a plan. I think it probably was just ill-unified.

scala/scala.bzl
johnynek: I'm not really sure what needs to be there actually. Since `scala_binary` creates a default executable, maybe it should only be the executable? Maybe it doesn't matter. Honestly, some of the bazel stuff is not super well documented and I wrote some of this about 2 years ago now and it might have both changed in bazel and also I may have misunderstood it at the time.

I'm afraid I don't have a great answer for you as to what should be in there.

scala/scala.bzl
johnynek: yeah, it looks like `outputs.class_jar` is always `ctx.outputs.jar` in all cases. We could probably remove it and only use `ctx.outputs.jar` to be more clear.

scala/scala.bzl
```diff
@@ -614,31 +670,33 @@ def _scala_binary_common(ctx, cjars, rjars, transitive_compile_time_jars, jars2l
...
       jvm_flags = ctx.attr.jvm_flags,
   )
-  return _scala_binary_common(ctx, cjars, transitive_rjars, jars.transitive_compile_jars, jars.jars2labels)
+  return out
 
 def _scala_repl_impl(ctx):
   # need scala-compiler for MainGenericRunner below
   jars = _collect_jars_from_common_ctx(ctx, extra_runtime_deps = [ctx.attr._scalacompiler])
   (cjars, transitive_rjars) = (jars.compile_jars, jars.transitive_runtime_jars)
-  transitive_rjars += [ctx.outputs.jar]
```
johnynek: this is removed because `_scala_binary_common` is putting it in also, is that right?

scala/scala.bzl
```diff
@@ -795,13 +861,16 @@ library_attrs = {
...
 
-library_outputs = {
+common_outputs = {
   "jar": "%{name}.jar",
   "deploy_jar": "%{name}_deploy.jar",
-  "ijar": "%{name}_ijar.jar",
   "manifest": "%{name}_MANIFEST.MF",
 }
 
+library_outputs = common_outputs + {
```
johnynek: nice cleanup.

johnynek:
Yeah, other than the ijar, comments and type annotation, yes.

scala/scala.bzl
ittaiz: Decided to leave it for now. Since the outputs.class_jar (although the unclear name) adds symmetry to the collection of the java jar.
We'll handle it when we'll do a bigger refactoring and re-alignment

scala/scala.bzl
ittaiz: This is what the java binary does so I'm guessing it's modeled after that.
We found a workaround so the current status quo is ok

src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
ittaiz: done

src/java/io/bazel/rulesscala/test_discovery/DiscoveredTestSuite.scala
ittaiz: done

scala/scala.bzl
ittaiz: done

scala/scala.bzl
ittaiz: ended up doing it for a different reason, it allowed me to pass a more generic out argument from the _common_binary_common

scala/scala.bzl
ittaiz: yes

ittaiz:
@johnynek I've handled the return type, the comment, the use of the scala ijar to java siblings and exposing java jars to intellij

ittaiz:
Build is green on 0.5.3

johnynek:
pro-style work sir!

Thanks for pushing this all the way through.

ittaiz:
:) Sure thing. Thank you for your inputs, I think they improved the end result.
Next up I'll open a PR to support strict java deps, I'm just waiting for https://github.com/bazelbuild/bazel/issues/3626 since currently warn/default aren't supported so it will break the build

ittaiz:
btw, I actually think squashing here wasn't the best approach since I tried keeping the commits pretty separated. The benefit from not squashing is smaller commits to comb when going through the history.

johnynek:
you and I seem to disagree about squashed commits. :)

Ultimately, I like one atomic commit of the reviewed and accepted change, and I like to keep each change very small (ideally much under 400 lines).

In your case, for instance, we could have done all the clean up changes in a separate PR (which may have been 1/3 of the diff at the end of the day).

Maybe we should hash out some standard and document it in the repo.

johnynek:
To be clear: this diff was a good size, but by making each reviewed PR atomic and as small as possible, we never have any confusion that any sha in master was not previous state which was reviewed.  I know people disagree about this.

Honestly, merge commits might be fine if the git tooling was clearer and easier to use to make it clear that there is no valid view of master that steps into these commits.

ittaiz:
I guess we'll do it at some point :)
To be honest I guess the truth (as always) is somewhere in the middle since
I could have squashed it into 3 commits.
BTW,
I meant to use rebase and not merge (which github ui supports), this
sidesteps the merge commit issue.

PR #271, opened 2017-09-02
Just test most recent two releases

johnynek:
Testing against HEAD is very brittle and frustrates many new contributors. Until bazel supplies a standard URL to grab the HEAD builds, we should probably rely on the bazel ci bot for HEAD testing.

We still have issue #76 to make more of our tests sh_tests. We have still not tackled them, but if we did, it would probably make the bazel ci bot happier (since they want to just run bazel commands, I think).

cc @damienmg

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,9 @@ os:
   - osx
 
 env:
-  - V=HEAD
+  # we want to test the most recent few releases
   - V=0.5.3
+  - V=0.5.4
 
 before_install:
   - |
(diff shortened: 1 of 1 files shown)


ittaiz:
Sounds good but why test a few latest releases?
Mainly asking since we keep moving the needle and I guess we'll require 0.6.0 as the basis very soon.
I suggest to just keep one version until we get a stable HEAD url.

Also,
@damienmg is working on enabling some better bazel e2e testing so once that's in we (Wix) might be able to shift some resources towards #76

johnynek:
Not everyone updates immediately. If we can possibly keep a wider range of compatibility, I'd like to.

Honestly, a better approach is something like the oldest version we are compatible with and the newest (even though in between we might still hit intermittent issues).

We are on a pretty old version of bazel and rules_scala now because simply bumping the version caused our builds to break. We didn't see a compelling issue to run it down yet, so we have not. Other users may be in the same boat (note occasionally people report that the latest bazel broke their build).

ittaiz:
latest and oldest supported sounds good. Sometimes they'll just be the same.

PR #273, opened 2017-09-04
exposing transitive_compile_time_jars to java_common.provider

ittaiz:
exposing transitive_compile_time_jars to java_common.provider to use strict java deps
We still can't run this with `warn` until https://github.com/bazelbuild/bazel/issues/3626 is resolved but it's already a step forward.
Additionally if you're using `java_library` targets with strict deps this currently doesn't work and this fixes it as well


johnynek:
Looks good.

PR #274, opened 2017-09-10
Strict deps classpaths fixes

ittaiz:
Apparently we had some classpath issues when strict deps was on which I couldn't find until I ran it on our internal codebase with the java strict deps.
Some details:
fixing transitive compile time jars for binary on binary
fixing transitive runtime jars for lib and binary
fixing scala_test when strict deps is on

BTW,
If anyone is interested this [branch](https://github.com/wix/rules_scala/tree/strict_deps_java_post_general_fixes) on our fork relies on this unreleased commit in bazel which enables the usage of warn/default in the java compilation we do and adds buildozer printing like we have in rules_scala.

cc @natansil / @johnynek  would appreciate your review


johnynek:
Quick scan on the phone looks good. Will look in detail tomorrow.

ittaiz:
:thumbsup:

test_run.sh
```diff
@@ -127,6 +127,13 @@ test_scala_binary_expect_failure_on_missing_direct_deps() {
   test_scala_library_expect_failure_on_missing_direct_deps ${dependency_target} ${test_target}
 }
 
+test_scala_binary_expect_failure_on_missing_direct_deps_through_other_binary() {
```
natansil: @ittaiz, how about `test_scala_binary_expect_failure_on_missing_direct_deps_located_in_dependency_binary_deps`?

ittaiz: interesting.
How about:
`test_scala_binary_expect_failure_on_missing_direct_deps_located_in_dependency_which_is_scala_binary`?

ittaiz:
@johnynek any comments? I'd like to merge this tomorrow

johnynek:
👍

test_run.sh
natansil: 👍

PR #278, opened 2017-09-14
wrap full jars in a struct for intellij plugin

### LINKED ISSUE #277: `object of type 'File' has no field 'class_jar'.` error from intellij plugin when syncing scala targets
I see this error in the Intellij Bazel console for every `scala_library` target in my project, during Sync:

```
ERROR: /REDACTED/common/BUILD.bazel:3:1: in @intellij_aspect//:intellij_info.bzl%intellij_info_aspect aspect on scala_library rule //REDACTED/common:common: 
	File "/REDACTED/common/BUILD.bazel", line 3
		@intellij_aspect//:intellij_info.bzl%intellij_info_aspect(...)
	File "/private/var/tmp/_bazel_kgessner/a607cbff8886d2ecb81399f157ed2535/external/intellij_aspect/intellij_info.bzl", line 18, in _aspect_impl
		intellij_info_aspect_impl(target, ctx, semantics)
	File "/private/var/tmp/_bazel_kgessner/a607cbff8886d2ecb81399f157ed2535/external/intellij_aspect/intellij_info_impl.bzl", line 691, in intellij_info_aspect_impl
		collect_java_info(target, ctx, semantics, ide_info, <2 more arguments>)
	File "/private/var/tmp/_bazel_kgessner/a607cbff8886d2ecb81399f157ed2535/external/intellij_aspect/intellij_info_impl.bzl", line 386, in collect_java_info
		library_artifact(output)
	File "/private/var/tmp/_bazel_kgessner/a607cbff8886d2ecb81399f157ed2535/external/intellij_aspect/intellij_info_impl.bzl", line 130, in library_artifact
		java_output.class_jar
object of type 'File' has no field 'class_jar'.
```

It prevents intellij from analyzing the scala targets, and breaks most code analysis -- e.g. intellij can't find any of the imports.  The symptoms look like https://github.com/bazelbuild/intellij/issues/96, but I'm running versions of everything that should include those fixes:

- bazel version: 0.5.4-homebrew (I'm on OS X)
- rules_scala: 23d94aec640ebfdcf862d9dcc20ab14c4eb448c9
- intellij: IntelliJ IDEA 2017.2.4
- bazel for intellij plugin: 2017.08.29.0.2

I've traced the problem as far as the [`jars` value exported to intellij](https://github.com/bazelbuild/rules_scala/blob/23d94aec640ebfdcf862d9dcc20ab14c4eb448c9/scala/scala.bzl#L592) being incorrect -- it should be a list of structs like `[outputs]`, not a list of files.

I'll try to put together a small test project I can share.  Let me know if you need anything else, or if I should report this to bazelbuild/intellij instead.  Thanks!

ittaiz:
this aims to fix #277 
@kevingessner any chance you can take this branch out for a manual spin on your repo?

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -551,6 +551,9 @@ def _collect_jars_from_common_ctx(ctx, extra_deps = [], extra_runtime_deps = [])
 
     return struct(compile_jars = cjars, transitive_runtime_jars = transitive_rjars, jars2labels=jars2labels, transitive_compile_jars = transitive_compile_jars)
 
+def _format_full_jars_for_intellij_plugin(full_jars):
+    return [struct (class_jar = file, ijar = None) for file in files if not _is_source_jar(file)]
+
 def _lib(ctx, non_macro_lib):
     # Build up information from dependency-like attributes
 
@@ -585,11 +588,13 @@ def _lib(ctx, non_macro_lib):
     next_cjars += exports_jars.compile_jars
     transitive_rjars += exports_jars.transitive_runtime_jars
 
+
+
     rule_outputs = struct(
         ijar = outputs.ijar,
         class_jar = outputs.class_jar,
         deploy_jar = ctx.outputs.deploy_jar,
-        jars = outputs.full_jars #needed by intellij plugin
+        jars = _format_full_jars_for_intellij_plugin(outputs.full_jars)
     )
     # Note that, internally, rules only care about compile_jars and transitive_runtime_jars
     # in a similar manner as the java_library and JavaProvider
@@ -658,7 +663,7 @@ def _scala_binary_common(ctx, cjars, rjars, transitive_compile_time_jars, jars2l
       ijar=outputs.class_jar,
       class_jar=outputs.class_jar,
       deploy_jar=ctx.outputs.deploy_jar,
-      jars = outputs.full_jars #needed by intellij plugin
+      jars = _format_full_jars_for_intellij_plugin(outputs.full_jars)
   )
 
   scalaattr = struct(
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -552,7 +552,7 @@ def _collect_jars_from_common_ctx(ctx, extra_deps = [], extra_runtime_deps = [])
     return struct(compile_jars = cjars, transitive_runtime_jars = transitive_rjars, jars2labels=jars2labels, transitive_compile_jars = transitive_compile_jars)
 
 def _format_full_jars_for_intellij_plugin(full_jars):
-    return [struct (class_jar = file, ijar = None) for file in files if not _is_source_jar(file)]
+    return [struct (class_jar = jar, ijar = None) for jar in full_jars]
 
 def _lib(ctx, non_macro_lib):
     # Build up information from dependency-like attributes
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -789,7 +789,7 @@ def _gen_test_suite_flags_based_on_prefixes_and_suffixes(ctx, archives):
 
 def _serialize_archives_short_path(archives):
   archives_short_path = ""
-  for archive in archives: archives_short_path += archive.short_path + ","
+  for archive in archives: archives_short_path += archive.class_jar.short_path + ","
   return archives_short_path[:-1] #remove redundant comma
 
 def _scala_junit_test_impl(ctx):


scala/scala.bzl
```diff
@@ -551,6 +551,9 @@ def _collect_jars_from_common_ctx(ctx, extra_deps = [], extra_runtime_deps = [])
 
     return struct(compile_jars = cjars, transitive_runtime_jars = transitive_rjars, jars2labels=jars2labels, transitive_compile_jars = transitive_compile_jars)
 
+def _format_full_jars_for_intellij_plugin(full_jars):
+    return [struct (class_jar = file, ijar = None) for file in files if not _is_source_jar(file)]
```
johnynek: weird spacing here, right. Why not `struct(class_jar = file, ijar = None)`?

kevingessner: I think `files` here should be `full_jars`; this doesn't run right now.

johnynek: yep. Looks like CI caught that.

ittaiz:
Embarrassing:( that's what I get for coding out of my sleep. Sorry. Will
fix soon

scala/scala.bzl
ittaiz: Oscar,
Didn't understand what you said about spacing. Can you take a look at the current version and comment again if it's still an issue?

ittaiz:
@kevingessner can you try it out?

ittaiz:
ok, I checked out the sample project and confirmed everything works.
@kevingessner the imports aren't recognized since you're using `//jar:file` which isn't supported.
I have a prototype for `scala_import` which supports this but it's without tests and I'm still trying to understand if it's the right solution.
I put it in a [gist](https://gist.github.com/ittaiz/f83d145c14414a7427ab9c393a3d57e6) in case you wanted to use it for now

kevingessner:
@ittaiz Thanks for the fix!

ittaiz:
Sure thing, thanks for the repro.
Glad to know other people are using the IntelliJ integration.
Do you have interest in a baked scala_import rule?

kevingessner:
The `//jar:file` import is copied out of a larger project, where we are using bazel-deps to manage dependencies.  I don't totally understand why it [uses `//jar:file`](https://github.com/johnynek/bazel-deps/blob/6d1498cb8ed10f5a1ade78698790b8bddc8e6303/src/scala/com/github/johnynek/bazel_deps/Label.scala#L30), and I'm concerned that it's causing problems with the intellij plugin.  Would using the `//jar` targets fix that?  What would it break?

bazel-deps generates a `scala_library` for that `exports` the jar and its dependencies.  So I'd only be interested in scala_import if bazel-deps was going to use it.  johnynek/bazel-deps#63 proposes switching to java_import so maybe scala_import would be welcome there too.

@johnynek Can you please weigh in on the above?

ittaiz:
As far as I understand IntelliJ isn't that used at Stripe and so Bazel-deps
doesn't care about using //jar:file.
The reason to use it is to avoid problems ijar causes with scala macros. On
the other hand this syntax breaks the IntelliJ plugin.

johnynek:
I am happy to use `scala_import` rule if we make it. Stripe also has intellij users.

I may refactor that code somewhat allow `lang: scala/macro` or something so it will use `java_import` by default, but `scala_import` (which skips ijar) if the jar is declared to contain macros.

ittaiz:
The plan for scala_import is to have a Boolean flag of whether to skip
ijars.
I've had a few conversations with Irina from the Bazel team and she says
that java_import just calls ijar and creates a java provider.
This means that scala_import will be a superset of java_import's features.
I actually wanted java_import to just add this Boolean but they felt it's
too scala specific.
Kevin,
Can you use "//jar" for now? Do you use macros from external code?
Just trying to prioritize.

kevingessner:
We do use some external scala code with macros, yes.  So changing the imports from `bazel-deps` to be just `//jar` causes compile errors.

My understanding is that the `//jar` target is an autogenerated `java_import`, and `//jar:file` is a filegroup pointing to the jar.  To use `scala_import` and make this all work, my workspace would need to generate a `scala_import` for each scala jar's `//jar:file` target, and then use that new target as the dependency of my code.  Right?  I can give that a try if so.

ittaiz:
Exactly.
My plan is to integrate with java_import_external/bazel_deps/both so that
scala_import will be automatically defined for you

kevingessner:
@ittaiz That would be great.  I've hacked my way with some custom workspace rules to use scala_import in my project, and intellij is now detecting my scala dependencies correctly!  Thanks so much!

Now I just need to figure out why intellij isn't finding the classes from my scrooge'd libraries...

ittaiz:
Awesome! Glad to hear. I'll add it to rules_scala in the near future.
Re Scrooge- Weird, @chaoren fixed something related to Scrooge libraries in
the latest release but maybe it's a different issue. A repro with an issue
in their github will probably go a long way

PR #279, opened 2017-09-18
Strip -source's jars from scala's compiler classpath for dependencies

ianoc-stripe:
This won't effect compiling source jars, but they shouldn't be in the dependency tree.

The issue here is that for maven_jar type imports we need to use the `jar:file` to avoid the ijar behavior for importing scala macros. But since scala 0.6.0 
https://github.com/bazelbuild/bazel/commit/7a7c41d7d342cd427e74f091b55690eed13e280d
Means that the source jars are also being included into this filegroup. Either we need a custom maven fetching rule, or this will strip `-sources.jar` from the dependencies tree. Since we wouldn't expect class files to be in sources jars already compiled this seems safe?

(First push here should fail build with error all going well, then the new code can be activated to pass build)
r? @johnynek

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,11 +21,12 @@ before_install:
       OS=darwin
     else
       sysctl kernel.unprivileged_userns_clone=1
-      sudo apt-get update -q 
+      sudo apt-get update -q
       sudo apt-get install libxml2-utils -y
       OS=linux
     fi
-    URL="https://github.com/bazelbuild/bazel/releases/download/${V}/bazel-${V}-installer-${OS}-x86_64.sh"
+    URL="https://storage.googleapis.com/bazel/0.6.0/rc1/bazel-0.6.0rc1-installer-${OS}-x86_64.sh"
+    # URL="https://github.com/bazelbuild/bazel/releases/download/${V}/bazel-${V}-installer-${OS}-x86_64.sh"
     wget -O install.sh "${URL}"
     chmod +x install.sh
     ./install.sh --user
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ env:
   # we want to test the most recent few releases
   - V=0.5.3
   - V=0.5.4
+  - V=0.6.0rc2
 
 before_install:
   - |
@@ -25,8 +26,13 @@ before_install:
       sudo apt-get install libxml2-utils -y
       OS=linux
     fi
-    URL="https://storage.googleapis.com/bazel/0.6.0/rc1/bazel-0.6.0rc1-installer-${OS}-x86_64.sh"
-    # URL="https://github.com/bazelbuild/bazel/releases/download/${V}/bazel-${V}-installer-${OS}-x86_64.sh"
+    if [[ $V =~ .*rc[0-9]+.* ]]; then
+      PRE_RC=$(expr "$V" : '\([0-9.]*\)rc.*')
+      RC_PRC=$(expr "$V" : '[0-9.]*\(rc.*\)')
+      URL="https://storage.googleapis.com/bazel/${PRE_RC}/${RC_PRC}/bazel-${V}-installer-${OS}-x86_64.sh"
+    else
+      URL="https://github.com/bazelbuild/bazel/releases/download/${V}/bazel-${V}-installer-${OS}-x86_64.sh"
+    fi
     wget -O install.sh "${URL}"
     chmod +x install.sh
     ./install.sh --user
@@ -35,3 +41,4 @@ before_install:
 
 script:
   - bash test_run.sh ci
+
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -481,8 +481,7 @@ def dep_target_contains_ijar(dep_target):
 # import cats.implicits._
 
 def filter_not_sources(deps):
-  return deps
-  # return depset([dep for dep in deps.to_list() if "-sources.jar" not in dep.basename ])
+  return depset([dep for dep in deps.to_list() if "-sources.jar" not in dep.basename ])
 
 def _collect_jars_when_dependency_analyzer_is_off(dep_targets):
   compile_jars = depset()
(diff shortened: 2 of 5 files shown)


bazel-io:

.travis.yml
```diff
@@ -14,18 +14,25 @@ env:
   # we want to test the most recent few releases
   - V=0.5.3
   - V=0.5.4
+  - V=0.6.0rc2
```
ianoc-stripe: Adding this seemed like the best way to avoid possibly regressing in future since the 0.5.x's don't have this behavior.

scala/scala.bzl
```diff
@@ -487,7 +498,10 @@ def _collect_jars_when_dependency_analyzer_is_off(dep_targets):
         compile_jars += dep_target.files
         runtime_jars += dep_target.files
 
-  return struct(compile_jars = compile_jars, transitive_runtime_jars = runtime_jars, jars2labels = {}, transitive_compile_jars = depset())
+  return struct(compile_jars = filter_not_sources(compile_jars),
```
ittaiz: Why not remove them from runtime jars as well? Essentially I think this filter should be where we do "dep.files"

ittaiz: Or rather only in the branch of "http file" this encapsulates the "mess" better and when we'll introduce scala import we can remove it

ianoc-stripe: I don't know if someone would want to do runtime reflection and access them tbh was the reason not to do it. Its mostly been accidental, and the likely hood of using/needing them at compile time seems amazingly low. But I don't mind dropping them from runtime, this just seemed smaller possible wake of unexpected behavior.

I don't think http_file branch would do it for us alone since we refer to maven_jar, which generates a build file. I'm not sure how easily we'll be able to remove it without saying that scala rules are incompatible with the native maven jar? (though this will only hit on some libraries where the sources path layout causes this path to fire too of course which is probably relatively uncommon -- this particular case cats are patching to remove).

ittaiz:
When you use maven_jar //jar:file it translates in rules scala to the http
file branch. It was created for this purpose.

ianoc-stripe:
Ahhh! didn't realize that -- i'll change this around to move the filter there, it would be much cleaner if that just works. Thanks!

ianoc-stripe:
I moved it to where we do the files dance, we do it 3 places right now though. So it doesn't look much different to before. But probably a bit more tidy?

johnynek:
test this please

(ci bot message)

johnynek:
test this please

scala/scala.bzl
```diff
@@ -472,6 +472,17 @@ def dep_target_contains_ijar(dep_target):
...
+# When import mavan_jar's for scala macros we have to use the jar:file requirement
+# since bazel 0.6.0 this brings in the source jar too
+# the scala compiler thinks a source jar can look like a package space
+# causing a conflict between objects and packages warning
+#  error: package cats contains object and package with same name: implicits
+# one of them needs to be removed from classpath
+# import cats.implicits._
+
+def filter_not_sources(deps):
+  return depset([dep for dep in deps.to_list() if "-sources.jar" not in dep.basename ])
```
ittaiz: Might sound picky but any chance you can factor out the if so that it's shared between the two places?

johnynek: yeah maybe:

```python
def _is_source_jar(filename):
  return "-sources.jar" in filename
```
then we call that twice? `if not _is_source_jar(dep.basename)`?

johnynek:
👍 

once @ittaiz is happy.

ittaiz:
BTW, I hope to merge one of the codepaths and just turn the strict deps on.

ianoc-stripe:
Cool -- the strict deps stuff does sound nice i've not used it yet. But its very nice in the java rules

ittaiz:
Awesome, thanks! I'll merge once build finishes.

ittaiz:
test this please

johnynek:
huh, bazel ci thinks this failed.

johnynek:
does this pass locally for you @ianoc-stripe ?

ianoc-stripe:
No sorry, just pushed thinking it was tiny. ofc I always forget the need to do `return` in python. (passes locally now)

johnynek:
test this please

ittaiz:
Forgetting the return happens to me every time :)

ianoc-stripe:
Looks like travis hasn't started ? ci from bazel looks green

ianoc-stripe:
Looks like travisci is having a sad day fyi -- https://www.traviscistatus.com/incidents/4qj8t3220jy8

johnynek:
seems like this failed on `0.6.0.rc2`: https://travis-ci.org/bazelbuild/rules_scala/jobs/276986474

johnynek:
test this please

ianoc-stripe:
the bazel ci failure looks transient i think, might need to re-request the test?

johnynek:
test this please

ianoc-stripe:
I'm not sure there is a single point here unless i'm missing it? We refer to files in a few places to add them to different parts of the struct. I could make a function to operate on the struct and then do it further out too.


The failures for both travis and bazel look transient and not code related :(

johnynek:
this this please

johnynek:
@damienmg looks like bazel ci is not working for us. Fails immediately with a channel closed exceptions. You probably already know, but I wanted to share.

johnynek:
@ianoc-stripe I'm re-running the travis build.

damienmg:
We had a downtime yesterday, rerunning should go ok: retest this please.

ittaiz:
@ianoc-stripe not critical. I can do the change later and cc you so you'll see what I meant, how does that sound?

ianoc-stripe:
@ittaiz That would be great, thanks for reviewing

ianoc-stripe:
This is finally green and should be good to go!

johnynek:
Thanks @ianoc-stripe

PR #286, opened 2017-09-29
Use zipper rather than jdk/jar in the thrift rules

### LINKED ISSUE #282: use bazel_tools zip instead of a call to jar in thrift rules
here:

https://github.com/bazelbuild/rules_scala/blob/master/thrift/thrift.bzl#L54

and we do a dance to try to avoid letting timestamps in.

But in fact, bazel tools has a tool that does this:
https://github.com/bazelbuild/bazel/blob/master/tools/zip/BUILD#L9

by using zipper, it should get cleaner and less likely to have weird timestamp issues.

johnynek:
Bazel has a built in command for building zips deterministically. Use that rather than the find + timestamp setting trick.

closes #282

### DIFF
--- a/thrift/thrift.bzl
+++ b/thrift/thrift.bzl
@@ -45,29 +45,26 @@ def _thrift_library_impl(ctx):
       actual_prefix = common_prefix[0:endpos]
       zipper_args = " ".join([ "%s=%s" % (src[endpos+1:], src) for src in src_paths ])
 
-  zipper_arg_path = ctx.actions.declare_file("%s_zipper_args" % ctx.outputs.libarchive.path)
-  ctx.file_action(zipper_arg_path, zipper_args)
-  _valid_thrift_deps(ctx.attr.deps)
-  # We move the files and touch them so that the output file is a purely deterministic
-  # product of the _content_ of the inputs
-  cmd = """
+  if len(src_paths) > 0:
+    zipper_arg_path = ctx.actions.declare_file("%s_zipper_args" % ctx.outputs.libarchive.path)
+    ctx.file_action(zipper_arg_path, zipper_args)
+    _valid_thrift_deps(ctx.attr.deps)
+    # We move the files and touch them so that the output file is a purely deterministic
+    # product of the _content_ of the inputs
+    cmd = """
 rm -f {out}
 {zipper} c {out} @{path}
 """
 
-  cmd = cmd.format(out=ctx.outputs.libarchive.path,
-                   #zipper_args=zipper_args,
-                   path=zipper_arg_path.path,
-                   zipper=ctx.executable._zipper.path)
+    cmd = cmd.format(out=ctx.outputs.libarchive.path,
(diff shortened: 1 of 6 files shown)


thrift/thrift.bzl
```diff
@@ -135,8 +148,7 @@ thrift_library = rule(
       "absolute_prefixes": attr.string_list(),
       # This is a list of JARs which only contain Thrift files
       "external_jars": attr.label_list(),
-      "_jar": attr.label(executable=True, cfg="host", default=Label("@bazel_tools//tools/jdk:jar"), allow_files=True),
-      "_jdk": attr.label(default=Label("//tools/defaults:jdk"), allow_files=True),
+      "_zipper": attr.label(executable=True, cfg="host", default=Label("@bazel_tools//tools/zip:zipper"), allow_files=True)
```
softprops: TIL this exsited! This kind of thing needs better publicity. I've cargo culted around the restamped archive code in like every archive rule I've written

johnynek: right?

johnynek:
I tested this final version on our largest internal repo and it seemed to work (and get rid of an annoying warning we saw when we had targets with no thrifts that only had external jars).

ittaiz:
@johnynek looks good.
A) Can you squash the 6 commits which are related to this so I'll be able to rebase? I think I can't squash due to the shared commits and I'd rather not add all these intermediate commits.
B) Shouldn't we use the zipper also in the `scala.bzl`? I seem to remember we also do a song and dance there.

johnynek:
@ittaiz if we squash, then the 2 branches that depend on this will likely have a painful merge process. I would really like to just regular merge these as is to avoid that. Then the other two can be easily squash merged into the repo saving precious time.

johnynek:
oh, wait, you did merge regular, so I think I can squash merge this now into one commit in the github UI.

Trying.

Thanks!

johnynek:
yeah, that worked: https://github.com/bazelbuild/rules_scala/commit/ecbe5e7833651875810b4bad803430b06d2706ff

ittaiz:
Glad to hear. Are you saying you chose `squash` in the github UI and it was smart enough to only squash the last six commits? Didn't know that.
Also, WDYT about **B** from above?

johnynek:
@ittaiz I agree: #288

ittaiz:
Great. I might be confusing things here but can we use the zipper to
replace the JarCreator used by ScalacProcessor?

johnynek:
@ittaiz I don't think we can because each action needs to make one output. We use JarCreator to produce single outputs from actions, but maybe I'm wrong. Maybe the action can write to a directory, and a subsequent action can read that directory.

That said, I don't see the big win there. That java code was also from bazel, it works and can be compiled with JIT so it is fast. I think spinning up a jvm to make an empty jar is definitely a bummer.

ittaiz:
You're right on both counts. I agree.

PR #291, opened 2017-10-02
Make the Scala rules work with Bazel at HEAD.

lberki:
The incompatible change was 655a529db693e0d3c89a68ca81fc4858147b90f0
("Expose full compile time jars in Skylark")


ittaiz:
Thanks!
WDYT about extracting a common method there?
 It will remove code duplication and hopefully hide a bit of the noise away for those who care about the nitty gritty of provider creation

lberki:
Actually, some of the tests failed because missed two lines. Fixed, commit amended.

iirina:
Which two lines? I cannot see the difference.

ittaiz:
@lberki please please don't amend the commits since it makes it harder for others (@iirina and myself) to understand what changed.
Also,
Can you extract to the shared method?

PR #298, opened 2017-10-10
Update protobuf to 3.4.1

vladmos:
The change is required for compatibility with the upcoming Bazel 0.8.0

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -56,14 +56,14 @@ maven_jar(
 
 http_archive(
     name = "com_google_protobuf",
-    urls = ["https://github.com/google/protobuf/archive/b4b0e304be5a68de3d0ee1af9b286f958750f5e4.zip"],
-    strip_prefix = "protobuf-b4b0e304be5a68de3d0ee1af9b286f958750f5e4",
-    sha256 = "ff771a662fb6bd4d3cc209bcccedef3e93980a49f71df1e987f6afa3bcdcba3a",
+    urls = ["https://github.com/google/protobuf/archive/b04e5cba356212e4e8c66c61bbe0c3a20537c5b9.zip"],
+    strip_prefix = "protobuf-b04e5cba356212e4e8c66c61bbe0c3a20537c5b9",
+    sha256 = "cf4a434ac3a83040e9f65be92e153d00d075d8cd25e3f6c6d8670879f5796aa0",
 )
 
 http_archive(
     name = "com_google_protobuf_java",
-    urls = ["https://github.com/google/protobuf/archive/b4b0e304be5a68de3d0ee1af9b286f958750f5e4.zip"],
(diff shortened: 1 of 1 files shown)


ittaiz:
I don't understand something. Is this something each user will need to do? Don't we have a "protobuf_repositories" method?

vladmos:
Users will need to update their local copies of rules_scala, because the current version will at some point become incompatible with Bazel. I'm not sure what function you're talking about, the proto repository is defined in the WORKSPACE file, that's why I updated it there.

I'll reopen the pull request to see if the Travis failure is flaky or not.

ittaiz:
I'll try to explain- given a user updates their rules_scala version they'll also need to update their protobuf http_archive declarations right? 
@johnynek are you aware of this setup? we try to hide these details for most users so they don't care about it.
@vladmos we hide it by adding methods which use `native.http_archive` and then `native.bind` these archives. This allows users to just **use** our method if convenient or ignore it, define their own and then bind to the agreed namespace

johnynek:
Yeah, I’m confused why this matters. This code does not bleed into dependencies since it is only used by the compiler worked.

That said, it is a shame to have such tight coupling. It would be nice to publish the protos and us just use published and stable protos for talking to the worker protocol.

vladmos:
Yes, users will have to update their local protobuf declarations if they use have them. And if they use rules_scala they'll need to update it as well (assuming that this pull request will get merged). This is happening because we're trying to make as many planned breaking changes as possible before Bazel 1.0 is released to freeze the API after that (or at least not to break it often).

vladmos:
The incompatible change I'm trying to fix now is: `+=` on lists in .bzl files will become equivalent to `.extend` (it will modify lists inplace) instead of copying lists (at the moment `a += b` is equivalent to `a = a + b` for all types). There was a problem with that in the protobuf repository which has been fixed, now all repositories that depend on it should update too, and the sooner the better (Bazel 0.8 will be released approximately in the middle of November, but rules repositories should become compatible in advance so that users have time to update their code).

ittaiz:
So,
I don't object to the upgrade itself, what seems strange is that we need to
upgrade and implicitly require every user to have these definitions in
their WORKSPACE file and if they'd already exist be upgraded.
Maybe there isn't currently a better way.
Oscar,
I'm leaning towards a merge since it sounds to me like a rough edge in
Bazel in general.
Wdyt?

vladmos:
A friendly ping. We're going to push the change to Bazel in the next couple of days, it won't appear in released versions until 0.8 (mid-November), but it would be nice to have clean nightly builds on our CI so that we can detect other problems if they occur.

Once users install Bazel 0.8 they'll have to update their local protobuf repositories anyway because if it's not fixed it won't be compatible with Bazel.

ittaiz:
I'll merge it tomorrow if Oscar won't voice an objection before that

vladmos:
Thanks!

ittaiz:
Sure thing. Sorry it took so long.

Issue #295, opened 2017-10-05
jmh rules uses jmh hook that doesn't support Java 8 bytecode


gkossakowski:
While upgrading the rules to Scala 2.12, we've run into mysterious:

```
ERROR: /Users/gkk/stripe/rules_scala/test/jmh/BUILD:23:1: Couldn't build file test/jmh/test_benchmark_codegen.srcjar: Generating benchmark code for //test/jmh:test_benchmark_codegen failed (Exit 1).
Exception in thread "main" java.lang.RuntimeException
	at org.objectweb.asm.MethodVisitor.visitParameter(Unknown Source)
	at org.objectweb.asm.ClassReader.b(Unknown Source)
	at org.objectweb.asm.ClassReader.accept(Unknown Source)
	at org.objectweb.asm.ClassReader.accept(Unknown Source)
	at org.openjdk.jmh.generators.asm.ASMGeneratorSource.processClass(ASMGeneratorSource.java:60)
	at io.bazel.rules_scala.jmh_support.BenchmarkGenerator$.$anonfun$generateJmhBenchmark$3(BenchmarkGenerator.scala:140)
	at io.bazel.rules_scala.jmh_support.BenchmarkGenerator$.$anonfun$generateJmhBenchmark$3$adapted(BenchmarkGenerator.scala:139)
	at scala.collection.immutable.List.foreach(List.scala:389)
	at io.bazel.rules_scala.jmh_support.BenchmarkGenerator$.$anonfun$generateJmhBenchmark$2(BenchmarkGenerator.scala:139)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
	at io.bazel.rules_scala.jmh_support.BenchmarkGenerator$.withClassLoader(BenchmarkGenerator.scala:97)
	at io.bazel.rules_scala.jmh_support.BenchmarkGenerator$.$anonfun$generateJmhBenchmark$1(BenchmarkGenerator.scala:134)
	at io.bazel.rules_scala.jmh_support.BenchmarkGenerator$.$anonfun$generateJmhBenchmark$1$adapted(BenchmarkGenerator.scala:127)
	at io.bazel.rules_scala.jmh_support.BenchmarkGenerator$.withTempDirectory(BenchmarkGenerator.scala:107)
	at io.bazel.rules_scala.jmh_support.BenchmarkGenerator$.generateJmhBenchmark(BenchmarkGenerator.scala:127)
[...]

PR #300, opened 2017-10-11
Scala 2.12 support

gkossakowski:
Adds support for Scala 2.12.3. In the history, you can see that I tested Scala 2.12.4 nightly builds to verify the issue with the flat classpath caching.

Notes:
  - Scala 2.12.3 has a bug related to classpath caching that breaks bazel's incremental compilation when scalac worker is enabled: https://github.com/bazelbuild/rules_scala/pull/251#issuecomment-333246940
  - this PR disables the jmh test, the rule doesn't support Java 8 bytecode, see #295


bazel-io:

johnynek:
adding CLA yes since Grzeg is at Stripe and we have signed the CLA.

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -9,7 +9,7 @@ load("//scala:scala.bzl",
 def twitter_scrooge():
   native.maven_server(
     name = "twitter_scrooge_maven_server",
-    url = "http://mirror.bazel.build/repo1.maven.org/maven2/",
+    url = "http://repo1.maven.org/maven2/",
```
ittaiz: Why can't we use the Bazel mirror?

WORKSPACE
```diff
@@ -23,23 +23,23 @@ specs2_junit_repositories()
 # test adding a scala jar:
 maven_jar(
   name = "com_twitter__scalding_date",
-  artifact = scala_mvn_artifact("com.twitter:scalding-date:0.16.0-RC4"),
-  sha1 = "659eb2d42945dea37b310d96af4e12bf83f54d14"
+  artifact = scala_mvn_artifact("com.twitter:scalding-date:0.17.0"),
```
ittaiz: Can we do the version upgrades in master to minimize the diff?

src/java/io/bazel/rulesscala/scalac/BUILD
```diff
@@ -6,6 +6,12 @@ java_binary(
         "ScalacProcessor.java",
     ],
     main_class = "io.bazel.rulesscala.scalac.ScalaCInvoker",
+    # this probably should be set globally for the entire rules_scala
```
ittaiz: Maybe this needs to be in the scala.bzl?
If so we need to also test that the user gets a clear error message when they request a lower version. 
And also they can request a higher version and it's ok

WORKSPACE
johnynek: +1

specs2/specs2.bzl
```diff
@@ -4,36 +4,36 @@ def specs2_repositories():
 
   native.maven_jar(
       name = "io_bazel_rules_scala_org_specs2_specs2_core",
-      artifact = "org.specs2:specs2-core_2.11:" + specs2_version(),
-      sha1 = "495bed00c73483f4f5f43945fde63c615d03e637",
+      artifact = "org.specs2:specs2-core_2.12:" + specs2_version(),
```
johnynek: if we used `scala_mvn_artifact` we wouldn't have this change here.

scala_proto/scala_proto.bzl
```diff
@@ -81,8 +81,8 @@ def scala_proto_repositories():
 
     native.maven_jar(
         name = "scala_proto_rules_scalapb_lenses",
-        artifact = "com.trueaccord.lenses:lenses_2.11:0.4.12",
-        sha1 = "c5fbf5b872ce99d9a16d3392ccc0d15a0e43d823",
+        artifact = "com.trueaccord.lenses:lenses_2.12:0.4.12",
```
johnynek: if we used `scala_mvn_artifact` here we wouldn't have this change and others in the file.

twitter_scrooge/twitter_scrooge.bzl
gkossakowski: i can't remember the reason to switch, switched back

WORKSPACE
gkossakowski: #303

specs2/specs2.bzl
gkossakowski: I went wholesale on this in #303

src/java/io/bazel/rulesscala/scalac/BUILD
gkossakowski: Is there a pattern of enforcing minimal Java version on clients of the rule?

johnynek:
I'm going to close this in favor of #304

PR #303, opened 2017-10-12
Dependencies cleanup

gkk-stripe:
* migrate from hardcoded _2.11 to `scala_mvn_artifact`
* upgrade to scalding-date 0.17.0

See commits for details.

### DIFF
--- a/scala_proto/scala_proto.bzl
+++ b/scala_proto/scala_proto.bzl
@@ -1,4 +1,5 @@
 load("//scala:scala.bzl",
+  "scala_mvn_artifact",
   "scala_library",
   "collect_jars",
   "create_java_provider")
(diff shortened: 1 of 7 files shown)


bazel-io:

scala_proto/scala_proto.bzl
```diff
@@ -1,4 +1,5 @@
 load("//scala:scala.bzl",
+  "scala_mvn_artifact",
```
johnynek: how does this build? I thought you moved this to a new file?

gkossakowski: I was puzzled by this too. It looks like if you load a symbol in `bzl` files, it's available to both you and downstream `bzl` files. Not sure if it's intended that way.

PR #304, opened 2017-10-12
Scala 2.12 support

gkk-stripe:
(follow up on #300 from my Stripe account that has cla signed bit on)

Adds support for Scala 2.12.3. In the history, you can see that I tested Scala 2.12.4 nightly builds to verify the issue with the flat classpath caching.

Notes:
* Scala 2.12.3 has a bug related to classpath caching that breaks bazel's incremental compilation when scalac worker is enabled:https://github.com/bazelbuild/rules_scala/pull/251#issuecomment-333246940
 * this PR disables the jmh test, the rule doesn't support Java 8 bytecode, see #295


bazel-io:

gkk-stripe:
I tried switching back to bazel mirror for dependencies and I've got some dependency failures. See f903c9d for details.

I always `YdisableFlatCpCaching` option to scalac as workaround for #305. It's a bit ugly but we'll able to drop this as soon as 2.12.4 is out.
I want to see if this option is the last missing bit for tests to pass on CI. They pass on my machine.

johnynek:
Kind of worryingly if the bazel jar cache is somehow changing the shas.

gkossakowski:
I think the bazel mirror is not resolving the dependency at all. It returns 403.

gkk-stripe:
the travis failure seems to be a transient network problem, can we kick travis build again?

gkk-stripe:
pushed nop change to a commit, let's see then

johnynek:
I think you meant to make this into 2.12 branch, no?

gkk-stripe:
yes, ty!

Issue #308, opened 2017-10-14
intellij integration is broken 


ittaiz:
[The failure with rules_scala master](https://gist.github.com/greggdonovan/7dcef8eaf809f01a97051ae0e88e51fa#file-rules_scala_master_fails_aspect-txt) (relevant to 7b9b89b70bfd6935ccfc9886105471ba4b733fd1)   looks related to scala_macro_library and ijars expectations.
@johnynek did we ever generate an ijar for scala_macro_library? I don't remember and it doesn't make a lot of sense
@chaoren These tests passed at some point right? Any idea how?

@greggdonovan all other tests pass?

greggdonovan:
@ittaiz Four of five tests fail with master. The testScalaBinary result actually looks unrelated to macros at first glance.

Here's a [quick diff](https://gist.github.com/greggdonovan/cd9596ac2bd74b6d6d597991a6f7d437) for adding intellij aspect testing to `test_run.sh` that you can use to reproduce it.

The approach of cloning or updating the intellij repo is ugly enough that I'm tempted to recommend a [git submodule that tracks intellij master](https://github.com/git/git/blob/master/Documentation/RelNotes/1.8.2.txt#L186:L189).

What sort of approach should we take to managing the dependency on intellij? Using a WORKSPACE dependency seems like the most Bazel-friendly way. Could we then bind `io_bazel_rules_scala` to ourself rather than using `--override_repository`?

ittaiz:
Why do we need to manage the dependency on IntelliJ?

greggdonovan:
I was hoping we could keep the local CI times reasonable. I'm not sure about travis runtimes, as I don't know what the caching options are. 

But, locally, doing a clean checkout of `intellij.git` and running the scala aspect integration tests takes five minutes.[1] Most of it is building protobuf and downloading large (300mb+) IntelliJ IDE artifacts for the tests.

If you're fine the approach in my diff above -- manually maintaining the intellij clone locally and creating it fresh every time in CI -- that seems like it could be fine.

[1] `time { git clone git@github.com:bazelbuild/intellij.git; cd intellij; bazel test --profile ~/intellij_test_profile.out --test_output=errors --incompatible_disallow_set_constructor=false --override_repository io_bazel_rules_scala=/Users/gdonovan/development/rules_scala //aspect/testing/tests/src/com/google/idea/blaze/aspect/scala/...; }`

Here's the result of `bazel analyze-profile ~/intellij_test_profile.out` 
[...]

PR #309, opened 2017-10-17
Integrate intellij scala aspect integration testing into rules_scala

greggdonovan:
Related to #308 

As we want to test multiple versions of the intellij plugin (master and a release tag) we are managing a manual clone of intellij instead of using a git submodule. I'm not sure how to get submodules to handle a test matrix of intellij versions reasonably.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -594,8 +594,8 @@ def _collect_jars_from_common_ctx(ctx, extra_deps = [], extra_runtime_deps = [])
 
     return struct(compile_jars = cjars, transitive_runtime_jars = transitive_rjars, jars2labels=jars2labels, transitive_compile_jars = transitive_compile_jars)
 
-def _format_full_jars_for_intellij_plugin(full_jars):
-    return [struct (class_jar = jar, ijar = None) for jar in full_jars]
+def _format_full_jars_for_intellij_plugin(full_jars, ijar):
+    return [struct (class_jar = jar, ijar = ijar) for jar in full_jars]
 
 def create_java_provider(scalaattr, transitive_compile_time_jars):
     # This is needed because Bazel >=0.7.0 requires ctx.actions and a Java
@@ -657,10 +657,9 @@ def _lib(ctx, non_macro_lib):
 
 
     rule_outputs = struct(
-        ijar = outputs.ijar,
         class_jar = outputs.class_jar,
         deploy_jar = ctx.outputs.deploy_jar,
-        jars = _format_full_jars_for_intellij_plugin(outputs.full_jars)
+        jars = _format_full_jars_for_intellij_plugin(outputs.full_jars, outputs.ijar)
     )
     # Note that, internally, rules only care about compile_jars and transitive_runtime_jars
     # in a similar manner as the java_library and JavaProvider
@@ -725,7 +724,7 @@ def _scala_binary_common(ctx, cjars, rjars, transitive_compile_time_jars, jars2l
       ijar=outputs.class_jar,
       class_jar=outputs.class_jar,
       deploy_jar=ctx.outputs.deploy_jar,
-      jars = _format_full_jars_for_intellij_plugin(outputs.full_jars)
+      jars = _format_full_jars_for_intellij_plugin(outputs.full_jars, outputs.class_jar)
   )
 
   scalaattr = struct(
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -594,8 +594,8 @@ def _collect_jars_from_common_ctx(ctx, extra_deps = [], extra_runtime_deps = [])
 
     return struct(compile_jars = cjars, transitive_runtime_jars = transitive_rjars, jars2labels=jars2labels, transitive_compile_jars = transitive_compile_jars)
 
-def _format_full_jars_for_intellij_plugin(full_jars, ijar):
-    return [struct (class_jar = jar, ijar = ijar) for jar in full_jars]
+def _format_full_jars_for_intellij_plugin(full_jars):
+    return [struct (class_jar = jar, ijar = None) for jar in full_jars]
 
 def create_java_provider(scalaattr, transitive_compile_time_jars):
     # This is needed because Bazel >=0.7.0 requires ctx.actions and a Java
@@ -657,9 +657,10 @@ def _lib(ctx, non_macro_lib):
 
 
     rule_outputs = struct(
+        ijar = outputs.ijar,
         class_jar = outputs.class_jar,
         deploy_jar = ctx.outputs.deploy_jar,
-        jars = _format_full_jars_for_intellij_plugin(outputs.full_jars, outputs.ijar)
+        jars = _format_full_jars_for_intellij_plugin(outputs.full_jars)
     )
     # Note that, internally, rules only care about compile_jars and transitive_runtime_jars
     # in a similar manner as the java_library and JavaProvider
@@ -724,7 +725,7 @@ def _scala_binary_common(ctx, cjars, rjars, transitive_compile_time_jars, jars2l
       ijar=outputs.class_jar,
       class_jar=outputs.class_jar,
       deploy_jar=ctx.outputs.deploy_jar,
-      jars = _format_full_jars_for_intellij_plugin(outputs.full_jars, outputs.class_jar)
+      jars = _format_full_jars_for_intellij_plugin(outputs.full_jars)
   )
 
   scalaattr = struct(
--- a/test_run.sh
+++ b/test_run.sh
@@ -742,7 +742,7 @@ test_intellij_aspect() {
     cd intellij && git fetch && git pull
   fi
   git checkout "${intellij_git_tag}"
-  bazel test --test_output=errors --incompatible_disallow_set_constructor=false --override_repository io_bazel_rules_scala=${rules_scala_dir} //aspect/testing/tests/src/com/google/idea/blaze/aspect/scala/...
+  bazel test --test_output=errors --incompatible_disallow_set_constructor=false --override_repository io_bazel_rules_scala="${rules_scala_dir}" //aspect/testing/tests/src/com/google/idea/blaze/aspect/scala/...
 }
 
 if [ "$1" != "ci" ]; then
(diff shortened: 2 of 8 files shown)


bazel-io:

johnynek:
test this please

greggdonovan:
Sorry --  the git clone failed. PR updated.

I'll try setting up my own travis account to iterate on this tomorrow.

ittaiz:
Looks good now!
I looked at only one cell in the matrix and it's currently broken because our integration is broken :)
I think we'll need to keep this PR open until we have another PR which branched out of this branch and fixes the current tests.
Thanks Gregg!

test_run.sh
```diff
@@ -525,6 +525,23 @@ revert_change() {
...
+    # ci: intellij is checked out with full git history in `.travis.yaml`
+    cd intellij
+  else
+    # local: update or checkout a sibling dir.
+    cd "${rules_scala_dir}/../"
+    test -d "intellij/.git" || git clone git@github.com:bazelbuild/intellij.git
+    cd intellij && git fetch && git pull
+  fi
+  git checkout "${intellij_git_tag}"
+  bazel test --test_output=errors --incompatible_disallow_set_constructor=false --override_repository io_bazel_rules_scala=${rules_scala_dir} //aspect/testing/tests/src/com/google/idea/blaze/aspect/scala/...
```
kevingessner: you should quote `${rules_scala_dir}` in case it has whitespace

test_run.sh
```diff
@@ -525,6 +525,23 @@ revert_change() {
   mv $1/$2.bak $1/$2
 }
 
+test_intellij_aspect() {
+  intellij_git_tag=$1
+  rules_scala_dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+
+  if [[ "$runner" == "run_test_ci" ]]; then
```
kevingessner: this feels brittle.  can you detect CI another way?  maybe `test -d intellij`?

ittaiz:
Why is this brittle? That's how the script works

test_run.sh
promiseofcake: It doesn't rely on any Travis specific env variables, `ci` is [passed to the script](https://github.com/bazelbuild/rules_scala/pull/309/files#diff-354f30a63fb0907d4ad57269548329e3R44) and then `runner` is set based upon that value so I can't see a situation where it would break nor be able to be handled across multiple ci environments.

test_run.sh
greggdonovan: Good idea, I'll update the PR.

test_run.sh
greggdonovan: @kevingessner How about:

`  if  [[ "$1" == "ci" ]]; then`

That is one fewer level of indirection, at least.

kevingessner: "brittle" in that, if someone changes the name of the `run_test_ci` runner function, this will not error but will start behaving badly -- it'd be better if this was a reference to the function that was safe against changes.

ittaiz: oh yeah, missed that. thanks!
I think @greggdonovan's latest is better.

johnynek:
@greggdonovan I want to merge this, but it is still red on travis. Any ideas on how to get it green?

greggdonovan:
@johnynek Thanks! It seems like we have to understand the exact nature of the failures and resolve them to make this green.

For example, [`testScalaLibrary` expects both a JAR and an ijar](https://github.com/bazelbuild/intellij/blob/master/aspect/testing/tests/src/com/google/idea/blaze/aspect/scala/scalalibrary/ScalaLibraryTest.java#L52:L53) but the test failure claims we're only supplying the JAR. My understanding is that we supplied both JAR and ijar for libraries, so this is surprising to me. I'll take a deeper look and see if I can provide more context.

greggdonovan:
The remaining test failure appears to be due to a transient failure to connect to Maven Central. 

Otherwise, everything else is green. As long as we're happy with the ijar change and a re-run on CI passes, I think this should be OK to commit.

As far as the transient failure goes, I'll open a separate ticket to switch `native.maven_jar` usage to `java_import_external` or see if it makes sense to incorporate it into the `scala_import` PR.

ittaiz:
Thanks!
I'll take a look at the change tomorrow.
The plans with respect to repository rule and java_import_external is to
add a jvm_import_external which gets a string template of the rule it needs
to serialize and then add a scala_import_external that unsurprisingly uses
scala_import.
If you want to advance it on the Bazel side that will be great.
Shouldn't be a lot of work even

test_run.sh
ittaiz: @greggdonovan can you change the above to `if [[ "$1" == "ci" ]]; then` like you suggested?

test_run.sh
```diff
@@ -604,3 +621,5 @@ $runner test_scala_library_expect_failure_on_missing_direct_java
 $runner bazel run test:test_scala_proto_server
 $runner test_scala_library_expect_failure_on_missing_direct_deps_warn_mode_java
 $runner test_scala_library_expect_better_failure_message_on_missing_transitive_dependency_labels_from_other_jvm_rules
+$runner test_intellij_aspect ac2bb7 # TODO(https://github.com/bazelbuild/intellij/issues/146): replace with release tag when fixed.
```
ittaiz: should we remove this? IIRC intellij team said they're not going to have tags any time soon

scala/scala.bzl
```diff
@@ -595,8 +595,8 @@ def _collect_jars_from_common_ctx(ctx, extra_deps = [], extra_runtime_deps = [])
 
     return struct(compile_jars = cjars, transitive_runtime_jars = transitive_rjars, jars2labels=jars2labels, transitive_compile_jars = transitive_compile_jars)
 
-def _format_full_jars_for_intellij_plugin(full_jars):
-    return [struct (class_jar = jar, ijar = None) for jar in full_jars]
+def _format_full_jars_for_intellij_plugin(full_jars, ijar):
+    return [struct (class_jar = jar, ijar = ijar) for jar in full_jars]
```
ittaiz: @chaoren what does the plugin do with the `ijar` field?
I set it to None since I didn't see any drawbacks from setting it to None and on the other hand it becomes a bit more tricky to match the java-code ijar to the java-code full jar and the scala-code ijar to the scala-code full jar. It's not rocket science but a bit more cumbersome than I want and without any value I can see.

@greggdonovan the above implementation might be a problem since it passes the scala-code ijar for the java-code full jar. Depending on @chaoren's response we'll need to:
1. Change the intellij tests to expect None (if he agrees we don't need it).
2. Change our code to pass the right tuples so that we pass the right ijar to the full jar (if he surfaces reasons to using the ijar).

ittaiz: btw, if @chaoren says this can be ignored we can merge this PR and later remove it once he fixes the test

chaoren: Not needed. But we'll need to change the intellij aspect to accommodate the divergence from java.

chaoren: It's possible to output multiple jars now?

chaoren: We actually just use the `class_jar` if the ijar is not available, so you might see some performance hits.

ittaiz: What performance hit? In loading bigger jars to memory?
@greggdonovan this sounds like the current change pleases the test but will break the actual production

chaoren: Yeah pretty much. class_jar can always replace ijar, except it's much larger.

ittaiz:
@chaoren:
"It's possible to output multiple jars now?"- not quite, we output one
scala jar and one java jar (if both exist).
If you're asking on whether you can remove the single support from your bzl
then I'm not sure if other scala rules still use that struct.

scala/scala.bzl
ittaiz: I see. 
@greggdonovan do you want to take a crack at passing the matching ijar to the relevant full jar?

ittaiz:
@greggdonovan I think this fell off of your radar. Any chance you can carry it through?

greggdonovan:
@ittaiz Yes, sorry about that! I'll take a look this week. Thanks for catching the green test not being the whole story. ;-)

Can you give me more details on the difference between the Scala JARs and the Java JARs? I'm a bit fuzzy on the distinction. Thanks!

ittaiz:
Sure no problem.
Essentially a scala_library can generate either a jar for the scala sources, a jar for the java sources or both. If we supply intellij with the ijars we need to propagate both ijars and pass the full with the ijar of each language. Alternatively we can just decide to punt on this for now, change the intellij tests to not expect the ijar for the time being and continue to pass None

ittaiz:
Hi,
I know you're swamped but just another (hopefully) friendly ping

greggdonovan:
Good news, bad news. With the ijar `.bzl` changes reverted, the build passes without any runtime changes. 

However, on the OSX Travis builders, we're hitting the [travis.org timeout of 50 minutes](https://docs.travis-ci.com/user/customizing-the-build/#Build-Timeouts) with the added time for this test. The test downloads about ~800mb of IntelliJ IDEs and other artifacts and spends ~3 minutes compiling `protoc`, so there may not be much to do to speed it up.

I can think of a few options to work around this:
- We can add rows in the build matrix just for this test, keeping us under 50 minutes.
- We can kick off this test at the beginning of the build in parallel.
- Just run this test periodically via a [Travis cron](https://docs.travis-ci.com/user/cron-jobs/).

ittaiz:
Option 1 is to synthetically simulate multiple travis configurations?
Sounds good actually.
Can you try it and see how much it complicated the script

greggdonovan:
I think this is ready for another look, @ittaiz. The Travis build matrix now has separate rows for the IntelliJ aspect test.

greggdonovan:
Thanks, @ittaiz!

This now passes because intellij stopped expecting the ijars [here](https://github.com/bazelbuild/intellij/commit/29b78dd208cb942f480f6aa5c5f43acf174e3306#diff-d1d73e7a6f1d17e746683b9d1fdad823L57). 

We can revisit re-adding them to the intellij tests and scala.bzl if we feel the ijars are useful for the Intellij plugin.

ittaiz:
Well done!
Important contribution in intellij scala stability :)
I think next step should probably be to add a few tests over in the intellij plugin repo for covering scala_import
The current implementation is untested with respect to the plugin.
I actually think we can simplify it somewhat.
@hmemcpy the intellij plugin now supports the option of having only java provider, right?
maybe we should remove the "intellij metadata" from scala_import and manually test it.

PR #311, opened 2017-10-18
Upgrade scalatest version to 3.0.4

dflemstr:
I didn't use the bazel.build mirror since the dependency does not exist there yet.

This was also a bit tricky since scalactic is an additional mandatory dependency of scalatest nowadays.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -777,7 +777,7 @@ def _scala_test_impl(ctx):
       jars.transitive_compile_jars, jars.jars2labels)
     # _scalatest is an http_jar, so its compile jar is run through ijar
     # however, contains macros, so need to handle separately
-    scalatest_jars = collect_jars([ctx.attr._scalatest]).transitive_runtime_jars
+    scalatest_jars = collect_jars(ctx.attr._scalatest).transitive_runtime_jars
     cjars += scalatest_jars
     transitive_rjars += scalatest_jars
 
@@ -863,6 +863,7 @@ _test_resolve_deps = {
   "_scala_toolchain" : attr.label_list(default=[
     Label("//external:io_bazel_rules_scala/dependency/scala/scala_library"),
     Label("//external:io_bazel_rules_scala/dependency/scalatest/scalatest"),
+    Label("//external:io_bazel_rules_scala/dependency/scalatest/scalactic"),
   ], allow_files=False),
 }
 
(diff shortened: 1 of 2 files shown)


bazel-io:

johnynek:
In general, I'm slightly negative on our current approach to versioning. Users should be able to configure the rule to set dependencies.

That is the intent of using bind. Users in their repos set the version they use. We use scalatest 3.0.4 at Stripe and we do it by not loading the scala_repositories method in this repo and controlling our own. In there, we bind the scalatest name to a `scala_library` that exports the scalatest and scalactic jars.

I'm a bit concerned that this change will break us and anyone else (since we now have two bind names to attach to) but not actually get us closer to a more easily configurable system (such as perhaps a toolchain rule like the rust_rules added: https://github.com/bazelbuild/rules_rust/commit/7b1ba1f2a89006fbe358e97011cb1c1516435806

dflemstr:
@johnynek I think that direction makes sense! For now this way of doing things works for us, and I don't have time to build a general solution for now, so we will keep using this branch in our `WORKSPACE` until it gets merged or there is a better and more generic solution available.

ittaiz:
@dflemstr I think what @johnynek is saying is that this isn't likely to get merged and there is a better (not perfect) solution and that is to not call `scala_repositories` and instead define such a function for yourselves which will bind to the scalatest namespace a scala_library which exports both jars.
@johnynek I think `scala_repositories` shouldn't bind `scalatest`.
Two reasons:
1. Users who wish to override only scalatest but not others now need to redefine scala library if they upgrade scalatest version.
2. Users who use junit/specs2 but not scalatest still have it defined

dflemstr:
Yes I realized that it's unlikely that this will get merged, I just wanted to exhaustively list the conditions for when we'd switch back to master.

I would also be in favor of this not binding the library in the first place, or passing it in as a parameter with a default value perhaps?

ittaiz:
I might be missing something.
Why do you need to be off of master?
Just don't use the supplied repositories method and declare one of your own.
That should give you master with an upgraded scalatest

dflemstr:
@ittaiz Currently, scalatest is assumed to be a single JAR, while with this new version, scalactic is also needed to make most practical uses of `scala_test` rules compile.  I don't see a simple way to add both scalatest and scalactic as dependencies without modifying the original rule?

softprops:
Small note on version  (probably related to the above). With the current state of rules versioning it's rough to update to the latest has of HEAD on the master and realize there's some breaking change between the sha your were and the sha of where your now at. The process of stepping though shas and checking diffs is not ideal for a new user. I wonder if there's room or time here to start thinking about versioning rules themselves.

ittaiz:
David,
I think you might be mistaken. AFAICT the current state assumes scalatest is a single *target* and not a single file.
Did you try binding a different target to the namespace? And have that target depend on both scalatest and scalactic? I think Oscar said that's exactly what they do internally

dflemstr:
Okay, I haven't had time to try different approaches but as soon as I get time I will try to do what you're proposing.  I guess the approach would be to have a `java_library` which `exports` both libraries...

ittaiz:
Yes

dflemstr:
I managed to get this to work with some trickery.  We implemented a generic solution involving using Apache Ivy for dependency resolution; will aim to open source the tools around that as soon as possible.  The gist is that we used the `bind()` trick to refer to a `java_library` that exports all the actual dependency JARs correctly.

dflemstr:
(Restored branch because we still have some internal feature branches without this fix merged yet)

Issue #312, opened 2017-10-18
scalapb_proto_library does not support well known protobufs


BlueDragonX:
So I've got this build file:
```
load("@io_bazel_rules_scala//scala_proto:scala_proto.bzl", "scalapb_proto_library")

proto_library(
    name = "scalapb_proto",
    srcs = ["scalapb.proto"],
    visibility = ["//visibility:public"],
    deps = ["@com_google_protobuf//:descriptor_proto"],
)

java_proto_library(
    name = "scalapb_java_library",
    deps = [":scalapb_proto"],
)

scalapb_proto_library(
    name = "scalapb_scala_library",
    deps = [":scalapb_proto"],
    with_flat_package = True,
)
```

Building it results in the following output:
```
INFO: Found 1 target...
INFO: From scala @io_bazel_rules_scala//src/scala/scripts:scalapb_generator_lib [for host]:
warning: there were two deprecation warnings; re-run with -deprecation for details
one warning found
INFO: From creating scalapb files //protocols/scalapb:scalapb_scala_library_srcjar:
protoc-jar: protoc version: 320, detected platform: linux/amd64
protoc-jar: executing: [/tmp/protocjar7612841753549369523/bin/protoc.exe, --plugin=protoc-gen-scala=/tmp/protocbridge137921835608203876, --scala_out=flat_package:/tmp/bazelscalapb6289116713956964561, -Iprotocols/scalapb/scalapb.proto=protocols/scalapb/scalapb.proto, -Ifiles/external/com_google_protobuf/google/protobuf/descriptor.proto=external/com_google_protobuf/files/external/com_google_protobuf/google/protobuf/descriptor.proto, protocols/scalapb/scalapb.proto, external/com_google_protobuf/files/external/com_google_protobuf/google/protobuf/descriptor.proto]
external/com_google_protobuf/files/external/com_google_protobuf/google/protobuf/descriptor.proto: warning: directory does not exist.
external/com_google_protobuf/files/external/com_google_protobuf/google/protobuf/descriptor.proto: No such file or directory
ERROR: /home/sadpengu/Work/src/github.com/zenreach/platypus/protocols/scalapb/BUILD:32:1: scala //protocols/scalapb:scalapb_scala_library failed (Exit 1).
java.lang.RuntimeException: Must have input files from either source jars or local files.
	at io.bazel.rulesscala.scalac.ScalacProcessor.processRequest(ScalacProcessor.java:60)
[...]

PR #321, opened 2017-10-23
Prepend generated file root to protobuf input root path.

BlueDragonX:
Protobufs generated in a separate repository do not have their root set properly. This causes builds to fail when using well known protos from the latest https://github.com/google/protobuf.

Fixes https://github.com/bazelbuild/rules_scala/issues/312.

I could not find documentation on the right way to join paths in `bzl` files so have resorted to `'/'.join(...)`. Open to suggestions on a better way to do that.

### DIFF
--- a/scala_proto/scala_proto.bzl
+++ b/scala_proto/scala_proto.bzl
@@ -310,8 +310,13 @@ def scala_proto_repositories():
         actual = '@scala_proto_rules_netty_handler_proxy//jar'
     )
 
+def _root_path(f):
+    if f.is_source:
+        return f.owner.workspace_root
+    return '/'.join([f.root.path, f.owner.workspace_root])
+
 def _colon_paths(data):
-  return ':'.join(["{root},{path}".format(root=f.owner.workspace_root, path=f.path) for f in data])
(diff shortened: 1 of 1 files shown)


bazel-io:

BlueDragonX:
I signed it!

johnynek:
I don't know a better way to fix this.

Looks good to me.

ittaiz:
Not sure this will work for windows, I'll merge since I think it's better than before
@damienmg is there a better way to handle this problem?

damienmg:
Unfortunately not at the moment.

ittaiz:
I see. So if we wanted to support windows you'd suggest just doing an `if` there and use relevant separators?

damienmg:
I believe people have open sourced skylib (
https://github.com/bazelbuild/bazel-skylib) for exactly those kind of base
functions. Cc @dslomov also for windows and Skylark.

ittaiz:
Thanks, skylib definitely looks interesting though they declare they don't support windows yet

ittaiz:
Thanks, skylib definitely looks interesting though they declare they don't support windows yet

Issue #315, opened 2017-10-19
The with_java option to scalapb_proto_library appears to be broken.


BlueDragonX:
I have tried and failed to get `scalapb_proto_library` to work with the `with_java` option enabled. With this `BUILD`:
```
load("@io_bazel_rules_scala//scala_proto:scala_proto.bzl", "scalapb_proto_library")

proto_library(
    name = "scalapb_proto",
    srcs = ["scalapb.proto"],
    visibility = ["//visibility:public"],
    deps = [
        "//google/protobuf:descriptor_proto",
    ],
)

java_proto_library(
    name = "scalapb_java_proto",
    visibility = ["//visibility:public"],
    deps = [":scalapb_proto"],
)

scalapb_proto_library(
    name = "scalapb_scala_proto",
    visibility = ["//visibility:public"],
    with_java = True,
    with_flat_package = True,
    deps = [":scalapb_java_proto"],
)
```

I receive the following output when running `bazel build --verbose_failures scalapb:scalapb_scala_proto`:
```
WARNING: /home/sadpengu/.cache/bazel/_bazel_sadpengu/d2d05fc8812ceca219eece2a81acf47e/external/com_google_protobuf_java/WORKSPACE:1: Workspace name in /home/sadpengu/.cache/bazel/_bazel_sadpengu/d2d05fc8812ceca219eece2a81acf47e/external/com_google_protobuf_java/WORKSPACE (@com_google_protobuf) does not match the name given in the repository's definition (@com_google_protobuf_java); this will cause a build error in future versions.
INFO: Found 1 target...
ERROR: /home/sadpengu/Work/src/github.com/zenreach/platypus/protocols/scalapb/BUILD:20:1: creating scalapb files //scalapb:scalapb_scala_proto_srcjar failed (Exit 1): scalapb_generator failed: error executing command 
  (cd /home/sadpengu/.cache/bazel/_bazel_sadpengu/d2d05fc8812ceca219eece2a81acf47e/execroot/com_github_zenreach_platypus_protocols && \
  exec env - \
  bazel-out/host/bin/external/io_bazel_rules_scala/src/scala/scripts/scalapb_generator @bazel-out/local-fastbuild/bin/scalapb/scalapb_scala_proto_srcjar_worker_input)

Use --sandbox_debug to see verbose messages from the sandbox.
@bazel-out/local-fastbuild/bin/scalapb/scalapb_scala_proto_srcjar_worker_input
[...]

PR #322, opened 2017-10-24
Allow java libraries as deps of scalapb_proto_library.

BlueDragonX:
This fixes https://github.com/bazelbuild/rules_scala/issues/315 by allowing the inclusion of `java_library` in the deps of `scalapb_proto_library`.

### DIFF
--- a/scala_proto/scala_proto.bzl
+++ b/scala_proto/scala_proto.bzl
@@ -330,7 +330,7 @@ def _gen_proto_srcjar_impl(ctx):
             jvm_deps.append(target)
 
     if ctx.attr.with_java and len(jvm_deps) == 0:
-        fail("must have at leat one jvm dependency if with_java is True")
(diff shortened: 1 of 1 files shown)


bazel-io:

ittaiz:
do you want to add a test which has an example with java dependency?
Mainly for future regression.
I can merge as is but for your own sake I recommend adding a small test

johnynek:
seems good, but +1 to a test.

johnynek:
ping: any chance you can add an example to the build so we have some test of this functionality? I'd love to merge this.

BlueDragonX:
I don't have time to get to it at the moment unfortunately. Possibly next week.

johnynek:
I'm going to merge it and we can add the test later. I don't want to block people that are already cherry picking, but it would be very nice to have a test.

ittaiz:
👍🏽

PR #324, opened 2017-10-26
strict-deps clearer message when no transitive jars2labels

ittaiz:
strict-deps clearer message when no transitive jars2labels
drive-by (mostly optimizations):
iterate over deps one (via current_dep_*_jars)
for runtime_deps skip collecting labels
@johnynek:
This PR came from me troubleshooting our strict deps move with our scala_import. The need comes from a problem in the design of java_common where each rule will have to calculate the transitive labels on their own. I hope to address it via bazel but it will take time. The current solution is the simplest I could find without going into the plugin and introducing more concepts (which I hope will be redundant once Bazel supports them natively).

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -433,12 +433,17 @@ def collect_srcjars(targets):
             srcjars += [target.srcjars.srcjar]
     return srcjars
 
-def add_labels_of_jars_to(jars2labels, dependency, all_jars):
+def add_labels_of_jars_to(jars2labels, dependency, all_jars, direct_jars):
+  for jar in direct_jars:
+    add_label_of_direct_jar_to(jars2labels, dependency, jar)
   for jar in all_jars:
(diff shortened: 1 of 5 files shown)


ittaiz:
Once this is merged I'll get to work on the 0.7.0 break

ittaiz:
@johnynek would love to get your feedback about this? I want to merge it and continue to the Bazel 0.7.0 issue on Sunday

ittaiz:
Sure thing and appreciate your review.
I've started work on the 0.7.0 issue and it's a tricky one since the new API of `create_provider` incorporates into it the `ijar` creation.
In general this is going to a more encapsulated phase which is good since we have a lot of code which should be jvm generic.
The bad part is that the move requires attention to see we haven't missed anything (though we have pretty thorough e2es in addition).
Will you be able to review it?
I'll obviously get @or-shachar and/or @natansil to also look but your skilled eye would be very valuable :)

PR #326, opened 2017-10-30
adding scala_import

ittaiz:
Motivation:
The need for `scala_import` stems from the lack of support for scala's macros in ijar (see https://github.com/bazelbuild/bazel/issues/632 fro more details). This means we need a way to import external jars which doesn't pass through ijar. The current workaround is to use `maven_jar` and then use the `//jar:file` target which exposes a raw filegroup target.
Unfortunately that doesn't play nicely with bazel-intellij plugin which needs metadata about the target.
Current state:
Does:
Supports intellij
Creates java_provider
has deps and runtime_deps
Partial support for "aggregators" (you can pass None)

Does not (yet):
Support neverlink
Support many code jars
Supports exports (not sure will be supported)
Allow to give a raw jar and have the rule create an ijar if needed for it. This will enable working against ijars of 3rd party. When supporting this we'll probably add another flag to say if to use ijar or not for scala macros. 
Have tests

This will hopefully take more form and especially get some coverage but in the meantime it's here for people to see, think and maybe even contribute tests

### DIFF
--- a/scala/scala_import.bzl
+++ b/scala/scala_import.bzl
@@ -1,8 +1,10 @@
+#intellij part is tested manually, tread lightly when changing there
+#if you change make sure to manually re-import an intellij project and see imports
+#are resolved (not red) and clickable
 def _scala_import_impl(ctx):
     target_data = _code_jars_and_intellij_metadata_from(ctx.attr.jars)
-    (code_jars, intellij_metadata) = (target_data.code_jars, target_data.intellij_metadata)
-    jars2labels = _add_labels_of_current_code_jars(code_jars, ctx.label)
-    code_jars_depset = depset(code_jars)
+    (current_target_compile_jars, intellij_metadata) = (target_data.code_jars, target_data.intellij_metadata)
+    jars2labels = _add_labels_of_current_code_jars(current_target_compile_jars, ctx.label)
     transitive_runtime_jars = _collect_runtime(ctx.attr.runtime_deps)
     jars = _collect(ctx.attr.deps, jars2labels)
     return struct(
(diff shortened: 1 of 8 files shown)


johnynek:
cc @gkk-stripe this may be the solution to using the standard intellij plugin.

johnynek:
why do you say exports cannot be supported? I think we will want that ideally.

johnynek:
Also, it would be nice to have an `http_scala_jar` or something that uses this along with a repository rule to fetch the jar.

ittaiz:
Exports- didn't say we cannot but that we do not and maybe will not. Not sure about the value of exports when you have strict deps but I guess will add it for parity. Btw, i did say there is a problem with exports on a different PR related to java_common.provider
Re repository rule- the general plan is to indeed create one which uses java_import_external rule where the approach was to parameterize it to accept java_import/scala_import/whatever_import

greggdonovan:
@ittaiz Is the plan to use `scala_import` with `java_import_external` inside of `rules_scala` in place of `native.maven_jar`?

I had a transient CI failure in a `rules_scala` PR that reminded me that we likely want to replace `native.maven_jar` with `java_import_external` where we can. Should we address that in this PR or in a separate issue?

ittaiz:
Replied in the IntelliJ thread but I'll repost here:
The plans with respect to repository rule and java_import_external is to
add a jvm_import_external which gets a string template of the rule it needs
to serialize and then add a scala_import_external that unsurprisingly uses
scala_import.
If you want to advance it on the Bazel side that will be great.
Shouldn't be a lot of work even

scala/scala_import.bzl
```diff
@@ -36,18 +40,19 @@ def _add_labels_of_current_code_jars(code_jars, label):
...
+  for jar in jars:
+    current_jar_code_jars = _filter_out_non_code_jars(jar.files)
+    code_jars += current_jar_code_jars
+    for current_class_jar in current_jar_code_jars:
+      intellij_metadata.append(struct(
+           ijar = None,
+           class_jar = current_class_jar,
+       )
+     )
+  print(intellij_metadata)
```
johnynek: this is still wip right?

ittaiz:
indeed. sorry...

ittaiz:
Ok. TL;DR no longer, WIP!!!
Longer:
Good news:
All code branches apart from IntelliJ are covered.
Bad news:
Because these are tests after and because the API is very implicit the tests are complicated. Hopefully once @iirina finishes the work on the new create_provider API I'll be able to elevate a lot of the pain here.

Stuff in further PRs: (by presumed order)
Exports
NeverLink
Introduce "use ijar" property and use it
Expose sources jar to IntelliJ 
Ask @chaoren to Add scala_import integration tests to the plugin so we'll be covered

ittaiz:
Cc @gkk-stripe it's ready

ittaiz:
ok,
also added support for exports

ittaiz:
@johnynek can you review me?

gkossakowski:
this is great! i switched my https://github.com/gkossakowski/bazel-scala-example/tree/gkk-intellij-import to use this and intellij import works with `scala_import` \o/
i also switched our internal repo to point at the head of this pr to see if it breaks anything, all is good
my next step is to hack bazel-deps to emit `scala_import` and run it on our largest repo to see if there are any uknown issues that pop up

ittaiz:
Happy this helps.
Still waiting for feedback from Oscar :)
I hope next week to get to the rest of the bullets on my list (though I
guess they don't interest you that much)

scala/scala_import.bzl
```diff
@@ -0,0 +1,110 @@
...
+  return struct(transitive_runtime_jars = runtime_jars, transitive_compile_jars = transitive_compile_jars)
+
+def _transitively_accumulate_labels(dep_target, java_provider, jars2labels):
+  if hasattr(dep_target, "jars_to_labels"):
+    jars2labels.update(dep_target.jars_to_labels)
+  else:  #untested, left for semi working backward compatibility with java_library and java_import
+    for jar in java_provider.compile_jars:
+      jars2labels[jar.path] = dep_target.label
+
+def _collect_runtime(runtime_deps):
```
johnynek: really annoying the lack of abstraction power in skylark....
```scala
def collect(deps: Iterable[Dependency])(fn: Dependency => Iterable[Dependency]): Set[Dependency] =
  deps.flatMap(fn).toSet
```

gkossakowski:
I tested this on our code base. The good news: the import works!

The bad news: The tests are broken due runtime test dependencies being ijar
so they have no code to run. The tests are broken both in IntelliJ and when
executed from cmd line.
Also sync takes almost 400s spent in Directory sync which is not a surprise
but still a bummer.

ittaiz:
This is strange since I added tests in my PR which use the classes to make
sure it's full jars and not ijars.
Can you make a repro?
Also,
We don't see that problem in our own repo using this rule.
Lastly,
Intelllij Bazel people did some work on DirectoryStructure so hopefully it
will be better. Release scheduled for 8 days from now.

johnynek:
It could also be a bazel version issue. @gkk-stripe can you share the version you tested with? I guess 0.6.1?

gkk-stripe:
Yes, it is with 0.6.1. I'm giving 0.7.0 a whirl.

I tried to reproduce the problem in my small repo but the problem didn't
show up.

gkk-stripe:
The 0.7.0 on our internal repo didn't help with the issue. I hope to get
around to a repro next week.

johnynek:
We missed adding a java test that depends on scala_import right? We should do that to make sure we preserve the ability of java to depend on these.

ittaiz:
We can (and I'll try to not forget :) ) but it's important to say that apart from the labels bit scala_import treats all java_provider rules the same...

johnynek:
I know it should work. Just want to exercise java_library depending on scala_import.

Issue #235, opened 2017-06-28
Strict scala deps and unused_deps


ittaiz:
So,
I met with @cgrushko a few days ago and apparently the way we (rules_scala) currently do strict deps (allowing a target to only use what it directly depends on) is very different from the way java rules do it and IMHO not as good.

**Problem:**
We currently pass to scalac only the direct jars and let scalac emit an error to the user.
These errors are quite often [cryptic](https://github.com/scalacenter/advisoryboard/pull/12) and unuseful for developers to be able to easily fix the problem.
Moreover because of how the scala compiler works there are many cases where it needs transitive dependencies even if they aren't mentioned in the source code of the unit we're compiling[1]. This leads to having to add many transitive targets without any reasoning just to appease the compiler or  worse to start using `exports` loosely around the codebase. 

We suffered from the above transitive dependencies so much that we're just adding the whole transitive closure to the `deps` of a target. This is in the onboarding of a project so we're ok with it but this solution can help greatly minimize it.

**Proposed solution:**
Pass direct and indirect jars to scalac's classpath (like java compilation does).  
Add a scala compiler plugin which essentially will walk the AST after the type annotation phase, will stop at each 'type expression' and check that the jar from which the type was loaded is a direct_dependency. If not, it is an --indirect_dependency, and the plugin generates an error message. The message includes an 'add_dep' command to add the missing direct dep.

We will also support a lenient mode where this error is outputted as a warning.
add_dep is a command which [buildozer](https://github.com/bazelbuild/buildtools/tree/master/buildozer) can take and apply it to your BUILD file.
I think (not sure) that [ibazel](https://github.com/bazelbuild/bazel-watcher) can pick up these warnings and automatically apply them to your BUILD file. Probably [rebazel](https://github.com/meetup/rebazel) will be able to do this as well if they'll want to.

**Value:**
Separation of concerns (scalac won't enforce graph concerns)
Allows higher level messages and so call to actions
Might allow finer grain modeling of when indirect dependencies are ok (for an example where scalac "messes" up but we might be able to improve see this [repo](https://github.com/ittaiz/scalac-troubleshooting))  
Better cycle for developers since they can run with lenient mode and iterate quickly and fix the deps before pushing (or use ibazel/rebazel). The caveat is obviously the fact people can have a local green build which fails on CI.

**Open questions:**
It's unclear how smart the compiler plugin can be given scalac's limitations and so unclear on how much noise will still be added to a target's `deps`.

[...]

PR #329, opened 2017-11-06
Update docs

### LINKED ISSUE #283: strict deps is undocumented in the readme
I think the strict deps work is in now? @ittaiz is that right?

If so, it is not clear to me what I am supposed to do to use it. Could we add some explanation of how to do that to the README?

ittaiz:
hopefully fixes #283
related to #235


ittaiz:
@johnynek @natansil @or-shachar would really appreciate your feedback

PR #337, opened 2017-11-13
[wip] Testing travis running cache in nginx webdav

ianoc-stripe:


bazel-io:

ianoc-stripe:
@damienmg @johnynek This looks to be a version in CI that might be useful wrt the failures.

ittaiz:
What are we trying to achieve here?

ianoc-stripe:
Broken build when build with cache + clean + build with cache again

johnynek:
@ianoc @ianoc-stripe you are a champ! thanks.

@damienmg Not the error message here shows that caching fails with the scala rules now: 
```
---8<---8<--- Start of log, file at /home/travis/.cache/bazel/_bazel_travis/7f7e07d167ded9af0372bc305ceff99e/bazel-workers/worker-4-Scalac.log ---8<---8<---

/home/travis/.cache/bazel/_bazel_travis/7f7e07d167ded9af0372bc305ceff99e/execroot/io_bazel_rules_scala/bazel-out/host/bin/src/java/io/bazel/rulesscala/scalac/scalac: line 350: /home/travis/.cache/bazel/_bazel_travis/7f7e07d167ded9af0372bc305ceff99e/execroot/io_bazel_rules_scala/bazel-out/host/bin/src/java/io/bazel/rulesscala/scalac/scalac.runfiles/local_jdk/bin/java: No such file or directory

---8<---8<--- End of log ---8<---8<---.
```
Who do we route this to? It is strange that it only comes up with clean? Do the runfiles interact with the CAS differently?

ittaiz:
Amazing, well done!

johnynek:
@damienmg ping: any suggestion on how to triage this? This is a repro for remote execution not working correctly for the scala rules. I don't see what we are doing wrong, or why we should see this failure.

damienmg:
And also cc @nlopezgi that might have an idea.

nlopezgi:
When I have seen this type of error before, its because an absolute path to a file inside the bazel cache is being sent to the remote worker. 
It looks like the command that bazel spawns is expecting a file to exist in
"/home/travis/.cache/bazel/_bazel_travis/7f7e07d167ded9af0372bc305ceff99e/execroot/io_bazel_rules_scala/bazel-out/host/bin/src/java/io/bazel/rulesscala/scalac/scalac.runfiles/local_jdk/bin/java"
But this path likely will not exist on the worker.
We saw a very similar error in the past for go rules that had an env variable that pointed to the folder for the go external repo. Not sure what about scala rules could be using an absolute path to a file inside the bazel cache (maybe its a script that gets hardcoded with the path of the runfiles tree and then shipped to the remote worker?). If you can send a pointer to the source of the problem (java/io/bazel/rulesscala/scalac/scalac: line 350 - could not find it after a quick search in the scala repo https://github.com/bazelbuild/rules_scala) or to any scripts that scala rules will run remotely that interact with the runfiles tree (specifically with ...src/java/io/bazel/rulesscala/scalac/scalac.runfiles... ) I can help you debug further.

ianoc-stripe:
Hi @nlopezgi ,

the scalac definition is @ https://github.com/bazelbuild/rules_scala/blob/master/src/java/io/bazel/rulesscala/scalac/BUILD#L1

Its a java_binary, we just invoke this from the scala rules as a wrapper around the scalac compiler. 

I don't think in the scala_rules we are doing anything that could change its runfiles of scalac

https://github.com/bazelbuild/rules_scala/blob/master/scala/scala.bzl#L258
is how we invoke this

with _scalac being defined @ https://github.com/bazelbuild/rules_scala/blob/master/scala/scala.bzl#L881

ianoc-stripe:
the launcher script (bazel-bin/src/java/io/bazel/rulesscala/scalac/scalac when compiled) has the reference:
```
# Set JAVABIN to the path to the JVM launcher.
JAVABIN=${JAVABIN:-${JAVA_RUNFILES}/local_jdk/bin/java}
```

johnynek:
yeah, to reiterate: our script is a direct copy of the java one from bazel. Would love to find out if we are putting in a path we shouldn't, but I can't seem to find it.

nlopezgi:
Looks like bazel is resolving ${JAVABIN:-${JAVA_RUNFILES}/local_jdk/bin/java} to /home/travis/.cache/bazel/_bazel_travis/7f7e07d167ded9af0372bc305ceff99e/execroot/io_bazel_rules_scala/bazel-out/host/bin/src/java/io/bazel/rulesscala/scalac/scalac.runfiles/local_jdk/bin/java which is a path that does not exist in the remote worker. It should probably be resolving this to something like ($pwd)/..../bin/src/java/io/bazel/rulesscala/scalac/scalac.runfiles/local_jdk/bin/java which would work with remote execution. I'll create a new issue to follow up on fixing that, as, it seems to me, this is a bug in how Bazel is resolving the build variable.

ittaiz:
Thanks!
I'd recommend making sure it's not already fixed since I think we're using
a launcher version from 0.4.5

nlopezgi:
Hmm, looking a bit more at what you are doing in scala rules, it looks like you are setting the JAVABIN for this template (https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt#L230) in this rule (https://github.com/bazelbuild/rules_scala/blob/master/scala/scala.bzl#L434) using the runfiles_root which is being set here (https://github.com/bazelbuild/rules_scala/blob/master/scala/scala.bzl#L401). I think this is the source of the error you are seeing. I'm not sure what the right solution is, but this path to the runfiles_root needs to be generated in a way that will be compatible with remote execution (i.e., needs to be a relative path or one that includes ($pwd) to properly resolve from the location where the script runs). So this is not a Bazel error, its an error in the scala rules.

ianoc-stripe:
That does look likely broken :( but from my reading of the code I don't think this itself can come into play for the error above, since we see it trying to invoke a java_binary and not a scala_binary. 
But ic could be totally missing something there

nlopezgi:
So are you saying that the scalac.txt file you sent me was produced by a java_binary rule and not by a scala_binary rule? Whoever produced that scalac.txt file is the culprit for the error. I thought it was here https://github.com/bazelbuild/rules_scala/blob/master/scala/scala.bzl#L428, but if you know the right  place where the scalac.txt is being generated please point me there to debug.

nlopezgi:
Also, did I understand (a bit too late) Ittai's comment that you are using bazel 0.4.5? If so, switch to using Bazel 0.7.0 could indeed solve the issue if its related to java_binary rules and not scala_binary rules.

johnynek:
we are using a recent bazel, but that launcher script is pointing at bazel 0.4.5 by default. But maybe that launcher script hasn't changed.

Can you suggest how we should correctly generate the path to javabin? This has never been clear to us and always a bit of a hack.

ianoc-stripe:
If that launcher script was wrong then i don't think its too related to the scala rules, but to i guess ensure we aren't talking at cross purposes:
```
03:09:48 (ianoc/testCacheFailure) ~/github/rules_scala$ bazel clean; bazel version && bazel build src/java/io/bazel/rulesscala/scalac
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.
Build label: 0.7.0-homebrew
Build target: bazel-out/darwin_x86_64-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Sun Nov 12 03:11:31 2017 (1510456291)
Build timestamp: 1510456291
Build timestamp as int: 1510456291
INFO: Found 1 target...
Target //src/java/io/bazel/rulesscala/scalac:scalac up-to-date:
  bazel-bin/src/java/io/bazel/rulesscala/scalac/scalac.jar
  bazel-bin/src/java/io/bazel/rulesscala/scalac/scalac
INFO: Elapsed time: 3.745s, Critical Path: 2.25s
```
[scalac.txt](https://github.com/bazelbuild/rules_scala/files/1497364/scalac.txt)

my laptop had been using 0.6.1 in this repo, though ci is set to 0.7.0 here -- but attached file/commands i just did now try clarify things. (From what i can tell the scala rules themselves might be broken, but i don't think it comes into the issue we have).

Isolating commands a little bit:
```
03:13:34 {ianoc/testCacheFailure} ~/github/rules_scala$ bazel clean
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.
03:13:37 {ianoc/testCacheFailure} ~/github/rules_scala$ bazel run --experimental_remote_spawn_cache --remote_rest_cache=http://localhost:9090/bazel-cache src/java/io/bazel/rulesscala/scalac
INFO: Found 1 target...
Target //src/java/io/bazel/rulesscala/scalac:scalac up-to-date:
  bazel-bin/src/java/io/bazel/rulesscala/scalac/scalac.jar
  bazel-bin/src/java/io/bazel/rulesscala/scalac/scalac
INFO: Elapsed time: 5.995s, Critical Path: 3.06s

INFO: Running command line: bazel-bin/src/java/io/bazel/rulesscala/scalac/scalac
java.lang.RuntimeException: Missing required arg JarOutput
	at io.bazel.rulesscala.scalac.CompileOptions.getOrError(CompileOptions.java:130)
	at io.bazel.rulesscala.scalac.CompileOptions.<init>(CompileOptions.java:32)
	at io.bazel.rulesscala.scalac.ScalacProcessor.processRequest(ScalacProcessor.java:47)
	at io.bazel.rulesscala.worker.GenericWorker.run(GenericWorker.java:125)
ERROR: Non-zero return code '1' from command: Process exited with status 1.
03:13:44 {ianoc/testCacheFailure} ~/github/rules_scala$ bazel clean
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.
^[[A
03:13:48 {ianoc/testCacheFailure} ~/github/rules_scala$ bazel run --experimental_remote_spawn_cache --remote_rest_cache=http://localhost:9090/bazel-cache src/java/io/bazel/rulesscala/scalac
INFO: Found 1 target...
Target //src/java/io/bazel/rulesscala/scalac:scalac up-to-date:
  bazel-bin/src/java/io/bazel/rulesscala/scalac/scalac.jar
  bazel-bin/src/java/io/bazel/rulesscala/scalac/scalac
INFO: Elapsed time: 1.774s, Critical Path: 0.15s

INFO: Running command line: bazel-bin/src/java/io/bazel/rulesscala/scalac/scalac
ERROR: Error running program: java.io.IOException: Cannot run program "/private/var/tmp/_bazel_ianoc/1c83cb230054428e22566b8d0a2fa63a/execroot/io_bazel_rules_scala/_bin/process-wrapper" (in directory "/private/var/tmp/_bazel_ianoc/1c83cb230054428e22566b8d0a2fa63a/execroot/io_bazel_rules_scala/bazel-out/darwin_x86_64-fastbuild/bin/src/java/io/bazel/rulesscala/scalac/scalac.runfiles/io_bazel_rules_scala"): error=2, No such file or directory.
```

Not sure if  `--experimental_remote_spawn_cache` is important here, we need it to use local worker + cache

nlopezgi:
I'm not an expert in how build variables work, but it looks like you are using TEST_SRCDIR which according to the documentation (https://docs.bazel.build/versions/master/test-encyclopedia.html) is an absolute path, which means that we're back to a couple of messages before, where it looks like this is a Bazel bug, as using TEST_SRCDIR with remote execution would not be possible. I'll start a thread to discuss with Bazel folk about this issue.

nlopezgi:
Thanks for the additional info ianoc-stripe, I'm almost 100% sure that the problem is that the build variable you are using here (https://github.com/bazelbuild/rules_scala/blob/master/scala/scala.bzl#L401) will be mapped by bazel to an absolute path, so yes, the problem is related to your rules (not to java rules) but it needs to be solved on the Bazel side not on your rules.

ianoc-stripe:
@nlopezgi I don't think the issue (the error here is a little different so possibly different issues?) can purely be in the scala rules. I made a little test @

https://github.com/bazelbuild/rules_scala/pull/350

I removed ~all of the scala rules and indeed pretty much the whole repo here.

```
bazel run src/java/io
bazel clean
bazel run src/java/io
```


https://travis-ci.org/bazelbuild/rules_scala/builds/306012403?utm_source=github_status&utm_medium=notification
is the output

nlopezgi:
As I said, its likely due to the use of the build variables that have as value absolute paths. I doubt these variables are being used in java rules, but they could be used in many other places in your build other than the scala rules. That is why the fix needs to be done on the Bazel side, these variables should work with remote execution, and if they do, then your rules will work just fine w/o modifications.

ianoc-stripe:
Thats great! look forward to the fix :) (tbc that branch I linked to had nothing else in it, its just a java hello world for bazel -- as a repo [here](https://github.com/ianoc-stripe/bazel-helloworld) ,and output [here](https://travis-ci.org/ianoc-stripe/bazel-helloworld/builds/306020247) )

nlopezgi:
hmm, interesting, you are running with a remote cache, right? My only hypothesis is that the cache might be poisoned, i.e., the path /home/travis/.cache/bazel/_bazel_travis/558b24af2033a53fab9e3ebdf9db3d96/execroot/io_bazel_hello_world/bazel-out/local-fastbuild/bin/src/java/io/io.runfiles/io_bazel_hello_world might correspond to something that was valid for a previous run but not for the current run and you're getting it from the cache? Can you clear the cache and try again to verify? I never use the remote cache directly, so no idea what sort of edge case behavior you're looking at here related to java rules. Please ping back about what the result is after you clear the cache, I want to know if java rules might not be compatible with remote caching.

ianoc-stripe:
These are running in travis-ci, so it should be a completely clean set state for each test run/invocation.

My best guess as to the behavior is that not all of the runfiles get stored in the cache (maybe the options we use to bazel controls this somehow). 

taking the hello world repo case:

`bazel-bin/src/java/io/io`
gets generated as a script 
[io.txt](https://github.com/bazelbuild/rules_scala/files/1497556/io.txt)

Asking it for the javabin lookslike:
```
04:48:17 {master} ~/github/bazel-helloworld$ bazel-bin/src/java/io/io --print_javabin
/Users/ianoc/github/bazel-helloworld/bazel-bin/src/java/io/io.runfiles/local_jdk/bin/java
```
it gets its JAVA_RUNFILES from:
```
if [[ "$SINGLEJAR" != 1 || "1" == 1 ]]; then
  if [[ -z "$JAVA_RUNFILES" ]]; then
    while true; do
      if [[ -e "$self.runfiles" ]]; then
        JAVA_RUNFILES="$self.runfiles"
        break
      fi
```

that winds up meaning the JAVABIN setup @ 
```
# Set JAVABIN to the path to the JVM launcher.
JAVABIN=${JAVABIN:-${JAVA_RUNFILES}/local_jdk/bin/java}
```

does that, but JAVA_RUNFILES from above is :   
```
/Users/ianoc/github/bazel-helloworld/bazel-bin/src/java/io/io.runfiles
```
which looks like:
```
04:50:06 {master} ~/github/bazel-helloworld$ ls /Users/ianoc/github/bazel-helloworld/bazel-bin/src/java/io/io.runfiles
MANIFEST
```
(so nothing but a manifest file)
[MANIFEST.txt](https://github.com/bazelbuild/rules_scala/files/1497565/MANIFEST.txt)
I've attached the manifest, so the manfiest refers to all of these files being in the runfiles tree that aren't there. 

so best guess is that the caching behavior rn doesn't include everything in the runfiles?

ulfjack:
${TEST_SRCDIR} is only evaluated during the execution of the test, not during analysis. I don't think that using it here is a problem.

I see an error message trying to execute process-wrapper:
```
ERROR: Error running program: java.io.IOException: Cannot run program "/home/travis/.cache/bazel/_bazel_travis/7f7e07d167ded9af0372bc305ceff99e/execroot/io_bazel_rules_scala/_bin/process-wrapper" (in directory "/home/travis/.cache/bazel/_bazel_travis/7f7e07d167ded9af0372bc305ceff99e/execroot/io_bazel_rules_scala/bazel-out/local-fastbuild/bin/src/java/io/io.runfiles/io_bazel_rules_scala"): error=2, No such file or directory.
```

Where does the error message with /java come from?

ianoc-stripe:
The other one comes if you invoke the script directly:
```
~/github/bazel-helloworld$ bazel-bin/src/java/io/io
```
style or if a rule tries to use the java produced binary (which i guess are in some way equiv):
```
---8<---8<--- Start of log, file at /home/travis/.cache/bazel/_bazel_travis/7f7e07d167ded9af0372bc305ceff99e/bazel-workers/worker-4-Scalac.log ---8<---8<---
/home/travis/.cache/bazel/_bazel_travis/7f7e07d167ded9af0372bc305ceff99e/execroot/io_bazel_rules_scala/bazel-out/host/bin/src/java/io/bazel/rulesscala/scalac/scalac: line 350: /home/travis/.cache/bazel/_bazel_travis/7f7e07d167ded9af0372bc305ceff99e/execroot/io_bazel_rules_scala/bazel-out/host/bin/src/java/io/bazel/rulesscala/scalac/scalac.runfiles/local_jdk/bin/java: No such file or directory
---8<---8<--- End of log ---8<---8<---.
____Building complete.
____Elapsed time: 1.257s, Critical Path: 0.04s
```

ulfjack:
We need to narrow down where this is coming from. Let's treat these as two different bugs.

Can you do a with / without build and tell us if there are any differences in the bazel-bin/src/java/io/io file or its runfiles tree? Also, can you confirm that you're doing bazel build //src/java/io && bazel-bin/src/java/io/io?

ianoc-stripe:
Hi @ulfjack, sorry about the delay, long holiday weekend here --
tldr for all this the runfiles tree in cache hits is the same for build and run, but are much different than no-cache hit.

I have a simple test repo if its useful linked above @ https://github.com/ianoc-stripe/bazel-helloworld 
the test its doing is just :
```
set -e

rm -rf /webdav/bazel-cache/ac/*
rm -rf /webdav/bazel-cache/cas/*
bazel build src/java/io
bazel clean
bazel build //src/java/io 
./bazel-bin/src/java/io/io
```
https://travis-ci.org/ianoc-stripe/bazel-helloworld/builds/307893856?utm_source=github_status&utm_medium=notification

(so for any settings or other things i might have strange thats a good place to look)

I've attached i think what you were asking about, `ls -lR` on both run files tree's, doing bazel build, vs bazel run, with both a cache it + no cache it.
`--experimental_remote_spawn_cache --remote_rest_cache=http://localhost:9090/bazel-cache` used in all tests 
[cache_hit_build.txt](https://github.com/bazelbuild/rules_scala/files/1506047/cache_hit_build.txt)
[cache_hit_run.txt](https://github.com/bazelbuild/rules_scala/files/1506048/cache_hit_run.txt)
[no_cache_hit_build.txt](https://github.com/bazelbuild/rules_scala/files/1506049/no_cache_hit_build.txt)
[no_cache_hit_run.txt](https://github.com/bazelbuild/rules_scala/files/1506050/no_cache_hit_run.txt)

ulfjack:
Ok, I can reproduce the issue.

ulfjack:
Ok, I have a fix. The problem is that the runfiles tree is not being generated. The reason for that is that we create a symlink tree spawn, which we mark as local, but which is still getting cached by the experimental spawn cache. Unfortunately, this spawn doesn't declare its output files, so the cache result is actually incorrect. Duh!

This is the same issue as #4041.

Unfortunately, I can't think of a workaround, but I'll submit the fix ASAP.

ianoc-stripe:
Excellent! Thats great news. if you let me know when its somewhere public i'll grab that diff and start testing our internal builds/CI with it.

Thanks for the quick fix!

ulfjack:
https://bazel-review.googlesource.com/c/bazel/+/23350 - I decided to throw in some cleanup on top of the actual fix. The primary fix is in SymlinkTreeHelper.java.

ianoc-stripe:
@ulfjack Thats great! thanks for the fix, fixes any internal repo cases I had. Much appreciated!

ittaiz:
@ianoc-stripe can we close this and #350?

ianoc-stripe:
Yep!

buchgr:
FYI: We have recently added support for Google Cloud Storage as a caching backend [1]. Users have found this useful, because they don't need to run their own remote cache.

[1] https://github.com/bazelbuild/bazel/commit/dd11a0e2e7a143f89bca1be6e4cd56df0640dbe0

PR #340, opened 2017-11-15
Depending on java_proto_library exclusively causes reference to nonexistent object

djudd-stripe:
This PR demonstrates what is either a bug, or a misunderstanding on my part. When I build locally with these changes, I get errors like:

```
bazel-out/darwin_x86_64-fastbuild/bin/test/proto/tmp8951578470537701453/test/proto/TestResponse1.scala:98: error: object Test2 is not a member of package test.proto
object TestResponse1 extends com.trueaccord.scalapb.GeneratedMessageCompanion[test.proto.TestResponse1] with com.trueaccord.scalapb.JavaProtoSupport[test.proto.TestResponse1, test.proto.Test2.TestResponse1] {
```

If I understand correctly, what I'm doing here shouldn't break the test, because I'm just removing redundancy. In the real-world example that motivated this, I have a java_proto_library where I can't also straightforwardly define a corresponding proto_library, because it's actually coming from another repository.

If that's not something that I should expect to work / I'm doing something wrong, please let me know.

cc @johnynek @azymnis

### DIFF
--- a/test/proto/BUILD
+++ b/test/proto/BUILD
@@ -30,12 +30,12 @@ scalapb_proto_library(
 
 java_proto_library(
     name = "test_proto_java_lib",
-    deps = [":test2", "//test/proto2:test"],
+    deps = ["//test/proto2:test"],
 )
 
 scalapb_proto_library(
     name = "test_proto_java_conversions",
-    deps = [":test2", "//test/proto2:test", ":test_proto_java_lib"],
+    deps = [":test2", ":test_proto_java_lib"],
     with_java = True,
     with_flat_package = True,
     visibility = ["//visibility:public"],


bazel-io:

azymnis:
@djudd-stripe the target `test2` is defining a new set of proto classes. Why do you think it is redundant?

azymnis:
@djudd-stripe regarding proto files pulled from external repos: unfortunately the only way to actually do that is to define bazel proto library targets in that external repo and then depend on them from the other repo. If you try to depend on the java or scala proto libraries the compilation path gets screwed up

azymnis:
@djudd-stripe oh also one more thing: depending on a proto library won't actualluy recursively pull its deps. AFAIK you need to specify them all explicitly (this is true for java and scala targets)

djudd-stripe:
The reason I believe(d) there was redundancy was:
- `test_proto_java_lib` is only used in `test_proto_java_conversions`, where it is included as a dependency
- `test2` and `"//test/proto2:test"` were both present in the deps of both `test_proto_java_lib` and `test_proto_java_conversions`; I removed one from each, so each is now only present in the _transitive_ dependencies of `test_proto_java_conversions` once

I understand that (unfortunately IMHO), proto_library deps don't work recursively, and I take your word for it that a java_proto_library isn't enough to make things work with an external repo. But it doesn't seem like either of those things is strictly relevant here. If having a Java library isn't enough to let us create corresponding Scala classes, and we also need to depend directly on the Java library's source protos, why do we allow depending on one?

azymnis:
The reason we need the source proto files is that these are needs by
scalapb to generate the scala source files. We cannot generate Scala code
from the java proto lib. You only need to include java_proto_libs in deps
if you have the with_java option enabled. What this will do is that it will
tell scalapb to generate automatic converters from java protos to Scala
protos and vice versa. If you don’t use java protos in your codebase and
only need the Scala classe, just set this flag to false and don’t pass any
java proto lib deps.

Does this make sense? If it’s not clear, I can update the documentation

djudd-stripe:
Thanks - that makes sense! Wasn't clear to me from the documentation, so an update would probably be useful. But I'll close this out.

azymnis:
Sounds good. I'll send a PR with updated docs

PR #341, opened 2017-11-15
Add RunWith for JunitJavaTest.

chaoren:


ittaiz:
Why? If I remember correctly the whole point was to show this is supported without explicit RunWith

chaoren:
Whoops. I was updating our internal rules_scala repo, and this thing refused to compile because of the missing RunWith, probably because we have stricter checks than external bazel.

Issue #327, opened 2017-10-31
no-src jars do not add entries for directories


ittaiz:
#289 changed the behavior of zipping a no-src jar in the sense that the zipper doesn't create entries for the directories. I guess that we're just using it wrong and that we need to also feed it the directories entries themselves but haven't gotten around to verifying and fixing it.
This is a problem for example when you want to get all files from a directory on the classpath. In that case the directory isn't found.
Current workaround: use java_library.

johnynek:
which directories? The paths to items?

Would be nice to produce a failing test.

ittaiz:
yes.
I know. I didn't report it for the past two weeks since I preferred submitting it with a failing test but I saw I'm not getting around to it so I decided to at least post it as is

PR #342, opened 2017-11-17
Add failing test for jar with only resources

djudd-stripe:
Context: when we generate a `scala_library` with only resources, and no srcs, we end up unable to load those resources. This appears to mean that there's a bug in this method: https://github.com/bazelbuild/rules_scala/blob/master/scala/scala.bzl#L81

I tried reverting @johnynek's commits 830ff628e693de9c172bdaa261fe1fadef673b76 and 65ef1009346955c3513025f4c3938aa121790fc0, the last major change to that method, but doing so didn't appear to fix the bug.

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -160,6 +160,14 @@ scala_library(
 )
 
 scala_library(
+    name = "ResourcesOnlyLib",
+    resources = [
+        "//test/data:bar.txt",
+        "//test/src/main/resources/scala/test:welcomes",
+    ],
+)
+
+scala_library(
     name = "ResourcesStripScalaLib",
     srcs = ["src/main/scala/scala/test/ResourcesStripScalaLib.scala"],
     resources = [
@@ -176,7 +184,7 @@ scala_binary(
     name = "ScalaLibBinary",
     srcs = ["src/main/scala/scala/test/ScalaLibBinary.scala"],
     main_class = "scala.test.ScalaLibBinary",
-    deps = ["ScalaLibResources"],
+    deps = ["ScalaLibResources", "ResourcesOnlyLib"],
 )
 
 scala_binary(
--- a/test/data/BUILD
+++ b/test/data/BUILD
@@ -1,1 +1,1 @@
-exports_files(["some.txt", "more.txt", "foo.txt"])
+exports_files(["some.txt", "more.txt", "foo.txt", "bar.txt"])
--- a/test/data/bar.txt
+++ b/test/data/bar.txt
@@ -1,0 +1,1 @@
+baz, of course!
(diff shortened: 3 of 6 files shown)


bazel-io:

ittaiz:
@djudd-stripe can you update the description with context please?
(bot) test this please

djudd-stripe:
@ittaiz is that better?

BTW, I should be covered under the Stripe CLA. Unsure what the process is to teach the bot that.

ittaiz:
Could this be a repro for #327?
I'm a nutshell the new tool needs explicit directory entries or otherwise you can't load directories (though single files can be loaded).
We worked around our problem by using java_library since it was a distinct use case.

djudd-stripe:
Looks related, although here I think we're only dealing with single files and still seeing a problem?

ittaiz:
I think you might have a mistake.
1. Your test is incorrect, `TestResourceJars` which belongs to the target `resource_jars` doesn't depend on `welcomes` so it can't see that resource. You added `welcomes` to `ScalaLibBinary` which wasn't changed and isn't really relevant here.
2. There is an existing e2e test `scala_library_jar_without_srcs_must_include_direct_file_resources` which checks that the resource file exists in the jar.
3. I've pushed a [PR](https://github.com/bazelbuild/rules_scala/pull/344) with a passing test which demonstrates successful loading of a resource from a no src library

ittaiz:
It could be great if you could give a really small snippet of how you're actually loading the resources.
You might be falling into one of the caveats mentioned here (http://www.thinkplexx.com/learn/howto/java/system/java-resource-loading-explained-absolute-and-relative-names-difference-between-classloader-and-class-resource-loading) or just to the other bug I mentioned with the directory entries

djudd-stripe:
Sorry for the false positive! I no longer remember why I thought ScalaLibBinary was relevant.

Context: I'm trying to diagnose a regression in upgrading rules_scala from e7791fb6bcf006427968d6620fcb2e2b84197220 to 85308acbd316477f3072e033e7744debcba4f054

It's entirely possible that we're doing something invalid, but for whatever reason it used to work anyway.

We have a bit in WORKSPACE that looks like this:

```
new_git_repository(
    name = "some_resource_repo",
    remote = "...",
    commit = "...",
    build_file_content = """
filegroup(
    name = "lib_data",
    srcs = glob(["lib/data/**/*"]),
)
"""
)
```

... then we do in a BUILD file:
```
scala_library(
    name = "utils",
    srcs = ...,
    resources = [
        ...,
        "@some_resource_repo//:lib_data",
    ],
```

... finally, in code in that "utils" library, we do:

```
getClass.getResourceAsStream("/external/some_resource_repo/lib/data/a_yaml_file.yaml")
```

... but getResourceAsStream returns null.

djudd-stripe:
I'll try to iterate on your branch and get an actual reproduction.

ittaiz:
We merged it to master already.
If you'll get a failing test it will be easy to see if it's related to the
directories problem I mentioned.

djudd-stripe:
Closing in favor of https://github.com/bazelbuild/rules_scala/pull/345 - which I think has an actual reproduction? The issue seems only to occur with resources from an external dependency. I'm continuing to try to pin it down

PR #345, opened 2017-11-19
Add passing & failing tests for getting resources

djudd-stripe:
I believe this is an actual reproduction for the issue discussed in https://github.com/bazelbuild/rules_scala/pull/342

Context: I'm trying to diagnose a regression in upgrading rules_scala from e7791fb6bcf006427968d6620fcb2e2b84197220 to 85308acbd316477f3072e033e7744debcba4f054

It's entirely possible that we're doing something invalid, but for whatever reason it used to work anyway.

We have a bit in WORKSPACE that looks like this:

```
new_git_repository(
    name = "some_resource_repo",
    remote = "...",
    commit = "...",
    build_file_content = """
filegroup(
    name = "lib_data",
    srcs = glob(["lib/data/**/*"]),
)
"""
)
```

... then we do in a BUILD file:
```
scala_library(
    name = "utils",
    srcs = ...,
    resources = [
        ...,
        "@some_resource_repo//:lib_data",
    ],
```

... finally, in code in that "utils" library, we do:

```
getClass.getResourceAsStream("/external/some_resource_repo/lib/data/a_yaml_file.yaml")
```

... but getResourceAsStream returns null.

@ittaiz suggested the issue might be #327, but my test in the previous PR turned out to be broken


bazel-io:

djudd-stripe:
Hmm, I suspect there's something wrong with my tests here too, because one also fails if cherry-picked and added back off the commit that works with our production code. :(

azymnis:
@djudd-stripe I suspect that all that has happened is that the resource path has changed. Mind building the jar with external resources and listing its contents to see where the resource lives?

```
jar tf my_jar.jar | grep a_yaml_file.yaml
```

johnynek:
The jar is missing the deps it looks like:

```
bazel build test/src/main/scala/scala/test/resources:TestScalaLibResourcesFromExternalDep

st-oscar1:rules_scala oscar$ jar tf bazel-bin/test/src/main/scala/scala/test/resources/TestScalaLibResourcesFromExternalDep.jar
META-INF/
META-INF/MANIFEST.MF
scala/
scala/test/
scala/test/resources/
scala/test/resources/ScalaLibResourcesFromExternalDepTest$$anonfun$1$$anonfun$apply$1$$anonfun$apply$2.class
scala/test/resources/ScalaLibResourcesFromExternalDepTest$$anonfun$1$$anonfun$apply$1$$anonfun$apply$3$$anonfun$apply$4.class
scala/test/resources/ScalaLibResourcesFromExternalDepTest$$anonfun$1$$anonfun$apply$1$$anonfun$apply$3.class
scala/test/resources/ScalaLibResourcesFromExternalDepTest$$anonfun$1$$anonfun$apply$1.class
scala/test/resources/ScalaLibResourcesFromExternalDepTest$$anonfun$1.class
scala/test/resources/ScalaLibResourcesFromExternalDepTest.class
```

johnynek:
This is different from our internal case, which I think is not using a filegroup. Also our internal case is using a jar as a dependency, I think, and not directly in the test.

Lastly, we don't use the the specs rules which exercise a different code path (cc @ittaiz who may care about remote resources in specs rules).

The next thing to see is; does this repro with the `scala_test` rule. I'll do that now (between meetings).

johnynek:
with `scala_test` we get a fail:
```
INFO: Found 391 targets and 37 test targets...
ERROR: /Users/oscar/oss/rules_scala/test/src/main/scala/scala/test/resources/BUILD:43:1: scala //test/src/main/scala/scala/test/resources:TestScalaLibResourcesFromExternalScalaTest failed (Exit 1): scalac failed:
 error executing command
  (cd /private/var/tmp/_bazel_oscar/cff0ec2cf472ebc5cb4402685dc7edf3/execroot/io_bazel_rules_scala && \
  exec env - \
  bazel-out/host/bin/src/java/io/bazel/rulesscala/scalac/scalac @bazel-out/darwin_x86_64-fastbuild/bin/test/src/main/scala/scala/test/resources/TestScalaLibResourcesFromExternalScalaTest_worker_input).
java.nio.file.FileAlreadyExistsException: bazel-out/darwin_x86_64-fastbuild/bin/test/src/main/scala/scala/test/resources/tmp4896264759761054167/../test_new_local_repo/resource.txt
        at sun.nio.fs.UnixCopyFile.copy(UnixCopyFile.java:551)
        at sun.nio.fs.UnixFileSystemProvider.copy(UnixFileSystemProvider.java:253)
        at java.nio.file.Files.copy(Files.java:1274)
        at io.bazel.rulesscala.scalac.ScalacProcessor.copyResources(ScalacProcessor.java:305)
        at io.bazel.rulesscala.scalac.ScalacProcessor.processRequest(ScalacProcessor.java:74)
        at io.bazel.rulesscala.worker.GenericWorker.runPersistentWorker(GenericWorker.java:52)
        at io.bazel.rulesscala.worker.GenericWorker.run(GenericWorker.java:121)
INFO: Elapsed time: 0.969s, Critical Path: 0.51s
```

There is definitely something janky about resources. If you look in:
`bazel-out/darwin_x86_64-fastbuild/bin/test/src/main/scala/scala/test/resources` there is another nested `bazel-out` directory hanging out. That does not seem like it should be the case. Also the `../` out of the tmp directory is suspect to me.

johnynek:
The issue about the file existing seems like a hermeticity issue to me. The file is around from a previous run and has been copied to location on disk. We then suck it up into a jar, but we don't delete the old files. I would have thought bazel would sandbox this so it can't be a problem.

When we do:
```
cp resource.txt bazel-out/darwin_x86_64-fastbuild/bin/test/src/main/scala/scala/test/resources/tmp4896264759761054167/../test_new_local_repo/resource.txt
```
the `../` puts us in the same output as the previous build, which causes the error. @damienmg am I wrong to expect that writing outside of a temporary directory should be caught by bazel sandboxing?

johnynek:
This seems related to paths interacting with external repos.

johnynek:
More digging, after instrumenting the ScalacProcessor to print out the resources we see:
```
source: test/data/more.txt
dest: test/data/more.txt
short: test/data/more.txt
source: test/data/foo.txt
dest: test/data/foo.txt
short: test/data/foo.txt
source: bazel-out/darwin_x86_64-fastbuild/genfiles/test/src/main/resources/scala/test/generated-hello.txt
dest: /scala/test/generated-hello.txt
short: test/src/main/resources/scala/test/generated-hello.txt
source: test/src/main/resources/scala/test/more-byes
dest: /scala/test/more-byes
short: test/src/main/resources/scala/test/more-byes
source: test/src/main/resources/scala/test/more-hellos
dest: /scala/test/more-hellos
short: test/src/main/resources/scala/test/more-hellos
ERROR: /Users/oscar/oss/rules_scala/test/src/main/scala/scala/test/resources/BUILD:43:1: scala //test/src/main/scala/scala/test/resources:TestScalaLibResourcesFromExternalScalaTest failed (Exit 1): scalac failed: error executing command
  (cd /private/var/tmp/_bazel_oscar/cff0ec2cf472ebc5cb4402685dc7edf3/execroot/io_bazel_rules_scala && \
  exec env - \
  bazel-out/host/bin/src/java/io/bazel/rulesscala/scalac/scalac @bazel-out/darwin_x86_64-fastbuild/bin/test/src/main/scala/scala/test/resources/TestScalaLibResourcesFromExternalScalaTest_worker_input).
source: external/test_new_local_repo/resource.txt
dest: ../test_new_local_repo/resource.txt
short: ../test_new_local_repo/resource.txt
```
so notice the source, destination and short path are almost always the same. In the case of external resources we wind up getting a `..` in the path. See the path generation code here:
https://github.com/bazelbuild/rules_scala/blob/master/scala/scala.bzl#L219

johnynek:
I have a fix building on this branch in #347

johnynek:
these commits were pulled into #347

ittaiz:
Just to make sure the specs2 and scala_test ended up being the same issue, right?
Also,
Is there a hermeticity issue?

johnynek:
as far as I have found, it is only the issue that the `.path` on a file from an external repo starts with `../` for some reason.

I don't really know why that should be.

PR #356, opened 2017-12-03
Handling special (regex) characters and multiple test names in specs2

### LINKED ISSUE #355: Support test names with special characters and braces in specs2
The specs2 test runner does not currently handle unescaped regex input (which may be invalid due to any unsafe characters may appear as part of the test name).

This is already handled by the Bazel plugin - the test filter is escaped (using `Pattern.quote`) before sent to bazel.

hmemcpy:
This fixes #355

### DIFF
--- a/src/java/io/bazel/rulesscala/specs2/Specs2RunnerBuilder.scala
+++ b/src/java/io/bazel/rulesscala/specs2/Specs2RunnerBuilder.scala
@@ -34,7 +34,28 @@ object Specs2FilteringRunnerBuilder {
 class Specs2ClassRunner(testClass: Class[_], testFilter: Pattern)
   extends org.specs2.runner.JUnitRunner(testClass) {
 
-  private def toDisplayName(d: Description) = d.getMethodName.split("::").reverse.headOption.getOrElse(d.getMethodName)
+  /** Taken from specs2: replaces () with [] because it cause display issues in JUnit plugins */
+  private def sanitize(testName: String) = {
+    val sanitized = testName.trim.replace('(', '[').replace(')', ']')
+    if (sanitized.isEmpty) " "
+    else sanitized
+  }
+
+  /**
+    * Retrieves an original (unsanitized) text of an example fragment,
+    * used later as a regex string for specs2 matching.
+    *
+    * This is done by matching the actual (sanitized) string with the sanitized version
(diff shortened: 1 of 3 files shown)


bazel-io:

hmemcpy:
@ittaiz please review

ittaiz:
Test this please

src/java/io/bazel/rulesscala/specs2/Specs2RunnerBuilder.scala
```diff
@@ -66,15 +87,15 @@ class Specs2ClassRunner(testClass: Class[_], testFilter: Pattern)
 
   private def specs2ExamplesMatching(testFilter: Pattern, junitDescription: Description): Option[String] =
     flattenDescription(junitDescription)
-      .filter(matching(testFilter))
```
ittaiz: why the change from filter to find?

hmemcpy: Hmm, a leftover from trying different things. Will fix, thanks!

ittaiz: Maybe we need another test?

hmemcpy: Perhaps. The test filter string doesn't currently support passing more than one test method, so currently there isn't a difference between `find` and `filter`. To be honest, I don't even know how to specify multiple test names in intellij...

ittaiz:
can't a regex utilize this? to match multiple ones?
Also this interface is for intellij and command line

hmemcpy:
Yeah, the regex should be able to handle multiple test methods, e.g. `foo should::(do bar|do baz)`, but with the escaping, they have to be individually escaped, i.e.: `foo should::(\Qdo bar\E|\Qdo baz\E)`.

Most of this is already supported (the intellij plugin now escapes individual test names). It needs to be tested and tweaked, if necessary, but perhaps this should be an explicit issue.

ittaiz:
Ok, if so then please leave it as find and open a detailed issue about this
missing feature so people will know it's only partly supported

ittaiz:
test this please

ittaiz:
test this please

PR #357, opened 2017-12-05
Fix for scala_import's export/ijar bug

### LINKED ISSUE #354: scala_import mixes ijars into the runtime classpath
@gkk-stripe may follow up more but we saw an issue with scala_import preventing us from using it.

The issue seems to be pretty easy to fix (sorry I’m on the phone and can’t easily link). There is a spot where you explicitly put a compile jar onto the runtime path. This can break. Since compile jars may be ijars and they don’t have code sections.

We didn’t see all the ijars on the runtime path, only a subset and we didn’t guess right away which they were.

gkk-stripe:
Fix a runtime error due to ijar dependendency sneaking in. See #354 for
details.

Fixes #354

### DIFF
--- a/test/src/main/scala/scala/test/scala_import/BUILD
+++ b/test/src/main/scala/scala/test/scala_import/BUILD
@@ -49,10 +49,23 @@ scala_specs2_junit_test(
     suffixes = ["Test"],
 )
 
+java_import(
+  name = "guava_and_commons_lang_java_import",
+  jars = ["@com_google_guava_guava_21_0//jar:file", "@org_apache_commons_commons_lang_3_5//jar:file"],
+)
+
 #Exports
 scala_import(
   name = "guava_and_commons_lang_as_exports",
   exports = [":guava_and_commons_lang"],
+)
+
+# we're testing the scenario:
+# jar <- java_import <- scala_import (exports) <- scala_specs2_junit_test
+# see for details https://github.com/bazelbuild/rules_scala/issues/354
+scala_import(
+  name = "guava_and_commons_lang_as_exports_from_java_import",
+  exports = [":guava_and_commons_lang_java_import"],
 )
 
 scala_specs2_junit_test(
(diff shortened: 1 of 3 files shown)


bazel-io:

gkk-stripe:
@ittaiz I went ahead with the original fix that improves over the current state. This unblocks Stripe to adopt the official IntelliJ plugin and retire hacky scripts we have internally. This is a great milestone for us.

ittaiz:
test this please

ittaiz:
@gkk-stripe this fixes it and if the build passes I'll merge it but it creates another problem and doesn't resolve an existing one.
The created problem is depending on full jars instead of ijars for exports (minor)
The existing one is that exports don't propagate their dependencies.
If you use scala_import with exports a bit like one uses maven pom as an aggregator then this might be a problem

johnynek:
seems worthwhile to merge and address the other issues in follow ups.

gkossakowski:
https://github.com/bazelbuild/rules_scala/issues/359

test/src/main/scala/scala/test/scala_import/BUILD
```diff
@@ -49,10 +49,23 @@ scala_specs2_junit_test(
...
+java_import(
+  name = "guava_and_commons_lang_java_import",
+  jars = ["@com_google_guava_guava_21_0//jar:file", "@org_apache_commons_commons_lang_3_5//jar:file"],
+)
+
 #Exports
 scala_import(
   name = "guava_and_commons_lang_as_exports",
-  exports = [":guava_and_commons_lang"],
+  exports = [":guava_and_commons_lang_java_import"],
```
ittaiz: should we be changing this line? because i see you added an example below

gkossakowski: oops, I forgot to remote that during cleanup, let me push a commit

gkossakowski:
pushed fix, i don't have merge rights, though

gkossakowski:
The 0.6 build failed: https://travis-ci.org/bazelbuild/rules_scala/jobs/312088834
given other chats, is the plan to ditch support for 0.6?

johnynek:
I think moving to only test on 0.7 and 0.8 is fine personally.

ittaiz:
I've added a commit which updates the travis config to use 0.7 as minimum and latest as 0.8.1
Messed up cla bot but it's still me and gkk

ittaiz:
test this please

gkk-stripe:
\o/
Thanks

PR #361, opened 2017-12-06
Fix bug in the provider we create in scala_library/binary/rest

ittaiz:
scala_* rules create java_provider with compile_jars in transitive_compile_jars

@johnynek we saw internally that with a scala_library depending on a scala_import depending on scala_library  (somewhat similar to bazel-deps `replacements` mode) the compile jars of the root scala library dep aren't propagated. Only its transitive compile ones.
I initially fixed it in scala_import but then I suspected the root problem is with the scala rules themselves.
I added two tests that show it and actually removed the crutches we had (scala.bzl and custom_jvm_rule) which did it themselves. This was a problem since every passing rule type needed to implement it themselves (like scala_import which is how I caught this).
@damienmg @lberki @dslomov @iirina 
FYI- The current state of the create_provider is causing a massive amount of pain and bugs.
We have so much code and tests around it and still sometimes fall over...

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -541,7 +541,7 @@ def _collect_jars_when_dependency_analyzer_is_on(dep_targets):
     if java_common.provider in dep_target:
         java_provider = dep_target[java_common.provider]
         current_dep_compile_jars = java_provider.compile_jars
-        current_dep_transitive_compile_jars = java_provider.transitive_compile_time_jars + java_provider.compile_jars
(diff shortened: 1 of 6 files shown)


ittaiz:
@johnynek
Can you review?
This is a big bug so I'd like to merge this as soon as possible

johnynek:
Sorry I lost track of this.

ittaiz:
Sure.
If you have feedback I'd still love to hear it.
Take into account though that the whole create_provider api is being
completely revamped right now so the smartest move is probably to hang
tight and wait for it to land.

PR #369, opened 2017-12-21
Scala import exports fixes

### LINKED ISSUE #359: scala_import's exports doesn't propagate its dependencies
> The existing one is that exports don't propagate their dependencies.
> If you use scala_import with exports a bit like one uses maven pom as an aggregator then this might be a problem

https://github.com/bazelbuild/rules_scala/pull/357#issuecomment-349423855

ittaiz:
fixes #359 
Additionally it fixes label propagation of transitive deps of exports
@johnynek this has no tests since quite honestly I'm tired from all of the permutations we need to do here. Given that https://github.com/bazelbuild/bazel/issues/3769 is being worked on and it will remove 90% of the code in this rule I feel ok with this.
We use this code as is today.
cc @gkk-stripe

### DIFF
--- a/scala/scala_import.bzl
+++ b/scala/scala_import.bzl
@@ -4,10 +4,12 @@
 def _scala_import_impl(ctx):
     target_data = _code_jars_and_intellij_metadata_from(ctx.attr.jars)
     (current_target_compile_jars, intellij_metadata) = (target_data.code_jars, target_data.intellij_metadata)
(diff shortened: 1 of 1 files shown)


gkossakowski:
thanks for the fix!
I'm working on integrating the scala_import into our build and hitting some snags. I'll open an issue here in rules to discuss it.

johnynek:
looks good to me.

merge when green.

This stuff is in flux. Let's get experience running it and add tests as issues are found. By the time we add tests, bazel will probably change as you say. :)

PR #373, opened 2017-12-24
Specs2 produced junit xml report as additional test output

hmemcpy:
This is needed to use the report with external tools, e.g. comparing against junit xml reports produced by other runners (e.g. Surefire)

@ittaiz please review

### DIFF
--- a/specs2/BUILD
+++ b/specs2/BUILD
@@ -1,13 +1,14 @@
+load("//scala:scala_import.bzl", "scala_import")
+
 package(default_visibility = ["//visibility:public"])
 java_import(
     name = "specs2",
-    jars = [],
-    exports = [
-      "@io_bazel_rules_scala_org_specs2_specs2_core//jar",
-      "@io_bazel_rules_scala_org_specs2_specs2_common//jar",
-      "@io_bazel_rules_scala_org_specs2_specs2_matcher//jar",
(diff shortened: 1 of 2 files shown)


bazel-io:

specs2/BUILD
```diff
@@ -1,13 +1,14 @@
+load("//scala:scala_import.bzl", "scala_import")
+
 package(default_visibility = ["//visibility:public"])
-java_import(
+scala_import(
```
ittaiz: Why is this change needed?

hmemcpy:
Hmm, I saw this, but I was sure this was just for scalatest?

specs2/BUILD
hmemcpy: I think it's a leftover from our change with the jars, I'll undo and see what happens.

ittaiz:
build fails...

hmemcpy:
Maybe it just doesn't like me? :)
I'll check...

specs2/BUILD
hmemcpy: Ok, I reverted it back to java_import, still works!

hmemcpy:
After further discussion with @ittaiz, I'm closing this PR in favor of a local cherry-pick for this. We need this temporarily to verify the migration was done correctly, and there's no reason to keep this in rules_scala.

PR #374, opened 2017-12-25
Specs2: fixing a NPE when trying to match against a "should" descripition

hmemcpy:
This happens if a give test filter matches everything (e.g. `.*`), including the intermediate `should` scopes, which are not real test method descriptions.


bazel-io:

ittaiz:
Test this please

hmemcpy:
Yeah :(

PR #378, opened 2018-01-03
Remove Skylark dictionary concatenation.

benjaminp:
As described in
https://docs.bazel.build/versions/master/skylark/backward-compatibility.html#dictionary-concatenation,
the ability to + dictionaries in Skylark is going away soon. This
change removes this practice for rules_scala. Alas, it makes the rule
attr code a bit wordier.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -928,12 +928,13 @@ _common_attrs_for_plugin_bootstrapping = {
   "print_compile_time": attr.bool(default=False, mandatory=False),
 }
 
-_common_attrs = _common_attrs_for_plugin_bootstrapping + {
+_common_attrs = dict(_common_attrs_for_plugin_bootstrapping)
+_common_attrs.update({
   # using stricts scala deps is done by using command line flag called 'strict_java_deps'
   # switching mode to "on" means that ANY API change in a target's transitive dependencies will trigger a recompilation of that target,
   # on the other hand any internal change (i.e. on code that ijar omits) WON’T trigger recompilation by transitive dependencies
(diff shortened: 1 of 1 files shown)


bazel-io:

scala/scala.bzl
```diff
@@ -946,68 +947,98 @@ common_outputs = {
   "manifest": "%{name}_MANIFEST.MF",
 }
 
-library_outputs = common_outputs + {
+library_outputs = dict(common_outputs)
```
johnynek: Can you use the same pattern as starting empty then use update as you do below?

scala/scala.bzl
```diff
@@ -928,12 +928,13 @@ _common_attrs_for_plugin_bootstrapping = {
   "print_compile_time": attr.bool(default=False, mandatory=False),
 }
 
-_common_attrs = _common_attrs_for_plugin_bootstrapping + {
+_common_attrs = dict(_common_attrs_for_plugin_bootstrapping)
```
ittaiz: Why do you use dict() here but below (for example _scala_test_attrs you don't use it?

johnynek:
Test this please

ittaiz:
Why did you revert the travis yml change?

benjaminp:
I reverted `.travis.yml` change because Bazel 0.8 has internal code that uses dictionary concatenation, which causes the whole build to fail. (This is not an infrequent problem with the incompatible change flags.)

Latest update always uses the  `.update()` pattern.

johnynek:
Test this please

johnynek:
Thank you!

PR #384, opened 2018-01-10
Depend on @bazel_tools//tools/jdk:current_java_runtime for the JDK.

lberki:
//tools/defaults:jdk is a filegroup and won't be compatible with Bazel 0.10 .

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -889,7 +889,7 @@ _implicit_deps = {
   "_java": attr.label(executable=True, cfg="host", default=Label("@bazel_tools//tools/jdk:java"), allow_files=True),
   "_zipper": attr.label(executable=True, cfg="host", default=Label("@bazel_tools//tools/zip:zipper"), allow_files=True),
   "_java_toolchain": attr.label(default = Label("@bazel_tools//tools/jdk:current_java_toolchain")),
-  "_host_javabase": attr.label(default = Label("//tools/defaults:jdk"))
+  "_host_javabase": attr.label(default = Label("@bazel_tools//tools/jdk:current_java_runtime"))
 }
 
 # Single dep to allow IDEs to pickup all the implicit dependencies.
@@ -1234,4 +1234,3 @@ def scala_specs2_junit_test(name, **kwargs):
    suite_label = Label("//src/java/io/bazel/rulesscala/specs2:specs2_test_discovery"),
    suite_class = "io.bazel.rulesscala.specs2.Specs2DiscoveredTestSuite",
    **kwargs)
-


johnynek:
👍

scala/scala.bzl
```diff
@@ -889,7 +889,7 @@ _implicit_deps = {
   "_java": attr.label(executable=True, cfg="host", default=Label("@bazel_tools//tools/jdk:java"), allow_files=True),
   "_zipper": attr.label(executable=True, cfg="host", default=Label("@bazel_tools//tools/zip:zipper"), allow_files=True),
   "_java_toolchain": attr.label(default = Label("@bazel_tools//tools/jdk:current_java_toolchain")),
-  "_host_javabase": attr.label(default = Label("//tools/defaults:jdk"))
+  "_host_javabase": attr.label(default = Label("@bazel_tools//tools/jdk:current_java_runtime"))
```
katre: Shouldn't this either be current_host_java_runtime, or have cfg = "host"?

lberki:
@katre : yes, it should be current_host_java_runtime, but that doesn't exist in Bazel 0.9 and the Scala rules need to work with that. Since the host and target JDKs are usually the same in Bazel, it's not that big of a lie.

katre:
Using current_java_runtime with cfg="host" would still work, wouldn't it? I ask because I was testing rules_scala earlier today with some in-progress changes to JavaRuntimeInfo, and I saw errors around dependencies in this exact scenario.

lberki:
Error message is `ERROR: Analysis of target '//test/tut:some_example' failed; build aborted: no such package '@java_stub_template//file': Error downloading`, which doesn't seem related so I'll merge.

ittaiz:
@lberki what about @katre's question?

lberki:
Erm, I managed to not notice that one -- I think that one would actually work (although I remember being unexpectedly tripped up in the area by something, but I can't remember what it was).

Is the current setup in your way?

katre:
It's not causing me any problems currently but will need to be changed when JavaRuntimeInfo is a full fledged toolchain. I'll open a new PR when that's closer.

ittaiz:
@katre- we need to define a custom java toolchain. We can hold off a bit
until we have to do so; any idea when it will be "new" toolchain type and
hence easier to override?

katre:
It's something I want to do soon, currently working out where the priorities are. The tricky point won't be defining the java toolchains, it'll be a) updating the existing java rules to use it, and b) updating the java skylark API to use it.

ittaiz:
I meant tricky part for me as from what I read in the discussion list defining a custom java toolchain is not fun (although easier than a custom CROSSTOOL)
I'll wait a week or two and ping you. thanks

katre:
I've opened https://github.com/bazelbuild/bazel/issues/4592 to track this work, but please be warned that it's not currently scheduled and will be done only on a best-effort basis.

PR #387, opened 2018-01-11
add expand_location to scalac_jvm_flags, fix #386

### LINKED ISSUE #386: scalac_jvm_flags not expanding locations like jvm_flags
I need to pass in a flag with path to a resource file at compile time, something like this:

```
scala_library(
    ...
    resources = "//path:file",
    scalac_jvm_flags = ["-Dresource=$(location //path:file)"],
)
```

This doesn't work and I got the following error:

`java.io.FileNotFoundException: $(location /path:file) (No such file or directory)`

Replacing `scalac_jvm_flags` with `jvm_flags` fixed the issue.

Looking at `scala.bzl` I noticed this under `def _write_launcher`
`jvm_flags = " ".join([ctx.expand_location(f, ctx.attr.data) for f in jvm_flags])`

But this under `def _compile`
`arguments=["--jvm_flag=%s" % flag for flag in ctx.attr.scalac_jvm_flags] + ["@" + argfile.path],`

So there might be some discrepancy in flag handling?

nevillelyh:


bazel-io:

ittaiz:
Hi,
Any chance you can add a test case which demonstrates what you're trying to achieve? It will help understand why the code is needed over time and hopefully keep from regressions.

ittaiz:
Test this please (ci bot)

nevillelyh:
Sure. First time working on a build rule so just wanna make sure I'm going in the right direction. Will look into the tests.

nevillelyh:
Just added some tests. PTAL @ittaiz thanks.

ittaiz:
Looks good thanks!
Did you see the tests fail and for the right reason? Asking since these bash tests are so easy to mess up and I've had many mistakes with them in the past

ittaiz:
Test this please (ci bot)

ittaiz:
It actually seems it's not passing right now

nevillelyh:
Should be fixed. I forgot to run `./test_rules_scala.sh` with `ci` argument.

ittaiz:
Did you amend the commit? I don't see another one there (and so it's harder
for me to see the change from what I've reviewed already)

nevillelyh:
Sorry, bad habit :(
Fixed as a separate commit.

ittaiz:
Test this please

ittaiz:
Thanks for splitting it back!
Two questions:
1. You did see it fail and for the right reason right? Like I said above these bash tests have a way of surprising up :(
2. Why did you need to run the script with ci argument?

ittaiz:
I'm sorry if I'm being a bit dense but I'm pretty sure that's not how the tests work. You can either run them with `ci` or with `local`. 
Results should be the same and IIRC the main difference between them was to circumvent a travis nonsense. This is done once and shouldn't have any bearing on your specific tests or why the failed on travis.
The main reason I'm asking is because I'd like to ensure that contributors have the easiest experience we can give them and it sounds it wasn't for you if you ran the tests, they passed locally but broke on CI.

ittaiz:
Merged it anyway as the question isn't about your code but about our tests/process. Thanks for your contribution and time.

ittaiz:
@nevillelyh do you have a bit more info about:

> I'm sorry if I'm being a bit dense but I'm pretty sure that's not how the tests work. You can either run them with ci or with local.
> Results should be the same and IIRC the main difference between them was to circumvent a travis nonsense. This is done once and shouldn't have any bearing on your specific tests or why the failed on travis.

I want to make sure the contribution experience is easy and it doesn't sound like it if passes when you run it locally but breaks on ci

nevillelyh:
Before the last commit, `./test_rules_scala.sh` passes in my local env (Mac High Sierra) but fails in Travis, I figured it might have something to do with Travis running it with the `ci` argument.

But I also missed the `exit 0` branch in this "fix tests" commit, so that could be related.
https://github.com/bazelbuild/rules_scala/pull/387/commits/99a0db5c4ff8db5bf6a96c972a1f6b2f09dc0d65#diff-9e850e1e0893c9ea027f87c0b45dffabR270

Don't have access to a Mac now but on Linux the test script passes with or without `ci` arg and that fix.

ittaiz:
Strange...
Thanks for the info

PR #395, opened 2018-01-17
scala_import keeps labels of direct deps

ittaiz:
Currently scala_import doesn't keep labels of direct deps, only of transitive deps.
This causes "unknown label" warnings for some use-cases (as the test shows).
A more thorough revamp is needed for label propagation but I'm trying to sync with the bazel team on it since it sounds a bazel issue


ittaiz:
cc @iirina @dbabkin for strict deps labels issues which I referenced in the java common doc

ittaiz:
@or-shachar can you please review?

PR #396, opened 2018-01-17
jmh normalizes jar timestamps

ittaiz:
additionally build_is_identical e2e test spits out differences in specific jar md5 files and not only the change in all of them. Actually useful for debugging reproducibility problems


johnynek:
👍 
Great changes!

ittaiz:
@johnynek thanks!
Can you please review again?
I've thought about it again and I think that the better change is to default to normalizing the timestamps so that if someone adds another usage they won't have to cargo-cult this setter. WDYT?

ittaiz:
only mac builds haven't run due to travis ongoing mac issues, I've run locally on a mac and merging

PR #398, opened 2018-01-18
build_is_identical stabilization

ittaiz:


ittaiz:
@johnynek I was worried on how come the jmh problem never surfaced on CI and only locally and I have a theory that it's because `build_is_identical` ran before jmh and that inside the test itself it was "too quick". I moved `build_is_identical` to the end with a (hopefully) useful comment as well as added a 2 seconds sleep inside it to try and win different timestamps.
I also reordered a bit some of the tests like #219 so we can close it.

ittaiz:
@johnynek any objections?

johnynek:
no, this is great. Just frustrating that travis behaves so poorly.

ittaiz:
Indeed.
They've been having a rough few days with respect to mac builds

PR #406, opened 2018-01-23
Emit a the Scala compilation time for downstream metrics

andyscott:
As it stands there isn't clean path to getting target level metrics from Bazel. The build event protocol currently doesn't have timing information for non-test targets [0]. But that doesn't have to stop us from being the first to get this kind of metric from Scala [1].

This adds an output file, the "statsfile" with a very basic property file inspired syntax. The values can be read easily with basic text manipulation utilities, for example `awk -F '=' '$1 == "property_name" { print $2 }' 'path_to.statsfile'` to read the value for `property_name`.

One downside is that this statsfile is non-deterministic. As long as this file is only used for downstream metrics, it shouldn't be problematic. The long term hope is that timing information can be added to the build event protocol and that this PR can be reverted.

### Sample Usage

At Stripe, we're proxying scala rules with macros of the same name. The macro adds additional metrics related target that use the statsfile output. This looks roughly like:

```python
def scala_binary(
        name,
        srcs = [],
        # ...
):
    upstream_scala_binary(...)
    _add_timed_target(name, srcs)

# Adds an extra target named `{name}-timed` that prints out
# timing information to the console
def _add_timed_target(
        name,
        srcs
):
    cmd = """
# ...
TIME_MS=`awk -F '=' '$$1 == "build_time" {{ print $$2 }}' '{statsfile}'` 
# ...
"""
    cmd = cmd.format(statsfile = "$(location :%s.statsfile)" % name)
    native.genrule(
        name = "%s-timed" % name,
        tools = [":%s.statsfile" % name, "//tools:cloc"],
        srcs = srcs, cmd = cmd,
        ...)
```

We're also using https://github.com/AlDanial/cloc to derive the LOC/sec compilation speed. I'd be happy to share the fuller code once all of the kinds are worked out.

[0] Timing support for BEP asked here https://groups.google.com/forum/#!topic/bazel-discuss/WnWq1b9l8Es, by @johnynek 
[1] See open Scala issue https://github.com/scala/bug/issues/10601

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -96,7 +96,7 @@ def _build_nosrc_jar(ctx, buildijar):
     cmd = """
 rm -f {jar_output}
 {zipper} c {jar_output} @{path}
-echo 0 > {statsfile}
+touch {statsfile}
 """ + ijar_cmd
 
     cmd = cmd.format(
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -96,15 +96,17 @@ def _build_nosrc_jar(ctx, buildijar):
     cmd = """
 rm -f {jar_output}
 {zipper} c {jar_output} @{path}
+echo 0 > {statsfile}
 """ + ijar_cmd
 
     cmd = cmd.format(
         path = zipper_arg_path.path,
         jar_output=ctx.outputs.jar.path,
         zipper=ctx.executable._zipper.path,
+        statsfile=ctx.outputs.statsfile.path,
         )
 
-    outs = [ctx.outputs.jar]
+    outs = [ctx.outputs.jar, ctx.outputs.statsfile]
     if buildijar:
         outs.extend([ctx.outputs.ijar])
 
@@ -189,9 +191,7 @@ CurrentTarget: {current_target}
     compiler_classpath = ":".join([j.path for j in compiler_classpath_jars])
 
     toolchain = ctx.toolchains['@io_bazel_rules_scala//scala:toolchain_type']
-    scalacopts = toolchain.scalacopts + ctx.attr.scalacopts
-
-    statsfile = ctx.new_file(ctx.outputs.jar, "%s.statsfile" % ctx.label.name)
+    scalacopts = toolchain.scalacopts + ctx.attr.scalacopts    
 
     scalac_args = """
 Classpath: {cp}
@@ -234,16 +234,16 @@ StatsfileOutput: {statsfile_output}
         resource_strip_prefix=ctx.attr.resource_strip_prefix,
         resource_jars=",".join([f.path for f in ctx.files.resource_jars]),
         dependency_analyzer_mode = dependency_analyzer_mode,
-        statsfile_output = statsfile.path
+        statsfile_output = ctx.outputs.statsfile.path
         )
     argfile = ctx.new_file(
       ctx.outputs.jar,
       "%s_worker_input" % ctx.label.name
     )
 
-    ctx.file_action(output=argfile, content=scalac_args + optional_scalac_args)    
+    ctx.file_action(output=argfile, content=scalac_args + optional_scalac_args)
 
-    outs = [ctx.outputs.jar, statsfile]
+    outs = [ctx.outputs.jar, ctx.outputs.statsfile]
     if buildijar:
         outs.extend([ctx.outputs.ijar])
     # _java_toolchain added manually since _java doesn't currently setup runfiles
@@ -958,6 +958,7 @@ common_outputs = {
   "jar": "%{name}.jar",
   "deploy_jar": "%{name}_deploy.jar",
   "manifest": "%{name}_MANIFEST.MF",
+  "statsfile": "%{name}.statsfile",
 }
 
 library_outputs = {}
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -96,6 +96,7 @@ def _build_nosrc_jar(ctx, buildijar):
     cmd = """
 rm -f {jar_output}
 {zipper} c {jar_output} @{path}
+# ensures that empty src targets still emit a statsfile
 touch {statsfile}
 """ + ijar_cmd
 
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -191,6 +191,8 @@ CurrentTarget: {current_target}
     toolchain = ctx.toolchains['@io_bazel_rules_scala//scala:toolchain_type']
     scalacopts = toolchain.scalacopts + ctx.attr.scalacopts
 
+    statsfile = ctx.new_file(ctx.outputs.jar, "%s.statsfile" % ctx.label.name)
+
     scalac_args = """
 Classpath: {cp}
 EnableIjar: {enableijar}
@@ -210,6 +212,7 @@ ResourceStripPrefix: {resource_strip_prefix}
 ScalacOpts: {scala_opts}
 SourceJars: {srcjars}
 DependencyAnalyzerMode: {dependency_analyzer_mode}
+StatsfileOutput: {statsfile_output}
 """.format(
         out=ctx.outputs.jar.path,
         manifest=ctx.outputs.manifest.path,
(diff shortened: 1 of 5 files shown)


bazel-io:

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
```diff
@@ -247,6 +248,13 @@ private static void compileScalaSources(CompileOptions ops, String[] scalaSource
       System.err.println("Compiler runtime: " + (stop - start) + "ms.");
     }
 
+    try {
+      PrintWriter writer = new PrintWriter(ops.statsfile, "UTF-8");
+      writer.println(Long.toString(stop - start));
+      writer.close();
+    } catch (Throwable ex) {
```
gkossakowski: catching throwable is a bad idea because it catches JVM errors
you need to catch IOException, right?

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
```diff
@@ -247,6 +248,13 @@ private static void compileScalaSources(CompileOptions ops, String[] scalaSource
...
+      throw new RuntimeException("YOLO", ex);
```
johnynek: can we have a better message here?

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
```diff
@@ -247,6 +248,13 @@ private static void compileScalaSources(CompileOptions ops, String[] scalaSource
       System.err.println("Compiler runtime: " + (stop - start) + "ms.");
     }
 
+    try {
+      PrintWriter writer = new PrintWriter(ops.statsfile, "UTF-8");
```
ittaiz: WDYT about using `java.nio.Files#write` instead of these three lines?

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
andyscott: At the time of writing this I figured that if the file fails to write then there must be a show-stopping issue with the build.
- Should the build continue if we can't write the file?
- YOLO might not cut is as the error message :)

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
```diff
@@ -247,6 +248,13 @@ private static void compileScalaSources(CompileOptions ops, String[] scalaSource
...
+      writer.println(Long.toString(stop - start));
```
johnynek: should this be a bit more structured? Maybe some hand crafted json even:
```json
{
  build_time: 123123
}
```
so we could add other examples of this?

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
johnynek: could we catch IOException instead?

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
andyscott: There's no built in way to process JSON downstream, either from common shell scripts, or via Skylark APIs. What about a simple properties style file and I drop the assumption that the build time appears on a given line?
```
baz=false
build_time= 123123
foo=bar
```
This can be read with the one liner:
```bash
BUILD_TIME=`awk -F '=' '$1 == "build_time" { print $2 }' some_target.statsfile`
```

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
andyscott: Sounds good to me

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
andyscott: I think so. Do we want to catch and continue the build or do we want to rethrow it with a better message?

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
johnynek: sure. Happy to do the above

johnynek:
I'm +1. I'll let @ittaiz take a peek before we merge.

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
```diff
@@ -247,6 +247,16 @@ private static void compileScalaSources(CompileOptions ops, String[] scalaSource
       System.err.println("Compiler runtime: " + (stop - start) + "ms.");
     }
 
+    String[] lines = {
+        "build_time=" + Long.toString(stop - start)
+    };
+    Path statsfilePath = FileSystems.getDefault().getPath(ops.statsfile);
```
ittaiz: Nit- wdyt about Paths.get? More canonical

src/java/io/bazel/rulesscala/scalac/ScalacProcessor.java
```diff
@@ -247,6 +247,16 @@ private static void compileScalaSources(CompileOptions ops, String[] scalaSource
...
+    try {
+        Files.write(statsfilePath, Arrays.asList(lines));
```
ittaiz: Why do we need an array of we convert it to a list? Maybe just create a list two lines above?

andyscott:
@ittaiz I've addressed the two small items and added a test.

re: nitpick-- this is totally welcomed. I haven't written plain Java in a long time and I don't remember/know the best patterns or JDK APIs to get the job done.

test_rules_scala.sh
```diff
@@ -734,6 +734,8 @@ $runner bazel run test:MixJavaScalaLibBinary
 $runner bazel run test:MixJavaScalaSrcjarLibBinary
 $runner bazel run test:ScalaBinary
 $runner bazel run test:ScalaLibBinary
+$runner bazel build test:ScalaBinary_statsfile
```
ittaiz: Are you sure we need these? I might be mistaken but I think that `bazel build //test/...` will also catch it and we're doing it in the beginning

ittaiz:
Thanks! liked the test with the genrule. I definitely need to improve my genrule-fu :)
Other than my question about whether or not we need to touch the test_run.sh I'm good.

test_rules_scala.sh
andyscott: Oh, good point!

scala/scala.bzl
```diff
@@ -96,15 +96,17 @@ def _build_nosrc_jar(ctx, buildijar):
     cmd = """
 rm -f {jar_output}
 {zipper} c {jar_output} @{path}
+touch {statsfile}
```
ittaiz: I'm sorry for being too iterative and I appreciate your patience but I just re-read this and wanted to ask- you're doing this (`touch`) since sometimes (only java sources / no sources at all) scala compilation doesn't happen and bazel complained otherwise? I think there's a better way to solve it by only adding it to the `outs`if we need it but since I don't want to complicate you (even more ;) ) I suggest a comment on why we're `touch`ing here?
I just think that it's not that idiomatic and so might be unclear down the line

andyscott: No problem. Scala targets that have no source files get short circuited and don't run the scalac invoker. Creating the statsfile regardless of the number of source files makes it easier to write rules/macros for generating downstream metrics related targets, IMHO.

ittaiz: Ok, sounds good. 
If you think `touch` is enough for these rules/macros, I'm good.
WDYT about adding a comment to explain why we `touch`?

katre:
I don't have any details about the build event protocol, so I can't help you any there, unfortunately.

scala/scala.bzl
andyscott: I added a brief comment 🙂. Let me know if you want a more thorough one.

ittaiz:
Looks good, thanks!
Test this please (for ci bot)
@johnynek
Travis Mac is stuck. I think that if Bazel Build passes we should merge. Wdyt?

johnynek:
fracking travis....

johnynek:
Thanks for the PR @andyscott !

ittaiz:
Indeed, thanks!

PR #407, opened 2018-01-24
[Do not merge] Scala2 12

ittaiz:
@gkk-stripe @johnynek 
I've opened this PR to see what is the actual delta for me to be able to use `master` with 2.12.
It seems to me that if someone re-implement the `*_repositories` methods that they need (in my case scala, proto and specs2 repositories methods) then the only logic that exists only here is:
```
# this probably should be set globally for the entire rules_scala
    # but I don't know how to do it
    javacopts = [
    "-source 1.8",
    "-target 1.8"
    ],
```
and 
```
        # always append -YdisableFlatCpCaching, workaround for
        # https://github.com/bazelbuild/rules_scala/issues/305
        # ~remove once we upgrade to Scala 2.12.4~
        # ^^ turns out that 2.12.4 didn't fix the issue, see:
        # https://github.com/bazelbuild/rules_scala/pull/310#issuecomment-337466097
    scalacopts = toolchain.scalacopts + ctx.attr.scalacopts + ["-YdisableFlatCpCaching"]
```
Right?
Am I missing something?
Reason I'm asking is maybe the better solution (from this branch) is to merge these bits to `master` and even introduce on `master` *_repositories_212` methods so people can use.
Pro- we don't have to keep updating this branch which already contains a bug for some reason (see scala_import)
Con- This doesn't promote us towards toolchains which we agree is the correct solution.

To me this really depends on whether or not we can use toolchains in repository rules or not (which hopefully @katre will be able to answer). If not then IMHO we should implement my proposal but if it's an option we should invest in `scala_import_external` and run with it.

WDYT?


gkossakowski:
I'm in favor of merging the two branches and having _2.12 variant of repository rules for setting up scala dependencies. I considered doing this while getting started with 2.12 support but other priorities took over.

Have we considered making scala version a some kind of a global setting that repository rules, regular rules and everything else would have an access to?

scala/scala_import.bzl
```diff
@@ -105,7 +105,7 @@ def _collect_exports(exports):
 
   for dep_target in exports:
       java_provider = dep_target[java_common.provider]
-      exported_jars += java_provider.full_compile_jars
+      exported_jars += java_provider.compile_jars
```
johnynek: I think this might be a revert here. Do you intend this?

src/java/io/bazel/rulesscala/scalac/BUILD
```diff
@@ -7,6 +7,12 @@ java_binary(
         "Resource.java",
     ],
     main_class = "io.bazel.rulesscala.scalac.ScalaCInvoker",
+    # this probably should be set globally for the entire rules_scala
```
johnynek: this can probably be added to the toolchain to pass javacopts.

johnynek:
So as I see it we have two techniques for users to configure these rules:

1) bind
2) toolchains.

It looks like toolchains can't handle everything we want (e.g. selecting an tool label that we would use ctx.executable.foo on in a current rule). We will likely soon (if not already) have differences in the 2.11 and 2.12 tools since they may not be source compatible.

Anyway, I'm happy for us to have the version technique or maybe take a scala version as an arg: `scala_repositories("2.11")` or something... maybe does not matter much.

It is a shame that all the plugins have the same issue and we don't have a clean way to deal with it except verbosity (e.g. all the hashes we need for scrooge, tut, scalapb, specs, scalatest...)

test/src/main/scala/scala/test/scala_import/BUILD
```diff
@@ -49,25 +49,12 @@ scala_specs2_junit_test(
     suffixes = ["Test"],
 )
 
-java_import(
```
ittaiz: @hmemcpy this looks incorrect

PR #409, opened 2018-01-30
Adds the _java_runtime attribute to fix toolchain use in Scala.

sethkoehler:
When executing non-locally, the wrong version of Java would be used,
referencing a path that didn't exist. This change adds the
_java_runtime attribute to have to locate the java runtime on the
machine where the Scala code is being executed, and selects the
correct configuration for the host_javabase.

As a minor cleanup, this also removes the now no-longer-necessary
_java attribute.

TESTED=Ran unit tests and also ran a custom Scala unit test through:
previously, the Scala unit test failed with an incorrect java path
error, now it passes.

Change-Id: I9c2829bded04ec2d9d014094bd9e72cd984fe6e6

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -247,9 +247,6 @@ StatsfileOutput: {statsfile_output}
     outs = [ctx.outputs.jar, ctx.outputs.statsfile]
     if buildijar:
         outs.extend([ctx.outputs.ijar])
-    # _java_toolchain added manually since _java doesn't currently setup runfiles
-    # _scalac, as a java_binary, should already have it in its runfiles; however,
-    # adding does ensure _java not orphaned if _scalac ever was not a java_binary
     ins = (list(compiler_classpath_jars) +
            list(dep_srcjars) +
            list(srcjars) +
@@ -259,10 +256,9 @@ StatsfileOutput: {statsfile_output}
            dependency_analyzer_plugin_jars +
            ctx.files.resources +
            ctx.files.resource_jars +
-           ctx.files._java_toolchain +
+           ctx.files._java_runtime +
            [ctx.outputs.manifest,
             ctx.executable._ijar,
-            ctx.executable._java,
             argfile])
     ctx.action(
         inputs=ins,
@@ -409,13 +405,34 @@ def write_manifest(ctx):
         output=ctx.outputs.manifest,
         content=manifest)
 
+def _path_is_absolute(path):
+    # Returns true for absolute path in Linux/Mac (i.e., '/') or Windows (i.e.,
+    # 'X:\' or 'X:/' where 'X' is a letter), false otherwise.
+    if len(path) >= 1 and path[0] == "/":
+        return True
+    if len(path) >= 3 \
+            and path[0].isalpha() \
+            and path[1] == ":" \
+            and (path[2] == "/" or path[2] == "\\"):
+        return True
+
+    return False
+
 def _write_launcher(ctx, rjars, main_class, jvm_flags, args="", wrapper_preamble=""):
     runfiles_root = "${TEST_SRCDIR}/%s" % ctx.workspace_name
     # RUNPATH is defined here:
     # https://github.com/bazelbuild/bazel/blob/0.4.5/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt#L227
     classpath = ":".join(["${RUNPATH}%s" % (j.short_path) for j in rjars])
     jvm_flags = " ".join([ctx.expand_location(f, ctx.attr.data) for f in jvm_flags])
-    javabin = "%s/%s" % (runfiles_root, ctx.executable._java.short_path)
+    # TODO: Replace the following if/else with just .java_executable_runfiles_path
+    # when that becomes generally available in Bazel (submitted in
+    # https://github.com/bazelbuild/bazel/commit/f2075d27ca124156fcd7c01242c552175c0cf145).
+    java_path = str(ctx.attr._java_runtime[java_common.JavaRuntimeInfo].java_executable_exec_path)
+    if _path_is_absolute(java_path):
+      javabin = java_path
+    else:
+      javabin = "%s/%s" % (runfiles_root, java_path)
+
     template = ctx.attr._java_stub_template.files.to_list()[0]
 
     exec_str = ""
@@ -728,10 +745,8 @@ def _scala_binary_common(ctx, cjars, rjars, transitive_compile_time_jars, jars2l
 
   java_wrapper = ctx.new_file(ctx.label.name + "_wrapper.sh")
 
-  # _java_toolchain added manually since _java doesn't currently setup runfiles
   runfiles = ctx.runfiles(
-      files = list(rjars) + [ctx.outputs.executable, java_wrapper] + ctx.files._java_toolchain,
-      transitive_files = _get_runfiles(ctx.attr._java),
+      files = list(rjars) + [ctx.outputs.executable, java_wrapper] + ctx.files._java_runtime,
       collect_data = True)
 
   rule_outputs = struct(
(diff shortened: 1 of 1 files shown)


bazel-io:

sethkoehler:
I think the bazel-io comment is automated, but I believe the admins have already okayed this change, but let me know if that's not the case!

ittaiz:
Test this please (ci bot)

sethkoehler:
Hi there!  Sorry, not sure *how* to test this?  Is there a doc that indicates what tests I need to run that I missed?  I tried "bazel test ..." but that gave errors from the beginning :(

sethkoehler:
Note, I should have mentioned, I did run "bazel test test/..." already, if that is what you're referring to (all tests pass)?

ittaiz:
Sorry, that was for the ci bot to run.

sethkoehler:
Ah no worries, I'd misunderstood, thanks!

scala/scala.bzl
```diff
@@ -409,13 +409,33 @@ def write_manifest(ctx):
...
+    return False
+
 def _write_launcher(ctx, rjars, main_class, jvm_flags, args="", wrapper_preamble=""):
     runfiles_root = "${TEST_SRCDIR}/%s" % ctx.workspace_name
     # RUNPATH is defined here:
     # https://github.com/bazelbuild/bazel/blob/0.4.5/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt#L227
     classpath = ":".join(["${RUNPATH}%s" % (j.short_path) for j in rjars])
     jvm_flags = " ".join([ctx.expand_location(f, ctx.attr.data) for f in jvm_flags])
-    javabin = "%s/%s" % (runfiles_root, ctx.executable._java.short_path)
+    # TODO: Replace the following if/else with just .java_executable_runfiles_path
```
ittaiz: do we know when this will be available? is there a relevant issue or PR we can link to it?

test/BUILD
```diff
@@ -214,6 +214,12 @@ scala_library(
     name = "MixJavaScalaLib",
     srcs = glob(["src/main/scala/scala/test/mix_java_scala/*.scala"]) +
            glob(["src/main/scala/scala/test/mix_java_scala/*.java"]),
+    javac_jvm_flags = [
```
ittaiz: Was this added for demonstration/regression or is this needed?
If this is needed I'd love to understand why.
If this is for the former reason then it's also unclear to me how this protects from regression

scala/scala.bzl
```diff
@@ -730,8 +750,8 @@ def _scala_binary_common(ctx, cjars, rjars, transitive_compile_time_jars, jars2l
 
   # _java_toolchain added manually since _java doesn't currently setup runfiles
```
ittaiz: is this comment incorrect anymore?

johnynek:
@sethkoehler did you use this with remote execution?

I think @ianoc has seen similar bugs, but we didn't track it all the way down.

cc @gkk-stripe @andyscott

scala/scala.bzl
sethkoehler: Would the bazel commit be acceptable?  The property is already in, but I felt it a bit premature to depend on it without it being in a bazel release...and even then, I suspect waiting for a while until folks are generally using those releases would be good:

https://github.com/bazelbuild/bazel/commit/f2075d27ca124156fcd7c01242c552175c0cf145

sethkoehler:
@johnynek 

Yep, this indeed gets tests to work with remote execution (I tested that path explicitly).

ianoc:
@sethkoehler what is the error you saw?

scala/scala.bzl
ittaiz: sounds good

sethkoehler:
@ianoc 

I got this error when trying to run a Scala unit test:

<PATH>/bazel-out/k8-fastbuild/bin/src/it/scala/root.runfiles/other/src/it/scala/root_wrapper.sh: line 4: <PATH>/bazel-out/k8-fastbuild/bin/src/it/scala/root.runfiles/other/../local_jdk/bin/java: No such file or directory

Basically, the wrong version of Java (the host java) was being used rather than the copy in the toolchain.

ianoc:
Yeah i'm afraid thats actually a bug in the remote execution stuff not in the rules, we shouldn't have absolute JVM paths in the rules i believe.

The bug has been i guess fixed on master, but you need to enable a few experimental options to have good times with the cache :
```
build --spawn_strategy=sandboxed --genrule_strategy=sandboxed --experimental_strict_action_env --experimental_multi_threaded_digest --strategy=Scalac=worker --jobs=30 --worker_max_instances 10 --experimental_remote_spawn_cache --remote_rest_cache=http://mycache
```
are the configs i'm using now. Post 0.9.0 i think this should fix it for you

ianoc:
https://bazel-review.googlesource.com/c/bazel/+/23350
is the CL for handling this. this effects other rules too than the scala ones, but from what i've seen this fix should make all of them work

sethkoehler:
@ianoc 

Ah, so you're saying this change is likely not needed at all once the next version of bazel is released?  If so, I can withdraw this change!

ianoc:
I think the change itself might be a bug to try get around the bazel bug and might cause issues later. Since our outputs would depend on absolute local jvm paths that sounds like it would cause caching issues in future?

Bazel's current release has the fix in it, but its behind an experimental flag, unfortunately bazel seems to keep things behind these flags for a long time. But if you are on >= 0.9.0 and try your repo with the `--experimental_remote_spawn_cache` i think it should clear up for you.

sethkoehler:
Thanks for the heads up about that!  Let me look into this and get back with you...using that flag seems to force me to run the test locally, which isn't what I was aiming for.

ianoc:
I've mostly been using remote caching rather than execution, if the remote execution starts breaking then i suspect thats a new bug. ulf @ google is very responsive about this and good for a fix!

sethkoehler:
Awesome, I'll discuss with bazel folks, thanks!

lberki:
I think this change is the right thing to do; the way the `_java` attribute was handled before in the Scala rules was incorrect, because they just used whatever JDK Bazel was running under (`@bazel_tools//tools/jdk:java` is an alias to `@local_jdk//:java`). This change fixes that and uses the Java runtime in the target configuration (that's what `@bazel_tools//tools/jdk:current_java_runtime` is), which is exactly how it should be.

lberki:
https://bazel-review.googlesource.com/c/bazel/+/23350 is orthogonal to this change; it deals with execution requirements for action, which is unrelated to the JVM paths set in the action. As far as I am concerned, this PR does the right thing: uses the path to the JVM in the target configuration and not the path to the JVM Bazel is running under.

ianoc:
They are orthogonal changes, but the error mentioned/what folks have ran into before has been caused by the runfiles not being unpacked correctly when being restored from a cache. So in previous debugging the path for the jvm itself was a red-herring. 

if the code here is wrong then by all means we should update it.

johnynek:
Well, if @lberki says this is the right way, and tests pass, it's good enough for me.

ittaiz:
I agree with Oscar.
@sethkoehler any chance you can reply on the two pending comments?

scala/scala.bzl
sethkoehler: Good catch, this comment is incorrect (as is one above it as well).  I'm actually checking into whether I need _java_toolchain here at all...will update when I have an answer (should be by tomorrow).

test/BUILD
sethkoehler: Thanks for catching this as well, this was not needed after all (I'd mistakenly thought it was).  I reverted the file.

scala/scala.bzl
sethkoehler: Just to make things clear, I've updated the two comments I intend to update once I've resolved the _toolchain question.

scala/scala.bzl
sethkoehler: I added this to the comment, thanks!

johnynek:
👍

johnynek:
test this please

sethkoehler:
Ok, sorry for the delay...after further discussion with lberki and trying out an example Scala test with remote execution, we don't believe the comment (nor the _java_toolchain entry) is necessary, so I've removed them from the two locations where I found them.

ittaiz:
Test this please (ci bot)

johnynek:
seems like the google ci isn't responding...

@ittaiz I'm happy to merge as is. What do you think?

ittaiz:
Ditto

PR #411, opened 2018-01-31
Allow files for scala_import

jjudd:
Hi everyone. Wanted to get people's thoughts on this before I added tests around it and prepared it to actually be merged.

I'm not sure if there is a reason the rules do not allow files for `scala_import`. We made this change and have been using files from `scala_import`. It also doesn't fail any tests locally.

If this change is something everyone is ok with, then I'll add a test for it and get it ready for merging.


bazel-io:

ittaiz:
Hi @jjudd,
Thanks for this! I think there was a need in previous iterations of this rule (and there were plenty of those) before we had exports but now that we have exports it indeed looks as if we don't need it.
Adding a test sounds great

ittaiz:
@jjudd gentle ping

jjudd:
Thanks for the ping. I should be able to add a test for this in the next couple days. I got sidetracked with the Scala IntelliJ plugin :)

ittaiz:
Sure thing, thank you for your contributions!

jjudd:
I haven't forgotten about this. It's next on my list.

ittaiz:
@jjudd friendly ping

jjudd:
Alright. I took a crack at adding a test for this. The jar I checked in to the repo comes from https://github.com/lucidsoftware/relate on Maven: http://central.maven.org/maven2/com/lucidchart/relate_2.11/2.1.1/relate_2.11-2.1.1.jar

I'm happy to use whatever jar people want. That was just the first one that came to my mine.

johnynek:
this is great. thanks!

test/src/main/scala/scala/test/scala_import/ScalaImportExposesFileJarsTest.scala
```diff
@@ -0,0 +1,23 @@
+package scala.test.scala_import
+
+import com.google.common.cache.Cache
+import org.apache.commons.lang3.ArrayUtils
+import org.specs2.mutable.SpecificationWithJUnit
+import com.lucidchart.relate.SqlRow
+
+class ScalaImportExposesJarsTest extends SpecificationWithJUnit {
```
ittaiz: @jjudd are you sure this runs/compiles? I don’t understand what target it belongs to?

test/src/main/scala/scala/test/scala_import/ScalaImportExposesFileJarsTest.scala
```diff
@@ -0,0 +1,23 @@
...
+
+  "scala_import" should {
+    "enable using the jars it exposes" in {
```
ittaiz: Do we need this here? I think it was copied from another test and wasn’t removed right?

test/src/main/scala/scala/test/scala_import/ScalaImportExposesFileJarsTest.scala
jjudd: Nope. My bad. I copied it, but then decided I wanted to add it to the original test instead. I accidentally added it to this new file. I'll make a PR real quick.

jjudd: I have a commit that fixes this and am running tests locally to make sure things are good. I'll open a PR as soon as it is done. Sorry about that.

PR #413, opened 2018-02-02
Update Scrooge to 18.1.0

travisbrown:
We're trying to update to Scrooge 18.1.0, which seems to have changed such that generated code now depends on util-core, which means that we're seeing failures like this:

```
error: object io is not a member of package com.twitter
import com.twitter.io.Buf
                   ^
```

This change adds the necessary dependencies to `scala_library` for `scrooge_scala_library` and also bumps the Scrooge version from 4.6.0 (which dates back to March 2016). I've confirmed that the tests here fail to compile with the version updates but not the `deps` and `exports` lines and succeeds with them.

### DIFF
--- a/twitter_scrooge/twitter_scrooge.bzl
+++ b/twitter_scrooge/twitter_scrooge.bzl
@@ -18,7 +18,7 @@ _jar_filetype = FileType([".jar"])
 def twitter_scrooge():
   native.maven_server(
       name = "twitter_scrooge_maven_server",
-      url = "http://mirror.bazel.build/repo1.maven.org/maven2/",
+      url = "https://repo1.maven.org/maven2/",
   )
 
   native.maven_jar(
@@ -33,8 +33,8 @@ def twitter_scrooge():
 
   native.maven_jar(
       name = "scrooge_core",
-      artifact = scala_mvn_artifact("com.twitter:scrooge-core:4.6.0"),
-      sha1 = "84b86c2e082aba6e0c780b3c76281703b891a2c8",
+      artifact = scala_mvn_artifact("com.twitter:scrooge-core:18.5.0"),
+      sha1 = "ee2450c394448786a0c6cec4695251866c0a7391",
       server = "twitter_scrooge_maven_server",
   )
   native.bind(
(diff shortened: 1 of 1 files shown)


johnynek:
you can make these changes in your own repo using custom binds. You don't need to do these binds.

I would only want to make a change in the rules when the current source does not work correctly with the latest scrooge.

travisbrown-stripe:
@johnynek Okay, makes sense, I guess. Is this a definite policy for the project? It seems like a shame to have off-the-shelf usage provide 2+ year-old versions of dependencies like ScalaTest and Scrooge, so that everyone on more recent versions has to pay this bind tax.

bazel-io:

ittaiz:
@johnynek I think he does have a point. WDYT?

ittaiz:
@johnynek wdyt?

ittaiz:
@johnynek I'd rather upgrade it but from a project policy standpoint. We are not scrooge users...
I suggest either merging or closing.

pvcnt:
I confirm it is a real blocker for me to have an outdated version of Scrooge which prevents me from adopting Bazel on one of my projects.

On the other hand, as Scrooge is now updated every month, it may be difficult to follow the pace here, and may break backwards compatibility for some users if Scrooge does so from one release to another.

travisbrown:
@pvcnt You can use a custom bind in your workspace file to make rules_scala work with a more recent Scrooge version (or a newer ScalaTest version, etc.). I originally thought that this wasn't possible for Scrooge specifically because of the way the dependencies were set up, but I was wrong—you definitely can use 18.1.0, 18.3.0, etc. without any changes here.

I've left this issue open because I still think we should keep these versions more up-to-date for the sake of convenience, especially for newcomers, but it shouldn't be a blocker in any situation.

pvcnt:
Oh, thanks for the tip. :) I also thought it would not be possible.

By the way, wouldn't registering a toolchain be the right way to go? Because it seems a bit brittle to have to rebind dependencies to make this work, and using bind() is not recommended.

ittaiz:
We're indeed considering this.
Would you like to take a stab at this?

travisbrown:
Just rebased this and updated to 18.5.0 in case @oscar-stripe changes his mind now that we're about to check in thousands of lines of `"//3rdparty/jvm/com/twitter:util_core"`… :smile:

PR #415, opened 2018-02-06
Add struct members that IntelliJ plugin expects to exist

jjudd:
The IntelliJ plugin assumes that source_jar and source_jars are both
present on the jars struct. This sets those to empty values, so the
IntelliJ plugin does not fail.


ittaiz:
test this please (ci bot)

ittaiz:
test this please (ci bot)
ci.bazel.build timed out in bazel fetch

jjudd:
Thanks for looking at it

ittaiz:
Thank you for tackling it :)

Issue #343, opened 2017-11-17
Android support


gregghz:
I would love to be able to use Bazel to create android apps with Scala. I know this isn't a common use case for Scala but at Lucid all of our Android code is Scala.

I've created a sample repo with a [`scala_android_library` rule](https://github.com/gregghz/android-scala-bazel/blob/master/rules/scala-android.bzl) (lifted mostly from the [kotlin rules](https://github.com/pubref/rules_kotlin/blob/master/kotlin/rules.bzl#L192)). Would this be something that we could get included in rules_scala?

ittaiz:
Re general request- definitely. This sounds like a nice feature even if it's uncommon.
Concretely- 
Re java_import- I suggest you take a look at our *_repositories methods to see how we add dependencies (in a nutshell we use bind to provide indirection).
Additionally we really try to set a strong testing bar to minimize regression so having relevant tests (which you've seen fail correctly) when you submit the PR would go a long way.

gregghz:
From what I understand bind is used for static imports that you want to set up once in the WORKSPACE file (like scala_respositories). The uses of java_import in my scala_android_library rule happen in a BUILD file and are dynamic based on the inputs to scala_android_library so I can't use bind because it errors out with:

```
bind must be in the WORKSPACE file (used by //app:io_bazel_rules_scala/dependency/internal/app-lib_scala)
```

Anyway, this might be the wrong way to discuss this without a PR. I appreciate your feedback and I'll work on creating a proper PR with tests :)

ittaiz:
Sure thing.
I'll just clarify that I was referring to this dependency which seems static: name + "_sdk",
The other java_import is indeed dynamic but isn't needed here. We have strong interoperability with java so the name target can just depend on the name+ "_scala" target

PR #418, opened 2018-02-11
Adds scala_android_library

gregghz:
https://github.com/bazelbuild/rules_scala/issues/343


bazel-io:

gregghz:
Signed the CLA

gregghz:
The build is currently failing because android_sdk_repository needs either ANDROID_HOME set or an explicit path to the android sdk on the build server. I'm not exactly sure how this will work with Travis, I'd love some input on this.

gregghz:
Alright, I made some progress on the android sdk. Now some builds are failing with "android-update-sdk" not found while others run that command fine.

scala/scala_android.bzl
```diff
@@ -0,0 +1,66 @@
...
+            name = name + "_res",
+            custom_package = custom_package,
+            manifest = manifest,
+            resource_files = resource_files,
+            deps = deps,
+            **kwargs
+        )
+        res_deps.append(name + "_res")
+
+    native.android_library(
```
johnynek: what is this doing?

scala/scala_android.bzl
```diff
@@ -0,0 +1,66 @@
...
+
+    scala_library(
+        name = name + "_compile",
+        srcs = srcs,
+        plugins = plugins,
+        deps = deps + ["//external:io_bazel_rules_scala/dependency/android_sdk", name + "_aar"] + res_deps,
+        **kwargs
+    )
+
+    native.java_import(
```
johnynek: why do we need this? does android_library not support the java-sandwich/providers stuff? I thought the point of that was that we could use scala interchangeably with java? Can you add a comment?

scala/scala_android.bzl
gregghz: This was unnecessary. I'll remove it.

scala/scala_android.bzl
gregghz: android_library can't export a scala_library. This is working around that restriction.

What I actually want is to add <name>_compile as one of the exports for the call to android_library. However, if I do that directly I get this error:

    BUILD:462:1: in exports attribute of android_library rule //test:ScalaAndroidLibrary: scala_library rule '//test:ScalaAndroidLibrary_compile' is misplaced here (expected aar_import, android_library, cc_library, java_import, java_library or java_lite_proto_library).

I get the same error using deps (except when I use deps instead of exports I also get an error about how I should use exports instead).

The java_import is a proxy to get around this issue. I'm open to better ways to accomplish this, I admittedly don't have much experience with bazel :)

johnynek:
I'm still interested in merging this.

@gregghz do you think you will have time to get the CI green?

Thank you again for the contribution!

ittaiz:
closing since stale. @gregghz if you find the time for this we'd appreciate it.

Issue #367, opened 2017-12-20
proto version: taking from scalapb vs taking from protoc 


ittaiz:
@azymnis I see that the current rules take the proto version from the `scalapb` config.
Our setup (being migrated from maven to bazel) omits that and lets `protoc` decide the version.
Can you say why you think `scalapb` is better? 
I think `protoc` is also versioned so reproducibility should be similar AFAIU.

ittaiz:
cc @simuons

PR #422, opened 2018-02-14
Upgrade protoc-jar from 3.2.0 to 3.5.0.

iphydf:
FreeBSD binaries don't exist for 3.2.0, but they do for various other
versions including 3.5.0.

Note that 3.5.1 does not work:
https://github.com/os72/protoc-jar/issues/43

Also, I had to remove the explicit version setting that ScalaPBC does:
https://github.com/scalapb/ScalaPB/blob/master/scalapbc/src/main/scala/scalapb/ScalaPBC.scala#L8
This seems a bit pointless, given that we provide our own version of
protoc anyway, but also there is no 3.1.0 for FreeBSD.

### DIFF
--- a/scala_proto/scala_proto.bzl
+++ b/scala_proto/scala_proto.bzl
@@ -12,8 +12,8 @@ def scala_proto_repositories():
 
     native.maven_jar(
         name = "scala_proto_rules_protoc_jar",
-        artifact = "com.github.os72:protoc-jar:3.2.0",
-        sha1 = "7c06b12068193bd2080caf45580b0a00d2a31638",
+        artifact = "com.github.os72:protoc-jar:3.5.0",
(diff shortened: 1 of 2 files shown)


bazel-io:

ittaiz:
Thanks!
Omitting the version from scalapb will also fix #367 which is good
Why did you need to upgrade scalapb as well?

ittaiz:
Test this (message to ci bot)

ittaiz:
@iphydf can you sign the CLA?
I think (not sure) it's also stopping the ci bot from running the additional check

ittaiz:
@iphydf ping about the cla

SirCmpwn:
This pull request is obsolete, ref #571

PR #425, opened 2018-02-20
use transitive_proto_path_flags if exist

ittaiz:
following https://github.com/bazelbuild/bazel/issues/4544 I'm adding an optional propagation of transitive_proto_path_flags to protoc.
It's optional since it's not currently in a released version and in fact the exposition to skylark will only land in an hour or two but I've tested it manually.
For the above reason the test is under `test_expect_failure` and not part of CI since we can't run only that test against bazel head.
@azymnis @johnynek I'd like to merge this soon so I'd appreciate your review

### DIFF
--- a/scala_proto/scala_proto.bzl
+++ b/scala_proto/scala_proto.bzl
@@ -327,6 +327,7 @@ def _gen_proto_srcjar_impl(ctx):
         if hasattr(target, 'proto'):
             proto_deps.append(target)
             acc_imports += target.proto.transitive_sources
+            #inline this if after 0.12.0 is the oldest supported version
             if hasattr(target.proto, 'transitive_proto_path_flags'):
               transitive_proto_path_flags += target.proto.transitive_proto_path_flags
         else:
--- a/scala_proto/scala_proto.bzl
+++ b/scala_proto/scala_proto.bzl
@@ -320,12 +320,15 @@ def _colon_paths(data):
 
 def _gen_proto_srcjar_impl(ctx):
     acc_imports = depset()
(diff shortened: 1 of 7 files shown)


azymnis:
@ittaiz LGTM

scala_proto/scala_proto.bzl
```diff
@@ -320,12 +320,15 @@ def _colon_paths(data):
...
 def _gen_proto_srcjar_impl(ctx):
     acc_imports = depset()
+    transitive_proto_path_flags = depset()
 
     proto_deps, jvm_deps = [], []
     for target in ctx.attr.deps:
         if hasattr(target, 'proto'):
             proto_deps.append(target)
             acc_imports += target.proto.transitive_sources
+            if hasattr(target.proto, 'transitive_proto_path_flags'):
```
johnynek: we should have either an issue or a code comment on these version compatibility check things so we don't lose track.

Like after version `0.11` is the oldest supported version, remove this check or something. What do you think?

I'm worried about the code building up layers of these checks we don't recall why the exist.

ittaiz:
@johnynek added the comment, wdyt?

ittaiz:
I’ll wait at least for the ci.bazel.build one to pass
One caveat- @iirina hasn’t committed the change to bazel that exposes transitive_proto_path_flags to skylark. If for whatever reason she decides to use a different name I’ll need to do another PR but I highly doubt there will be a problem since I came up with name according to the convention

ittaiz:
Merging since I think ci is stuck. The additional commit was just comment addition

PR #431, opened 2018-03-01
Add default javac opts when compiling java srcs

anchlovi:
Scala rules compile java sources using `java_common.compile`. When building the required input for `java_common.compile` scala rules uses only `javacopts` and `javac_jvm_flags` attributes while ignoring any global/default `javac_opts` such as the ones from `.bazelrc` file, e.g., `build --javacopt=...` or `build --java_toolchain=//:some_tool_chain`

This PR uses `java_common.default_javac_opts` to add the default `javac_opts` to `java_common.compile` `javac_opts`


johnynek:
This looks good to me. Thanks for the PR.

PR #434, opened 2018-03-03
adjust transitive_proto_path_flags skylark name to transitive_proto_path

ittaiz:
@johnynek Bazel ended up using a different name for the skylark field.
It landed [here](https://github.com/bazelbuild/bazel/commit/98ed9af727ec6342206d181fcd465e3e83980c52)


ittaiz:
Thanks!

PR #435, opened 2018-03-04
Scala2_12 rebased on top of master

or-shachar:
might need to force push it. @johnynek - do I have permissions to do it?


johnynek:
If we force push we will break shas that other people are using. I really don’t want to force push a publicly advertised branch.

Can we not use merge and avoid force push?

or-shachar:
@johnynek so just regular merge?

ittaiz:
Or,
I suggest just to use a new branch for now each time.
2.12_date

PR #438, opened 2018-03-07
[WIP] Add detection for unused classpath entries as part of strict mode

jjudd:
Based on what I read in #235, I hacked together a quick `unused_deps` implementation that is currently bundled together with `strict_java_deps`. 

This is a work in progress. I would like to get people's feedback on this before continuing.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -171,17 +171,20 @@ def _compile(ctx, cjars, dep_srcjars, buildijar, transitive_compile_jars, labels
         compiler_classpath_jars = transitive_compile_jars
 
         direct_jars = ",".join([j.path for j in cjars])
+        direct_targets = ",".join([labels[j.path] for j in cjars])
         indirect_jars = ",".join([j.path for j in transitive_compile_jars])
         indirect_targets = ",".join([labels[j.path] for j in transitive_compile_jars])
         current_target = str(ctx.label)
 
         optional_scalac_args = """
 DirectJars: {direct_jars}
+DirectTargets: {direct_targets}
 IndirectJars: {indirect_jars}
 IndirectTargets: {indirect_targets}
 CurrentTarget: {current_target}
         """.format(
               direct_jars=direct_jars,
+              direct_targets=direct_targets,
               indirect_jars=indirect_jars,
               indirect_targets=indirect_targets,
               current_target = current_target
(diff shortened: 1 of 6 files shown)


jjudd:
This should be it's own flag. I bundled it with `strict_java_deps` for the sake of speed in building a prototype.

jjudd:
Wanted to provide an update here: we're still working on this and should hopefully have something within the next couple weeks. 

We found an issue with macros and `classpath-shrinker`, which I filed an issue about: https://github.com/scalacenter/classpath-shrinker/issues/4

I like their suggestion of adding a whitelist for deps that are actually used, but the plugin marks as unused. That way `unused_deps` doesn't become totally blocked on bugs in the classpath shrinking mechanism.

We're moving into using our internal version of this to remove unused deps from our build files in combination with ibazel. If that goes well, we will polish it up and update this PR.

ittaiz:
Sounds good!
Sorry I haven’t had the chance to go over the code.
How do you think of handling the whitelist?

johnynek:
really excited about this! Thank you!

jjudd:
I was planning on adding an attribute to the scala rules for the whitelist. The attribute lists the deps that should not be marked as unused.

I'm open to other ways of solving this. Ideally I'd be able to tag deps in the actual deps list, but I'm I don't think Bazel supports that.

ittaiz:
Does the bug mean you remove “macro code”?
Meaning if you were to automatically whitelist code that was tagged as
having macro would that solve the problem?
Because we have scala_macro_library (or similar name) and scala_import will
soon also have this distinction (since it will use ijars for non macro code)

jjudd:
I'm not totally certain if that macro classpath bug happens for all macros or just some. I was able to repro with those two jars, but there are others that I can't repro the problem with.

Automatically whitelisting scala_macro_library and scala_import would likely fix the issue. It may whitelist too many things, though. 

As I work on optimizing our internal build files over the next few weeks, I'll see what happens and keep people updated.

ittaiz:
Sounds good.
To be clear- I don’t think we need to whitelist all of scala_import but we
want to turn on ijars for scala_import by default and have the ability to
turn them off. For those where we turn them off we can take that as a cue
for macro existence and whitelist them

ittaiz:
Closing this since we have unused deps. Thanks.

PR #440, opened 2018-03-07
Fix mixup that occured when adding scala_import files test

jjudd:
When adding a test for scala_import files support, a test was added that
is never run. This fixes that mixup.

### DIFF
--- a/test/src/main/scala/scala/test/scala_import/BUILD
+++ b/test/src/main/scala/scala/test/scala_import/BUILD
@@ -9,16 +9,24 @@ scala_import(
 
 # Jars as files
 scala_import(
-  name = "relate",
-  jars = [
-   "relate_2.11-2.1.1.jar",
-  ],
+    name = "relate",
+    jars = [
+      "relate_2.11-2.1.1.jar",
+    ],
 )
 
 scala_specs2_junit_test(
     name = "scala_import_exposes_jars",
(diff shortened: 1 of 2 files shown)


jjudd:
@ittaiz here is the fix to that issue you found earlier

test/src/main/scala/scala/test/scala_import/BUILD
```diff
@@ -9,16 +9,24 @@ scala_import(
...
+    jars = [
+      "relate_2.11-2.1.1.jar",
+    ],
 )
 
 scala_specs2_junit_test(
     name = "scala_import_exposes_jars",
     srcs = ["ScalaImportExposesJarsTest.scala"],
-    deps = [":guava_and_commons_lang"],
+    deps = [":guava_and_commons_lang", "relate"],
```
ittaiz: Do we need relate here?

jjudd: I'm on a roll. Thanks for catching that.

jjudd: Fixed.

ittaiz:
Travis failure seems to be from network error. I’m on mobile so can’t rerun.
Merging as is

ittaiz:
Thanks!

PR #442, opened 2018-03-13
Use commons-io to copy bytes in ScalacProcessor.extractJar

gkk-stripe:
This greatly improves the performance of that method. It was spotted in
the wild to take up to 50s for relatively small input source jar.


gkk-stripe:
The lack of buffering was slowing down `scala_library` target with a source jar as input by 50%: unpacking the jar was taking as much time as actually compiling the scala code.

ittaiz:
Let's wait until it's green

PR #449, opened 2018-03-24
Refactor write_launcher

johnynek:
This is a minor refactor of write_launcher to prepare for a change to make a repl script for all targets.

One cleanup here is that we had repeated ourselves with the name of a launcher script. Here we explicitly plumb this through so we can define it in a single place and be more honest about the dependencies.

Secondly, we make it clear that we only need a single wrapper script for each target (really each directory I think), so we can share a wrapper between the repl launcher and, for instance, the binary executable.

ptal @ittaiz

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -425,12 +425,16 @@ def _path_is_absolute(path):
 
     return False
 
-def _write_launcher(ctx, rjars, main_class, jvm_flags, args="", wrapper_preamble=""):
-    runfiles_root = "${TEST_SRCDIR}/%s" % ctx.workspace_name
+def _runfiles_root(ctx):
+    return "${TEST_SRCDIR}/%s" % ctx.workspace_name
+
+def _write_java_wrapper(ctx, args="", wrapper_preamble=""):
+    """This creates a wrapper that sets up the correct path
+       to stand in for the java command."""
+
+    runfiles_root = _runfiles_root(ctx)
     # RUNPATH is defined here:
     # https://github.com/bazelbuild/bazel/blob/0.4.5/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt#L227
-    classpath = ":".join(["${RUNPATH}%s" % (j.short_path) for j in rjars])
-    jvm_flags = " ".join([ctx.expand_location(f, ctx.attr.data) for f in jvm_flags])
     # TODO: Replace the following if/else with just .java_executable_runfiles_path
     # when that becomes generally available in Bazel (submitted in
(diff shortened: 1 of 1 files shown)


johnynek:
I wish I understood runfiles better. :(

So, I tried to add a default repl rule, but it does not work yet and I don't know why.

I can't seem to get a runfiles directory created for the library target. Without a runfiles directory, the script we write fails. I'm confused because return runfiles in the rule:
https://github.com/bazelbuild/rules_scala/blob/master/scala/scala.bzl#L722

we aren't using DefaultInfo, which seems to be the current way of doing things, but it seems that returning a struct should also work. I don't see what the difference is here: 
https://github.com/bazelbuild/rules_scala/blob/master/scala/scala.bzl#L787

which *does* work. As a start, I tried to just copy as much of the scala_binary as possible, but like I said, nothing I did seemed to make it work...

I have to say, I am very sympathetic to people who get frustrated about bazel because issues like this are pretty hard. You can google for related confusion on runfiles, so clearly I'm not the only one who has issues with this.

If anyone has time to look at what is going on, I would love any help.

PS: I'm sure the first thing any bazel team member will say is we are using like 10 deprecated things. Yes, that's what happens when you release software that deprecates APIs every month.

cc @dslomov I have asked you previously about runfiles, and I *though* the answer was just return a runfiles attribute from the rule implementation. Am I making any obvious errors?

johnynek:
@ittaiz despite not yet being able to get the default repl support working I still think this is worth merging. At least it removes the duplication of the wrapper file name.

johnynek:
@ittaiz what about this one? Can we merge?

ittaiz:
I'm sorry :( didn't get around to it yet.
if you're confident and you need it you can merge

johnynek:
take your time... (to a point. ;)

scala/scala.bzl
```diff
@@ -425,12 +425,16 @@ def _path_is_absolute(path):
...
-    runfiles_root = "${TEST_SRCDIR}/%s" % ctx.workspace_name
+def _runfiles_root(ctx):
+    return "${TEST_SRCDIR}/%s" % ctx.workspace_name
+
+def _write_java_wrapper(ctx, args="", wrapper_preamble=""):
+    """This creates a wrapper that sets up the correct path
+       to stand in for the java command."""
+
+    runfiles_root = _runfiles_root(ctx)
     # RUNPATH is defined here:
```
ittaiz: shouldn't this comment move to `_write_executable`?

ittaiz:
ping @johnynek about the comment and you have a conflict

johnynek:
I'll try to get back to this, I was very busy at work this week and I got a new computer, so I have to setup the machine still to be able to push.

johnynek:
@ittaiz okay, I have merged with master and moved the comment. See what you think.

PR #451, opened 2018-03-24
Use 0.10 in travis

johnynek:
make a bazel_0_10 branch which will hold the current master before we start being more aggressive about removing incompatible bazel features.


ittaiz:
I have no objection but why 0.10.1? Not sure why you picked that one specifically

ittaiz:
I still don't understand why not move to 0.11.1 but you can merge if you want

Issue #430, opened 2018-02-28
rules_scala fails under --all_incompatible_changes: `+` operator on a depset is forbidden


iphydf:
```
$ bazel build //jvm-toxcore-c/... --all_incompatible_changes
Loading:
Loading: 0 packages loaded
ERROR: /private/var/tmp/_bazel_iphy/178403b66d944445b84ccba31dd03db8/external/io_bazel_rules_scala/third_party/plugin/src/main/BUILD:5:1: in scala_library_for_plugin_bootstrapping rule @io_bazel_rules_scala//third_party/plugin/src/main:dependency_analyzer:
Traceback (most recent call last):
	File "/private/var/tmp/_bazel_iphy/178403b66d944445b84ccba31dd03db8/external/io_bazel_rules_scala/third_party/plugin/src/main/BUILD", line 5
		scala_library_for_plugin_bootstrapping(name = 'dependency_analyzer')
	File "/private/var/tmp/_bazel_iphy/178403b66d944445b84ccba31dd03db8/external/io_bazel_rules_scala/scala/scala.bzl", line 733, in _scala_library_impl
		_lib(ctx, True)
	File "/private/var/tmp/_bazel_iphy/178403b66d944445b84ccba31dd03db8/external/io_bazel_rules_scala/scala/scala.bzl", line 660, in _lib
		_collect_jars_from_common_ctx(ctx)
	File "/private/var/tmp/_bazel_iphy/178403b66d944445b84ccba31dd03db8/external/io_bazel_rules_scala/scala/scala.bzl", line 621, in _collect_jars_from_common_ctx
		collect_jars(((ctx.attr.deps + auto_deps) + e...), ...)
	File "/private/var/tmp/_bazel_iphy/178403b66d944445b84ccba31dd03db8/external/io_bazel_rules_scala/scala/scala.bzl", line 597, in collect_jars
		_collect_jars_when_dependency_analyzer_is_off(dep_targets)
	File "/private/var/tmp/_bazel_iphy/178403b66d944445b84ccba31dd03db8/external/io_bazel_rules_scala/scala/scala.bzl", line 549, in _collect_jars_when_dependency_analyzer_is_off
		compile_jars += java_provider.compile_jars
`+` operator on a depset is forbidden. See https://docs.bazel.build/versions/master/skylark/depsets.html for recommendations. Use --incompatible_depset_union=false to temporarily disable this check.
[...]

PR #452, opened 2018-03-25
WIP: test with all_incompatible_changes

johnynek:
this is an attempt to address #430 

but of course, I broke things.

It seems related to scala_binary somehow. I can build items in test, but anything that itself uses a scala binary seems broken. So tut, protobuf and thrift are broken.

I tried to just translate to the non-deprecated form, and it runs, but it just doesn't work, which is pretty hard to debug.

I'd welcome any help here and certainly a review would be great.

cc @iphydf @ittaiz

### DIFF
--- a/.bazelrc.travis
+++ b/.bazelrc.travis
@@ -9,6 +9,7 @@ build --local_resources 2048,.5,1.0
 
 # This is so we understand failures better
 build --verbose_failures
+build --all_incompatible_changes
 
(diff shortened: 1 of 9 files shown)


ittaiz:
@johnynek thanks for breaking it down! it's really helpful
can we close this one?
I've merged two and added a few questions on the remaining two

johnynek:
We can close.

PR #454, opened 2018-03-25
fix bazel incompatibilities in proto

johnynek:
just the proto part of #452 to address #430

### DIFF
--- a/scala_proto/scala_proto.bzl
+++ b/scala_proto/scala_proto.bzl
@@ -319,20 +319,21 @@ def _colon_paths(data):
     return ':'.join(["{root},{path}".format(root=_root_path(f), path=f.path) for f in data])
 
 def _gen_proto_srcjar_impl(ctx):
-    acc_imports = depset()
+    acc_imports = []
     transitive_proto_paths = depset()
 
     proto_deps, jvm_deps = [], []
     for target in ctx.attr.deps:
         if hasattr(target, 'proto'):
             proto_deps.append(target)
-            acc_imports += target.proto.transitive_sources
+            acc_imports.append(target.proto.transitive_sources)
             #inline this if after 0.12.0 is the oldest supported version
             if hasattr(target.proto, 'transitive_proto_path'):
               transitive_proto_paths += target.proto.transitive_proto_path
         else:
             jvm_deps.append(target)
 
+    acc_imports = depset(transitive = acc_imports)
     if "java_conversions" in ctx.attr.flags and len(jvm_deps) == 0:
         fail("must have at least one jvm dependency if with_java is True (java_conversions is turned on)")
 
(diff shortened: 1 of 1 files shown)


scala_proto/scala_proto.bzl
```diff
@@ -319,38 +319,39 @@ def _colon_paths(data):
...
             proto_deps.append(target)
-            acc_imports += target.proto.transitive_sources
+            acc_imports.append(target.proto.transitive_sources)
             #inline this if after 0.12.0 is the oldest supported version
             if hasattr(target.proto, 'transitive_proto_path'):
               transitive_proto_paths += target.proto.transitive_proto_path
         else:
             jvm_deps.append(target)
 
+    acc_imports = depset(transitive = acc_imports)
```
ittaiz: Can we move changing acc_imports to a depset after 44? 
In it you’re changing it back to a list which seems redundant

scala_proto/scala_proto.bzl
```diff
@@ -319,38 +319,39 @@ def _colon_paths(data):
...
-    acc_imports = depset()
+    acc_imports = []
     transitive_proto_paths = depset()
 
     proto_deps, jvm_deps = [], []
     for target in ctx.attr.deps:
         if hasattr(target, 'proto'):
             proto_deps.append(target)
-            acc_imports += target.proto.transitive_sources
+            acc_imports.append(target.proto.transitive_sources)
```
ittaiz: Isn’t transitive_sources a depset? If so isn’t it better to accumulate them in depsets?

scala_proto/scala_proto.bzl
johnynek: so, note that `acc_imports` as the arg to transitive is a `List[Depset]` not a `List[File]` or something, so when we do `depset(transitive = acc_imports).to_list()` converts a `List[Depset]` to a `List[File]`.

scala_proto/scala_proto.bzl
johnynek: there is no way to accumulate depsets that is not deprecated. You accumulate a list and then make a `depset(transitive = your_list_here)`

ittaiz: @laurentlb is this correct? Why can’t we accumulate depsets? Turning depsets to lists so we accumulate them as depsets sounds counter intuitive

scala_proto/scala_proto.bzl
ittaiz: 😱 that's so implicit JavaConversions feels explicit 😉

scala_proto/scala_proto.bzl
laurentlb: Yes, it's correct (https://docs.bazel.build/versions/master/skylark/depsets.html).
* depset is immutable, so the `+=` operator was misleading
* we can create only one `depset`, with the correct structure, and it's much better for performance (adding depset inside a loop could lead to `O(n^2)` complexity
* it's more explicit: you see that the depset has n transitive depsets at the same level (which corresponds to the internal representation)

Depsets are not turned into lists. We just collect all the depsets first, and then merge them all at the same time.

ittaiz: Can we pass a depset to the “transitive” argument of the constructor?
What Oscar did is convert them to lists for this ctor which sounds wrong

johnynek: I didn’t convert them to lists. I added them to a list. The argument to transitive is a list of Depsets. I constructed that list so I could pass the list of depsets to transitive.

ittaiz: I see, my bad.
Thanks for the clarification

PR #456, opened 2018-03-25
Fix thrift bazel incompatibilities

johnynek:
This is the thrift part of #452 to address #430

### DIFF
--- a/twitter_scrooge/twitter_scrooge.bzl
+++ b/twitter_scrooge/twitter_scrooge.bzl
@@ -1,11 +1,11 @@
-_jar_filetype = FileType([".jar"])
-
 load("//scala:scala.bzl",
   "scala_mvn_artifact",
   "scala_library",
   "write_manifest",
   "collect_srcjars",
   "collect_jars")
+
+_jar_filetype = FileType([".jar"])
 
 def twitter_scrooge():
   native.maven_server(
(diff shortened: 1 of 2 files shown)


twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -130,16 +139,16 @@ def _gen_scrooge_srcjar_impl(ctx):
   # These are the thrift sources in the dependency graph. They are necessary
   # to generate the code, but are not "owned" by this target and will not
   # be in the resultant source jar
-  transitive_thrift_srcs = transitive_owned_srcs + _collect_transitive_srcs(ctx.attr.deps)
+  transitive_thrift_srcs = depset(transitive = [transitive_owned_srcs, _collect_transitive_srcs(ctx.attr.deps)]).to_list()
```
ittaiz: does this (depset and then to_list) make sense?

johnynek: We need to walk the depset eventually, but we do need it as a list. So, I don't know another way.

Note this is semantically a union of two depsets, but union is deprecated now...

I really don't know why they deprecated union when you can implement with `depset(transitive = [a, b])`... They should be optimizing depset to handle that case...

I'm concerned the bazel folks have not looked at the FP literature on fast immutable types, but maybe they have. I fear they may just be doing immutability with copying.

PR #458, opened 2018-03-27
Unconditionally use java_executable_runfiles path.

lberki:
This is now in Bazel in the latest release, so it works this way.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -196,7 +196,7 @@ CurrentTarget: {current_target}
     compiler_classpath = separator.join([j.path for j in compiler_classpath_jars])
 
     toolchain = ctx.toolchains['@io_bazel_rules_scala//scala:toolchain_type']
-    scalacopts = toolchain.scalacopts + ctx.attr.scalacopts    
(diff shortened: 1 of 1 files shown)


lberki:
ping @iirina @travisbrown

ittaiz:
@lberki thanks! 
@johnynek with respect to your plans of having master be up to date and keep version branches. Do you want to to bump the minimum version to 0.11.1? (I think this is what's effectively happening here)

johnynek:
I’m fine with that, and we can merge the other PR into the bazel_0_10 branch for people on earlier versions.

We should fix the .travis.yml file so travis can go green @lberki

ittaiz:
Sounds good

johnynek:
@lberki it does not look like you fixed the travis build so now our master is broken.

We use travis because we have a number of negative tests that we have not yet put into bazel, i.e. tests where we expect targets to fail.

lberki:
@johnynek I misunderstood you -- I thought Travis was already red and you are recommending an independent change to fix that :( Let me just try my hand at it...

PR #461, opened 2018-03-29
Support custom runners (@RunWith) on filtered Scala junit tests

### LINKED ISSUE #376: Custom test runners (via @RunWith) are not supported together with a --test_filter
Given a scala (junit) test:

```scala
@RunWith(classOf[MyCustomRunner])
class MyTest {
  @Test
  def myTest() = ...
}
```
```scala
public class MyCustomRunner extends BlockJUnit4ClassRunner {

    public static String mutableState;

    @Override
    protected List<TestRule> getTestRules(Object target) {
        mutableState = "Hello from test rules!"; // <--- this doesn't get executed from scala_junit_test
        return super.getTestRules(target);
    }
}
```

The code inside `MyCustomRunner` will not be executed, if `--test_filter` is passed.

Note that the same scenario works fine in a `java_test`.

hmemcpy:
Fixes #376

This is a bit hacky, since it requires replacing a private object instance of the current test runner using private reflection.
Unfortunately, there doesn't seem to be a way to set this object from the outside using normal API.

This PR moves the filtering logic to the actual `TestClass` instance that the JUnit runners query internally, allowing us to perform external filtering on custom runner instances. 

Background info:
The actual logic for querying test methods lives in `org.junit.runners.model.TestClass`, which is, unfortunately, a private field, created by the runner.
To be able to support our custom filtering on other runners, this `TestClass` instance has to be modified.

We're currently handling scala junit tests by "intercepting" the runner as a custom suite, and creating a new runner with additional filtering logic. This doesn't work for custom runners, because their `TestClass` instance is already initialized. Using reflection, this PR creates a new instance of `TestClass` with the filtering logic, and replaces it on the runner.

### DIFF
--- a/src/java/io/bazel/rulesscala/test_discovery/FilteredRunnerBuilder.scala
+++ b/src/java/io/bazel/rulesscala/test_discovery/FilteredRunnerBuilder.scala
@@ -44,10 +44,10 @@ private[rulesscala] class FilteredTestClass(testClass: Class[_], pattern: Patter
 object JUnitFilteringRunnerBuilder {
   val f: FilteringRunnerBuilder = {
     case (runner: BlockJUnit4ClassRunner, testClass: Class[_], pattern: Pattern) =>
-      hackRunner(runner, testClass, pattern)
+      replaceRunnerTestClass(runner, testClass, pattern)
   }
 
-  private def hackRunner(runner: BlockJUnit4ClassRunner, testClass: Class[_], pattern: Pattern) = {
+  private def replaceRunnerTestClass(runner: BlockJUnit4ClassRunner, testClass: Class[_], pattern: Pattern) = {
     allFieldsOf(runner.getClass)
       .find(f => f.getName == "testClass" || f.getName == "fTestClass")
       .foreach(field => {
--- a/src/java/io/bazel/rulesscala/test_discovery/FilteredRunnerBuilder.scala
+++ b/src/java/io/bazel/rulesscala/test_discovery/FilteredRunnerBuilder.scala
@@ -42,6 +42,9 @@ private[rulesscala] class FilteredTestClass(testClass: Class[_], pattern: Patter
 }
 
 object JUnitFilteringRunnerBuilder {
+  private final val TestClassFieldPreJUnit4_12 = "fTestClass"
+  private final val TestClassField = "testClass"
+
   val f: FilteringRunnerBuilder = {
     case (runner: BlockJUnit4ClassRunner, testClass: Class[_], pattern: Pattern) =>
       replaceRunnerTestClass(runner, testClass, pattern)
@@ -49,7 +52,7 @@ object JUnitFilteringRunnerBuilder {
 
   private def replaceRunnerTestClass(runner: BlockJUnit4ClassRunner, testClass: Class[_], pattern: Pattern) = {
     allFieldsOf(runner.getClass)
-      .find(f => f.getName == "testClass" || f.getName == "fTestClass")
+      .find(f => f.getName == TestClassField || f.getName == TestClassFieldPreJUnit4_12)
       .foreach(field => {
         field.setAccessible(true)
         field.set(runner, new FilteredTestClass(testClass, pattern))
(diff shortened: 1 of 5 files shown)


ittaiz:
Looks good (considering the constraints) @johnynek any objections?

src/java/io/bazel/rulesscala/test_discovery/FilteredRunnerBuilder.scala
```diff
@@ -21,29 +28,40 @@ class FilteredRunnerBuilder(builder: RunnerBuilder, filteringRunnerBuilder: Filt
...
-    super
-      .getChildren
-      .asScala
-      .filter(method => methodMatchesPattern(method, pattern))
-      .asJava
-
-  private def methodMatchesPattern(method: FrameworkMethod, pattern: Pattern): Boolean = {
-    val testCase = method.getDeclaringClass.getName + "#" + method.getName
-    pattern.matcher(testCase).find
+  private def hackRunner(runner: BlockJUnit4ClassRunner, testClass: Class[_], pattern: Pattern) = {
```
johnynek: Can we call this something else? Perhaps setRunnerFieldsAccessible?

test/BUILD
```diff
@@ -380,6 +380,14 @@ scala_junit_test(
     print_discovered_classes = True
 )
 
+scala_junit_test(
+    name = "JunitCustomRunner",
+    srcs = ["src/main/scala/scala/test/junit/JunitCustomRunnerTest.scala",
```
johnynek: Can we show an example where the runner is in a stand alone scala_library and then put that in deps? It should work but how to write the junit dependency may not be clear.

src/java/io/bazel/rulesscala/test_discovery/FilteredRunnerBuilder.scala
hmemcpy: Sure :)

test/BUILD
```diff
@@ -400,7 +408,13 @@ scala_library(
...
     srcs = ["src/main/scala/scala/test/utf8/JavaClassWithUtf8.java",
             "src/main/scala/scala/test/utf8/ScalaClassWithUtf8.scala"]
-    )
+)
+
+scala_library(
+    name = "customJunitRunner",
+    srcs = ["src/main/scala/scala/test/junit/JunitCustomRunner.java"],
+    deps = ["//external:io_bazel_rules_scala/dependency/junit/junit"]
+)
```
hmemcpy: @johnynek is this what you had in mind?

src/java/io/bazel/rulesscala/test_discovery/FilteredRunnerBuilder.scala
```diff
@@ -21,29 +28,40 @@ class FilteredRunnerBuilder(builder: RunnerBuilder, filteringRunnerBuilder: Filt
...
-      .asScala
-      .filter(method => methodMatchesPattern(method, pattern))
-      .asJava
-
-  private def methodMatchesPattern(method: FrameworkMethod, pattern: Pattern): Boolean = {
-    val testCase = method.getDeclaringClass.getName + "#" + method.getName
-    pattern.matcher(testCase).find
+  private def replaceRunnerTestClass(runner: BlockJUnit4ClassRunner, testClass: Class[_], pattern: Pattern) = {
+    allFieldsOf(runner.getClass)
+      .find(f => f.getName == "testClass" || f.getName == "fTestClass")
```
johnynek: Can we comment why runner has these fields and why we are testing two different names? Is that to deal with different versions?

ittaiz: @hmemcpy can you please add the comment @johnynek asked for?

hmemcpy: Oops, missed it. Yes, it has to do with JUnit version, the field was renamed in JUnit 4.12.

hmemcpy: Done, I extracted those values into named fields, instead of a comment. @johnynek please review :)

PR #462, opened 2018-03-31
Use some more non deprecated APIs

johnynek:
This is part of #452 in the scala.bzl part. Not all of the changes. I want to see them pass on CI, since the laptop I am using is a bit slow. They seem to pass, but to be extra sure we might want a clean build from travis.


johnynek:
cc @andyscott can you take a look?

andyscott:
LGTM

PR #463, opened 2018-03-31
Support JMH for scala 2.12 #295

dkomanov:
Use reflection-based generator (RFGeneratorSource) instead of ASM-based, it does support Java8 byte-code.

Also version of JMH is upgraded.

### DIFF
--- a/src/scala/io/bazel/rules_scala/jmh_support/BenchmarkGenerator.scala
+++ b/src/scala/io/bazel/rules_scala/jmh_support/BenchmarkGenerator.scala
@@ -4,15 +4,15 @@ import java.net.URLClassLoader
 
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
-
-import org.openjdk.jmh.generators.core.{ BenchmarkGenerator => JMHGenerator, FileSystemDestination }
+import org.openjdk.jmh.generators.core.{FileSystemDestination, GeneratorSource, BenchmarkGenerator => JMHGenerator}
 import org.openjdk.jmh.generators.asm.ASMGeneratorSource
-import org.openjdk.jmh.runner.{ Runner, RunnerException }
-import org.openjdk.jmh.runner.options.{ Options, OptionsBuilder }
-
+import org.openjdk.jmh.generators.reflection.RFGeneratorSource
+import org.openjdk.jmh.runner.{Runner, RunnerException}
+import org.openjdk.jmh.runner.options.{Options, OptionsBuilder}
 import java.net.URI
+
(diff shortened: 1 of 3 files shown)


src/scala/io/bazel/rules_scala/jmh_support/BenchmarkGenerator.scala
```diff
@@ -47,17 +55,18 @@ object BenchmarkGenerator {
...
-        "Usage: BenchmarkGenerator INPUT_JAR RESULT_JAR RESOURCE_JAR [CLASSPATH_ELEMENT] [CLASSPATH_ELEMENT...]"
+        "Usage: BenchmarkGenerator GENERATOR_TYPE INPUT_JAR RESULT_JAR RESOURCE_JAR [CLASSPATH_ELEMENT] [CLASSPATH_ELEMENT...]"
       )
       System.exit(1)
     }
     val fs = FileSystems.getDefault
 
     BenchmarkGeneratorArgs(
-      fs.getPath(argv(0)),
+      if ("asm".equalsIgnoreCase(argv(0))) AsmGenerator else ReflectionGenerator,
```
johnynek: can we error if it is not asm or reflection?

src/scala/io/bazel/rules_scala/jmh_support/BenchmarkGenerator.scala
```diff
@@ -131,17 +141,33 @@ object BenchmarkGenerator {
...
+        val source: GeneratorSource = generatorType match {
+          case AsmGenerator =>
+            val generatorSource = new ASMGeneratorSource
+            try {
+              generatorSource.processClasses(collectClassesFromJar(benchmarkJarPath).map(_.toFile).asJavaCollection)
+            } catch {
+              case _: ArrayIndexOutOfBoundsException =>
+                // this would fail due to https://github.com/bazelbuild/rules_scala/issues/295
+                // let's throw a useful message instead
+                sys.error("jmh in rules_scala doesn't work with Java 8 bytecode: https://github.com/bazelbuild/rules_scala/issues/295")
```
johnynek: can we change the error to say asm mode does not work with java 8 bytecode and instead suggest using reflection?

johnynek:
Thank you so much for tackling this! This will be a big help.

I think the 2.12 branch is a bit out of date sadly. For instance, I think it may have an out of date `.bazelci/presubmit.yml`

Can you make this as a PR against master and then we can cherry-pick it and re-enable the jmh test?

If you do this, we can probably merge very quickly.

dkomanov:
Closing in favor of #465

Issue #450, opened 2018-03-24
Update to non-legacy providers


johnynek:
The returns from most of our rules are what this page:

https://docs.bazel.build/versions/master/skylark/rules.html

calls "legacy" providers.

This may be related to the confusion about creating runfiles noted in this comment:
https://github.com/bazelbuild/rules_scala/pull/449#issuecomment-375925292

This is also related to the issue about incompatible changes: #430

PR #464, opened 2018-03-31
Add a ThriftInfo provider

johnynek:
This updates thrift to use a modern provider instead of an anonymous struct.

This is part of #450 

ptal @andyscott @ittaiz

### DIFF
--- a/thrift/thrift.bzl
+++ b/thrift/thrift.bzl
@@ -2,6 +2,8 @@
 
 _thrift_filetype = FileType([".thrift"])
 
+ThriftInfo = provider(fields=["srcs", "transitive_srcs", "external_jars", "transitive_external_jars"])
+
 def _common_prefix(strings):
   pref = None
   for s in strings:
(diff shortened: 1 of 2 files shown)


thrift/thrift.bzl
```diff
@@ -2,6 +2,8 @@
 
 _thrift_filetype = FileType([".thrift"])
 
+ThriftInfo = provider(fields=["srcs", "transitive_srcs", "external_jars", "transitive_external_jars"])
```
andyscott: nit: format this with each field on a new line
also consider adding a one line comment to each provider field to document what it's for

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -73,13 +75,12 @@ def _collect_owned_srcs(targets):
...
-        for jar in thrift.external_jars:
-          r.append(depset(_jar_filetype.filter(jar.files)))
-      r.append(depset(_jar_filetype.filter(thrift.transitive_external_jars)))
-  return depset(transitive = r)
+    if ThriftInfo in target:
+      thrift = target[ThriftInfo]
+      for jar in thrift.external_jars:
+        r.extend(_jar_filetype.filter(jar.files))
+      r.extend(_jar_filetype.filter(thrift.transitive_external_jars))
+  return depset(r)
```
andyscott: consider using the transitive parameter on the depset constructor. Something like this might work:
```
depset(transitive =
  [_jar_filetype.filter(jar.files) for jar in thrift.external_jars] +
  [_jar_filetype.filter(thrift.transitive_external_jars)]
)
```

andyscott: Also note that `FileType` is deprecated and we should move away from it (maybe in another PR).

johnynek: I don't think that works because transitive is a list of depsets while the arg to depset is a list of items (non-depset). I could make singleton depsets, but I assume that is slower than just putting it in the first arg since it is adding more nesting.

andyscott: 👍 makes sense. I'm not terribly familiar with the thrift code so it wasn't immediately apparent what's a depset and what's not.

johnynek: yeah, me neither. Especially due to the lack of types... :(

andyscott:
LGTM

PR #465, opened 2018-03-31
Support JMH for scala 2.12 #295

dkomanov:
Use reflection-based generator (RFGeneratorSource) instead of ASM-based, it does support Java8 byte-code.

Also version of JMH is upgraded.


johnynek:
This looks great. When CI is green, will merge.

Thank you very much! This was a blocker on 2.12 for us.

johnynek:
closes #295

dkomanov:
What's about scala2_12 branch? How to merge it there?

PR #467, opened 2018-04-01
Remove unreferenced private methods

andyscott:
These methods don't seem to be used anymore. I haven't tested it locally; just testing via CI.


johnynek:
LGTM

Shame skylark can’t lint this but it requires some static analysis to do so and I guess skylark does none.

PR #470, opened 2018-04-01
refactor to add a create_scala_provider method

johnynek:
This is a refactor to get the place where we create our ad-hoc scala provider, which is just a struct with certain data.

This will change to returning a ScalaInfo when we move to providers.

ptal @ittaiz


johnynek:
cc @andyscott

PR #473, opened 2018-04-04
[WIP] Scala maven import external

natansil:
@ittaiz currently java_import_external does not support depending on the jar file directly. in order to not do major changes to justine's code, I've had to change the usage label structure from `@com_google_guava_guava_21_0//jar:file` to `@com_google_guava_guava_21_0//:file`.
Is this legitimate or should I re-write the code to have the same folder structure as `maven_jar`?

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -42,11 +42,11 @@ maven_jar(
   sha1 = "e5b3e2753d0817b622c32aedcb888bcf39e275b4")
 
 # test of strict deps (scalac plugin UT + E2E)
-maven_jar(
-    name = "com_google_guava_guava_21_0",
-    artifact = "com.google.guava:guava:21.0",
-    sha1 = "3a3d111be1be1b745edfa7d91678a12d7ed38709"
-)
+#maven_jar(
+#    name = "com_google_guava_guava_21_0",
+#    artifact = "com.google.guava:guava:21.0",
+#    sha1 = "3a3d111be1be1b745edfa7d91678a12d7ed38709"
+#)
 
 maven_jar(
     name = "org_apache_commons_commons_lang_3_5",
@@ -83,3 +83,14 @@ filegroup(
 
 load("@io_bazel_rules_scala//scala:toolchains.bzl","scala_register_toolchains")
 scala_register_toolchains()
+
+load("//scala:scala_maven_import_external.bzl", "java_import_external")
+java_import_external(
+    name = "com_google_guava_guava_21_0",
+    licenses = ["notice"],  # Apache 2.0
+    jar_urls = [
+        "https://mirror.bazel.build/repo1.maven.org/maven2/com/google/guava/guava/21.0/guava-21.0.jar",
+        "https://repo1.maven.org/maven2/com/google/guava/guava/21.0/guava-21.0.jar",
+    ],
+    jar_sha256 = "972139718abc8a4893fa78cba8cf7b2c903f35c97aaf44fa3031b0669948b480",
+)
--- a/scala/scala_maven_import_external.bzl
+++ b/scala/scala_maven_import_external.bzl
@@ -1,0 +1,279 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Rules for defining external Java dependencies.
+
+java_import_external() replaces `maven_jar` and `http_jar`. It is the
+recommended solution for defining third party Java dependencies that are
+obtained from web servers.
+
+This solution offers high availability, low latency, and repository
+scalability at the cost of simplicity. Tooling can be used to generate
+the WORKSPACE definitions from Maven metadata.
+
+The default target in this BUILD file will always have the same name as
+the repository itself. This means that other Bazel rules can depend on
+it as `@repo//:repo` or `@repo` for short.
+
+### Setup
+
+Add the following to your `WORKSPACE` file:
+
+```python
+load("@bazel_tools//tools/build_defs/repo:java.bzl", "java_import_external")
+```
+
+### Best Practices
+
+#### Downloading
+
+The recommended best practices for downloading Maven jars are as follows:
+
+1. Always follow release versions or pinned revisions.
+2. Permanently mirror all dependencies to GCS or S3 as the first URL
+3. Put the original URL in the GCS or S3 object name
+4. Make the second URL the original repo1.maven.org URL
+5. Make the third URL the maven.ibiblio.org mirror, if it isn't 404
+6. Always specify the sha256 checksum
+
+Bazel has one of the most sophisticated systems for downloading files of any
+build system. Following these best practices will ensure that your codebase
+takes full advantage of the level of reliability that Bazel able to offer. See
+https://goo.gl/uQOE11 for more information.
+
+#### Selection
+
+Avoid using jars that bundle their dependencies. For example, a Maven jar for
+the artifact com.initech:tps:1.0 should not contain a classes named
+com.fakecorp.foo. Try to see if Initech distributes a tps jar that doesn't
+bundle its dependencies. Then create a separate java_import_external() for each
+one and have the first depend on the second.
+
+Sometimes jars are distributed with their dependencies shaded. What this means
+is that com.initech.tps will contain classes like
+com.initech.tps.shade.com.fakecorp.foo. This is less problematic, since it
+won't lead to mysterious classpath conflicts. But it can lead to inefficient
+use of space and make the license of the the end product more difficult to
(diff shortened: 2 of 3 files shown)


WORKSPACE
```diff
@@ -83,3 +83,14 @@ filegroup(
 
 load("@io_bazel_rules_scala//scala:toolchains.bzl","scala_register_toolchains")
 scala_register_toolchains()
+
+load("//scala:scala_maven_import_external.bzl", "java_import_external")
+java_import_external(
+    name = "com_google_guava_guava_21_0",
+    licenses = ["notice"],  # Apache 2.0
+    jar_urls = [
```
johnynek: are these alternative urls for the same sha?

scala/scala_maven_import_external.bzl
```diff
@@ -0,0 +1,295 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Rules for defining external Java dependencies.
+
+java_import_external() replaces `maven_jar` and `http_jar`. It is the
```
johnynek: is this just a temporary copy paste? can we not get this with a remote dependency?

ittaiz:
@natansil I don't have time now to do a complete review but WDYT about sharing here a bit more detail on what problem we're trying to solve and what are the gains we're trying to achieve? Also maybe a bit on intermediate solution vs longer term

test/src/main/scala/scala/test/scala_import/BUILD
```diff
@@ -4,7 +4,7 @@ load("//scala:scala_import.bzl", "scala_import")
 #Many jars
 scala_import(
   name = "guava_and_commons_lang",
-  jars = ["@com_google_guava_guava_21_0//jar:file", "@org_apache_commons_commons_lang_3_5//jar:file"],
+  jars = ["@com_google_guava_guava_21_0//:file", "@org_apache_commons_commons_lang_3_5//jar:file"],
```
natansil: @ittaiz re-posting my question right next to the example of the consumer:

currently java_import_external does not support depending on the jar file directly. in order to not do major changes to justine's code, I've had to change the usage label structure from `@com_google_guava_guava_21_0//jar:file` to `@com_google_guava_guava_21_0//:file`.
Is this legitimate or should I re-write the code to have the same folder structure as `maven_jar`?

Alternatively, you've stated that `scala_import` can depend on full jar files without needing the `file` filegroup target.

so with the example above, I can just change it to 
jars = ["@com_google_guava_guava_21_0//jar", "@org_apache_commons_commons_lang_3_5//jar"],

and that will solve the issue (and still not take ijars)?

WORKSPACE
jart: I can mirror files if you need it. Not sure I follow.

scala/scala_maven_import_external.bzl
jart: I can make the existing java_import_external serve your use case, if you'd prefer.

natansil:
@ittaiz , @johnynek please review.
The changes made:
- [X] make the repo_rule write scala_import instead of java_import
- [X] introduce artifact + server
- [ ]  discuss possibility to extract a common "template" repo rule out of java_import_external and scala_maven_import_external

jjudd:
We at Lucid are currently working on the external dependency issue and were just about to implement a `scala_import_external` by extending `java_import_external` so this is a very well timed PR :) Thanks for your work on this!

I have a couple questions:

1. The original `java_import_external` uses `jar_urls` instead of `server_urls` and `artifact`. Do we want to use `jar_urls` here? My thought is that we could use [bazel-deps](https://github.com/johnynek/bazel-deps) to resolve the maven dependencies, find the `jar_urls` and then produce a `scala_import_external`.

It seems like using `bazel-deps` would be quite nice: you can specify your dependencies in a file somewhere. Then import rules for those dependencies and all their transitive dependencies are generated.

This approach also works well with strict deps and unused deps. By using `scala_import` and `java_import` the generated import rules can declare their dependencies as `deps` instead of `runtime_deps` or `exports`.

2. How challenging would it be to create a generic `jvm_import_external`, which would be used to create `java_import_external` and `scala_import_external`? I'm hoping we could wrap `java_import_external`, as it is now, but pass in the rule you want it to produce: either `java_import` or `scala_import`.

3. If we create the `jvm_import_external`, do people want to use `bazel-deps` or is there a plan for Bazel to do something with Maven style dependencies without the need for a separate `bazel-deps` binary.

natansil:
@jjudd,

Regarding 1. + 2. 

As I see it this is part of the modularity effort we want to achieve.
At the base will be `jvm_import_external` that accepts the underlying import rule name (`java_import` or `scala_import` or `foo_import`) and also special args these rules needs.
It will also still have `jar_urls` attr, as a more basic feature.

In addition we are thinking about something like a `maven_import_external` that has `artifact` arg that can potentially wrap `jvm_import_external` and resolve the jar urls.
Hopefully these rules will be able to be part of bazel repo itself.

Regarding 3. I think that Bazel's long-term goal is to have transitive deps resolution built-in, like explained [here](https://bazel.build/designs/2016/09/19/recursive-ws-parsing.html)

scala/scala_maven_import_external.bzl
ittaiz: Yes this is a temporary copy paste.
We're starting out this way to validate the differences are indeed small instead of iterating between this repository and bazel itself.
The general plan is to generalize with as few changes as possible and then port 80%-90% of the code back to bazel

ittaiz:
I consent to my contributions going in

scala/scala_maven_import_external.bzl
```diff
@@ -307,3 +311,28 @@ scala_maven_import_external = repository_rule(
         "default_visibility": attr.string_list(default=["//visibility:public"]),
         "extra_build_file_content": attr.string(),
     })
+
+def scala_maven_import_external(artifact, server_urls, **kwargs):
+  jvm_maven_import_external(
+    rule_name = "scala_import",
+    rule_load = "load(\"@io_bazel_rules_scala//scala:scala_import.bzl\", \"scala_import\")",
```
ittaiz: this should probably be an attribute with a default value since some organizations (like us) add these in prelude and also some organizations (stripe?) might have aliases or stuff so they load from a different place

WORKSPACE
```diff
@@ -84,11 +84,19 @@ filegroup(
...
+load("//scala:scala_maven_import_external.bzl", "scala_maven_import_external", "java_import_external")
 scala_maven_import_external(
     name = "com_google_guava_guava_21_0",
     licenses = ["notice"],  # Apache 2.0
     artifact = "com.google.guava:guava:21.0",
     server_urls = ["https://mirror.bazel.build/repo1.maven.org/maven2"],
     jar_sha256 = "972139718abc8a4893fa78cba8cf7b2c903f35c97aaf44fa3031b0669948b480",
+)
+
+# for regression testing purposes
```
ittaiz: i don't understand this (even with the comment sorry!)

jjudd:
Thanks for generalizing things into `jvm_import_external`. Can we add a `scala_import_external` that looks like` java_import_external`, i.e., it takes `jarl_urls`, `jar_sha256`, `srcjar_urls`, and `srcjar_sha256`?

I ask because I imagine not everyone will want to rely on `scala_maven_import_external` to do the maven resolution, especially considering that the rule doesn't support transitive dependencies right now.

I'm currently modifying bazel-deps to use `java_import_external` and `scala_import_external (an internal version) instead of `java_library` and `scala_import`. bazel-deps resolves the jars, finds all the information it needs, and then writes the rule calls to a `.bzl` file.

I'm happy to add the rule to this PR. I just wanted to get people's feedback.

ittaiz:
Did you see what kind of maven resolution the rule currently does? Why
would you rather work with full URLs and not coordinates?

jjudd:
I have read through the code. I still see value in having a `scala_import_external` that is used in the same way as `java_import_external`. I would like to separate Maven resolution from artifact downloading until the Bazel community agrees on a Maven resolution solution.

Maven resolution is a pretty active area in Bazel without an agreed on solution. I know of bazel-deps, `transitive_maven_jar`, native `maven_jar`, skylark `maven_jar`, or `rules_maven`. This adds another option to that mix. I would like to be able to use `scala_import_external` without subscribing to a particular artifact resolution strategy. 

You can implement `scala_maven_import_external` using `scala_import_external` more easily than the other way around. The current commit does this for `jvm_maven_import_external` and `jvm_import_external`:

```
def jvm_maven_import_external(artifact, server_urls, **kwargs):
  jvm_import_external(
      jar_urls = _convert_to_url(artifact, server_urls),
      **kwargs
  )
```

I'm suggesting that we have the same thing for Scala.

ittaiz:
I know the change is small, I just don’t understand the value.
I don’t consider simple translation from coordinates to url as maven
resolution but I don’t object if you plan to capitalize on it.
@natansil,
Can you add it?

scala/scala_maven_import_external.bzl
natansil: done

WORKSPACE
natansil: changed a more detailed explanation. is this clear now?

jjudd:
Thanks so much for adding `scala_import_external` @natansil and @ittaiz!! I used it in our bazel-deps work today and so far things are working great! :smiley:

ittaiz:
@johnynek I'll merge in a couple of hours if you don't ping back

PR #478, opened 2018-04-10
Enable local cache by default

andyscott:
- Enables Bazel's local cache for all development by default
- Sets up Travis to cache the local cache directory across CI builds

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ bazel-*
 hash1
 hash2
 .DS_store
+.bazel_cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ osx_image: xcode8
 # Not technically required but suppresses 'Ruby' in Job status message.
 language: java
 
+cache:
+    directories:
+     - .bazel_cache
+
 os:
   - linux
   - osx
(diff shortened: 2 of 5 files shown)


ittaiz:
By Jenkins you mean travis?

andyscott:
Yeah, fixed my comment already!

johnynek:
I'm a little worried this invalidates our reproduciblity test.

We might as well remove that test if we ship this right?

andyscott:
Where is the reproducibility test? Depending on how it's being tested, it might be possible to disabled caching for those tests.

johnynek:
@andyscott https://github.com/bazelbuild/rules_scala/blob/master/test_rules_scala.sh#L32

andyscott:
@johnynek:
- I split out reproducibility test into a new job
- I also disabled the cache for second round reproducibility build

andyscott:
With caching enabled the primary job completes in about 5 minutes (as compared ~15 minutes). The new reproducibility job takes forever but the overall build time has gone down.

johnynek:
thanks!

ittaiz:
@johnynek thanks! I felt I was missing something...
@andyscott only problem is that people now running locally won’t run the reproducibility test. Can you add the new sh to the all.sh?

PR #479, opened 2018-04-11
Update test_all.sh to include test_reproducibility.sh

andyscott:
This is a quick follow up to #478


andyscott:
Yep, works for me locally.

PR #481, opened 2018-04-13
Upgrade to Bazel 0.12.0

pauldraper:


pauldraper:
Failing on `test_scala_library_expect_failure_on_missing_direct_java`. Looks like we're losing the label.

cushon:
ijars created by `java_common.compile` do not currently get stamped, and we can't fix that until the label is passed in to the API. This is partially what is being discussed in https://groups.google.com/forum/#!topic/bazel-discuss/mt2llfwzmac.

ittaiz:
Right...
Long and complicated thread.
Thanks for the clarification!

johnynek:
looks like the test for strict deps needs to be updated. It failed.

ittaiz:
We’re finalizing the behavior.
I’ll update here tomorrow or Thursday the latest.
How does that sound?

ittaiz:
ok, the situation here is a bit complicated.
Bazel 0.12.0 indeed broke strict-deps add_dep support for code that doesn't use ijar (for example if you use the current scala_import or the filegroup from maven_jar)
internally we've worked around it by having a script iterate the log and take various hints to fix the suggestions but this is heavily based on internal assumptions.
The solution for most issues (except for exports) will probably come from: https://docs.google.com/document/d/1ubah6phuvWnugShtVgSQnaopQ1BtKtNxQASVwGZA7k0/edit?usp=sharing

I think the pragamatic approach here is to loosen the assertion in the test so that it covers strict-deps error but not the add_deps message **until** the above is implemented and supported in rules_scala.
This will allow to move forward with rules_scala's CI versions.

WDYT?

cc @natansil @or-shachar for thoughts as well

natansil:
@ittaiz I'm fine with loosening the tests. it's just the `java` ones right?
I suppose an issue should be opened that states the `add_deps message` tests should be re-instated once the solution is implemented in bazel itself.

johnynek:
Note, we are using the rules with bazel 0.12 at Stripe. No issues that I know of yet.

PR #482, opened 2018-04-19
Split core rules into public API file and private implementation files

andyscott:
Part 1 of some substantial cleanup/refactoring of the core rules: splitting the main rules into a public facing API file and a series of implementation files.

I'd like to do some more cleanup to the public facing `scala.bzl` file to make it very readable for users of the rules. That will happen in a follow up PR.

No logic/indentation/formatting has been changed-- functions have only been moved around and renamed according to public/private scoping requirements.


andyscott:
I should add that this makes a bunch of previously private things public. The Kotlin and Go rules do something similar. I'd argue that anything in the `private` folder has no API stability guarantees and nobody should use things from there.

pauldraper:
Public/private means way more jumping around, but it is nice for end users, especially if it's complimented with docs.

Another data point: Skydoc does way better with public/private. (Though it's still a rough piece of work...don't use it; breaks on all sorts of stuff.)

scala/private/common.bzl
```diff
@@ -0,0 +1,174 @@
...
+      return java_common.create_provider(
+          compile_time_jars = scalaattr.compile_jars,
+          runtime_jars = scalaattr.transitive_runtime_jars,
+          transitive_compile_time_jars = transitive_compile_time_jars,
+          transitive_runtime_jars = scalaattr.transitive_runtime_jars,
+      )
+
+# TODO: this should really be a bazel provider, but we are using old-style rule outputs
+# we need to document better what the intellij dependencies on this code actually are
+def create_scala_provider(
```
johnynek: I don't think this should be private. We eventually want to have an API to create a ScalaInfo. Can we move to public?

scala/private/rule_impls.bzl
```diff
@@ -0,0 +1,756 @@
...
+    if java_common.provider in dep_target:
+        runtime_jars.append(dep_target[java_common.provider].transitive_runtime_jars)
+    else:
+        # support http_file pointed at a jar. http_jar uses ijar,
+        # which breaks scala macros
+        runtime_jars.append(filter_not_sources(dep_target.files))
+
+  return runtime_jars
+
+
```
johnynek: why all the space here?

scala/private/rule_impls.bzl
```diff
@@ -0,0 +1,756 @@
...
+            "%jvm_flags%": jvm_flags,
+            "%needs_runfiles%": "",
+            "%runfiles_manifest_only%": "",
+            "%set_jacoco_metadata%": "",
+            "%workspace_prefix%": ctx.workspace_name + "/",
+        },
+        is_executable = True,
+    )
+
+def dep_target_contains_ijar(dep_target):
```
johnynek: why is this public?

andyscott:
Re: Skydoc--

I tried Skydoc on `rules_scala_annex` and it was very difficult. `rules_kotlin` uses Skydoc and has lots of comments about the ugly workaround being made. With that in mind, I wouldn't recommend using Skydoc for rules_scala just yet.

jjudd:
I set up Skydoc for `rules_twirl` and `rules_play_routes`. It needs some work, and a few months ago it wasn't getting much support. Hopefully more people are working on it now.

scala/private/common.bzl
andyscott: Just `create_scala_provider`? or `create_scala_provider` and `create_java_provider`?

scala/private/rule_impls.bzl
andyscott: Whoops. Fixed.

scala/private/rule_impls.bzl
andyscott: I copied it as it was from the original code. It doesn't seem to be used anywhere, so I suppose I could remote it entirely.

johnynek: fine by me.

scala/private/common.bzl
johnynek: create_java_... is a compatability tool for our use, so I would say it is private, other callers should use the bazel API. So I would just make create_scala_provider the public thing.

andyscott:
- Unused method removed.
- Moved `create_scala_provider` to a new public `scala/providers.bzl` file for organizing providers and provider related utilities.

andyscott:
I'd ultimately like to break up `common.bzl` according to functionality. This PR was strictly a split along public API vs internal API lines.

pauldraper:
I had same experience as @andyscott.

Sky doc is bad and IDK if anyone is making it better.

I'm hoping the build environment interface proposal will eventually result in workable docs.

ittaiz:
Ok. Any chance for a small “layout design” which you have in mind? Similar
to what you and Paul discussed on Gitter?
Just so everyone (who cares) can agree the direction is good

andyscott:
I'd argue that `common.bzl` is currently a ball of mud, but it's a smaller ball of mud compared to `scala.bzl` as it currently stands.

The master plan is to split all rules into a public file and a private file and then have topic oriented utilities organized into private utility files as we see fit. I'd like to audit each method as this is done, ensuring that:
- it's actually used somewhere!
- it's not using deprecated patterns

ittaiz:
@andyscott thanks for this and sorry it took so long to merge! If you have a visual layout of what you have in mind I'd still very much appreciate it :)

PR #480, opened 2018-04-12
Fix remaining incompatible changes warnings

johnynek:
This fixes the remaining issues to resolve #430. With this, I can pass all tests using `--all_incompatible_changes`.

We test both strict deps and without strict deps to make sure we can pass with the all_incompatible_changes. Since the build should be exactly the same, the cache works when things pass, so it doesn't really increase CI time.

PR #484, opened 2018-04-24
`+` operator on a depset is forbidden

cushon:
Fixes:

```
ERROR: test/proto/BUILD:36:1: in scala_proto_srcjar rule //test/proto:test_proto_java_conversions_srcjar:
	File "test/proto/BUILD", line 36
		scala_proto_srcjar(name = 'test_proto_java_conversions_srcjar')
	File "scala_proto/scala_proto.bzl", line 332, in _gen_proto_srcjar_impl
		transitive_proto_paths += target.proto.transitive_proto_path
`+` operator on a depset is forbidden. See https://docs.bazel.build/versions/master/skylark/depsets.html for recommendations. Use --incompatible_depset_union=false to temporarily disable this check.
```

See also #480

### DIFF
--- a/scala_proto/scala_proto.bzl
+++ b/scala_proto/scala_proto.bzl
@@ -320,7 +320,7 @@ def _colon_paths(data):
 
 def _gen_proto_srcjar_impl(ctx):
     acc_imports = []
-    transitive_proto_paths = depset()
+    transitive_proto_paths = []
 
     proto_deps, jvm_deps = [], []
     for target in ctx.attr.deps:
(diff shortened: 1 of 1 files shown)


johnynek:
lgtm

confused why our test with `--all_incompatible_changes` didn't catch this.

johnynek:
we are still using 0.11 because the PR to bump to 0.12 failed CI: https://github.com/bazelbuild/rules_scala/pull/481 and no one has diagnosed/repaired it AFAIK.

cushon:
Makes sense. This is the only issue I saw with 0.12 related to `--all_incompatible_changes`, so I think it's worth fixing even though it isn't sufficient to get 0.12 working.

johnynek:
thanks!

PR #489, opened 2018-05-01
Format BUILD files and check format during CI

andyscott:
This formats everything using a slight variation of the scripts available in the `bazelbuild/bazel-integrationtesting` repository.


andyscott:
Addresses https://github.com/bazelbuild/rules_scala/issues/474

johnynek:
I see you are tying to cheat your way to the top of the changelog.

;)

Fine by me to add formatting.

andyscott:
> I see you are tying to cheat your way to the top of the changelog.

Just you wait.

andyscott:
There's a bug in buildifier formatting that I'll need to fix before we can get this merged.

ittaiz:
wdyt about following this [PR](https://github.com/bazelbuild/bazel-integration-testing/pull/72)?
We obviously don't have to follow the java formatting but the bzl/BUILD files can be checked every commit and one can run it locally

andyscott:
I've reformatted the files after switching to a `buildtools` with the fix in https://github.com/bazelbuild/buildtools/pull/293.

andyscott:
@ittaiz I believe this PR has roughly the same functionality as the one you've linked. FWIW the formatter is set to install and share the same local cache as the overall build so CI builds shouldn't be slowed down by having to install a ton of stuff.

I can look at migrating the external formatter scripts into targets within the main Bazel build, but I'd prefer to follow up with that in another PR.

johnynek:
@ittaiz how about now? Do you think we should merge?

ittaiz:
Sorry for dropping the ball here but I think there might be an issue here.
I think the script checks bzl files but doesn’t fix them, am I mistaken? If
so people will fail the check but will have a hard time fixing it.
The other script handles this by running python format *and* skylint (it
also has a lengthy explanation on why everything is needed).
If I’m correct then one thing we can do is remove the bzl check from this
PR and merge it.
Then add skylark formatting in a follow up PR.

andyscott:
`format.sh` will format all the files for you if you run it without any arguments. If you pass in the check flag, it does the CI format checking.

I can add some README documentation on this for contributors. Would that do the trick?

ittaiz:
I might have misread the code.
What fixes skylark code? Buildifier only checks it but doesn’t fix it

ittaiz:
From https://github.com/bazelbuild/buildtools/issues/109 and other issues it seems that buildifier only does a very partial job there

andyscott:
This isn't formatting or checking Skylark files yet. I may have misspoken earlier.

ittaiz:
I think the second check does check skylark files.
It looks for bzl files and checks them.
I suggest to either:
1. Use the script from the other PR minus java formatting since it’s very
robust and documented.
2. Only format BUILD files for now and drop the second check from the
current script

andyscott:
I'm currently back on this thread of work and actively porting over the linting scripts from the `bazel-integrationtesting` project.

ittaiz:
Yay 😀
Thanks!

andyscott:
Alright! I think this is good to go. The formatting was done on top of the changes in master as of this comment.

andyscott:
There was a build error that didn't make sense, so I reverted all the formatting changes. Right now I'm stuck running the build since I have Bazel 0.14 instead of Bazel 0.13. I'll post an update soonish.

andyscott:
I think this is _finally_ ready!

johnynek:
@andyscott this diff is really large. Did you update the README to explain how to run the linter to help people who send PRs? I didn't see that.

andyscott:
I'll add a quick note. Regarding the large diff-- yes! The best I could do to manage this was to isolate the diff to a single commit. I can split the formatting changes to separate commits based on file type, if needed.

johnynek:
Okay. I’m +1 when @ittaiz is happy. Thanks for adding the note to Contributing.

andyscott:
The giant diff is _only_ from running `./lint.sh fix` off of the most recent changes in master.

andyscott:
If you want to double check, I can first rebase this PR so that the giant diff is the last commit. Then you could test locally:
```

ittaiz:
Thank you for your perseverance!

andyscott:
Sure thing! And same to you!

PR #495, opened 2018-05-07
Remove tools/bazel.rc on CI as it's incompatible.

philwo:
Local disk caching does not work on Bazel's CI and actually causes the project to fail in our downstream tests at the moment:
https://buildkite.com/bazel/bazel-with-downstream-projects-bazel/builds/224#e8eab4eb-9baf-4503-8328-62d1ca2f454d

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -1,16 +1,25 @@
 ---
 platforms:
   ubuntu1404:
+    shell_commands:
+    # Disable local disk caching on CI.
+    - rm -f tools/bazel.rc
     build_targets:
     - "//test/..."
     test_targets:
     - "//test/..."
   ubuntu1604:
+    shell_commands:
+    # Disable local disk caching on CI.
+    - rm -f tools/bazel.rc
     build_targets:
     - "//test/..."
     test_targets:
     - "//test/..."
   macos:
+    shell_commands:
+    # Disable local disk caching on CI.
+    - rm -f tools/bazel.rc
     build_targets:
     - "//test/..."
     test_targets:


philwo:
Thank you! :)

johnynek:
I think this also has the effect of not testing using workers now since I think workers must be configured in the bazelrc.

Maybe we should have a separate bazelrc for buildkite?

ittaiz:
@philwo can we have that?

philwo:
The currently checked in [bazel.rc](https://github.com/bazelbuild/rules_scala/blob/master/tools/bazel.rc) doesn't enable workers, right?

It should be possible to have a bazelrc just for buildkite. This should work:

- Create a `tools/bazel.rc.buildkite` (or similar)
- Modify `.bazelci/presubmit.yml` and replace `rm -f tools/bazel.rc` with `mv tools/bazel.rc.buildkite tools/bazel.rc`.

ittaiz:
Oh actually the plot thickens...
@johnynek when we added the tools/bazel.rc we actually caused Bazel to
ignore the .travis.bazelrc (which we turn into .bazelrc in travis.yml).
This is because bazelrc isn’t yet hierarchical.
There is a way to import a bazelrc though.

Bottom line I can’t see anywhere in the code that says that we used to use
workers on Bazel’s CI. Only on Travis

ittaiz:
ok. I dug some more into it and apparently I was wrong since `tools/bazelrc` is a global rc and that doesn't contradict with the workspace bazelrc.
Additionally it sounds like @johnynek you were wrong since we were only running with workers on travis (it's defined in `.bazelrc.travis`) so this change didn't hurt us.
Bottom line:
buildkite- runs with workers thanks to #504 (which implicitly also means no local caching since the workers were the only instruction)
travis- runs with workers (thanks to [.bazelrc.travis](https://github.com/bazelbuild/rules_scala/blob/10564c6e3cd7e7c3fcce40d5a48b1b04c4382f5a/.bazelrc.travis#L20)) and with disk caching (thanks to #478 https://github.com/bazelbuild/rules_scala/blob/606ccaeb3de6ac88c73dba69f87767ab43e7707b/tools/bazel.rc)

PR #496, opened 2018-05-07
fix bug in jvm_import_external 

natansil:
when user adds a `generated_linkable_rule_name` attribute


third_party/plugin/src/test/BUILD
```diff
@@ -12,10 +12,10 @@ scala_junit_test(
                 "//external:io_bazel_rules_scala/dependency/scala/scala_compiler",
                 "//external:io_bazel_rules_scala/dependency/scala/scala_library",
                 "@com_google_guava_guava_21_0_with_file//jar",
-                "@org_apache_commons_commons_lang_3_5_without_file//jar"
+                "@org_apache_commons_commons_lang_3_5_without_file//:linkable_org_apache_commons_commons_lang_3_5_without_file"
```
natansil: @ittaiz please note that for `linkable` imports the label cannot use `jar` alias

PR #500, opened 2018-05-15
Prevent exception in static initializers in specs2 tests taking down the test runner

### LINKED ISSUE #499: Exception in initializer causes test runner to crash
Given this test (using specs2):

```scala
import org.specs2.mutable.SpecificationWithJUnit

class MyTest extends SpecificationWithJUnit {
    val myVal = {
        throw new Exception("boom!")  
    }
}
```

Will cause the test runner to crash, and 0-byte `test.xml` written to the disk by Bazel.
Specs2 is gracefully handing the initializer exception (turning it into a detailed `UserException`, but this `UserException` is not handled in the runner, and ends up crashing Bazel.

This only seems to affect Scala tests - Java JUnit exceptions are seem to be handled by Bazel's JUnit runner.

hmemcpy:
Fixes #499 

This produces an empty top-level (suite) description if there's an exception while retrieving the underlying test description (happens if exceptions are thrown from `val` fields).
This allows the bazel test runner to gracefully complete the run and write the exception to the `text.xml` file correctly.

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -390,6 +390,8 @@ def _build_deployable(ctx, jars_list):
     args.extend([j.path for j in jars_list])
     if getattr(ctx.attr, "main_class", ""):
         args.extend(["--main_class", ctx.attr.main_class])
+    if "main_class" in ctx.var:
+        args.extend(["--main_class", ctx.var["main_class"]])
     args.extend(["--output", ctx.outputs.deploy_jar.path])
     ctx.actions.run(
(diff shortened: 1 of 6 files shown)


johnynek:
Can we add a test for this?

Also it would be nice to test for this with scalatest based tests as well.

hmemcpy:
@johnynek sure, I just added a test for specs, will add another one for scalatest as well!

ittaiz:
@johnynek I'll merge tomorrow if you don't object by then

PR #501, opened 2018-05-16
Update tools/bazel.rc for 0.14.00

dslomov:
Fixes #498

PR #502, opened 2018-05-16
disable local cache

### LINKED ISSUE #498: `tools/bazel.rc` uses removed option `--experimental_local_disk_cache`
This fails on Bazel CI: https://buildkite.com/bazel/bazel-with-downstream-projects-bazel/builds/232#1cd8b9e4-c91f-4152-9c39-53b6e8a8032b

buchgr:
@buchgr Starting with Bazel 0.14 these flags no longer exist and have been replaced by
the --disk_cache flag. We want to maintain compatibility with older Bazel versions
and thus we have to remove these flags for now.

Fixes #498.

### DIFF
--- a/tools/bazel.rc
+++ b/tools/bazel.rc
@@ -1,2 +1,0 @@
-build --experimental_local_disk_cache
-build --experimental_local_disk_cache_path=.bazel_cache


ittaiz:
this is just from a github search, right? since @philwo removed our usage  (#495) of this in buildkite saying that: 

> Local disk caching does not work on Bazel's CI and actually causes the project to fail in our downstream tests at the moment

The other problem is that our travis is currently stuck on 0.11.1 which also means @dslomov's solution in #501 won't work.
If you're not seeing an actual problem but just want to clean it up then I'll close this one and keep #501 open until I fix the travis to work with at least 0.13.0

WDYT?

dslomov:
It actually causes failures on CI: https://buildkite.com/bazel/bazel-with-downstream-projects-bazel/builds/232#1cd8b9e4-c91f-4152-9c39-53b6e8a8032b (can you see that?)
I do not know why #495 does not remove tools/bazel.rc from rules_scala. @buchgr wdyt?

ittaiz:
can you please look now after #504 was merged?

buchgr:
Hi ittai,

it turned out that we had a bug in our CI setup. We would run `bazel version` before even reading the `presubmit.yml` CI configuration. `bazel version` would pick up flags from the `.bazelrc` file and fail as they no longer exist. I have updated our ci to do `bazel --nomaster_rc --bazelrc=/dev/null info` instead.

PR #503, opened 2018-05-17
Moving scala junit test from test_rules_scala.sh to RulesScalaIntegra…

dimagoltsman:
Moving scala junit test from test_rules_scala.sh to RulesScalaIntegrationTests.scala using bazel-integration-testing framework.

Note:
currently it is running the sandbox with a specific commit of rules_scala.
what is left to do is to switch using the local build of rules_scala and move rest of the tests to RulesScalaIntegrationTests.scala

cc: @ittaiz   @genrym


WORKSPACE
```diff
@@ -108,3 +116,38 @@ java_import_external(
...
 )
+
+build_bazel_integration_testing_version="36ffe6fe0f4bb727c1fe34209a8d6fd33d8d0d8e" # update this as needed
+http_archive(
+    name = "build_bazel_integration_testing",
+    url = "https://github.com/bazelbuild/bazel-integration-testing/archive/%s.zip"%build_bazel_integration_testing_version,
+    strip_prefix = "bazel-integration-testing-" + build_bazel_integration_testing_version,
+)
+load("@build_bazel_integration_testing//tools:repositories.bzl", "bazel_binaries")
+bazel_binaries(versions = ["0.12.0"])
```
johnynek: can we make a `bazel_version_notes.md` or something that lists all the place we need to update when we bump bazel versions?

WORKSPACE
```diff
@@ -108,3 +116,38 @@ java_import_external(
...
+bazel_java_integration_test_deps()
+
+load("@build_bazel_integration_testing//tools:import.bzl", "bazel_external_dependency_archive")
+BAZEL_JAVA_LAUNCHER_VERSION = "0.4.5"
+java_stub_template_url = ("raw.githubusercontent.com/bazelbuild/bazel/" +
+                          BAZEL_JAVA_LAUNCHER_VERSION +
+                          "/src/main/java/com/google/devtools/build/lib/bazel/rules/java/" +
+                          "java_stub_template.txt")
+
+rules_scala_version = "a8ef632b1b020cdf2c215ecd9fcfa84bc435efcb"
```
johnynek: this is pretty wild, we are pointing to a previous version of this repo. Can we comment why?

test/src/main/scala/scala/test/junit/RulesScalaIntegrationTests.scala
```diff
@@ -0,0 +1,87 @@
...
+      line.contains("FAIL") && line.contains(NoTestsFoundLabel)
+    })
+
+    Assert.assertTrue(hasCorrectOutput)
+  }
+
+  private def givenRulesScalaWorkspaceFile(): Unit = {
+    val workspace =
+      """
+        |rules_scala_version="a8ef632b1b020cdf2c215ecd9fcfa84bc435efcb" # update this as needed
```
johnynek: can we not somehow point to the current version? It feels weird to always be testing a previous version. Also, how updating this seems like a challenge especially if we have many.

ittaiz:
Hey,
Haven't had a chance to take a look but as the comment explicitly says the biggest remaining gap is:

> Note:
> currently it is running the sandbox with a specific commit of rules_scala.
> what is left to do is to switch using the local build of rules_scala

johnynek:
Sorry I overlooked the comment. I have a bad habit of just reading the code.

PR #505, opened 2018-05-18
Upgrade bazel travis to 0.12.0

ittaiz:
Hopefully this supersedes #481 (cc @pauldraper)
Like discussed on the ticket I've changed the assertion (due to the naming it was pointless to loosen the assertion so I changed it a bit).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,9 @@ os:
 
 env:
   # we want to test the last release
-  - V=0.13.0 TEST_SCRIPT=test_rules_scala.sh
-  - V=0.13.0 TEST_SCRIPT=test_intellij_aspect.sh
-  - V=0.13.0 TEST_SCRIPT=test_reproducibility.sh
+  - V=0.12.0 TEST_SCRIPT=test_rules_scala.sh
+  - V=0.12.0 TEST_SCRIPT=test_intellij_aspect.sh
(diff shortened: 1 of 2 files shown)


ittaiz:
also apologies for this dragging on for some time

ittaiz:
@johnynek it seems that we have a problem with 0.13.0 and protobuf repository.
```
The value 'REPOSITORY_NAME' has been removed in favor of 'repository_name()', please use the latter (https://docs.bazel.build/versions/master/skylark/lib/native.html#repository_name). You can temporarily allow the old name by using --incompatiblePackageNameIsAFunction=false
```
https://travis-ci.org/bazelbuild/rules_scala/jobs/380855273
On master of protobuf this is still the deprecated form so we should probably open an issue to them.
I want to see that it works with 0.12.0 and then I'll open the relevant issue

johnynek:
huh, are we using an old sha?

Pretty frustrating if google can't even keep up with the deprecation cycle they create.

ittaiz:
We're using an older sha but even on master it isn't fixed.
yeah, it is a bit frustrating.
I think I'll mention it (in a good way) in the issue.

ittaiz:
@buchgr the buildkite failed very strangely. Any chance you can take a look?
Funny enough the files changed don't relate to the bazel build on buildkite...

buchgr:
Sorry I missed this. Is there still a problem?

PR #506, opened 2018-05-20
flatten depset of transitive proto paths (currently broken accumulation)

ittaiz:
@johnynek #454 broke transitive proto paths. Because the test was not in CI (requires bazel 0.12.0 which we didn't have until yesterday) we didn't notice it.
I'm now running the test and fixed the issue.
This is a serious blocker for us so if you can take a look and sign off I'd appreciate it :)


ittaiz:
Those were my exact thoughts :)
Thanks for merging.

PR #507, opened 2018-05-23
update protobuf version

ittaiz:
to a version that includes a fix to the bazel incompatibility of PACKAGE_NAME.
Currently one of the things blocking the usage of 0.13.0 in travis

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -58,16 +58,16 @@ maven_jar(
 
 http_archive(
     name = "com_google_protobuf",
-    urls = ["https://github.com/google/protobuf/archive/b04e5cba356212e4e8c66c61bbe0c3a20537c5b9.zip"],
+    urls = ["https://github.com/google/protobuf/archive/ce57cccef913681c4a2de1c7703a29135df38ba8.zip"],
     strip_prefix = "protobuf-b04e5cba356212e4e8c66c61bbe0c3a20537c5b9",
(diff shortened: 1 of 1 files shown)


johnynek:
got some failures here.

Ping me when you want me to look.

ittaiz:
Just to make sure we’re aligned- we’re not going to merge it before it’s
green right?

johnynek:
@ittaiz yes, sorry I didn't clarify. I'm happy to bump to fix the issue, but of course, almost never should we merge a red PR, and so I think we should get this green.

ittaiz:
@buchgr I think the failure here indicates protobuf doesn't build with bazel@head.
Am I reading it right?
If so wdyt about adding a build for it on ci.bazel.build?
cc @davidstanke since protobuf feels like a critical building block which should be attended (when thinking of source dependencies)

buchgr:
@ittaiz we have pipeline for protobuf HEAD and it's green atm: https://buildkite.com/bazel/protobuf

johnynek:
I wonder if we should migrate the proto support to:

https://github.com/pubref/rules_protobuf

Thoughts @argyris @andyscott @ittaiz?

ittaiz:
First problem was my own 🤦‍♂️ 
Second problem was protobuf actually being broken on master for it seems 2 days (https://github.com/google/protobuf/commit/ccc56a3c1ebf10536c709cbc02164246a223267c#diff-96239ee297e0a92ac6ff96a6bc434ef0)
I'll update the pointer and we'll see

johnynek:
:+1:

PR #509, opened 2018-05-28
Remove some old java_common checks

johnynek:
I think we already require a bazel version that means we don't need to use these checks.

Just some small housekeeping on the way to using modern providers on all the rules now that https://github.com/bazelbuild/intellij/issues/279 is solved.


johnynek:
@ittaiz did you see this one?

ittaiz:
I did. Will comment later today. Sorry for the delay :(

PR #511, opened 2018-06-02
expose scala binary jars in build even protocol

ittaiz:
@johnynek apparently we only expose the wrapper script for scala binary targets.
Added tests for scala_binary/scala_test/scala_junit_test
There is some redundancy between the three but they felt important enough to protect if we re-implement them without a shared method


johnynek:
I wonder if build event protocol has enough info to remove the statsfile instrumentation we added.

johnynek:
@andyscott can you take a look to see if anything is currently missing from build event protocol for timing?

PR #514, opened 2018-06-08
Update Travis to run with Bazel 0.13.0

ittaiz:
Let’s see how this goes now that we’ve upgraded protobuf


ittaiz:
@johnynek I’ll merge this when green, ok?
Hard part will be 0.14.1
We’ll need to work on the JavaInfo part. I’ll start on it after this one I guess

PR #516, opened 2018-06-12
Fix the remote jars of the scala rules and make thrift + protobuf more stable

ianoc-stripe:
The updates around incompatible changes were actually broken for the scrooge rules for remote jars.

Also sort the args to improve the stability


scala_proto/scala_proto.bzl
```diff
@@ -323,7 +323,9 @@ def _root_path(f):
     return '/'.join([f.root.path, f.owner.workspace_root])
 
 def _colon_paths(data):
-    return ':'.join(["{root},{path}".format(root=_root_path(f), path=f.path) for f in data])
+    if type(data) != "list":
```
ittaiz: Why is this needed? Can we maybe have a test which demonstrates it?

scala_proto/scala_proto.bzl
```diff
@@ -351,7 +353,7 @@ def _gen_proto_srcjar_impl(ctx):
         # Command line args to worker cannot be empty so using padding
         flags_arg = "-" + ",".join(ctx.attr.flags),
         # Command line args to worker cannot be empty so using padding
-        packages = "-" + ":".join(depset(transitive = transitive_proto_paths).to_list())
+        packages = "-" + ":".join(sorted(depset(transitive = transitive_proto_paths).to_list()))
```
ittaiz: Strange that our reproducibility test didn’t catch it

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -127,7 +127,8 @@ def _list_to_map(items):
 def _gen_scrooge_srcjar_impl(ctx):
   remote_jars = []
   for target in ctx.attr.remote_jars:
-    remote_jars.append(_jar_filetype.filter(target.files))
+    remote_jars.append(depset(_jar_filetype.filter(target.files)))
```
ittaiz: Same thought about having a test for this

scala_proto/scala_proto.bzl
johnynek: Depsets are supposed to be in a consistent order. This is a bummer that we also need to sort.

scala_proto/scala_proto.bzl
johnynek: Yeah I’d rather have only one input type here: depset or list. Dynamically switching worries me.

twitter_scrooge/twitter_scrooge.bzl
ianoc-stripe: Adding a test for this is pretty involved I believe since we need to have a remote jar built/setup, and its broken now so i'd rather not do this as part of this PR.

scala_proto/scala_proto.bzl
ianoc-stripe: I mostly wanted to make these things more robust since we have more failures at runtime with no types... but i'll remove it.

twitter_scrooge/twitter_scrooge.bzl
ianoc-stripe: Umm nevermind, turns out we had enough stuff in place the test was pretty easy to add.

scala_proto/scala_proto.bzl
ianoc-stripe: I'm not quite sure how we can exactly hit this case in our tests, but from what i can see the input hash/args change which stops it being a cache hit. but the resulting jar is identical so our reproducibility tests wouldn't see anything wrong

PR #521, opened 2018-06-13
Fix issue building all outputs rather than just default

ianoc-stripe:
https://github.com/bazelbuild/rules_scala/pull/511 introduced a bug meaning we always build all non-default outputs for our binary rules. So we were always building the deploy jars (ultimately filling disks in our CI cluster :) )

@ittaiz can you take a look? i didn't remove any of the tests you added so this shouldn't be a regression i don't believe for what you were looking at.

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -614,7 +614,7 @@ def _scala_binary_common(ctx,
   java_provider = create_java_provider(scalaattr, transitive_compile_time_jars)
 
   return struct(
-      files=depset([ctx.outputs.executable, ctx.outputs.jar]),
+      files = depset([ctx.outputs.executable, ctx.outputs.jar]),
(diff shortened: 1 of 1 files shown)


ianoc-stripe:
@johnynek i think you'll need to merge this, i don't believe i've the bits

ittaiz:
I think this might actually break us (not sure java jars are declared this way) but I still think we should merge it since it sounds like the regression on your end is more severe then the one on our end.
I’ll merge when green.

ittaiz:
And sorry as well

ianoc-stripe:
@ittaiz i believe the java jars must be since the same optionality around building exists there. I added to the original definition of the jar (not deploy jar) from the original version. So its possible it should be good for you now?

ittaiz:
Maybe.
I think I’ll add a test to be sure

Issue #510, opened 2018-05-30
use aspects for scrooge / thrift support for scala.


johnynek:
we currently implement scrooge by adding a generic mechanism to see all dependencies in with a field called "extra_information" tacked onto the scala rules. By using that, we can see the full transitive dependencies of any target.

https://github.com/bazelbuild/rules_scala/blob/master/twitter_scrooge/twitter_scrooge.bzl#L76

We use this to make sure scrooge is called on a full covering of the thrift graph.

An alternative could be to use aspects to generate the scrooge:
https://docs.bazel.build/versions/master/skylark/aspects.html#advanced-example

in this picture, aspects help us generate 1 scrooge target for each thrift target. This way, we have a 1:1 mapping. This fixes the need to verify that we cover all the thrifts with a scrooge target, and potentially improves performance (but since scrooge parses so slowly, and we may need to parse many times now it could actually slow down performance).

It would be really nice for old code to continue to build, it may not be possible and may need a migration. In any case, using aspects seems like the right direction since this is what they were designed for.

ittaiz:
Looks good!

PR #523, opened 2018-06-15
Require JavaInfo on deps, move scrooge srcjar to srcs

johnynek:
since we put the scrooge srcjar into deps, we wind up putting all the srcjars as well as the thrift srcs on the classpaths. This is not what we want.

I am having trouble using this rule in a repo with a very large number of thrifts because the classpaths are getting too large. I will do a run with this and see if it bypasses.

cc @ianoc @ittaiz


johnynek:
while this is still probably good, this did not solve my problem. somehow, the thrift jars are still showing up on the classpath.

ittaiz:
Is there value in reviewing this or do you want to iterate a bit more until
you find your solution?

johnynek:
it is a short review, not much to it, but I don't think we should merge it if it does not improve matters.

I'll try to write a test that asserts that the thrifts don't end up on the classpath.

Meanwhile: https://github.com/bazelbuild/bazel/issues/5413 seems to let me bypass (at the expense of very long startup times due to the quadratic algorithm that bazel uses to construct the jars for long classpaths).

ittaiz:
I'll leave this to @ianoc since I'm not sure I understand the change (it seems you're just passing the srcjar in a different way)

ianoc:
this change lgtm

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -289,12 +289,11 @@ def scrooge_scala_library(name,
       with_finagle = with_finagle,
   )
 
-  # deps from macro invocation would come via srcjar
-  # however, retained to make dependency analysis via aspects easier
   scala_library(
       name = name,
+      srcs = [srcjar],
```
ianoc: does this work? i think the scala_library rule is restricted to only allow actual files in the srcs not targets?

ianoc: (or does it intelligently realize the output file?... interesting)

ittaiz:
Can you do that? Restrict to only files (in the declaration)?

scala/scala.bzl
```diff
@@ -12,6 +12,11 @@ load(
     "@io_bazel_rules_scala//specs2:specs2_junit.bzl",
     _specs2_junit_dependencies = "specs2_junit_dependencies")
 
+load(
```
ittaiz: Leftover?

scala/private/common.bzl
```diff
@@ -26,15 +26,12 @@ def _collect_jars_when_dependency_analyzer_is_off(dep_targets):
   runtime_jars = []
 
   for dep_target in dep_targets:
+    # we require a JavaInfo for dependencies
+    # must use java_import or scala_import if you have raw files
```
ittaiz: Why keep the “if”? I think you can require the divider on the signature of the attribute and so throw all of the conditionals away

test/src/main/scala/scala/test/twitter_scrooge/grep.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+if grep -q $1 $2 ; then
+  echo "ERROR: found $1"
```
ittaiz: Maybe add what was expected? Either in the assertion or in the test name since 6 months from now I won’t remember :)

scala/private/rule_impls.bzl
```diff
@@ -562,7 +560,9 @@ def _lib(ctx, non_macro_lib):
       # this information through. extra_information allows passing
       # this information through, and it is up to the new_targets
       # to filter and make sense of this information.
-      extra_information=_collect_extra_information(ctx.attr.deps),
+      # unfortunately, we need to see this for scrooge and protobuf to work,
+      # but those are generating srcjar, so they should really come in via srcs
+      extra_information=_collect_extra_information(ctx.attr.deps + ctx.attr.srcs),
```
ittaiz: I’m sorry but I don’t understand this. Where does scroog euse this?

scala/private/common.bzl
```diff
@@ -49,26 +46,19 @@ def _collect_jars_when_dependency_analyzer_is_on(dep_targets):
...
-    current_dep_transitive_compile_jars = None
-
+    # we require a JavaInfo for dependencies
+    # must use java_import or scala_import if you have raw files
     if JavaInfo in dep_target:
       java_provider = dep_target[JavaInfo]
       current_dep_compile_jars = java_provider.compile_jars
       current_dep_transitive_compile_jars = java_provider.transitive_compile_time_jars
       runtime_jars.append(java_provider.transitive_runtime_jars)
-    else:
```
ittaiz: Again the whole conditional as above but even if I’m wrong why silently ignore and not fail?

scala/private/common.bzl
johnynek: because scrooge puts everything in the deps sadly... we want to fix it, but one step at a time.

johnynek: scrooge macro puts all the thrift dependencies on the deps. :( we need #510

scala/private/rule_impls.bzl
johnynek: search for all the extra information in the file. For instance: https://github.com/bazelbuild/rules_scala/blob/master/twitter_scrooge/twitter_scrooge.bzl#L85

Really this is something that should be replaced with #510 but that is staged behind a few layers of tech debt removal (a supported skylark function to compile scala and get a JavaInfo, and a toolchain that sets up all the things we need for scala so we don't have to copy-paste the scala setup for the ctx).

johnynek:
okay, @ittaiz I think I've addressed your comments. Thanks for the quick look!

ittaiz:
I have a thought,
Wdyt about adding a Boolean attribute until #510 is solved which opts in to this new behavior and once 510 is solved we’ll just enforce this requirement (by explicitly adding required providers)

johnynek:
@ittaiz I’ll try that tomorrow. Maybe the same approach to not generate the java jar when we know we don’t need it.

johnynek:
actually, @ittaiz I'd like to not fail yet and plumb the booleans through.

This will be a nontrivial change to wire that optionality down to the bottom of several collection methods and it is throw away work. Next, several places in the code were already lax in that they probed for expected shape and ignored things that don't match the shape.

Let's steer towards requiring providers at the bazel level and not probe anywhere, directly access `dep[JavaInfo]`.

Can we merge this as is to unblock?

ittaiz:
I approved btw as a way to say that if you feel strongly about not adding the comment to the readme then you can merge as is

johnynek:
okay, @ittaiz one more round... I added comments and went ahead and locked down runtime_deps, which scrooge does not use.

ittaiz:
👍🏽

Issue #402, opened 2018-01-20
create a scrooge toolchain


johnynek:
relates to #14 #399 #400 #401

liucijus:
resolved by #1116

PR #524, opened 2018-06-16
Aspect-based scrooge implementation

### LINKED ISSUE #510: use aspects for scrooge / thrift support for scala.

johnynek:
this fixes #510 

It works, but there are a few issues currently:
.
1. it appears aspects can only have string attr arguments, we have previously been using booleans for with_finagle. See the string requirement here: https://docs.bazel.build/versions/master/skylark/aspects.html#aspect-definition-1
      ```Public aspect attributes are of type string and are called parameters. Parameters must have avalues attribute specified on them. In this case we have a parameter called extension that is allowed to have '*', 'h', or 'cc' as a value.```. I side stepped this by always enabling, but we can revisit by adding a flag on the thrift_library to enable or not if this becomes an issue.
2. there is a weird double wrapping of ScroogeInfo and ScroogeAspectInfo. This is because aspects and the rules themselves cannot return the same provider type. We want to support depending on scrooges in a scrooge because this is how it was implemented before, and it would be nice to minimize migration pains.
3. I can't handle the java flags for the same reason I can't handle with_finagle boolean: they were a list of strings, not a single string from a finite set. This one may be solvable if we use a toolchain to set the java flags, and it may be that the toolchain, which is yet another migration we have yet to finish.

I'd appreciate feedback at this point to raise any warning signs.

cc @ittaiz @ianoc


johnynek:
related to #170 #402

johnynek:
ptal @ianoc @ittaiz @andyscott 

I will try a build of some internal stripe repos with this tomorrow.

It does change the semantics slightly, but it makes them more bazel-y I would argue. Also, I think all the current uses at Stripe will either work directly, or will error (but most will work).

johnynek:
note this is almost a net 0 change in lines: it enables us to remove extra_information and once we get #534 green, we can convert to totally modern providers, which will be nice to finally get there.

andyscott:
I read through and it looks good. What's the subtle semantic change? It's not apparent from a read through the diff.

ittaiz:
btw, the failures seem to be related to the fact that you're using syntax/constructs from 0.14 while we're running in travis with 0.13 (which explains why buildkite is green)

test/src/main/scala/scalarules/test/twitter_scrooge/thrift/bare_jar_thrifts/BUILD
```diff
@@ -1,20 +1,23 @@
...
     name = "barejar",
-    srcs = ["bare-thrift.jar"],
+    jars = ["bare-thrift.jar"],
     visibility = ["//visibility:public"],
 )
 
 java_import(
     name = "barejar_java_import",
-    jars = ["bare-thrift.jar"],
+    jars = ["bare_jar_thrifts_scrooge.jar"],
```
ianoc-stripe: what is `thrifts_scrooge` ?

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -72,134 +75,49 @@ def twitter_scrooge():
...
+      worker_arg_pad + _colon_paths(ps)
+      for ps in [compile_thrifts, include_thrifts, [], []]
   ])
   worker_content = "{output}\n{paths}\n{flags}".format(
-      output = ctx.outputs.srcjar.path,
+      output = jar_output.path,
       paths = path_content,
       flags = worker_arg_pad + ':'.join([
-          '--with-finagle' if ctx.attr.with_finagle else '',
+          # '--with-finagle' if ctx.attr.with_finagle else '',
```
ianoc-stripe: commented code?

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -211,96 +129,193 @@ def _gen_scrooge_srcjar_impl(ctx):
...
+      all_srcjars = depset([scrooge_jar]),
+      buildijar = False,
+      transitive_compile_jars = merged_deps.transitive_compile_time_jars,
+      plugins = [],
+      resource_strip_prefix = "",
+      resources = [],
+      resource_jars = [],
+      labels = {},
+      in_scalacopts = [],
+      print_compile_time = False,
```
ianoc-stripe: not for this PR, but with stats files and other things, we should probably just remove this maybe to clean up/fewer options soon

ianoc-stripe:
small questions and such, otherwise i think lgtm when green. Might be worth firing it through an internal build once you get it green?

twitter_scrooge/twitter_scrooge.bzl
johnynek: I agree. cc @andyscott do we have a ticket to remove this and make sure we can get the data from BEP logs?

thrift/thrift.bzl
```diff
@@ -160,6 +165,9 @@ thrift_library = rule(
         "absolute_prefix": attr.string(default = '', mandatory = False),
         "absolute_prefixes": attr.string_list(),
         # This is a list of JARs which only contain Thrift files
+        # these files will NOT  be compiled as part of the current target,
+        # but the current thrifts do include these. It should also include
```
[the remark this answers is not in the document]
johnynek: so, I kept it segregated, but now we compile the jars. This allows you to do import style of remote items by using a target with only an external_jar. mixing them with the srcs ultimately seemed confusing to me since I don't think we want to rezip them into the output or apply the prefix rules to them. What do you think about this?

It has the side effect, I think, of making current uses just work, although slower if they include a giant jar over and over.

test/src/main/scala/scalarules/test/twitter_scrooge/thrift/bare_jar_thrifts/BUILD
johnynek: it is the compiled scrooge code.

johnynek:
with minor expected changes, this builds our giant thrift repo.

johnynek:
building our largest scala repo, which has more complex usage of thrift is somewhat painful, but manageable (maybe 2 hours of modification).

The problem was:
1. it didn't really matter if you made the deps correct on thrift_library. It was a dummy, so you could fix them in the scrooge_scala_library deps. That is currently not possible because we invoke one scrooge run per thrift_library
2. we removed remote_jar and no wrapping I could see could put that behavior back. Basically, by the time you have a scrooge_scala_library it is too late to add deps. I could work around with scrooge_scala_import, but it was work.
3. we often had thrift targets private, but public scrooge, but the way the rule worked, public scrooge made the thrift public too. That sucks. You had no way to just expose the compiled code, but not allow people to make thrifts that depend on your code, we now have that ability.

twitter_scrooge/twitter_scrooge.bzl
ianoc-stripe: Is this used for the stats file target? I believe originally this was just a println I added

johnynek: I think this is your println. The statsfile is always emitted I believe.

ianoc-stripe:
lgtm

PR #530, opened 2018-06-21
increase the use of scala_toolchain

### LINKED ISSUE #380: extend scala_toolchain to include the scala-sdk

johnynek:
closes #380 related to #170 

@ittaiz @ianoc-stripe please take a look.

I will test our largest repo with this build to make sure it doesn't fail when used as an external repo (which I think happened before).

TODO (maybe not in this PR): we should not use private attributes, I don't think, to get the java toolchain, instead, we should use `ctx.toolchains[...]` inside the rule, but I don't know the toolchain type to put there. I asked this question:

https://stackoverflow.com/questions/50977784/how-to-resolve-the-java-toolchain-in-bazel

cc @katre if you know the answer to how to get the java toolchain.

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -262,7 +262,7 @@ StatsfileOutput: {statsfile_output}
   ctx.actions.run(
       inputs = ins,
       outputs = outs,
-      executable = ctx.executable.scalac,
+      executable = ctx.executable._scalac,
       mnemonic = "Scalac",
       progress_message = "scala %s" % ctx.label,
       execution_requirements = {"supports-workers": "1"},
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -148,7 +148,10 @@ def _compile(ctx, cjars, dep_srcjars, buildijar, transitive_compile_jars,
   srcjars = _srcjar_filetype.filter(ctx.files.srcs)
   all_srcjars = depset(srcjars, transitive = [dep_srcjars])
   # look for any plugins:
-  plugins = _collect_plugin_paths(ctx.attr.plugins)
+  plugins = ctx.attr.plugins
+  if len(plugins) == 0:
+    plugins = toolchain.plugins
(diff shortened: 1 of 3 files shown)


johnynek:
getting

```
Cannot locate runfiles directory. (Set $JAVA_RUNFILES to inhibit searching.)
```
killing all the builds on the bazel CI. So, something is not right....

This passes locally, so it seems like something with remote execution.

johnynek:
I can repro the failure after a `bazel clean` on `bazel 0.14.1`. So, that's not great that it passed before I did the clean.

johnynek:
So, @katre is there any issue with using a label in the current repo in the rules? The toolchain is run in the workspace, but it is pointing to a label, so, I wonder if that's the issue.

If it is not okay, it seems like it should error when the WORKSPACE is parsed and we evaluate the toolchain rule and see the defaults.

I wonder if a work around would be more nesting: make an external repo that only holds the java rule which runs the scala compiler. That is bad for usability though.

It would be really nice to finally solve this issue.

johnynek:
Well, it failed again.... ugh... It is always some kind of experimental project for me trying to decode how this stuff works. Now we have location failing to give a value for `//external:io_bazel_rules_scala/dependency/scala/scala_library`.

It is interesting to note that go does not put any labels in their toolchain. I wonder if that is a hint that it doesn't work, is an untried pattern, or what.

It would be really nice to have it fail at load if you do something that is not supported. It's very expensive in terms of engineer time to try to make progress on this stuff.

johnynek:
superseded by #544

PR #535, opened 2018-06-23
remove cargo-culted extra info in protobuf rules

johnynek:
extra info is a wart from over 2 years ago to allow walking the dependency graph. Since then, it seems that aspects are the way bazel recommends doing things. We only currently use extra_info in the scrooge rules, but the protobuf rules seem to have copied the idea, but does not use it.

ptal @ittaiz 

cc @azymnis


johnynek:
the intellij test failure seems unrelated. Added #536.

PR #537, opened 2018-06-24
refactor the _compile function

johnynek:
The goal here is to produce a skylark function we can call to compile source or source jars. The first step is to stop getting all the arguments implicitly from the ctx which implies that they were set up by a rule.

Instead, we want to have a function we could call inside any other rule, which may not have the right attrs or files.

This does not get all the way there, maybe halfway, but I'd rather send the PR.

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -132,18 +132,15 @@ def _expand_location(ctx, flags):
 def _join_path(args, sep = ","):
   return sep.join([f.path for f in args])
 
-def _compile(ctx, cjars, dep_srcjars, buildijar, transitive_compile_jars,
-             labels, implicit_junit_deps_needed_for_java_compilation):
+def _compile_scala(ctx, output, sources, cjars, all_srcjars, buildijar,
+                   transitive_compile_jars, resource_strip_prefix, resources,
+                   resource_jars, labels, in_scalacopts):
   ijar_output_path = ""
   ijar_cmd_path = ""
   if buildijar:
     ijar_output_path = ctx.outputs.ijar.path
     ijar_cmd_path = ctx.executable._ijar.path
 
-  java_srcs = _java_filetype.filter(ctx.files.srcs)
-  sources = _scala_filetype.filter(ctx.files.srcs) + java_srcs
-  srcjars = _srcjar_filetype.filter(ctx.files.srcs)
-  all_srcjars = depset(srcjars, transitive = [dep_srcjars])
   # look for any plugins:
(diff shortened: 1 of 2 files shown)


johnynek:
@ittaiz @ianoc can you take a look?

scala/private/rule_impls.bzl
```diff
@@ -223,33 +219,31 @@ StatsfileOutput: {statsfile_output}
       ijar_out = ijar_output_path,
       ijar_cmd_path = ijar_cmd_path,
       srcjars = _join_path(all_srcjars.to_list()),
-      java_files = _join_path(java_srcs),
```
ianoc: Do we not need the java files in here?

johnynek: We don’t. Since we removed the java compiler shell-out, we don’t need a separate key. We do have them in the list of all sources.

johnynek: Which scalac needs to compile mixed java/scala targets.

scala/private/rule_impls.bzl
```diff
@@ -343,9 +328,23 @@ def _compile_or_empty(ctx, jars, srcjars, buildijar, transitive_compile_jars,
...
-                        implicit_junit_deps_needed_for_java_compilation)
+    in_srcjars = _srcjar_filetype.filter(ctx.files.srcs)
+    all_srcjars = depset(in_srcjars, transitive = [srcjars])
+    java_srcs = _java_filetype.filter(ctx.files.srcs)
+    sources = _scala_filetype.filter(ctx.files.srcs) + java_srcs
+    _compile_scala(ctx, ctx.outputs.jar, sources, jars, all_srcjars, buildijar,
+                   transitive_compile_jars, ctx.attr.resource_strip_prefix,
+                   ctx.files.resources, ctx.files.resource_jars, jars2labels,
+                   ctx.attr.scalacopts)
+    # compile the java now
```
ittaiz: should we add this `if` inside the `try_to_compile_java_jar` or add another indirection? looks more appropriate that after _compile_scala we'll see just a call to the java counterpart

johnynek: Well, I don't like passing three args (buildijar, ctx.outputs.ijar and ctx.outputs.jar) into compile java. Also, we want a lot more refactors. We want to use the API to create an ijar:

https://docs.bazel.build/versions/master/skylark/lib/java_common.html#run_ijar

so, I'd like to punt on this and continue on cleaning up after I have unblocked the ability to make proto and thrift generation a pure rule, and not a macro.

ittaiz: ok, merge away...

PR #538, opened 2018-06-26
Use java_common.run_ijar

johnynek:
here is a supported function to run ijar:

https://docs.bazel.build/versions/master/skylark/lib/java_common.html#run_ijar

I don't know at the moment when it was introduced because bazel docs are sadly unversioned. It does allow us to delete a fair amount of code, so I think it is worth it.

cc @ittaiz @ianoc

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -329,7 +329,7 @@ def _compile_or_empty(ctx, manifest, jars, srcjars, buildijar,
                   ctx.attr.print_compile_time, ctx.attr.expect_java_output,
                   ctx.attr.scalac_jvm_flags)
 
-    # compile the java now
+    # build ijar if needed
     if buildijar:
       ijar = java_common.run_ijar(
           ctx.actions,
@@ -340,6 +340,8 @@ def _compile_or_empty(ctx, manifest, jars, srcjars, buildijar,
       #  macro code needs to be available at compile-time,
       #  so set ijar == jar
       ijar = ctx.outputs.jar
+
+    # compile the java now
     java_jar = try_to_compile_java_jar(
         ctx, ijar, all_srcjars, java_srcs,
         implicit_junit_deps_needed_for_java_compilation)
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -70,12 +70,9 @@ def _add_resources_cmd(ctx):
     res_cmd.extend([line])
   return "".join(res_cmd)
 
-def _build_nosrc_jar(ctx, buildijar):
+def _build_nosrc_jar(ctx):
   resources = _add_resources_cmd(ctx)
   ijar_cmd = ""
-  if buildijar:
-    ijar_cmd = "\ncp {jar_output} {ijar_output}\n".format(
-        jar_output = ctx.outputs.jar.path, ijar_output = ctx.outputs.ijar.path)
 
   # this ensures the file is not empty
   resources += "META-INF/MANIFEST.MF=%s\n" % ctx.outputs.manifest.path
@@ -97,8 +94,6 @@ touch {statsfile}
   )
 
   outs = [ctx.outputs.jar, ctx.outputs.statsfile]
-  if buildijar:
-    outs.extend([ctx.outputs.ijar])
 
   inputs = ctx.files.resources + [
       ctx.outputs.manifest, ctx.executable._zipper, zipper_arg_path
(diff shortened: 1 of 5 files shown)


johnynek:
looks like run_ijar was added in 0.14.1.

scala/private/rule_impls.bzl
```diff
@@ -355,28 +322,28 @@ def _compile_or_empty(ctx, manifest, jars, srcjars, buildijar,
...
-    if buildijar:
-      scala_output = ctx.outputs.ijar
-    else:
-      scala_output = ctx.outputs.jar
-    java_jar = try_to_compile_java_jar(
-        ctx, scala_output, all_srcjars, java_srcs,
-        implicit_junit_deps_needed_for_java_compilation)
 
-    ijar = None
+    # compile the java now
```
ianoc-stripe: this comment from the diff looks to be around the ijar rather than the java?

ianoc-stripe:
small nit on the comment, otherwise lgtm

scala/private/rule_impls.bzl
johnynek: well, yeah, it is before the ijar which is before java. Due to CI slowness, I'll pick this up in the next PR (the merge of #524 with this).

test_rules_scala.sh
```diff
@@ -176,11 +176,9 @@ test_scala_library_expect_failure_on_missing_direct_deps_warn_mode() {
 
 test_scala_library_expect_failure_on_missing_direct_java() {
   dependency_target='//test_expect_failure/missing_direct_deps/internal_deps:transitive_dependency'
-  #since bazel 0.12.0 the labels are only emmitted if ijar is in play 
-  dependency_file='test_expect_failure/missing_direct_deps/internal_deps/transitive_dependency_ijar.jar'
   test_target='//test_expect_failure/missing_direct_deps/internal_deps:transitive_dependency_java_user'
 
-  expected_message="$dependency_file.*$test_target"
```
ittaiz: this doesn't sound like the right approach. Can you share what the error message was when it failed? this expected message is too generic

johnynek: it no longer prints the ijar. Which you shouldn't have as far as I can tell. Users shouldn't need to worry about it.

It still fails, and if you follow the recommendation, it still gives the correct next step.

johnynek: PS: I'm pretty convinced we need to move the jdeps approach and away from this custom approach that runs at analysis time (vs build time).

johnynek: here is the failure though: https://travis-ci.org/bazelbuild/rules_scala/jobs/397490178#L919

ittaiz: can you explain why you  shouldn't have the ijar?

Re approach:
1. As far as I understand they are trying to get rid of jdeps since the params passing is too expensive and instead just stamp each jar with a canonical label and if an infraction is spotted pay the cost of extracting the label from the jar.
2. Our approach is very similar to the jdeps one where we only warn/error in build time (in the compiler plugin); Our difference is in how we collect the list of labels/jars to pass.
3.  We want to align ourselves with the mainstream approach and I've been working with Google people about a shared design for this

johnynek: the remedy has nothing to do with the ijar, and the ijar is an implementation detail. So, that should not be an message to the users anyway.

Secondly, with this change, we only create the ijar when it is needed (not in binary, not in macro libraries), so that optional output nature probably changed.

Can you explain why the message should mention an ijar?

ittaiz: TL;DR change the assertion to `$dependency_target.*$test_target"`
Longer:
Unfortunately it's a long explanation with some of it here: https://github.com/bazelbuild/rules_scala/pull/481 and some of it is in bazel-discuss threads.
Basically bazel-java people changed the inner workings in Bazel 0.12.0 forward which rendered our previous assertion `expected_message="$dependency_target[ \t]*to[ \t]*$test_target"` useless.
After some discussion we decided to loosen the assertion and check for ijar (which is what they outputted in the message) and thus keep the core logic of the validation.
This was mainly done in https://github.com/bazelbuild/rules_scala/pull/505.
Some time passed and bazel-java people understood that they created a broken API and with some iteration with us they generated the [composable JavaInfo design](https://docs.google.com/document/d/1ubah6phuvWnugShtVgSQnaopQ1BtKtNxQASVwGZA7k0/edit) which contains `run_ijar`.
Apparently one of the (side?) effects of `run_ijar` is that it fixed the message they output and now they stop printing the filename (ijar implementation detail) and come back to outputting the label.

ittaiz: btw, also the test that fails now was changed in #505 and it looks like that for it too the above assertion should be used

ittaiz:
@or-shachar @natansil fyi re changes to our fix strict deps script

test_rules_scala.sh
johnynek: okay, I see how I goofed. I think this is better. What do you think of the current change?

ittaiz: Wrote on the commit. I don’t think it’s enough.
IMHO the point is to show the transitive_target *and* the test_target since they’re both related

johnynek: okay, I pushed one more time.

These tests don't work when run multiple times, which is annoying. It looks like you are relying on the side effect of building, which doesn't happen on the second build with a warn mode.

ianoc-stripe:
Given the assertion looks to be @ittaiz  one, lgtm

test_rules_scala.sh
ittaiz: I agree it’s annoying. There’s no way to re trigger warnings and this issue on CI only started happening with the caching which is pretty recent. The real solution is bazel integration testing since we have a reproducible sandbox for each scenario

PR #540, opened 2018-06-26
Make compile scala callable

johnynek:
this moves all the rest of the items I don't expect would be required out of the function accessing by ctx and instead by a parameter.

After this is merged, I will pick back up #510 which I think can be made to work using this PR.

Note, there are still some toolchain-ish stuff accessed on ctx, but I think that we will just have to add those to the thrift/proto generation rule until we have toolchains working correctly (see #530)

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -132,9 +132,25 @@ def _expand_location(ctx, flags):
 def _join_path(args, sep = ","):
   return sep.join([f.path for f in args])
 
-def _compile_scala(ctx, output, sources, cjars, all_srcjars, buildijar,
-                   transitive_compile_jars, resource_strip_prefix, resources,
-                   resource_jars, labels, in_scalacopts):
+def compile_scala(ctx,
+                  target_label,
+                  output,
+                  manifest,
+                  statsfile,
+                  sources,
+                  cjars,
+                  all_srcjars,
+                  buildijar,
+                  transitive_compile_jars,
+                  plugins,
+                  resource_strip_prefix,
+                  resources,
+                  resource_jars,
+                  labels,
+                  in_scalacopts,
+                  print_compile_time,
+                  expect_java_output,
+                  scalac_jvm_flags = []):
   ijar_output_path = ""
   ijar_cmd_path = ""
   if buildijar:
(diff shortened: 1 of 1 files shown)


scala/private/rule_impls.bzl
```diff
@@ -229,27 +246,26 @@ StatsfileOutput: {statsfile_output}
...
   if buildijar:
     outs.extend([ctx.outputs.ijar])
   ins = (
-      compiler_classpath_jars.to_list() + all_srcjars.to_list() +
-      list(sources) + ctx.files.plugins + dependency_analyzer_plugin_jars +
-      classpath_resources + resources + resource_jars + ctx.files._java_runtime
-      + [ctx.outputs.manifest, ctx.executable._ijar, argfile])
+      compiler_classpath_jars.to_list() + all_srcjars.to_list() + list(sources)
+      + plugins_list + dependency_analyzer_plugin_jars + classpath_resources +
+      resources + resource_jars + [manifest, ctx.executable._ijar, argfile])
```
ittaiz: What happened to ctx.files._java_runtime?

scala/private/rule_impls.bzl
```diff
@@ -132,17 +132,33 @@ def _expand_location(ctx, flags):
...
+                  transitive_compile_jars,
+                  plugins,
+                  resource_strip_prefix,
+                  resources,
+                  resource_jars,
+                  labels,
+                  in_scalacopts,
+                  print_compile_time,
+                  expect_java_output,
+                  scalac_jvm_flags = []):
```
ittaiz: Why does the scalac_jvn_flags have a default value?

scala/private/rule_impls.bzl
johnynek: That should have been removed a long time ago. It was for when we were calling the java compiler ourselves. It is not needed to call ijar which is a c++ program (which we shouldn’t be calling this way anyway).

scala/private/rule_impls.bzl
johnynek: We have virtually never used this in practice and I just put it in. Would you prefer I remove it?

johnynek: I meant removing the default. Of course we can (and I assume will) iterate.

ittaiz: I'd suggest to keep it as is (if it's alright with you) and remove the default

ianoc-stripe:
lgtm, remove the default in a follow up again if no one objects seems reasonable toof

PR #541, opened 2018-06-26
Bump to using min version of 0.14.1

ianoc-stripe:
This will be needed for @johnynek ijar build and it. Also should avoid having to comment out lines in the tools/bazelrc

### DIFF
--- a/test_rules_scala.sh
+++ b/test_rules_scala.sh
@@ -803,11 +803,11 @@ dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 runner=$(get_test_runner "${1:-local}")
 
 $runner bazel build test/...
-$runner bazel build "test/... --all_incompatible_changes"
+#$runner bazel build "test/... --all_incompatible_changes"
 $runner bazel test test/...
 $runner bazel test third_party/...
 $runner bazel build "test/... --strict_java_deps=ERROR"
-$runner bazel build "test/... --strict_java_deps=ERROR --all_incompatible_changes"
+#$runner bazel build "test/... --strict_java_deps=ERROR --all_incompatible_changes"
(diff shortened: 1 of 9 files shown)


test_rules_scala.sh
```diff
@@ -803,11 +803,11 @@ dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 runner=$(get_test_runner "${1:-local}")
 
 $runner bazel build test/...
-$runner bazel build "test/... --all_incompatible_changes"
+#$runner bazel build "test/... --all_incompatible_changes"
```
ittaiz: should we have an issue about reviving it? (will try to take a closer look later)

ianoc-stripe: @ittaiz Totally -- hopefully we can get it back in soon. I took care of a bunch of them, but lost steam when it came to doing the create_provider -> JavaInfo, it looked a little involved since it seems like we should be separating out the transitive compile time jars from the current target's compile jars for the newer API.

ianoc-stripe:
Added https://github.com/bazelbuild/rules_scala/issues/542

ianoc-stripe:
Current only failure here looks like a download failure on travis, could someone with the perms re-kick that job ?

ittaiz:
I’ll re kick in 10 minutes.

ianoc-stripe:
Thank you!

johnynek:
I'd like to hold just a bit on the 0.14 migration because our biggest scala repo is not yet there and I'm trying to get the aspect based thrift generation into production (as part of a big migration on our thrift generation).

johnynek:
I'm going to merge this when it goes green. It is blocking pretty much all the changes we want to make, and 0.13 is pretty old at this point.

Thanks @ianoc-stripe

PR #543, opened 2018-06-26
Add additional substitutions

johnynek:
the launcher in 0.14.1 has some additional jacoco variables. With these unset, things build, but you get a bash warning.

This fixes that.


johnynek:
ptal @ianoc @ittaiz

ianoc:
👍

PR #544, opened 2018-06-26
[WIP] Support for selecting scala version

jhnj:
Currently the scala version used is hardcoded and e.g. using rules_scala to build 2.12 requires pointing to a "2.12" branch. This PR aims to allow users of rules_scala to use any 2.11/2.12 version.

The original goal was to allow the user to specify specifc scala versions for specific targets but as
that would have required more structural changes we decided to focus on a single global Scala version. Due to this I might have made some decisions that don't make sense when just using 1 version.

This is tested with some internal Stripe repos.

### Changes:

- All deps related to a specific Scala version are passed around in a provider: `ScalacProvider`.
This could be a toolchain as in https://github.com/bazelbuild/rules_scala/pull/530, provider was
chosen due to "multiple versions, same workspace" legacy

- New scala versions are added using `new_scala_repository`. This will download scala, create a build file for the scalac worker and add all of this to a provider. The build file could be declared statically as well (done dynamically to allow for multiple versions in the future).

- Both 2.11 and 2.12 versions of libraries (scalatest, scrooge, specs2...) are specified and the correct version is selected using `native.bind`. The versions of Scrooge and scalatest are also bumped as the current ones don't support 2.12.

- Currently I've only tested it manually for 2.12 on a different branch, to test both 2.11/2.12 requires a new workspace for testing. I've got a partly working wip version of that as well.

### Potential problems:
This PR renames the labels for some scala libraries, e.g. `@scala//:scala-xml` -> `@scala_xml_2_12//jar` which
might infere with e.g. replacements when using bazel-deps.
```
replacements:
  org.scala-lang:
    scala-compiler:
      lang: scala/unmangled
      target: "@scala//:scala-compiler"
```

### TODO (maybe):
- Add the ability to select e.g. the typelevel compiler + specify where you want to download scala from
- Multiple scala versions in the same workspace? Create multiple `ScalacProvider`s and override the default to use different versions. Would also require changing some rules to macros

@johnynek @andyscott @travisbrown


johnynek:
This is pretty exciting!

A few comments.

scala/scala.bzl
```diff
@@ -257,56 +253,53 @@ scala_repl = rule(
...
+      "http://central.maven.org/maven2/org/scalactic/scalactic_2.11/3.0.5/scalactic_2.11-3.0.5.jar"
   )
 
-  # scalatest has macros, note http_jar is invoking ijar
   native.http_jar(
-      name = "scalatest",
+      name = "scalatest_2_12",
       url =
-      "https://mirror.bazel.build/oss.sonatype.org/content/groups/public/org/scalatest/scalatest_2.11/2.2.6/scalatest_2.11-2.2.6.jar",
-      sha256 =
```
johnynek: let's make sure we find a solution to lock these back down with shas.

Maybe a dict for `Map[Version, Map[Name, Sha]]` something like that.

scala/scala.bzl
```diff
@@ -257,56 +253,53 @@ scala_repl = rule(
...
+      url =
+      "http://central.maven.org/maven2/org/scalactic/scalactic_2.12/3.0.5/scalactic_2.12-3.0.5.jar"
+  )
+  native.http_jar(
+      name = "scala_xml_2_11",
+      url =
+      "http://central.maven.org/maven2/org/scala-lang/modules/scala-xml_2.11/1.0.5/scala-xml_2.11-1.0.5.jar"
+  )
+  native.http_jar(
+      name = "scala_xml_2_12",
```
johnynek: this changes using the sdk single jar approach to using individual jars. It's not a bad change, but maybe we should make that in the mainline and keep it separate from this?

scala_proto/scala_proto.bzl
```diff
@@ -32,85 +35,105 @@ def scala_proto_repositories():
...
-      sha1 = "73d38f045ea8f09cc1264991d1064add6eac9e00",
-      server = "scala_proto_deps_maven_server",
+      name = "scala_proto_rules_protoc_bridge_2_11",
+      artifact = "com.trueaccord.scalapb:protoc-bridge_2.11:0.3.0-M1",
+      #server = "scala_proto_deps_maven_server",
+  )
+  native.maven_jar(
+      name = "scala_proto_rules_protoc_bridge_2_12",
+      artifact = "com.trueaccord.scalapb:protoc-bridge_2.12:0.3.0-M1",
+      #server = "scala_proto_deps_maven_server",
```
johnynek: let's just remove it entirely vs comment.

scala/providers.bzl
```diff
@@ -24,3 +24,33 @@ def create_scala_provider(ijar, class_jar, compile_jars,
...
       transitive_exports = []  #needed by intellij plugin
   )
+
+ScalacProvider = provider(
+    doc = "ScalaProvider",
+    fields = [
+        "scalac", "scalalib", "scalareflect", "scalacompiler",
+    ])
+
+def _declare_scalac_provider(ctx):
```
johnynek: this is an interesting approach.

scala/providers.bzl
```diff
@@ -24,3 +24,33 @@ def create_scala_provider(ijar, class_jar, compile_jars,
       transitive_runtime_jars = transitive_runtime_jars,
       transitive_exports = []  #needed by intellij plugin
   )
+
+ScalacProvider = provider(
+    doc = "ScalaProvider",
+    fields = [
+        "scalac", "scalalib", "scalareflect", "scalacompiler",
```
johnynek: I think going this way vs toolchains i.e. #530 is something we need to think about.

@katre what do you think? This probably lets us sidestep the current toolchain issues blocking us. It also allows us to potentially have more than one build in a single repo (build for both scala 2.11 and 2.12), although I don't see how.

Side note: instead of enumerating the extra jars, can we have something like default_compile_classpath, default_runtime_classpath or something? That way we can control that a bit easier in the different cases (really only macro_library should get reflect by default, and only repl needs scalacompiler in the runtime jars.

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -294,12 +313,15 @@ def scrooge_scala_library(name,
   scala_library(
       name = name,
       deps = deps + remote_jars + [
-          srcjar, "//external:io_bazel_rules_scala/dependency/thrift/libthrift",
-          "//external:io_bazel_rules_scala/dependency/thrift/scrooge_core"
+          srcjar,
+          "//external:io_bazel_rules_scala/dependency/thrift/libthrift",
+          "//external:io_bazel_rules_scala/dependency/thrift/scrooge_core",
+          "//external:io_bazel_rules_scala/dependency/thrift/util_core",
```
travisbrown: Hurray!

tut_rule/tut.bzl
```diff
@@ -1,21 +1,29 @@
...
+      ".", "_")
   native.maven_server(
       name = "tut_repositories_maven_server",
       url = "https://dl.bintray.com/tpolecat/maven/",
   )
 
   native.maven_jar(
-      name = "io_bazel_rules_scala_org_tpolecat_tut_core",
+      name = "io_bazel_rules_scala_org_tpolecat_tut_core_2_11",
       artifact = scala_mvn_artifact("org.tpolecat:tut-core:0.4.8"),
```
travisbrown: I wish we could update this—0.4.8 is over a year and a half old, which is a looooong time in the life of a young-ish project like this.

scala/scala.bzl
```diff
@@ -70,6 +64,7 @@ _resolve_deps = {
         allow_files = False),
 }
 
+# TODO not version specific
```
travisbrown: Can you explain?

third_party/plugin/src/main/BUILD
```diff
@@ -1,6 +1,6 @@
 licenses(["notice"])  # 3-clause BSD
 
-load("//scala:scala.bzl", "scala_library", "scala_library_for_plugin_bootstrapping")
```
travisbrown: Not a big deal, but I don't understand why this change is necessary.

scala/scala.bzl
jhnj: Leftover, was relevant when trying to use multiple versions, I'll remove

scala/scala.bzl
jhnj: In this case it's possible to just add the shas, just haven't done that. As we only need one scalatest version for 2.11 and one for 2.12 we can hardcode it, could ofc also be done programmatically with something like that

scala/scala.bzl
jhnj: I can make a pr for that 👍 

The reasoning for the change was mainly that I wasn't sure whether the versions for the xml + parser-combinators jars found in `{scala_version}.tgz` are the same for every minor scala version

third_party/plugin/src/main/BUILD
jhnj: Nothing related to this PR, I think I played around with the file when getting tests to pass and noticed it wasn't used

scala_proto/scala_proto.bzl
jhnj: I noticed that in [twitter_scrooge.bzl](https://github.com/bazelbuild/rules_scala/blob/master/twitter_scrooge/twitter_scrooge.bzl#L21) a maven_server with bazels mirror is specified, do we want to use that one or just remove all `server = ...` from this file (I guess `maven_jar` defaults to maven central)

scala/providers.bzl
jhnj: Re multiple versions in a single repo:
You'd be able to add multiple scala versions (`ScalacProvider`s) in [scala_repositories](https://github.com/bazelbuild/rules_scala/pull/544/files#diff-9957b181001dad28e385437a86e99126R260) with custom labels, e.g. `scala_2_12_6`. Then you could override the default `ScalacProvider` passed to an invocation of `scala_library` etc.

However, this would require some rules (e.g. scala_test) that have dependencies that need to be either 2.11/2.12 to be changed to macros (so they can figure out which versions to use). I haven't come up with a better solution to that problem yet at least

scala/private/rule_impls.bzl
```diff
@@ -700,16 +712,18 @@ def scala_test_impl(ctx):
...
   cjars = depset(transitive = [cjars, scalatest_jars])
   transitive_rjars = depset(transitive = [transitive_rjars, scalatest_jars])
 
   if is_dependency_analyzer_on(ctx):
     transitive_compile_jars = depset(
         transitive = [scalatest_jars, transitive_compile_jars])
     scalatest_jars_list = scalatest_jars.to_list()
-    add_labels_of_jars_to(jars_to_labels, ctx.attr._scalatest,
-                          scalatest_jars_list, scalatest_jars_list)
+    for dep in ctx.attr._scalatest:
```
jhnj: As scalatest 3.0.5 requires scalactic I changed `ctx.attr._scalatest` to a `label_list` containing both scalatest and scalactic jars. This might interfere with how the scalatest version is overridden when using bazel-deps.

In `WORKSPACE`:
```
bind(
    name = "io_bazel_rules_scala/dependency/scalatest/scalatest",
    actual = "//3rdparty/jvm/org/scalatest",
)
```
and
```
  org.scalatest:
    scalatest:
      exports:
        - "org.scalactic:scalactic"
      lang: scala
      version: "3.0.4"
```

Not sure how to combine two jars under one label in the workspace

scala/providers.bzl
jhnj: @ittaiz re cross building: exactly, I have a version where the main rules are working like that (scala_library, ...) allowing you to build different graphs for different versions.

Decided to create a pr for this simpler version first as getting all rules to work with multiple versions might require some bigger changes

scala/private/rule_impls.bzl
```diff
@@ -240,18 +243,24 @@ StatsfileOutput: {statsfile_output}
...
   outs = [ctx.outputs.jar, ctx.outputs.statsfile]
   if buildijar:
     outs.extend([ctx.outputs.ijar])
   ins = (compiler_classpath_jars.to_list() + dep_srcjars.to_list() +
          list(srcjars) + list(sources) + ctx.files.srcs + ctx.files.plugins +
          dependency_analyzer_plugin_jars + classpath_resources +
          ctx.files.resources + ctx.files.resource_jars + ctx.files._java_runtime
-         + [ctx.outputs.manifest, ctx.executable._ijar, argfile])
+         + [ctx.outputs.manifest, ctx.executable._ijar, argfile
+           ] + [scalac_provider.scalac.files_to_run.runfiles_manifest])
```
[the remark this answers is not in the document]
jhnj: Works, will update

scala/private/rule_impls.bzl
```diff
@@ -240,18 +243,24 @@ StatsfileOutput: {statsfile_output}
...
+
   ctx.actions.run(
       inputs = ins,
       outputs = outs,
-      executable = ctx.executable._scalac,
+      executable = scalac_provider.scalac.files_to_run.executable,
```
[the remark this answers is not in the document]
jhnj: Doesn't work, command is `["/bin/bash", "-c", ""]` not an executable

scala/scala.bzl
```diff
@@ -217,17 +212,18 @@ _scala_test_attrs = {
     "suites": attr.string_list(),
     "colors": attr.bool(default = True),
     "full_stacktraces": attr.bool(default = True),
-    "_scalatest": attr.label(
-        default = Label(
-            "//external:io_bazel_rules_scala/dependency/scalatest/scalatest"),
-        allow_files = True),
+    "_scalatest": attr.label_list(
+        default = [
+            Label("//external:io_bazel_rules_scala/dependency/scalatest/scalatest"),
```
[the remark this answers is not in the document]
jhnj: Definitely should be like that! Just wasn't sure how to implement it, I'll add something similar

scala/scala.bzl
jhnj: Didn't actually think about 2.13. Makes sense to use a dict, to make it easier to add versions / allow the user to add a version without requiring changes in rules_scala

scala/scala_cross_version.bzl
```diff
@@ -25,3 +25,109 @@ def scala_mvn_artifact(artifact):
...
+      "ScalaCInvoker.java",
+      "ScalacProcessor.java",
+      "Resource.java",
+  ]
+
+  for src in scalac_worker_srcs:
+    path = Label(
+        "@io_bazel_rules_scala//src/java/io/bazel/rulesscala/scalac:{}".format(
+            src))
+    repository_ctx.symlink(path, "scalac_worker_srcs_symlinked/{}".format(src))
```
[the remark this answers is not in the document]
jhnj: Good, I'd like to know if this is supported.

@ittaiz the reason it is not in the main workspace is to allow us to add multiple scala versions (as the scalac_worker needs to be compiled with a specific version of the compiler). So for this PR it could actually be in the main workspace (as we only use one scala version)

scala/scala.bzl
```diff
@@ -217,17 +212,18 @@ _scala_test_attrs = {
...
+        ]),
     "_scalatest_runner": attr.label(
-        executable = True,
         cfg = "host",
         default = Label("//src/java/io/bazel/rulesscala/scala_test:runner.jar"),
+        executable = True,
         allow_files = True),
     "_scalatest_reporter": attr.label(
-        default = Label("//scala/support:test_reporter")),
+        default = "//scala/support:test_reporter"),
```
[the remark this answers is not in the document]
jhnj: I'll readd the `Label`. Probably accidentally changed, I've seen both no `Label` and `Label` used, not sure which is "correct".

scala/scala.bzl
```diff
@@ -356,28 +349,33 @@ def scala_repositories():
       actual = "@scalac_rules_commons_io//jar")
 
   native.bind(
-      name = "io_bazel_rules_scala/dependency/scala/parser_combinators",
-      actual = "@scala//:scala-parser-combinators")
+      name = "io_bazel_rules_scala/dependency/scalatest/scalatest",
+      actual = "@scalatest_{}//jar".format(major_version_underscore))
```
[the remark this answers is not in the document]
jhnj: If this way of selecting the dependencies sound reasonable to you I can definitely add some kind of utility to make it simpler

specs2/specs2.bzl
```diff
@@ -6,44 +6,78 @@ load(
...
-      sha1 = "ebf85118d0bf4ce18acebf1d8475ee7deb7f19f1",
-  )
-
-  native.bind(
-      name = 'io_bazel_rules_scala/dependency/specs2/specs2',
-      actual = "@io_bazel_rules_scala//specs2:specs2")
+def specs2_repositories(scala_version = "2.11.11"):
+  major_version = scala_version[:scala_version.find(".", 2)]
+
+  if major_version == "2.11":
```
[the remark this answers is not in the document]
jhnj: `//specs2:specs2` needs to export the correct versions of the stuff imported in `specs2_repositories`. Thus there is one branch for 2.11 versions and one for 2.12 versions.

This should probably be done in a more readably way by not duplicating everything, I'll update

jhnj:
Re testing:
I definitely need to add some tests, thus far I've mainly used the existing test suite + manual testing. To be able to test multiple versions my plan was to add a workspace specifically for tests as the scala version is specified in the workspace.

WORKSPACE
```diff
@@ -29,21 +29,21 @@ load("//scala:scala_cross_version.bzl", "scala_mvn_artifact")
 # test adding a scala jar:
 maven_jar(
     name = "com_twitter__scalding_date",
-    artifact = scala_mvn_artifact("com.twitter:scalding-date:0.17.0"),
+    artifact = scala_mvn_artifact("com.twitter:scalding-date:0.17.0", "2.11"),
```
ittaiz: thought (which can be deferred to later PR): 
WDYT about:
`artifact = scala_mvn_artifact("com.twitter:scalding-date:0.17.0", default_scala_major_version),`
where `default_scala_major_version` is defined as `extract_major_version_underscore(default_scala_version())`
Will ease changing the default version for rules_scala when we'll need it

scala/private/rule_impls.bzl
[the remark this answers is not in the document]
ittaiz: @laurentlb is this the expected behavior? Sounds strange we can get the inputs and manifests from `resolve_command` but not the executable. (@jhnj this comment isn't blocking the merge)

scala/scala.bzl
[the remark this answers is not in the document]
ittaiz: To me this sounds reasonable. what do **you** think? ;) @johnynek wdyt?

scala/scalatest/BUILD
```diff
@@ -0,0 +1,10 @@
+package(default_visibility = ["//visibility:public"])
+
+java_import(
```
ittaiz: currently java_import doesn't play well enough with our strict-deps mechanism.
We plan on changing this *and* I don't know how many scalatest users use this feature (we use specs2) but maybe it's better to change to `scala_import`

scala/scala.bzl
```diff
@@ -253,53 +252,60 @@ scala_repl = rule(
...
 
-def scala_repositories(scala_version = "2.11.11"):
-  major_version_underscore = scala_version[:scala_version.find(".", 2)].replace(
-      ".", "_")
+def scala_repositories(scala_version = _default_scala_version()):
+  major_version_underscore = _extract_major_version_underscore(scala_version)
 
   _new_scala_repository("scala_default", scala_version)
 
   # scalatest has macros, note http_jar is invoking ijar
```
ittaiz: @johnynek should I open an issue about changing these http_jar to use `scala_maven_import_external`?

scala/scala.bzl
```diff
@@ -349,13 +355,17 @@ def scala_repositories(scala_version = "2.11.11"):
...
   native.bind(
-      name = "io_bazel_rules_scala/dependency/scalatest/scalatest",
+      name = "io_bazel_rules_scala/dependency/scalatest/scalatest_library",
       actual = "@scalatest_{}//jar".format(major_version_underscore))
 
   native.bind(
       name = "io_bazel_rules_scala/dependency/scalactic/scalactic",
       actual = "@scalactic_{}//jar".format(major_version_underscore))
 
+  native.bind(
```
ittaiz: I think you only need this bind. Why do you need to bind scalactic and scalatest_library as well?

scala/private/rule_impls.bzl
```diff
@@ -240,18 +243,25 @@ StatsfileOutput: {statsfile_output}
   ctx.actions.write(
       output = argfile, content = scalac_args + optional_scalac_args)
 
+  scalac_provider = ctx.attr._scalac[_ScalacProvider]
```
ittaiz: @johnynek given your work trying to make `compile_scala` more explicit and less `ctx` dependent; Would you rather this be a parameter?

specs2/specs2.bzl
```diff
@@ -6,44 +6,78 @@ load(
 def specs2_version():
   return "3.8.8"
```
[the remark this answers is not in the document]
ittaiz: definitely agree but:
1. This PR is big as is.
2. There is always the question of what should rules_scala default to. Not sure. We (Wix) use latest specs2 version internally by binding our own target

scala/private/rule_impls.bzl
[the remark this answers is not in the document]
laurentlb: @c-parsons

scala/scala.bzl
jhnj: `scala/scalatest:scalatest` needs to export the correct versions of scalatest/scalactic and here the binds are used to select which versions it exports. I think I'll do it with an if statement instead, that's probably clearer

scala/scala.bzl
```diff
@@ -260,60 +253,57 @@ scala_repl = rule(
...
+
+  _scala_maven_import_external(
+      name = "io_bazel_rules_scala_scala_parser_combinators",
+      artifact = "org.scala-lang.modules:scala-parser-combinators_{major_version}:1.0.4".format(major_version = major_version),
+      jar_sha256 = scala_jar_shas[major_version]["scala_parser_combinators"],
+      licenses = ["notice"],
+      server_urls = maven_servers,
   )
 
   native.maven_server(
```
ittaiz: if we don't use maven_jar then we don't need maven_server anymore

scala/scala_cross_version.bzl
```diff
@@ -110,24 +101,38 @@ _declare_scalac_provider(
...
-      strip_prefix = "scala-{version}".format(version = version),
-      url =
-      "https://downloads.lightbend.com/scala/{version}/scala-{version}.tgz".
-      format(version = version),
-      build_file_content = _generate_scala_imports(major_version),
+def new_scala_repository(name, version, maven_servers):
+  version_underscore = version.replace(".", "_")
+
+  _scala_maven_import_external(
+      name = "io_bazel_rules_scala_scala_library_{}".format(version_underscore),
```
ittaiz: I lost a bit of context but why do we need to declare these twice as external dependencies. Once in scala.bzl and once here.
Also if there's a good enough reason then maybe we need the dict sha trick here as well

test/aspect/aspect.bzl
```diff
@@ -23,32 +23,32 @@ def _rule_impl(ctx):
   expected_deps = {
       "scala_library": [
           "//test/aspect:scala_library",
-          "@scala_default_imports//:scala-library",
+          "@io_bazel_rules_scala_scala_library_2_11_11//:io_bazel_rules_scala_scala_library_2_11_11",
```
ittaiz: you can and probably should replace with shorthand notation of @repo_name

jhnj: The full paths actually need to be here as the test does string comparison between this and what is returned by `test_aspect`

scala/scala.bzl
jhnj: That `maven_server` is still used by 
```
  native.maven_jar(
      name = "scalac_rules_protobuf_java",
      artifact = "com.google.protobuf:protobuf-java:3.1.0",
      sha1 = "e13484d9da178399d32d2d27ee21a77cfb4b7873",
      server = "scalac_deps_maven_server",
  )
```

Could remove it if we're happy just using maven central instead of the bazel mirror

scala/scala_cross_version.bzl
jhnj: In `scala.bzl` we only import scalatest, scala_xml and parser-combinators (which are only specific to major scala versions) and here we import everything that's specific to minor scala versions.

These imports could be in `scala.bzl` as well but if we want to be able to have multiple versions in the future they need to be here.

scala/scala.bzl
ittaiz: Let’s keep it. Maybe we should drop the maven jar entirely but that’s a different story

ittaiz: Let’s keep it. Maybe we should drop the maven jar entirely but that’s a different story

scala/scala.bzl
```diff
@@ -257,54 +254,66 @@ scala_repl = rule(
...
-      licenses = ["notice"],
-      server_urls = maven_servers,
-  )
+def scala_repositories(
+    scala_version = _default_scala_version(),
+    scala_version_jar_shas = _default_scala_version_jar_shas(),
+    maven_servers = ["http://central.maven.org/maven2"]):
+  major_version = _extract_major_version(scala_version)
+  version_underscore = scala_version.replace(".", "_")
+
```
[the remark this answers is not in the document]
jhnj: Not necessarily, I can call it when needed instead (`_extract_major_version` is also just a simple pure function)

WORKSPACE
```diff
@@ -24,26 +24,35 @@ load("//specs2:specs2_junit.bzl", "specs2_junit_repositories")
...
 
 # test adding a scala jar:
 maven_jar(
     name = "com_twitter__scalding_date",
-    artifact = scala_mvn_artifact("com.twitter:scalding-date:0.17.0"),
+    artifact = scala_mvn_artifact(
+        "com.twitter:scalding-date:0.17.0",
+        default_scala_major_version(),
+    ),
     sha1 = "420fb0c4f737a24b851c4316ee0362095710caa5",
```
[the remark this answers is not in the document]
jhnj: My understanding is that these maven_jars are just for tests and as they always use `default_scala_version()` (which is hardcoded) we shouldn't need a map here. The shas only need to be changed if we change `default_scala_version()`. Or should we add a map just to make the fact that they need to be changed more explicit?

scala/providers.bzl
```diff
@@ -24,3 +24,36 @@ def create_scala_provider(ijar, class_jar, compile_jars,
       transitive_runtime_jars = transitive_runtime_jars,
       transitive_exports = []  #needed by intellij plugin
   )
+
+ScalacProvider = provider(
+    doc = "ScalaProvider",
+    fields = [
+        "scalac",
+        "scalalib",
```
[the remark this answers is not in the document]
jhnj: Would the idea be to have `default_compile_classpath` etc. as a depset ( `depset([<source file external/io_bazel_rules_scala_scala_compiler_2_11_12/scala-compiler-2.11.12.jar>, ...], order = "preorder")` ) and combine that with the result of `collect_jars` or just have `default_compile_classpath` be a list of labels that is passed to `collect_jars` (basically replacing `auto_deps`)?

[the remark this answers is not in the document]
johnynek: I think attr.label_list is convenient.

jhnj:
Updated ScalacProvider to include `default_compile_classpath`, `default_runtime_classpath` ... instead of `scala_library`, `scala_reflect` ...

`_collect_jars_from_common_ctx` gets `default_compile_classpath` + `default_runtime_classpath` as parameters. Should `scala_binary` have scala_reflect on its classpath (should `default_compile_classpath` include scala_reflect)? Without it Tut fails.

unrelated: @johnynek is there anything I need to do to get cla/google to identify my stripe email?

jhnj:
There are now some tests run for both 2.11/2.12 by `test_version.sh`. Which tests need to be run still needs to checked, I removed some which don't need to be run for multiple versions but I'll go through that in detail next

scala/private/rule_impls.bzl
```diff
@@ -481,21 +480,27 @@ def is_dependency_analyzer_off(ctx):
...
+  deps_jars = collect_jars(ctx.attr.deps + extra_deps,
                            dependency_analyzer_is_off)
-  (cjars, transitive_rjars, jars2labels,
-   transitive_compile_jars) = (deps_jars.compile_jars,
-                               deps_jars.transitive_runtime_jars,
-                               deps_jars.jars2labels,
-                               deps_jars.transitive_compile_jars)
+
+  (cjars, transitive_rjars, jars2labels, transitive_compile_jars) = (
+      depset(transitive = [deps_jars.compile_jars] + default_cjars),
```
ittaiz: I think you're changing the semantics here because now one can get a warning about the usage of the default compile classpath. Might be wrong but worth taking another look

scala/private/rule_impls.bzl
```diff
@@ -559,10 +567,15 @@ def _lib(ctx, non_macro_lib):
...
-  return _lib(ctx, True)
+  scalac_provider = ctx.attr._scalac[_ScalacProvider]
+  return _lib(ctx, scalac_provider.default_compile_classpath,
+              scalac_provider.default_runtime_classpath, True)
 
 def scala_macro_library_impl(ctx):
-  return _lib(ctx, False)  # don't build the ijar for macros
+  scalac_provider = ctx.attr._scalac[_ScalacProvider]
+  return _lib(ctx, scalac_provider.default_macro_classpath,
+              scalac_provider.default_macro_classpath,
```
ittaiz: do macros need reflect at runtime as well?

johnynek: Macros generally would only need reflect at runtime if they generate code that uses reflect. In that case, I think they should add reflect as an export since it unusual.

johnynek:
@jhnj did you join the bazelers@stripe google group? I think that’s how google sees who is on a CLA.

jhnj:
At the moment the test code is basically a copy of `./tests` with most tests removed, only keeping the ones we want to run with multiple versions of scala. Those tests are run from `./test_version.sh`, still pretty WIP

scala/private/rule_impls.bzl
jhnj: I'll change the naming to `base_compile_classpath`, let me know if there are some bigger changes in the semantics that you'd like

jhnj:
Updated `ScalacProvider` to only include "default_classpath", "default_macro_classpath" and "default_repl_classpath" as having different classpaths for compiletime and runtime would require changes to `common.bzl:collect_jars` and the scope of this PR is big already.

`default_classpath` also includes both scala_library and scala_reflect as having only scala_library makes the dependency_analyzer fail if a scala_library rule depends on a macro library without specifying a dependency on `scala_reflect`. Should we force the user to explicitly define this dependency or always include scala_reflect. Might be that there is some other way and I'm missing something.

Version specific tests are now run using `test_version.sh` which makes a copy of the directory `test_version/version_specific_tests_dir` and adds a WORKSPACE file to it. The bash script is currently pretty hacky so I'd appreciate feedback on how to improve it (@travisbrown). Also let me know if I should remove/add tests from `test_version/version_specific_tests_dir`

travisbrown:
This `test_version.sh` looks reasonable to me, and in line with the other `test_` scripts. Maybe you could use `mktemp` instead of manually appending a timestamp but that's totally a nitpick.

scala/scala_cross_version.bzl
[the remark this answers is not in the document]
dslomov: In general I'll be very wary of creating symlinks pointing outside the external repo. Can this be a filegroup target instead?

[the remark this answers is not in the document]
jhnj: @dsmolov @ittaiz Filegroup doesn't work (or it works but logs warnings). The problem is that I want to compile a target in an external workspace with sources from the original workspace. This is required (or the only way I could come up with) if we want to have multiple dynamically defined scala versions in the same workspace.

If this is a no-go then I can update this PR to not do this (as this PR is only for choosing a single version), but this would make adding support for multiple versions harder.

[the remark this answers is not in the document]
jhnj: @ittaiz as I mentioned that gives me warnings: 
```
in srcs attribute of java_binary rule @io_bazel_rules_scala_scala_default//:scalac_worker: please do not import
'@io_bazel_rules_scala//src/java/io/bazel/rulesscala/scalac:CompileOptions.java' directly.
You should either move the file to this package or depend on an appropriate rule there
```

But if I don't dynamically generate a build file from a repository_rule as is currently done then I could use a filegroup. That would allow us to only have one scala version per workspace. For this PR I think I'll do that

jhnj:
@johnynek @ittaiz this PR should be ready for review, I've hopefully addressed all comments.

I removed the dynamic part of adding scala versions so now we don't do any symlinking between workspaces anymore.

The main thing that haven't been review yet is the testing: `test_version.sh` and `test_version/`

ittaiz:
I’ll take a look in the next 2-3 days.
Can’t we have dynamic versions with @dslomov’s suggestion of a file group?

jhnj:
@ittaiz @dslomov Unfortunately we can't. Commented about that but not sure where that whole comment chain went.

The problem is that we need to use sources from another workspace for a java target. Both linking the sources directly and using a filegroup leads to warnings:
```
WARNING: /private/var/tmp/_bazel_johan/df328ac0aa9f45e46f1da13619951d87/external/io_bazel_rules_scala_sc
ala_default/BUILD:8:12: in srcs attribute of java_binary rule 
@io_bazel_rules_scala_scala_default//:scalac_worker: please do not import 
'@io_bazel_rules_scala//src/java/io/bazel/rulesscala/scalac:CompileOptions.java' directly. You should  
either move the file to this package or depend on an appropriate rule there
```

I initially did it using a filegroup but then added the symlinks to work around those warnings

ittaiz:
I see. @dslomov can we work around this? Multiple versions is a feature
we’d like to support later on

scala/scala.bzl
```diff
@@ -61,7 +56,10 @@ _implicit_deps = {
         default = Label("@bazel_tools//tools/jdk:current_java_runtime"),
         cfg = "host"),
     "_java_runtime": attr.label(
-        default = Label("@bazel_tools//tools/jdk:current_java_runtime"))
+        default = Label("@bazel_tools//tools/jdk:current_java_runtime")),
+    "_scalac": attr.label(
```
johnynek: `_scalac` is a bit confusing, since it includes everything, what about `_scala_tools` or something?

Also, if we make this public, e.g. use `"scala_provider"` as the name with a default, users could change it at individual call sites, something that might be useful, not sure.

scala/scala_cross_version.bzl
```diff
@@ -5,23 +5,71 @@
...
 
-def scala_mvn_artifact(artifact):
+def default_scala_version_jar_shas():
+  return {
+      "scala_compiler": "3e892546b72ab547cb77de4d840bcfd05c853e73390fed7370a8f19acb0735a0",
+      "scala_library": "0b3d6fd42958ee98715ba2ec5fe221f4ca1e694d7c981b0ae0cd68e97baf6dce",
+      "scala_reflect": "6ba385b450a6311a15c918cf8688b9af9327c6104f0ecbd35933cfcd3095fe04",
+  }
+
+def scala_mvn_artifact(artifact, major_scala_version):
```
johnynek: do you want to give a `major_scala_version = extract_major_version(default_scala_version())` variant to avoid breaking people?

specs2/BUILD
```diff
@@ -4,11 +4,11 @@ java_import(
     name = "specs2",
     jars = [],
     exports = [
-        "@io_bazel_rules_scala_org_scalaz_scalaz_core//jar",
-        "@io_bazel_rules_scala_org_scalaz_scalaz_effect//jar",
-        "@io_bazel_rules_scala_org_specs2_specs2_common//jar",
-        "@io_bazel_rules_scala_org_specs2_specs2_core//jar",
-        "@io_bazel_rules_scala_org_specs2_specs2_matcher//jar",
+        "@io_bazel_rules_scala_org_scalaz_scalaz_core",
```
johnynek: this looks weird to me... I guess it is the default target somehow.

src/java/io/bazel/rulesscala/scalac/BUILD
```diff
@@ -1,20 +1,32 @@
 java_binary(
-    name = "scalac",
+    name = "scalac_worker",
```
johnynek: do we need to change the target name? The standard is to make the default target the name of the directory.

scala/BUILD
```diff
@@ -23,3 +27,22 @@ java_import(
     jars = ["@bazel_tools//tools/jdk:TestRunner_deploy.jar"],
     visibility = ["//visibility:public"],
 )
+
+_declare_scalac_provider(
+    name = "scala_default",
+    default_classpath = [
+        "@io_bazel_rules_scala_scala_library",
```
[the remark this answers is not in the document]
jhnj: I'll add the full paths

scala/scala.bzl
jhnj: Currently there's no point in making it public as we can only have one version of `@io_bazel_rules_scala//src/java/io/bazel/rulesscala/scalac:scalac_worker` (= only one version of scala) but for the future when we'd support multiple scala versions that's what I'd want to do.

I'll change the name to `_scala_provider` but keep it private for now

specs2/BUILD
jhnj: `scala_maven_import_external` creates the target `@io_bazel_rules_scala_org_scalaz_scalaz_core:io_bazel_rules_scala_org_scalaz_scalaz_core` and an alias to `@io_bazel_rules_scala_org_scalaz_scalaz_core//jar:jar`, I can add the full path

src/java/io/bazel/rulesscala/scalac/BUILD
jhnj: I'll change it back, this change is leftover from when I did it dynamically

jhnj:
I think github is having some problems at the moment, addressed the comments but my commits don't shown here. Here they are anyways https://github.com/jhnj/rules_scala/commits/johan/single-scala-version-selectable

scala/BUILD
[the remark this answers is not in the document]
ittaiz: honestly I disagree. This is the style promoted (though not enforced) by *_import_external.
I don't feel strongly enough about it to object but just wanted to add these 2c

scala/private/rule_impls.bzl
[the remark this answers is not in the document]
ittaiz: ping @laurentlb @c-parsons

test_version/WORKSPACE.template
```diff
@@ -0,0 +1,88 @@
...
+
+local_repository(
+    name = "io_bazel_rules_scala",
+    path = "../../"
+)
+
+load("@io_bazel_rules_scala//scala:scala_cross_version.bzl", "extract_major_version")
+
+scala_version = "${scala_version}"
+scala_major_version = extract_major_version(scala_version)
```
ittaiz: is this leftovers?

scala/scala.bzl
```diff
@@ -257,54 +255,67 @@ scala_repl = rule(
...
+    maven_servers = ["http://central.maven.org/maven2"]):
+  (scala_version, scala_version_jar_shas) = scala_version_shas
+  major_version = _extract_major_version(scala_version)
+
+  _new_scala_default_repository(
+      scala_version = scala_version,
+      scala_version_jar_shas = scala_version_jar_shas,
+      maven_servers = maven_servers)
+
+  scala_extra_jar_shas = {
```
ittaiz: shouldn't this also be a parameter like the `_default_scala_version_jar_shas`? I might be missing something here

ittaiz:
@jhnj one last thing- you **moved** the test code into the test_version, right? if so did you preserve history? I think it's super important...

scala/scala.bzl
jhnj: As these only need to be specified per major scala version I decided to not to have them as parameters. This means that we'll have to update them to support other major scala versions.

I'd rather leave it like this for now and do another PR to add that functionality

johnynek:
👍 

@ittaiz do you have any concerns?

I'm ready to (squash) merge. Note, the CLA is signed. @jhnj did this work at Stripe, and we have signed the CLA. For some reason the system is not detecting it.

ittaiz:
@jhnj you have conflicts again :( any chance you can merge master one last time?

jhnj:
@ittaiz It should be up to date with latest master, git log has `329543d78d2071ef8a54986f4cd2116e64c6c087` (latest commit on master) locally and github says "This branch has no conflicts with the base branch" for me

ittaiz:
🎉 🎈 well done and thanks!

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -36,46 +43,71 @@ def twitter_scrooge():
...
-  native.maven_jar(
-      name = "util_core",
-      artifact = scala_mvn_artifact("com.twitter:util-core:6.33.0"),
-      sha1 = "bb49fa66a3ca9b7db8cd764d0b26ce498bbccc83",
-      server = "twitter_scrooge_maven_server",
+      actual = '@io_bazel_rules_scala_scrooge_generator')
+
+  _scala_maven_import_external(
+      name = "io_bazel_rules_scala_util_core",
+      artifact = _scala_mvn_artifact("com.twitter:util-core:18.6.0",
```
ianoc-stripe: @johnynek This PR has a few decent changes into how folks must consider the bindings for both the scala libs, and also the version changed in scrooge. --- should we have some sort of upgrading readme/notes with these? 
e.g. if you use bazel deps you need to change X to Y, If you use scrooge you must add the bind section , etc..

johnynek: @jhnj would you be able to make a short .md file that describes how to set up in the normal case but also how to customize to point to different versions? This often comes up as a question users have but we only have the source code to point to and it can be hard to find for new users.

jhnj: @johnynek @ianoc There is a short section in the readme about selecting scala version (`Selecting Scala version
`), I can definitely add a more thorough explanation there (or another md file) if needed.

Regarding the updated thrift version + bazel-deps change, there is currently no versioning for rules_scala, would we want that? We can ofc add upgrading notes for commits that need action from the users, e.g. `76d4ab9f855f78113ee5990a84b0ad55d2417e4a requires you to change X if you're using bazel-deps` but I think that could quickly become noisy + it's not very readable.

(sidenote: the bazel-deps change actually came from an earlier PR)

johnynek: Some users don't use our provided scala_repositories, and instead wire them up from jars found by say bazel-deps. It would be good to add a section on using these rules with bazel-deps (either here, or in the bazel-deps repo).

johnynek: I think referencing commits is fine. I don't think we have the energy to do a more formal versioning for now.

jhnj: If the users just bind their own jars, basically replacing this:
``` 
native.bind(
      name = "io_bazel_rules_scala/dependency/scalatest/scalatest",
      actual = "@io_bazel_rules_scala//scala/scalatest:scalatest")
```
nothing should have changed in this PR, the name it's bound to hasn't changed.

There is a example on using jars from `scala_repositories` with bazel-deps in the bazel-deps readme but can add something here as well

ianoc-stripe: This for instance we need to call out would silently change the version of the scrooge generator used by people. It wasn't a custom binding before hand. Thats an important change we would want to call out since it would lead to strange/confusing errors. 

(in general changing the default version of the libraries is probably something we want to call out pretty loudly. could easily cause binary incompatibilities or other issues for folks even if it compiles inside their repo)

jhnj: Agree, adding notes about that 👍

PR #549, opened 2018-06-28
Add scala-xml and scala-parser-combinators jars individually

jhnj:
As part of a larger PR (https://github.com/bazelbuild/rules_scala/pull/544) for adding support for selecting scala versions I separated `scala-xml` and `scala-parser-combinators` to be imported individually using `native.http_jar`. @johnynek suggested making o separate PR for this smaller change so here it is.

The motivation behind the change is that the versions of `scala-xml` and `scala-parser-combinators` that are downloaded as part of the main `{scala_version}.tgz` differ between minor scala versions and this way we can select specific versions (https://github.com/bazelbuild/rules_scala/pull/544 would allow users to select any 2.11/2.12 minor version)

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -259,7 +259,7 @@ scala_repl = rule(
 
 def scala_repositories(maven_servers = ["http://central.maven.org/maven2"]):
   _scala_maven_import_external(
-      name = "scala_library",
+      name = "io_bazel_rules_scala_scala_library",
       artifact = "org.scala-lang:scala-library:2.11.12",
       jar_sha256 =
       "0b3d6fd42958ee98715ba2ec5fe221f4ca1e694d7c981b0ae0cd68e97baf6dce",
@@ -267,7 +267,7 @@ def scala_repositories(maven_servers = ["http://central.maven.org/maven2"]):
       server_urls = maven_servers,
   )
   _scala_maven_import_external(
-      name = "scala_compiler",
+      name = "io_bazel_rules_scala_scala_compiler",
       artifact = "org.scala-lang:scala-compiler:2.11.12",
       jar_sha256 =
       "3e892546b72ab547cb77de4d840bcfd05c853e73390fed7370a8f19acb0735a0",
@@ -275,7 +275,7 @@ def scala_repositories(maven_servers = ["http://central.maven.org/maven2"]):
       server_urls = maven_servers,
   )
   _scala_maven_import_external(
-      name = "scala_reflect",
+      name = "io_bazel_rules_scala_scala_reflect",
       artifact = "org.scala-lang:scala-reflect:2.11.12",
       jar_sha256 =
       "6ba385b450a6311a15c918cf8688b9af9327c6104f0ecbd35933cfcd3095fe04",
@@ -283,7 +283,7 @@ def scala_repositories(maven_servers = ["http://central.maven.org/maven2"]):
       server_urls = maven_servers,
   )
   _scala_maven_import_external(
-      name = "scalatest",
+      name = "io_bazel_rules_scala_scalatest",
       artifact = "org.scalatest:scalatest_2.11:2.2.6",
       jar_sha256 =
       "f198967436a5e7a69cfd182902adcfbcb9f2e41b349e1a5c8881a2407f615962",
@@ -292,7 +292,7 @@ def scala_repositories(maven_servers = ["http://central.maven.org/maven2"]):
   )
 
   _scala_maven_import_external(
-      name = "scala_xml",
+      name = "io_bazel_rules_scala_scala_xml",
       artifact = "org.scala-lang.modules:scala-xml_2.11:1.0.5",
       jar_sha256 =
       "767e11f33eddcd506980f0ff213f9d553a6a21802e3be1330345f62f7ee3d50f",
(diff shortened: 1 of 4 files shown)


johnynek:
@jhnj you need to sign the CLA. Stripe has already done it for all of stripe. If you use your stripe account, it *should* work, or I think you can personally sign.

scala/scala.bzl
```diff
@@ -251,18 +251,6 @@ scala_repl = rule(
...
-    visibility = ["//visibility:public"],
-)
-
-java_import(
-    name = "scala-parser-combinators",
-    jars = ["lib/scala-parser-combinators_2.11-1.0.4.jar"],
-    visibility = ["//visibility:public"],
-)
-
 java_import(
```
johnynek: can we kill the whole use of the sdk and just use jars? I think that might be the way to go so we can just use maven servers for all dependencies (or git repos I guess).

jhnj:
I'll sign the cla individually. Decided to use my personal account as I won't have access to the stripe one after the summer

scala/scala.bzl
jhnj: Don't see why we couldn't, I'll update the PR after lunch

jhnj:
I signed it!

jhnj:
@johnynek Updated the PR. One question though, currently using `native.http_jar` for the downloads (as scalatest was using it). That or `native.maven_jar`?

scala/scala.bzl
```diff
@@ -300,6 +278,18 @@ def scala_repositories():
...
+      url =
+      "http://central.maven.org/maven2/org/scala-lang/modules/scala-xml_2.11/1.0.5/scala-xml_2.11-1.0.5.jar",
+      sha256 = "767e11f33eddcd506980f0ff213f9d553a6a21802e3be1330345f62f7ee3d50f"
+  )
+  native.http_jar(
+      name = "scala_parser_combinators",
+      url =
+      "http://central.maven.org/maven2/org/scala-lang/modules/scala-parser-combinators_2.11/1.0.4/scala-parser-combinators_2.11-1.0.4.jar",
+      sha256 = "0dfaafce29a9a245b0a9180ec2c1073d2bd8f0330f03a9f1f6a74d1bc83f62d6"
+  )
```
johnynek: I think we need to use `http_file` then `scala_import` them because they could in principle have macros inside, so we don't want to use ijar.

ittaiz:
Please just use scala_import_external or better yet
scala_maven_import_external
They do the same but in a cleaner and more semantic fashion

jhnj:
@johnynek @ittaiz now with http_file + scala_import. Didn't know about `scala_maven_import_external`, I'll use that

johnynek:
can you look at the CI failures? seems like you still have some `@scala//` around.

jhnj:
@johnynek Updated now with `scala_maven_import_external`, I'll check the CI failures

jhnj:
Not sure what the CI failures where, didn't get those errors locally. We'll see if the new commits pass

jhnj:
@johnynek Could you take a look at the CI errors, can't figure out why it's failing.

`dependency_analyzer_test` is failing as it's not able to resolve `$(location //external:io_bazel_rules_scala/dependency/scala/scala_library)`. I couldn't get it working locally by changing to `@scala_library//jar:jar` either.

johnynek:
@jhnj I see:

```
bazel test third_party/...
```
failing. Does that work for you?

jhnj:
Doesn't work, `dependency_analyzer_test` is in third_party

jhnj:
`@scala_library` is added using `scala_maven_import_external` and works correctly elsewhere

ittaiz:
I’ll take a look later today

ittaiz:
I think, but haven’t confirmed it yet, that it’s because scala import doesn’t return the files in the DefaultInfo provider. Third time we’re stublin onto this but each time it’s part of a bigger PR so it doesn’t get fixed. I’ll see if I can send a small PR just for this. @jhnj if you’re blocked by it you can also pick it up

jhnj:
I'll try to make a PR today if I have the time

jhnj:
Cherry-picked commit from https://github.com/bazelbuild/rules_scala/pull/550

scala/scala.bzl
```diff
@@ -249,56 +257,56 @@ scala_repl = rule(
...
-)
-
-java_import(
-    name = "scala-reflect",
-    jars = ["lib/scala-reflect.jar"],
-    visibility = ["//visibility:public"],
-)
-"""
-
 def scala_repositories():
```
johnynek: Can we make server URLs maven_servers as a parameter and default to a single list with maven central?

scala/scala.bzl
```diff
@@ -249,56 +257,56 @@ scala_repl = rule(
...
-      sha256 =
+      artifact = "org.scalatest:scalatest_2.11:2.2.6",
+      jar_sha256 =
       "f198967436a5e7a69cfd182902adcfbcb9f2e41b349e1a5c8881a2407f615962",
+      licenses = ["notice"],
+      server_urls = ["http://central.maven.org/maven2"],
+  )
+
+  _scala_maven_import_external(
+      name = "scala_xml",
```
johnynek: These names are pretty short. Since they have to be globally unique what about io_bazel_rules_scala_scala_xml?

ittaiz: Doesn’t this contradict canonical naming? Why do you think here we should preface with rules_scala?
Not sure I disagree btw, trying to gather my thoughts about it

johnynek: We don’t own the global namespace. So who is to say someone else didn’t point scala_xml at something else. If you prefix with io_bazel_rules_scala I think you are asking for it. I think we can argue that we do own that namespace. If we could solve the problem of a canonical maven encoding, this could be a solved problem.

jhnj: 👍 on having longer names. Don't have strong opinions on whether it should be `io_bazel_rules_scala_scala_xml` or e.g. `org_scala_lang_modules_scala_xml`

PR #550, opened 2018-06-29
Add DefaultInfo provider to outputs of scala_import

jhnj:
https://github.com/bazelbuild/rules_scala/pull/549 is blocked as using a `$(location)` expression with a target created using `scala_import` fails. As @ittaiz suspected this was due to `scala_import` not returning a `DefaultInfo` provider.

Tested with https://github.com/bazelbuild/rules_scala/pull/549 confirming that it fixes the problem.


johnynek:
: +1: when buildkite goes green I’ll merge.

PR #555, opened 2018-07-05
Run "bazel run" tests from test/BUILD

jhnj:
Currently tests executed by `bazel run` are run both by `test_rules_scala.sh` directly:
```
$runner bazel run test/src/main/scala/scalarules/test/twitter_scrooge:justscrooges
$runner bazel run test:JavaBinary
...
```
and using `sh_test` in `test/BUILD`:
```python
# Make sure scala_binary works in test environment
[sh_test(
    name = "Run" + "".join([binary[idx] if binary[idx].isalnum() else "_" for idx in range(len(binary))]),
    srcs = ["test_binary.sh"],
    args = ["$(location %s)" % binary],
    data = [binary if (":" in binary) else (":%s" % binary)],
) for binary in [
    "JavaBinary",
    "JavaBinary2",
...
```

This PR moves all `bazel run` tests to be run using `sh_test` in `test/BUILD`

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -356,6 +356,12 @@ scala_specs2_junit_test(
     "ResourcesStripScalaBinary",
     "ScalaBinary_dependent_on_binary_with_java",
     "//test/src/main/scala/scalarules/test/classpath_resources:classpath_resource",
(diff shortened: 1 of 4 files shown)


johnynek:
@jhnj thanks!

PR #556, opened 2018-07-11
[WIP]Try 0.16.0rc3

ianoc-stripe:


ianoc-stripe:
This is really just to force CI against this

ianoc-stripe:
I think we can just close, the distinct host configuration stuff won't apply to 0.16.1 either -- but that might just be known to break/fail going forward anyway

PR #558, opened 2018-07-14
add scala stdlib to java compile classpath, relates to #526

ittaiz:
@johnynek I'd appreciate your review
@softprops I'm not sure if there are other issues in #526 remaining after this PR. Do you remember?

PR #559, opened 2018-07-14
repro of bug with java_common.compile

ittaiz:
repro of bug with java_common.compile and buildozer commands adding a self edge
@iirina @cushon I'd really appreciate your help with this issue.
I wanted to create an isolated repro but didn't manage to.
Essentially there are scenarios where java_common.compile emits buildozer commands with self edge.
To see:
`bazel build --strict_java_deps=warn //test:foo_dep`
you'll get:
```
INFO: From Compiling Java headers test/foo_java-hjar.jar (1 files):
test/JavaClass.java:8: error: package scala does not exist
test/JavaClass.java:4: warning: [strict] Using type scalarules.test.ScalaCase from an indirect dependency (TOOL_INFO: "//test:foo"). See command below **
test/JavaClass.java:4: warning: [strict] Using type scalarules.test.ScalaCase from an indirect dependency (TOOL_INFO: "//test:foo"). See command below **
 ** Please add the following dependencies:
r 196a2c3 WIP
repro of bug with java_common.compile and buildozer commands adding a self edg
  //test:foo to //test:foo
 ** You can use the following buildozer command:
buildozer 'add deps //test:foo' //test:foo

 ** Please add the following dependencies:
  //test:foo to //test:foo
 ** You can use the following buildozer command:
repro of bug with java_common.compile and buildozer commands adding a self edgbuildozer 'add deps //test:foo' //test:foo
```
where `buildozer 'add deps //test:foo' //test:foo` is the important part
For now I'm adding scala std lib as direct dependency of java targets but I don't think that's the root cause

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -121,3 +121,11 @@ java_import_external(
 load("//private:format.bzl", "format_repositories")
 
 format_repositories()
+load("@io_bazel_rules_scala//scala:scala_maven_import_external.bzl", "scala_maven_import_external")
+scala_maven_import_external(
+          name = "org_scala_lang_scala_library",
+          artifact = "org.scala-lang:scala-library:2.11.12",
+          licenses = ["notice"],  # Apache 2.0
+          jar_sha256 = "17824fcee4d3f46cfaa4da84ebad4f58496426c2b9bc9e341f812ab23a667d5d",
+          server_urls = ["https://repo.maven.apache.org/maven2/"],
+      )
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -263,6 +263,9 @@ def try_to_compile_java_jar(ctx, scala_output, all_srcjars, java_srcs,
   providers_of_dependencies = collect_java_providers_of(ctx.attr.deps)
   providers_of_dependencies += collect_java_providers_of(
       implicit_junit_deps_needed_for_java_compilation)
+# if we add scalalib on the compile classpath of java compilation the issue goes away
+#  providers_of_dependencies += collect_java_providers_of(
+#      [ctx.attr._scalalib, ctx.attr._scalareflect])
   scala_sources_java_provider = _interim_java_provider_for_java_compilation(
       scala_output)
   providers_of_dependencies += [scala_sources_java_provider]
--- a/test/BUILD
+++ b/test/BUILD
@@ -531,6 +531,24 @@ scalapb_proto_library(
     ],
 )
 
+#import is needed since then it's part of the transitive compile dependencies
+scala_library(
+    name = "import",
+    deps = ["@org_scala_lang_scala_library//jar"]
+)
+scala_library(
+    name = "foo",
+    srcs = ["ScalaCase.scala", "JavaClass.java"],
+    deps = [":import"]
+)
+
+#not sure why but when building foo this doesn't happen, only when building foo_dep
+scala_library(
+    name = "foo_dep",
+    srcs = ["Dep.java"],
+    deps = [":foo"]
+)
+
 load(":check_statsfile.bzl", "check_statsfile")
 
 check_statsfile("ScalaBinary")
--- a/test/Dep.java
+++ b/test/Dep.java
@@ -1,0 +1,3 @@
+public class Dep {
+	
+}
--- a/test/JavaClass.java
+++ b/test/JavaClass.java
@@ -1,0 +1,13 @@
+package scalarules.test;
+
+//import scala.Option; //having an import isn't sufficient
+import scalarules.test.ScalaCase; //must, if we just have a usage (commented out below) doesn't reproduce
+
+public class JavaClass {
+
+  public static void foo(scala.Option scalaType) {
+    //scala.Option.empty(); //this usage of scala type doesn't show the issue, only in method signature
+    //new ScalaCase();  //this usage of ScalaCase doesn't show the issue
+  }
+
+}
--- a/test/ScalaCase.scala
+++ b/test/ScalaCase.scala
@@ -1,0 +1,3 @@
+package scalarules.test
+
+class ScalaCase()


ittaiz:
internally even with #558 the problem persists. Will need to invest more time in triage

ittaiz:
Thanks!
Two more data points which might help:
1. Adding scala lib as direct dependency fixes the above case but still this happens in other cases (haven't been able to triage yet) in our internal codebase
2. This doesn't happen in rules_scala version of 22b63a9bfe2edce7454caf0174ccc3cfa8b77f3c (not sure why though)

ittaiz:
@iirina any thoughts? This is currently blocking us from upgrading to a newer rules_scala :(

iirina:
@ittaiz I'm trying to figure it out but still no luck yet.

New information: SJD looks at the file `scalarules/test/ScalaCase.class` and determines that it's included in `bazel-out/k8-fastbuild/bin/test/foo-ijar.jar`, which is not a direct dependency of `test:foo` and this is how it prints the wrong `add_dep` message. Is it correct that `ScalaCase.class` is included in this ijar?

I also noticed that both the ijar and the hjar for `test:foo` are created. That's incorrect, but I don't know yet why.

I'm going to investigate further.

iirina:
I understood why there are so many ijars/header jars created: one for the whole target (`test/foo.jar` -> `test/foo-ijar.jar`) and one for the jar that only has the java files (`test/foo_java.jar` -> `test/foo_java-hjar.jar`).

The `buildozer add_dep` commands are produced when building the header jar `test/foo_java-hjar.jar`. The classpath at that point is `[test/import.jar, test/foo-ijar.jar]` and there are no direct dependencies. SJD sees that `JavaClass.class` uses `ScalaClass.class` and finds it in the user jar `test/foo-ijar.jar`. Since this jar is not marked as a direct jar it fails assuming the user must add the originating target of `test/foo-ijar.jar` to its dependencies. Even though it is the same target. I have a simple [PR](https://github.com/bazelbuild/bazel/pull/5639) that solves this, but I'm not sure if it's correct. 

@tomlu Do you know why we never set `--direct_dependencies` when we create the header jars?

tomlu:
@cushon is definitely better equipped to answer this question.

ittaiz:
Thanks for taking a look!
The ijar isn’t for the whole target but for the scala code.
The hjar comes from java_common.compile obviously and is only for the java
code.
Irina,
How does SJD decide “bazel-out/k8-fastbuild/bin/test/foo-ijar.jar is not a
direct dependency of test:foo”?
I manually create a provider and add it to the deps for this purpose

iirina:
> How does SJD decide “bazel-out/k8-fastbuild/bin/test/foo-ijar.jar is not a
> direct dependency of test:foo”?
> I manually create a provider and add it to the deps for this purpose

When Bazel creates the header jar action it doesn't set the flag `--direct_dependencies` for invoking Turbine but rather it uses the direct jars as the classpath. I am not familiar with how Turbine works and should be used, I hope @cushon could help (he's currently OOO). 

If you have time, could you try building bazel with my PR and see if the issue persists or if there are other problems?

ittaiz:
I will.
Question- if it doesn’t set the flag then shouldn’t I get many warnings?

cushon:
A couple of notes:

Header compilation actions don't enforce strict deps, they leave that to the full compilation actions. If there's actually a problem javabuilder will report it, it isn't necessary to emit the errors in turbine to get enforcement.

> When Bazel creates the header jar action it doesn't set the flag --direct_dependencies for invoking Turbine but rather it uses the direct jars as the classpath.

There are a few code paths to set up turbine command lines. The fast path doesn't use any of the SJD infrastructure, and doesn't need `--direct_dependencies` to be set. The slower path uses jdeps to do the reduced classpath optimization. I think that right now the strict deps infrastructure is all kind of tied together, and we're accidentally enabling SJD warnings as a side-effect of getting the jdeps.

One option is to just drop the log output from passing turbine invocations, since we don't rely on it to emit warnings and it's safe to ignore the output as long as the action succeeded.

ittaiz:
Can you give an example to what you want to drop from the log?
We (Wix) currently work around the mismatch between java/scala rules and
the different SJD support between them by parsing the logs. Wouldn’t want
that to fail.
Cc @natansil @or-shachar

cushon:
> Can you give an example to what you want to drop from the log?

Warnings from the logs for "Compiling Java headers" actions, for example the warnings in https://github.com/bazelbuild/rules_scala/pull/559#issue-201438934.

The regular, non-header compilation actions will still enforce strict deps and report errors (or warnings under `--strict_java_deps=warn`). Emitting the same warnings again during header compilation is redundant.

ittaiz:
Thanks for the explanation!
What you’re saying is that currently I get two “buildozer add deps” log messages? One for headers (which can be wrong in certain code paths) and one for the actual compilation action (which is correct)?
Cc @natansil @or-shachar to make sure our log parsing won’t be hurt

cushon:
> What you’re saying is that currently I get two “buildozer add deps” log messages? One for headers (which can be wrong in certain code paths) and one for the actual compilation action (which is correct)?

Yep, that's the idea. The log message for the action that compiles `test/foo_java.jar` should have a correct buildozer command.

natansil:
So what I understood from this thread, is that wix's parsing of the SJD log will most probably not be affected by the needed change. 
@ittaiz please confirm

ittaiz:
I’m not sure. I’d appreciate it if you or Or can confirm :)

natansil:
An example of the current output of SJD that we use for deciding which buildozer commands to run is this one for example:
```
....java:22: warning: [strict] Using type javax.validation.Constraint from an indirect dependency (TOOL_INFO: "external/javax_validation_validation_api/validation-api-2.0.1.Final.jar").
@Constraint(validatedBy = MetaSiteNameConstraintValidator.class)
 ^
 ** Please add the following dependencies: 
   to //path/to/target 
 ** You can use the following buildozer command: 
buildozer 'add deps ' //path/to/target 
```
@or-shachar have I missed any other kinds of log entries?

@cushon will these types of log entries be affected?

ittaiz:
@cushon any chance this will get in to the 0.17.0 release?

ittaiz:
@cushon does [this commit](https://github.com/bazelbuild/bazel/commit/a4b28b5fc57d33a245e932eb4763c49a22e61115) contain the fix?

cushon:
I think it was mostly bc9b35b1856b1bd1fc565a8ee3b6de278f6e5c0a. All of the changes in this area should make it into 0.17.

ittaiz:
Thanks!

PR #562, opened 2018-07-17
Fix Intellij Bazel Plugin support for scrooge_scala_library rule

beala-stripe:
# Overview
This fixes intelliJ support for `scrooge_scala_library` targets using the new aspect based implementation (#524). Without this fix, symbol resolution and jump-to-def support does not work. 

Previously, the `scrooge_scala_library` rule would return a `scala` field on its provider struct, which contained output jars (full path: scala -> outputs -> [].jars -> class_jar). This IJ plugin would [inspect this field and collect the jars for IDE support](https://github.com/bazelbuild/intellij/blob/master/aspect/intellij_info_impl.bzl#L352-L366).

The new `scrooge_scala_library` does not return a `scala` field, breaking IJ support. It does return a `JavaInfo` provider, which the IJ plugin does pick up, but it lacks an `outputs` field. One seemingly straightforward solution would be to simply set the `outputs` field on the existing `JavaInfo` provider. There are two reasons this doesn't work:
1. There is at least one place several `JavaInfo`s need to be [merged together](https://github.com/beala-stripe/rules_scala/blob/72d8cd26b267eb88c41adbefbeb357868c1a0bac/twitter_scrooge/twitter_scrooge.bzl#L104), and the current implementation of `java_common.merge` strips the `outputs` field. There is a [fix forthcoming](https://github.com/bazelbuild/bazel/commit/4ff32da2f79f36fbb1e608830835a31f39772b8a) for this, but it's not [landing until 0.16.0](https://groups.google.com/d/msg/bazel-sig-jvm/THl2DRqyYW0/tEkt38SpBgAJ).
2. There is no way AFAIK for setting multiple output_jars on a `JavaInfo`. The `JavaInfo` [constructor](https://docs.bazel.build/versions/master/skylark/lib/JavaInfo.html#JavaInfo) only accepts one output jar and [`java_common.create_provider`](https://docs.bazel.build/versions/master/skylark/lib/java_common.html#create_provider) doesn't set the `outputs` field at all.

The solution in this PR threads the output jars through and puts them in the `scala` field that the IJ plugin expects. Concretely, this means using the `JavaInfo` constructor to set `output_jar` explicitly, and going back to old style providers, where the rule implementation returns a struct rather than a list of providers.

I believe this is an acceptable fix because once the `java_common.merge` fix is released in 0.16.0, the `scala` field can be removed in favor of putting the jars `JavaInfo`s outputs, and we can go back to new-style providers (returning a list rather than a struct).

# Testing
Tests pass (`./test_all.sh`) and I've verified the fix works with the IJ plugin on our repos at Stripe.

That said, this is my first bazel rules PR and would appreciate extra scrutiny, in particular in making sure I've constructed all the `JavaInfo`s correctly.

### DIFF
--- a/twitter_scrooge/twitter_scrooge.bzl
+++ b/twitter_scrooge/twitter_scrooge.bzl
@@ -185,15 +185,13 @@ def _compile_scala(ctx, label, output, scrooge_jar, deps_java_info,
       expect_java_output = False,
       scalac_jvm_flags = [])
 
-  return java_common.create_provider(
-      use_ijar = False,
-      source_jars = [scrooge_jar],
-      compile_time_jars = depset(
-          [output], transitive = [merged_deps.compile_jars]),
-      transitive_compile_time_jars = depset(
-          [output], transitive = [merged_deps.transitive_compile_time_jars]),
-      transitive_runtime_jars = depset(
-          [output], transitive = [merged_deps.transitive_runtime_jars]))
+  return JavaInfo(
+      source_jar = scrooge_jar,
+      deps = deps_java_info + implicit_deps,
+      runtime_deps = deps_java_info + implicit_deps,
+      exports = deps_java_info + implicit_deps,
+      output_jar = output,
+      compile_jar = output)
 
 def _empty_java_info(deps_java_info, implicit_deps):
   merged_deps = java_common.merge(deps_java_info + implicit_deps)
@@ -300,10 +298,11 @@ def _scrooge_scala_library_impl(ctx):
     all_java = java_common.merge(exports + [aspect_info.java_info])
   else:
     all_java = aspect_info.java_info
-  return [
-      DefaultInfo(files = aspect_info.output_files),
-      ScroogeInfo(aspect_info = aspect_info), all_java
-  ]
+
+  return struct(
+    scala = _create_scala_struct(ctx), # For IntelliJ support
+    providers = [all_java, ScroogeInfo(aspect_info = aspect_info), DefaultInfo(files = aspect_info.output_files)]
+  )
 
 scrooge_scala_library = rule(
     implementation = _scrooge_scala_library_impl,
@@ -313,6 +312,23 @@ scrooge_scala_library = rule(
     },
     provides = [DefaultInfo, ScroogeInfo, JavaInfo],
 )
+
+def _create_scala_struct(ctx):
+    """Create a scala provider in the shape expected by the IntelliJ bazel plugin."""
+    output_jars = []
+
+    if ctx.attr.exports:
+      for exp in ctx.attr.exports:
+        for j in exp[JavaInfo].outputs.jars:
+          output_jars.append(struct(class_jar = j.class_jar, ijar = None, source_jar = None, source_jars = []))
+
+    for dep in ctx.attr.deps:
+      for j in dep[ScroogeAspectInfo].java_info.outputs.jars:
+        output_jars.append(struct(class_jar = j.class_jar, ijar = None, source_jar = None, source_jars = []))
+
+    return struct(
+        outputs = struct(jars = output_jars)
+    )
 
 def _scrooge_scala_import_impl(ctx):
   scala_jars = depset(ctx.files.scala_jars)


twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -185,15 +185,13 @@ def _compile_scala(ctx, label, output, scrooge_jar, deps_java_info,
...
-          [output], transitive = [merged_deps.compile_jars]),
-      transitive_compile_time_jars = depset(
-          [output], transitive = [merged_deps.transitive_compile_time_jars]),
-      transitive_runtime_jars = depset(
-          [output], transitive = [merged_deps.transitive_runtime_jars]))
+  return JavaInfo(
+      source_jar = scrooge_jar,
+      deps = deps_java_info + implicit_deps,
+      runtime_deps = deps_java_info + implicit_deps,
+      exports = deps_java_info + implicit_deps,
```
ittaiz: May I ask why do you export everything? Doesn’t sound like the right behavior (though I agree that it looks like the existing code does the same)

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -185,15 +185,13 @@ def _compile_scala(ctx, label, output, scrooge_jar, deps_java_info,
...
+      output_jar = output,
+      compile_jar = output)
```
ittaiz: Just making sure you intentionally don’t want ijar here?

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -300,10 +298,14 @@ def _scrooge_scala_library_impl(ctx):
     all_java = java_common.merge(exports + [aspect_info.java_info])
   else:
     all_java = aspect_info.java_info
-  return [
-      DefaultInfo(files = aspect_info.output_files),
-      ScroogeInfo(aspect_info = aspect_info), all_java
-  ]
+
+  return struct(
+      scala = _create_scala_struct(ctx),  # For IntelliJ support
```
ittaiz: We’re trying to kill this integration in favor of JavaInfo. Is there a specific issue this solves and which JavaInfo doesn’t solve?

beala-stripe:
@johnynek: A splendid idea: https://github.com/bazelbuild/intellij/pull/366

twitter_scrooge/twitter_scrooge.bzl
beala-stripe: Sadly, yes. In order to use `JavaInfo` here, I need a way of setting multiple output jars on a `JavaInfo`, but neither the `JavaInfo` constructor nor the `java_common.create_provider` allow me to do this.

Once [this fix](https://github.com/bazelbuild/bazel/commit/4ff32da2f79f36fbb1e608830835a31f39772b8a) is released, we can ditch the `scala` field and go back to just returning a `JavaInfo` with multiple output jars.

twitter_scrooge/twitter_scrooge.bzl
beala-stripe: See this comment on line 164:
```
   # this only compiles scala, not the ijar, but we don't
   # want the ijar for generated code anyway: any change	   
   # in the thrift generally will change the interface and	   
   # method bodies	   
```

twitter_scrooge/twitter_scrooge.bzl
ittaiz: sounds good. WDYT about adding a TODO and an issue to track this?

twitter_scrooge/twitter_scrooge.bzl
ittaiz: @johnynek wdyt? As far as I'm concerned we can merge as is but just want to make sure this is indeed what you need

johnynek:
Regarding exports, my thinking is: for schemas, all dependencies should are things that show up on the API: if schema foo depends on schema bar, the user of foo is going to need bar to use foo, pretty much 100% of the time. Requiring explicit dependencies, seems like just adding more toil.

Secondly: the motivation for thin dependencies is so you can avoid recomputation, but when a schema changes, almost always the API changes, so you don't get to avoid a recomputation, so the motivation from a caching perspective also disappears.

So, that's why I put the transitive dependencies into exports.

twitter_scrooge/twitter_scrooge.bzl
johnynek: I added a comment to the ticket (schemas don't have implementations that users control, so the data types are generally needed by downstream, more-over if any transitive schema changes you will need a recomputation in most cases).

We can revisit if this intuition is wrong.

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -314,6 +318,31 @@ scrooge_scala_library = rule(
     provides = [DefaultInfo, ScroogeInfo, JavaInfo],
 )
 
+def _create_scala_struct(ctx):
+  """Create a scala provider in the shape expected by the IntelliJ bazel plugin."""
```
johnynek: I guess we are transitively putting things here?


This part is confusing to me because it seems some jars will end up in many targets.

Why doesn't this work to just put `class_jar` from the current target? Why do we need the transitive part?

beala-stripe: It's transitive in the sense that it pulls jars from the scrooge_scala_library's deps. Is that what you mean? I'm not sure I understand the question. Let me explain my understanding and see if that clears it up or if I'm mistaken:

My understanding is, suppose we have some scrooge target:
```
scrooge_scala_library(
    name = "my_scrooge",
    deps = [":thrift1", ":thrift2"]
)
```

The aspect will traverse the deps, call scrooge against each target's sources, produce a jar for each, and store a reference to those jars in each dep's `ScroogeAspectInfo.java_info`. So the result is `:thrift1` and `:thrift2` have jars that I need to return to IJ stored in their `ScroogeAspectInfo.java_info.outputs`.

This function gathers up all those jars by iterating through `:thrift1` and `:thrift2`, pulls the jars out of `ScroogeAspectInfo.java_info.outputs` and puts them in a place IJ can find them.

It also gets the jars for anything a scrooge_scala_library exports since those symbols might be used in the editor.

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -314,6 +318,31 @@ scrooge_scala_library = rule(
...
+  output_jars = []
+
+  if ctx.attr.exports:
```
johnynek: I don't see why we want the exports here. The exports should be another target, so can't the intellij plugin see those? I don't even think we are using exports in scrooge anywhere internally.

Can we remove this part?

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -314,6 +318,31 @@ scrooge_scala_library = rule(
...
+    for exp in ctx.attr.exports:
+      for j in exp[JavaInfo].outputs.jars:
+        output_jars.append(
+            struct(
+                class_jar = j.class_jar,
+                ijar = None,
+                source_jar = None,
+                source_jars = []))
+
+  for dep in ctx.attr.deps:
```
johnynek: okay, I guess this is correct, since the deps are actually the thrifts that we are going to trigger generation of, so pretending that this target created them, I guess is okay (rather than the aspect generating them).

twitter_scrooge/twitter_scrooge.bzl
beala-stripe: Yup I think you're right. Removed.

twitter_scrooge/twitter_scrooge.bzl
beala-stripe: Having the aspects generate a Java/Scala provider is an interesting idea. I tried implementing this, but it doesn't seem like aspects support returning an old style struct provider. I tried returning a struct with a `scala` field from the scrooge aspect, but it wasn't visible in the IJ plugin. 

Once the `java_common.merge` fix is released, the scrooge aspect could attach a JavaInfo with multiple outputs to the `thrift_library` targets.

johnynek: always a yak stack....

next year in Jerusalem.

PR #563, opened 2018-07-17
Make jarcreator compile with older java versions

jhnj:
I was updating a repo that was using an older 2.12 branch to the newest rules_scala and our CI failed with:
```
...
io_bazel_rules_scala/src/java/io/bazel/rulesscala/jar/JarCreator.java:125: error: local variable directory is accessed from within inner class; needs to be declared final
              if (!path.equals(directory)) {
                               ^
``` 

`JarCreator.java` compiles only with java 1.8 (and effectively final), this PR makes it work for older versions as well.

The alternative is to add
```
    javacopts = [
        "-source 1.8",
        "-target 1.8",
    ],
```
to the targets in `src/java/io/bazel/rulesscala/jar/BUILD`

### DIFF
--- a/src/java/io/bazel/rulesscala/jar/JarCreator.java
+++ b/src/java/io/bazel/rulesscala/jar/JarCreator.java
@@ -110,7 +110,7 @@ public class JarCreator extends JarHelper {
    *
    * @param directory the directory to add to the jar
    */
-  public void addDirectory(Path directory) {
+  public void addDirectory(final Path directory) {
     if (!Files.exists(directory)) {
       throw new IllegalArgumentException("directory does not exist: " + directory);
     }


ittaiz:
You’re using a java version older than 8?
I have no problem with the current change but does it mean that you’d rather we avoid java 8 syntax in the Java code of the rule? I think bazel itself already uses that internally.
Again if this is concretely to here then let’s merge (once green) but I just want to discuss it if you want to keep all Java code compatible to source 6/7

jhnj:
@ittaiz Looked a bit more into it and realized why we were running java 7 on CI (not intentional) for this specific repo and thus we can close this PR. I'd rather use java 8 than support 6/7, especially if bazel itself already uses that.

I'll close this, should I create a PR adding
```
    javacopts = [
        "-source 1.8",
        "-target 1.8",
    ],
```
to all java rules to force them to use java 8 or is just leaving them as is enough?

ittaiz:
I think the better approach might be to create a default_java_toolchain
which specifies these since they are global.
That’s what we do internally

PR #566, opened 2018-07-24
Add notes about updates and usage with bazel-deps to readme

jhnj:
Add notes to make upgrading rules_scala easier. Based on discussion here: https://github.com/bazelbuild/rules_scala/pull/544#discussion_r204822174

@johnynek @ianoc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
   # we want to test the last release
   - V=0.14.1 TEST_SCRIPT=test_lint.sh
   - V=0.14.1 TEST_SCRIPT=test_rules_scala.sh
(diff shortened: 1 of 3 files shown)


jhnj:
Seems like the intellij aspect tests are failing, https://download-cf.jetbrains.com/idea/ideaIC-2018.1.3.tar.gz returns 403

@beala-stripe

beala-stripe:
Discussed in slack, but wanted to copy it here for visibility.

That looks like a problem with https://github.com/bazelbuild/intellij. rules_scala clones that repo and runs its tests are part of `test_intellij_aspect.sh`. Those tests depend on that link working. I don't believe this is related to my changes. Could we open an issue/PR in the intellij repo and disable `test_intellij_aspect.sh` for now?

PR #568, opened 2018-07-27
Unused dependency checking

jhnj:
@johnynek last week we discussed adding a check for unused dependencies. This adds that to the current dependency-analyzer and will thus also be run with the `--strict_java_deps` option.

Currently there is no option to run either only the check for the usage of transative deps or only the check for unused direct dependencies, do we want that?

Also quickly chatted about this with @DarkDimius. The current approach [1] will flag some jars as used even if they're not. E.g. considering a jar during implicit search will count it as used.

In general the scala compiler assumes that it will have access to all transative jars (for e.g. errormessages). Although I guess we're already breaking this with ijars etc.

[1]https://github.com/bazelbuild/rules_scala/blob/master/third_party/plugin/src/main/io/bazel/rulesscala/dependencyanalyzer/DependencyAnalyzer.scala#L85


johnynek:
Two things:

1. Have you tried building a repo at Stripe with this?
2. I’m super concerned about the current strict deps killing the cache due to putting the full transitive classpath. The main motivation of this change is better classpath construction so we can get better caching.
3. Maybe the right place for this option is in the scala provider where you can select if you want just to error if the classpath is too big. If you do strict deps warn, maybe we can warn for too big or too small.

I’m worried this will interact badly with exports where people have set up exports that they don’t care if they get used everywhere.

Hard to know the right level to set this at. Maybe it should be an option on each target and by default the setting is controlled by the provider?

Hard to know the answer without trying to use it internally.

One thing I’m sure of: we probably don’t want to use this AND the current strict deps which puts he full transitive path in. One is to help caching one hurts it.

jhnj:
Thanks for the comments!

Makes sense to have it separate from the strict deps. I think having it as an option in the `scala_toolchain` and in the individual targets would be a good approach. Thus you could set the default in the toolchain and add settings for specific targets manually (as with `scalacopts`).

Haven't tried it with any stripe repos yet, I'll make it to a compiler plugin separate from the strict deps checker and then do that.

jhnj:
@johnynek This is now ready to review again.

Added the `unused_dependency_checker` option to the `scala_toolchain` and to `_common_attrs` (used by most scala rules). Setting the option in a rule will override the toolchain.

I created separate plugins for `dependency_analyzer` and `unused_dependency_checker`, for this I had to create different "trees" for the plugins (`third_party/unused_dependency_checker/src/main...`) as scalac plugins has to have a `scalac-plugin.xml` file in the root of the jar and didn't figure out how to do that from a single resource folder.

The plugins has some shared code in the `third_party/utils/...` folder. The shared code is however only for testing as the way plugins currently are added in rules_scala requires them to be a single jar (and not have dependencies)

Currently the plugin errors here [1]. Scala rules that only export other libraries don't work as the exporting rule still creates a jar (which is empty), this jar will then be flagged as unused and fail. Do we want this behavior?


[1] https://github.com/bazelbuild/rules_scala/pull/568/files#diff-c40927b6657800717d4b0b498e1bad55R118

ittaiz:
I’m on vacation so haven’t had a lot of connectivity but why would you like
this on a per target level? This setting sounds more like a general build
hygiene setting than a per target one. I can easily see someone writing a
macro which turns this off and just using that macro.

jhnj:
@ittaiz The default setting (for the whole build) can be set in a toolchain. This setting can then be overridden in individual targets.

Do you see people misusing this as a big risk? Couldn't the same be said for `scalacopts` in the toolchain as well?

And is setting it for individual targets actually needed? With this option you wouldn't have to turn it off completely if you for some reason need to violate it for a single target. Or you can have different modes (warn/error) for different targets

jhnj:
@johnynek, do you have any comments on the api (default in toolchain, can be overridden in individual targets). If you have any I can update that before a more thorough review.

scala/private/rule_impls.bzl
```diff
@@ -509,20 +576,34 @@ def _collect_jars_from_common_ctx(ctx,
       jars2labels = jars2labels,
       transitive_compile_jars = transitive_compile_jars)
 
-def _lib(ctx, base_classpath, non_macro_lib):
+def _lib(ctx,
+         base_classpath,
+         non_macro_lib,
+         unused_dependency_checker_mode = "off"):
```
johnynek: Can we remove defaults as much as possible on private methods? It makes it too easy to overlook them.

jhnj:
@johnynek Tried it out with a stripe repo today. Ran into some problems with bazel-deps targets that pull in a lot of stuff (e.g doobie-postgres [0])

I'll change the logic from:
`find any unused jars and warn about their targets`
to:
`find targets with all of their jars unused and warn about them`


[0]:
```
compile_jars = depset([<source file external/org_tpolecat_doobie_postgres_2_12/jar/doobie-postgres_2.12-0.5.0.jar>, <source file external/org_tpolecat_doobie_core_2_12/jar/doobie-core_2.12-0.5.0.jar>, <source file external/co_fs2_fs2_core_2_12/jar/fs2-core_2.12-0.10.1.jar>, <source file external/com_lihaoyi_sourcecode_2_12/jar/sourcecode_2.12-0.1.4.jar>, <source file external/org_typelevel_cats_core_2_12/jar/cats-core_2.12-1.0.1.jar>, <source file external/org_typelevel_cats_free_2_12/jar/cats-free_2.12-1.0.1.jar>], order = "preorder")
```

johnynek:
@jhnj external dependencies are challenging: most maven builds are not very precise, they will export a ton of stuff. We may want to have a mode to ignore extra deps if they don't come from a known target.

jhnj:
Added a ignored-targets param to the unused dependency checker. Scala-reflect is added to the classpath for all rules but is not necessarily used, this makes sure that doesn't cause an error

scala/scala_toolchain.bzl
```diff
@@ -1,8 +1,13 @@
...
+      unused_dependency_checker_mode = ctx.attr.unused_dependency_checker_mode)
   return [toolchain]
 
 scala_toolchain = rule(
-    _scala_toolchain_impl, attrs = {
+    _scala_toolchain_impl,
+    attrs = {
         'scalacopts': attr.string_list(),
+        'unused_dependency_checker_mode': attr.string(
+            default = "off", values = ["off", "warn", "error"]),
```
johnynek: since this has a default of "off", won't this make the setting on the scala_provider unused?

scala/private/rule_impls.bzl
```diff
@@ -561,29 +644,62 @@ def _lib(ctx, base_classpath, non_macro_lib):
       jars_to_labels = jars.jars2labels,
     )
 
+def get_unused_dependency_checker_mode(ctx):
+  if ctx.attr.unused_dependency_checker_mode:
```
johnynek: isn't this always set since you have a default?

third_party/unused_dependency_checker/src/test/BUILD
```diff
@@ -0,0 +1,26 @@
...
+
+load("//scala:scala.bzl", "scala_junit_test", "scala_test")
+
+scala_test(
+    name = "unused_dependency_checker_test",
+    size = "small",
+    srcs = [
+        "io/bazel/rulesscala/dependencyanalyzer/UnusedDependencyCheckerTest.scala",
+    ],
+    jvm_flags = [
```
johnynek: can you comment about these strings? I don't see them in your code. Seems like you copied the test, but I don't think you need these jvm_flags other than the plugin.jar.location. Am I wrong?

I guess this is the testing fra

jhnj: `-Dguava.jar.location` is not used so removed that.

The other ones are accessed in `TestUtil.scala` and used there and in `UnusedDependencyChecker.scala`


https://github.com/jhnj/rules_scala/blob/johan/dependencyanalyzer-check-unused/third_party/utils/src/test/io/bazel/rulesscala/utils/TestUtil.scala#L58

https://github.com/jhnj/rules_scala/blob/johan/dependencyanalyzer-check-unused/third_party/unused_dependency_checker/src/test/io/bazel/rulesscala/dependencyanalyzer/UnusedDependencyCheckerTest.scala#L39

scala/private/rule_impls.bzl
jhnj: 👍 left from when I didn't have the default

scala/scala_toolchain.bzl
jhnj: The setting in the params passed to individual targets (not in the scala_provider) are checked first and if they're not set then the toolchain is used

scala/scala.bzl
```diff
@@ -127,7 +129,17 @@ _common_attrs.update({
     # on the other hand any internal change (i.e. on code that ijar omits) WON’T trigger recompilation by transitive dependencies
     "_dependency_analyzer_plugin": attr.label(
         default = Label(
-            "@io_bazel_rules_scala//third_party/plugin/src/main:dependency_analyzer"
+            "@io_bazel_rules_scala//third_party/dependency_analyzer/src/main:dependency_analyzer"
+        ),
+        allow_files = [".jar"],
+        mandatory = False),
+    "unused_dependency_checker_mode": attr.string(
+        values = ["warn", "error", "off", ""],
```
jhnj: If this is not set then bazel handles that as an empty string. If that's not set as one of the possible values it fails

scala/private/rule_impls.bzl
johnynek: yep. got it.

jhnj:
@johnynek yea, should be ready. removed the wip

If we want to have an option to check for **all** unused jars and not just targets where with all jars unused I can add that.

This makes the code in `rule_impls.bzl` a bit more complicated as it adds more params to a lot of functions but not sure if there is any way around that without bigger refactorings

ittaiz:
Oscar I did send a comment but it fell through due to github email servers
somehow :(
“These modes sound to me like the complexity is greater than the actual
value of only finding these targets (if you feel otherwise I’d love to hear
why).

Re external dependencies- given a scala_library target foo which depends on
some_external which depends on commons-io (without compilation need). Will
I get a warn/error about foo?
I think the issue might be the serialization choices of bazel deps so if
you can link a small example here it would help me understand “

I’d appreciate waiting off a bit more as this sounds like an important
feature and like I mentioned I’m just getting back from a 2 week vacation

johnynek:
I'm confused about your concerns @ittaiz.

These are disabled by default. They can be useful for us even if only in CI. Do you oppose merging now and iterating as we add features?

We have a considerable issue with extra dependencies that are not really needed. I'd like to start using this to improve our caching performance.

To me, this is similar to "strict deps" which we don't use, but your team wants. It seems okay that you all iterate on it while we leave it disabled.

jhnj:
@ittaiz you won't get a warning in that case.

It only checks for dependencies (targets) that aren't used at all (none of the target's direct jars are used). The logic for which dependencies it warns about is here: https://github.com/bazelbuild/rules_scala/pull/568/files#diff-779f6c1f0135fefd9ca6c8e183957260R66

As an example of why just checking for any unused jars would be problematic together with bazel-deps:
The maven target `doobie-free` contains all of these direct jars if used as a dependency with `bazel-deps`:
```
List(external/org_tpolecat_doobie_postgres_2_12/jar/doobie-postgres_2.12-0.5.0.jar, external/org_tpolecat_doobie_core_2_12/jar/doobie-core_2.12-0.5.0.jar, external/co_fs2_fs2_core_2_12/jar/fs2-core_2.12-0.10.1.jar, external/com_lihaoyi_sourcecode_2_12/jar/sourcecode_2.12-0.1.4.jar, external/org_typelevel_cats_core_2_12/jar/cats-core_2.12-1.0.1.jar, external/org_typelevel_cats_free_2_12/jar/cats-free_2.12-1.0.1.jar)
```
All of those will probably not be used so it would warn if we would just check for unused jars

ittaiz:
@johnynek 
Re strict deps- IIRC you did several rounds of review over a period of time and we did not rush it since it was a big change.
I think this is a big change as well both feature wise and code complexity wise. We're adding more and more features and LOCs without significantly reducing complexity.
I know you've worked hard on removing technical debt like providers but I think that structure wise we're only going backwards.

If this is pressing for you time wise then we can merge but I really think we need to understand what's our plan going forward with respect to modularization.

My concrete concerns were:
1. Adding strict_error/strict_warn sounds like a lot of premature complexity- it wasn't added so I'm good.
2. I don't understand the problem @jhnj said with respect to bazel-deps targets. Is it because you encode scala_library targets and not scala_import? If so do you work around that with the explicit attribute?

One last note is that we also want to use this feature which is another smaller reason for my interest.

ittaiz:
@jhnj ok, after re-reading for a few times I think I understand.
I think the current implementation is very influenced (coupled?) to what you're doing internally two by using exports (which cause many coarse grain jars on a target).
Additionally I think, not sure, that when you use scala_library instead of scala_import for external dependencies you have different semantics which might cause an issue here but I'm not 100% sure (for example the need for the specific target attribute for unused deps).

ittaiz:
@johnynek what's mainly an open question to me is why you think strict-deps and unused-deps are opposites while AFAIU in the java-rules they are used together.
@cushon am I correct in my understanding that in java rules strict-deps and unused-deps complement one another?

cushon:
> in java rules strict-deps and unused-deps complement one another?

Yes, the infrastructure that records which classes are used when compiling a target is the same.

ittaiz:
Thanks!
Re the exports and “all unused” I’ve thought about it again and I think the
implementation makes sense given the exports use-case (we have it
internally for a few very selected targets). I still think it’s not the
best approach to model all of the 3rd party code in but YMMV.
@johnynek only my general points remain and my previous statement about not
blocking you if it’s urgent for you still stands.

jhnj:
strict-deps and unused-deps could work together here as well. The concern that @johnynek has was mainly that strict-deps adds all transative dependencies to the classpath while we'd like to check for unused deps without doing that.

Thus they're two different things atm

johnynek:
@ittaiz in terms of caching, unused deps improve cache performance, "strict-deps" worsens cache performance by putting the transitive deps onto the input list. That is why I mention that.

I don't see how this is coupled to exports, and we don't use exports in 3rdparty deps genereally, we put dependencies on runtime deps. There are cases where people create exports in rules (when they think you can't use a library without it) but not our third party deps.

It is an option to use exports for third party jars, so maybe some repos have done that and that was what @jhnj tested.

As far as the urgency:

1. we want to merge PRs as soon as possible. If you have concrete concerns we are happy to hear them, but we should bias towards shipping.
2. this work is being done by @jhnj who is an intern and will soon be leaving. We would like to merge ASAP so they can exercise it better internally.

So, if you have concrete concerns for us to address in the PR, let us know, but as far as I can see this basically mirrors the strict-deps implementation. We were willing to add the labels machinery (which we are reusing here) for that, so I don't quite see the issue currently.

PR #572, opened 2018-08-03
Create .bazelrc

cvcal:
See bazelbuild/bazel#5756

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -4,6 +4,7 @@ platforms:
     shell_commands:
     # Disable local disk caching on CI.
     - mv tools/bazel.rc.buildkite tools/bazel.rc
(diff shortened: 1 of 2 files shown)


ittaiz:
Thanks!
I think this change will conflict with our travis setup (https://github.com/bazelbuild/rules_scala/blob/master/.travis.yml#L45).
WDYT?

cvcal:
Ah, yes. Buildkite and travis run completely independently, yes? In that case I can modify https://github.com/bazelbuild/rules_scala/blob/master/.bazelci/presubmit.yml instead

cvcal:
Or we could make the travis import be to tools/bazel.rc, and for the .bazelrc file import it either way, but then if we're in neither travis or buildkite, it's got an import to nowhere.

PR #575, opened 2018-08-06
take version from protoc instead of scalapb config

### LINKED ISSUE #367: proto version: taking from scalapb vs taking from protoc

ittaiz:
fixes #367 
@azymnis any objections?


ittaiz:
@johnynek any objections?

johnynek:
It would be nice to have test that exercised what you are trying to get at here, but I guess this won't break us.

We can merge and if this unblocks you. If it does break us, we can follow up with a test or change.

ittaiz:
What I'm trying to get here is to have one source of truth where today we have two (as detailed in #367). This isn't actually for our needs (internally we use a different entry point) but for the good of the community.

PR #578, opened 2018-08-09
Add CI config to test on RBE.

xingao267:


ittaiz:
Exciting, thanks! Any idea why the build fails on RBE?

xingao267:
I got this error when if I turn on targets under //test/tut:

ERROR: /remote_build_test_container/project_src/tut_rule/tut.bzl:34:3: no such package '@io_bazel_rules_scala_org_tpolecat_tut_core//': java.io.IOException: Error downloading [https://dl.bintray.com/tpolecat/maven/org/tpolecat/tut-core_2.11/0.4.8/tut-core_2.11-0.4.8.jar] to /root/.cache/bazel/_bazel_root/7e958634aed2e0b9513fa7cce861a282/external/io_bazel_rules_scala_org_tpolecat_tut_core/tut-core_2.11-0.4.8.jar: GET returned 403 Forbidden and referenced by '//external:io_bazel_rules_scala/dependency/tut/tut_core'
ERROR: Analysis of target '//test/tut:some_example_compiler' failed; build aborted: Analysis failed

It seems also failing on other platforms due to the same reason, which I'm not sure why because this PR should affect those platforms.

xingao267:
It also failed locally with the same reason as well.

WORKSPACE
```diff
@@ -130,3 +131,13 @@ java_import_external(
 load("//private:format.bzl", "format_repositories")
 
 format_repositories()
+
+http_archive(
+    name = "bazel_toolchains",
```
johnynek: where is this being used? I don't see it?

ittaiz: `rbe_ubuntu1604` is the identifier of a runtime which runs Bazel on a build server while using RBE (remote build execution) for remote execution. Part of the flags which they pass to Bazel includes `bazel_toolchains` for example `extra_toolchains` and `crosstool_top` point there

ittaiz:
Re running the builds now that #581 is in

xingao267:
@ittaiz the tests are all good now!

PR #579, opened 2018-08-09
Compile with unused dependency checker

jhnj:
While trying to add the unused dependency checking to one of our larger repos at Stripe I realized that parts of `rules_scala` doesn't pass with unused_dependency_checker on. This PR adds a toolchain with `unused_dependency_checker="error"` to `rules_scala` and fixes stuff to work with that.

- Pass targets used by `scalapb_proto_library` and junit tests as unused_dependency_checker_ignored_targets
- Set `unused_dependency_checker_mode = "off"` for `scala_library` calls in `scala_benchmark_jmh` (alternative is to use `unused_dependency_checker_ignored_targets` but we probably don't care about checking unused deps for benchmarks)
- Don't check for unused deps if the target has java sources
- Set `unused_dependency_checker_mode = "off"` for tests with deps that only have resources
- Set `unused_dependency_checker_mode = "off"` targets with deps that are not used but exported (the other option here would be to add all `exports` to `unused_dependency_checker_ignored_targets`)

This PR also makes it so that if both `unused_dependency_checker_mode` and `strict_java_deps` are active `strict_java_deps` is chosen. This is mainly so that we can have `unused_dependency_checker_mode` set in the `WORKSPACE` to run with tests but also be able to test `strict_java_deps` in the integration tests.


src/scala/io/bazel/rules_scala/jmh_support/BUILD
```diff
@@ -8,8 +8,6 @@ scala_library(
         "//external:io_bazel_rules_scala/dependency/jmh/jmh_core",
         "//external:io_bazel_rules_scala/dependency/jmh/jmh_generator_asm",
         "//external:io_bazel_rules_scala/dependency/jmh/jmh_generator_reflection",
-        "//external:io_bazel_rules_scala/dependency/jmh/org_ows2_asm_asm",
```
johnynek: should these be in the runtime_deps?

src/scala/scripts/BUILD
```diff
@@ -34,7 +34,6 @@ scala_library(
     visibility = ["//visibility:public"],
     deps = [
         ":scala_proto_request_extractor",
-        "//external:io_bazel_rules_scala/dependency/com_google_protobuf/protobuf_java",
```
johnynek: maybe a runtime dep?

src/scala/io/bazel/rules_scala/jmh_support/BUILD
jhnj: Not sure, I'll take a closer look once I get tests passing. Still want to add some stuff to this PR that I noticed while working on one of our bigger repos at stripe. I'll make this WIP

src/scala/scripts/BUILD
jhnj: Same as above

johnynek:
looks like this is still failing due to #580 

Is there no resolution to this yet? ugh...

Maybe we will have to comment the tut tests out if we can't get that jar cached somewhere.

jhnj:
I'll comment out the test for now and at least make sure CI passes without it

ittaiz:
Jonah,
Please rebase on top of master. The test is commented out there.

jhnj:
:+1:, just noticed

jhnj:
@johnynek tests passing now. Also disabled unused-deps mode for `_suite` rules as currently it requires all dependencies of the suite to be used in all targets of the suite. Couldn't think of a way of getting around that without adding a lot of complexity

src/scala/io/bazel/rules_scala/jmh_support/BUILD
johnynek: Well, I guess if tests pass, let's keep it simple.

johnynek: that said, we may not actually be running a benchmark in the tests, can we move to runtime_deps for now and open an issue to see if it is really needed?

src/scala/scripts/BUILD
johnynek: same response

scala_proto/scala_proto.bzl
```diff
@@ -496,6 +496,7 @@ def scalapb_proto_library(name,
   scala_library(
       name = name,
       deps = [srcjar] + external_deps,
+      unused_dependency_checker_ignored_targets = [srcjar] + external_deps,
```
ittaiz: why do we need to ignore here?

scala/scala.bzl
```diff
@@ -440,7 +441,8 @@ def scala_test_suite(name,
         visibility = visibility,
         size = size,
         colors = colors,
-        full_stacktraces = full_stacktraces)
+        full_stacktraces = full_stacktraces,
+        unused_dependency_checker_mode = "off")
```
stefanobaghino-da, line 445: Apologies for commenting on an ancient PR, but I just dug through history do understand this. Am I right in my understanding that for `_suite`-like rules the dependency checker must be disabled as each individual "sub-target" might not depend on the whole set of dependencies for the overall build target? Is there a way in which dependency checking can be done at the target level while being disabled for "sub-target"s? (also apologies if I'm not using the correct wording, I hope that's understandable)

ittaiz: I think indeed suite like rules weren't in the focus of the design. I suggest opening a new issue and referencing this PR to discuss this with current maintainers

stefanobaghino-da: Thanks!

PR #583, opened 2018-08-15
Remove cfg = "data" no-op

jjudd:
This removes the need to use --incompatible_disallow_data_transition

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -108,7 +108,7 @@ _common_attrs_for_plugin_bootstrapping = {
     "deps": attr.label_list(),
     "plugins": attr.label_list(allow_files = [".jar"]),
     "runtime_deps": attr.label_list(providers = [[JavaInfo]]),
-    "data": attr.label_list(allow_files = True, cfg = "data"),
+    "data": attr.label_list(allow_files = True),
     "resources": attr.label_list(allow_files = True),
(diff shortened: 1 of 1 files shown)


jjudd:
I was updating our internal version of Bazel to 0.16  these errors showed up with `--all_incompatible_changes` enabled, so I fixed them.

ittaiz:
@jjudd are you sure? this bazel [issue](https://github.com/bazelbuild/bazel/issues/4316) seems to say that you need this setting if the attribute is called `data`. You also link to it in your commit to rules_scala_annex where you disable the check.

ittaiz:
This currently breaks because our Travis runs with an older Bazel and I think this restriction Yoon was lifted in 0.15 IINM

jjudd:
AFAIK cfg = "data" is now a no-op AFAIK. I found that Bazel issue, but didn't realize it was out of date. I later went back to annex and added this commit https://github.com/lucidsoftware/rules_scala_annex/commit/39bb7818f67d61f94dfc159175f44674f4b59c18 which deals with this properly.

Regarding 0.15 - are we good to update to 0.16 on Travis? I'm not sure what Bazel versions y'all support.

ittaiz:
@jjudd can you please rebase on top of master?
That will get this to green

jjudd:
This has been updated to master and builds are now passing.

PR #585, opened 2018-08-16
remove unused-deps attr from scala_macro

jhnj:
Set `unused_dependency_checker_mode = "off"` for all `scala_macro` rules.

As discussed internally with @johnynek, any dependency can be used by macro targets at compile time. This PR removes the ability to set `unused_dependency_checker_mode` for scala_macro rules. Also updates the README to match this.

Another option would be to set `unused_dependency_checker_mode` to default to `"off"` for scala_macro_library

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,21 +91,15 @@ for an example workspace using another scala version.
 [scala]: http://www.scala-lang.org/
 
 <a name="scala_library"></a>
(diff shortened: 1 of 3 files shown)


jhnj:
I think I'm just going to set unused_deps to off by default instead of disabling it completely for `scala_macro_library`. I'll create a new PR with that change

PR #586, opened 2018-08-16
Update bazel to 0.16.1

ittaiz:


ittaiz:
Needed by #583 cc @jjudd

jjudd:
Awesome. Thanks for upgrading this.

ittaiz:
Sure thing. It even seems it will pass. Unfortunately we needed to disable
our own strict checks for bazel issues due to much noise

johnynek:
Can you update the shas in the README.md to help people see which shas work with which versions?

ittaiz:
Not sure I understand which shas you’re referring to

johnynek:
https://github.com/bazelbuild/rules_scala#bazel-compatible-versions

I think we should set the 14.1 version to the current sha. If we aren’t testing it any longer we aren’t sure it works.

ittaiz:
@johnynek PTAL

ittaiz:
Oscar,
I’m planning of merging tomorrow morning if I don’t hear any objections now
that I’ve updated the readme.

PR #588, opened 2018-08-16
IntelliJ plugin support for -sources.jar from scala_import

ceason:
Propagating the source jars through to `source_jars` seems to get the job done; works with bazel-deps, raw jars and jars via filegroup. A [cursory glance](https://github.com/bazelbuild/intellij/blob/master/aspect/intellij_info_impl.bzl#L89-L94) at the intellij plugin indicates that `source_jars` has precedence over `source_jar`, so the latter can be left unpopulated.

This is a bit of "build tool+IDE" polish that will be nice to have back :)


ceason:
@johnynek updated, passes linter now.

PR #590, opened 2018-08-19
changed explicit references of external workspaces to bind

ittaiz:
@johnynek @jhnj I'm trying to upgrade to the latest rules_scala and am encountering some difficulties.
We are using our own custom versions of dependencies and so don't call `scala_repositories`. 
What  we've been doing up until now is `bind` to the required points by rules_scala.

After applying the below changes the project compiles. 
I'm not certain they should be merged into master since my gut tells me we're trying to phase out the `bind`s in favor of the `declare_scalac_provider` but I'm not sure we're there.
1. Do we want to phase out `bind`s in favor of custom scalac_provider/scala_provider?
2. How can one declare their own scalac_provider? Currently it's a private attribute
3. Sounds like `//src/java/io/bazel/rulesscala/scalac` needs to change one way or another. Use `bind`s or  somehow depend on the `scalac_provider` and take from it the `default_repl_classpath` (need to decide what to do about commons_io).

Would really appreciate your help on this,
Ittai

### DIFF
--- a/scala/BUILD
+++ b/scala/BUILD
@@ -44,17 +44,17 @@ java_import(
 _declare_scalac_provider(
     name = "scala_default",
     default_classpath = [
-        "@io_bazel_rules_scala_scala_library",
-        "@io_bazel_rules_scala_scala_reflect",
+        "//external:io_bazel_rules_scala/dependency/scala/scala_library",
+        "//external:io_bazel_rules_scala/dependency/scala/scala_reflect",
     ],
     default_macro_classpath = [
-        "@io_bazel_rules_scala_scala_library",
-        "@io_bazel_rules_scala_scala_reflect",
+        "//external:io_bazel_rules_scala/dependency/scala/scala_library",
+        "//external:io_bazel_rules_scala/dependency/scala/scala_reflect",
     ],
     default_repl_classpath = [
-        "@io_bazel_rules_scala_scala_library",
-        "@io_bazel_rules_scala_scala_reflect",
-        "@io_bazel_rules_scala_scala_compiler",
+        "//external:io_bazel_rules_scala/dependency/scala/scala_library",
+        "//external:io_bazel_rules_scala/dependency/scala/scala_reflect",
+        "//external:io_bazel_rules_scala/dependency/scala/scala_compiler",
     ],
     scalac = "@io_bazel_rules_scala//src/java/io/bazel/rulesscala/scalac",
     visibility = ["//visibility:public"],
--- a/src/java/io/bazel/rulesscala/scalac/BUILD
+++ b/src/java/io/bazel/rulesscala/scalac/BUILD
@@ -13,10 +13,10 @@ java_binary(
         "@io_bazel_rules_scala//src/java/com/google/devtools/build/lib:worker",
         "@io_bazel_rules_scala//src/java/io/bazel/rulesscala/jar",
         "@io_bazel_rules_scala//src/java/io/bazel/rulesscala/worker",
-        "@io_bazel_rules_scala_scala_compiler",
-        "@io_bazel_rules_scala_scala_library",
-        "@io_bazel_rules_scala_scala_reflect",
-        "@scalac_rules_commons_io//jar",
+        "//external:io_bazel_rules_scala/dependency/scala/scala_library",
+        "//external:io_bazel_rules_scala/dependency/scala/scala_reflect",
+        "//external:io_bazel_rules_scala/dependency/scala/scala_compiler",
+        "//external:io_bazel_rules_scala/dependency/commons_io/commons_io",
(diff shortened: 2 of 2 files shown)


jhnj:
Moving out today so just my quick thoughts:

I think doing this is fine for now. We rely on binds to use custom versions of e.g. scrooge as well so don't think we can phase out binds completely anyways (even if we'd want to).

`declare_scalac_provider` is usable from https://github.com/bazelbuild/rules_scala/blob/master/scala/providers.bzl#L47 but the scala rules use the the provider defined here: https://github.com/bazelbuild/rules_scala/blob/master/scala/BUILD#L44 by default and currently that is not configurable. You could change the `_scalac_provider` attr that all the scala rules use to be public if you want to define your own provider.

`//src/java/io/bazel/rulesscala/scalac` can't really depend on the ScalacProvider (as it's part of it). If you want to define your own ScalacProvider you can create your own `//src/java/io/bazel/rulesscala/scalac` (BUILD file in your workspace and symlinking or copying the sources from rules_scala, using e.g. https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#template )

jhnj:
I could create an example of what defining a own ScalacProvider would look like if you want to (Although would probably be end of next week when I'm back in Finland)

ittaiz:
@johnynek,
Tests are passing and I need this in. Any concerns?

johnynek:
Sorry @ittaiz I didn’t look till now.

So can I instead make a PR to the readme later today that shows you how to set things up with the scala provider? We also customize but do it using the provider. If we merge this we basically revert most of the scala provider but leave the mechanism around.

ittaiz:
That was my initial hunch.
This PR is to show what we needed to do.
A PR to the readme sounds great.
I’ll probably try it out on Sunday (AFK for a couple of days) and if it
works then we’ll close this one.

johnynek:
yeah. I think there are a couple things:

1. we should make the scala_provider public so you can configure in your own repo.
2. the question of how we connect the external jars to names the rules can use. We used to use bind, but we changed as part of the scala 2.12 support to use workspace names. You can customize by using the same names as this function: https://github.com/bazelbuild/rules_scala/blob/master/scala/scala.bzl#L278

If we revert back, you will break us. I don't want to flap on this, so we should decide how to move forward?

I thought that we wanted to move away from bind since we have heard so much disparaging about bind, but maybe not actual deprecation.

Can you change your code to use the workspace names that the rules expect now?

cc @andyscott @beala-stripe @travisbrown

ittaiz:
On Thu, 23 Aug 2018 at 3:29 P. Oscar Boykin <notifications@github.com>
wrote:

> yeah. I think there are a couple things:
>
>    1. we should make the scala_provider public so you can configure in
>    your own repo.
>    2. the question of how we connect the external jars to names the rules
>    can use. We used to use bind, but we changed as part of the scala 2.12
>    support to use workspace names. You can customize by using the same names
>    as this function:
>    https://github.com/bazelbuild/rules_scala/blob/master/scala/scala.bzl#L278
>
> If we revert back, you will break us. I don't want to flap on this, so we
> should decide how to move forward?
>
I definitely agree. I think today we’re half way there. How are you
(stripe) currently working around the fact that the provider is hidden?

> I thought that we wanted to move away from bind since we have heard so
> much disparaging about bind, but maybe not actual deprecation.
>
> Can you change your code to use the workspace names that the rules expect
> now?
>
> Dont know. I guess I can but I need to know what needs to be done to give
a concrete answer.
Are these the only steps needed? I think the PR to the documentation you
suggested can clarify a lot of the question marks.

ittaiz:
@johnynek any news?
How are you (stripe) working around the fact that the provider is hidden?
Also it sounds like the correct thing to do would be to somehow get `scalac` target to depend on the provider and have the provider only contain the jars. Need to mull it over.

ittaiz:
Also, there are many references in the code for `//external:io_bazel_rules_scala*` (92 matches!).
It feels like this isn't fully baked.
I can push a direction but it would probably be better that we agree on it before.

johnynek:
So, I was misinformed. We are only only configuring the maven server we fetch from.

I think depending all the way on the provider, is the right goal. Incidentally, can you just copy-paste the scala_repositories function and set up these external repos as you like in the mean time?

Should we open an issue on the two issues separately:

1. a single place to configure which jars to use (scalacprovider usage any time we are looking at the names of dependencies for the scala compiler)?
2. easier customization of scalac provider (e.g. making the attribute public on the scala rules?)

I think number 2 is easy, but maybe does not solve things until 1 is done.

Can we try to sort out what is blocking for people?

johnynek:
was this superceded by #605

ittaiz:
Yes

PR #605, opened 2018-08-31
Allow using a completely different scala version without bind 

ittaiz:
Main change:
Getting `ScalacProvider` via toolchain and not via attribute.
`scalac` attribute is no longer part of the above provider. In addition it gets the dependencies that it needs from the toolchain as well via a `export_scalac_repositories_from_toolchain_to_jvm` rule which does exactly what the name says.
Open issue:
Not sure I like the mismatch between "scalac provider" and "scalac repositories provider". I introduced it but I'm still deliberating. Would appreciate feedback about it (can use both like now or only one of them).

To use this internally we needed to:
1. Declare commons_io with the rules_scala name (can be improved but probably not critical since it changes very rarely and we use the same version so no classpath collision).
2. Declare a custom toolchain and a `declare_scalac_provider` target with the repositories our org defines.

@katre can you take a look at `export_scalac_repositories_from_toolchain_to_jvm`? This seems like a really big hammer to replace `bind` but I'm not sure I have an alternative.
@iirina I wanted to use `JavaInfo` but didn't remember how to do it (no output_jar). Any chance you can comment with a snippet of how I should have done it? Probably something with merge but I wasn't sure (also don't remember if fixed merge was released in 0.16.x or not).

cc @jhnj

### DIFF
--- a/scala/scala_toolchain.bzl
+++ b/scala/scala_toolchain.bzl
@@ -5,7 +5,7 @@ load(
 def _scala_toolchain_impl(ctx):
   toolchain = platform_common.ToolchainInfo(
       scalacopts = ctx.attr.scalacopts,
-      scalac_repositories = ctx.attr.scalac_repositories,
+      scalac_provider_attr = ctx.attr.scalac_provider_attr,
       unused_dependency_checker_mode = ctx.attr.unused_dependency_checker_mode)
   return [toolchain]
 
@@ -13,8 +13,8 @@ scala_toolchain = rule(
     _scala_toolchain_impl,
     attrs = {
         'scalacopts': attr.string_list(),
-        'scalac_repositories': attr.label(
-            default = "@io_bazel_rules_scala//scala:scala_repositories_default",
+        'scalac_provider_attr': attr.label(
+            default = "@io_bazel_rules_scala//scala:scalac_default",
             providers = [_ScalacProvider]),
         'unused_dependency_checker_mode': attr.string(
             default = "off", values = ["off", "warn", "error"]),
--- a/scala/BUILD
+++ b/scala/BUILD
@@ -42,7 +42,7 @@ java_import(
 )
 
 _declare_scalac_provider(
-    name = "scala_repositories_default",
+    name = "scalac_default",
     default_classpath = [
         "@io_bazel_rules_scala_scala_library",
         "@io_bazel_rules_scala_scala_reflect",
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -310,7 +310,7 @@ def _interim_java_provider_for_java_compilation(scala_output):
 
 def _scalac_provider(ctx):
   return ctx.toolchains[
-      '@io_bazel_rules_scala//scala:toolchain_type'].scalac_repositories[
+      '@io_bazel_rules_scala//scala:toolchain_type'].scalac_provider_attr[
           _ScalacProvider]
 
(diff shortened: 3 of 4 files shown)


johnynek:
Sorry for the latency. Will review in the next 24 hours

scala/providers.bzl
```diff
@@ -26,9 +26,8 @@ def create_scala_provider(ijar, class_jar, compile_jars,
   )
 
 ScalacProvider = provider(
-    doc = "ScalaProvider",
+    doc = "ScalaCProvider",
```
johnynek: We use ScalacProviser for the provider but capitalize the C here.

scala/providers.bzl
```diff
@@ -47,11 +45,6 @@ def _declare_scalac_provider(ctx):
 declare_scalac_provider = rule(
     implementation = _declare_scalac_provider,
     attrs = {
-        "scalac": attr.label(
-            executable = True,
```
johnynek: Why did you remove scalac from this?

scala/scala_toolchain.bzl
```diff
@@ -1,13 +1,21 @@
...
+      scalac_repositories = ctx.attr.scalac_repositories,
       unused_dependency_checker_mode = ctx.attr.unused_dependency_checker_mode)
   return [toolchain]
 
 scala_toolchain = rule(
     _scala_toolchain_impl,
     attrs = {
         'scalacopts': attr.string_list(),
+        'scalac_repositories': attr.label(
+            default = "@io_bazel_rules_scala//scala:scala_repositories_default",
```
johnynek: Interesting... if we are back to putting the repositories into the toolchain, which I actually like, since it seems closer to what bazel devs expect, can we not remove it entirely and put all the things the provider give directly on the toolchain?

scala/providers.bzl
ittaiz: Indeed. In general I mix scalac provider and scalac repositories provider.  Do you think that is ok?

ittaiz: done

ittaiz:
@johnynek re `scalac_provider` vs `scalac_repositories_provider` and such I made a call.
I made it `scalac_provider` across the board and in the toolchain it's `scalac_provider_attr` to signal it's only there for the provider it declares.
I think the current layout is really good in the sense that each provider we want to add (specs/scalatest/proto/etc) can be added to the toolchain in the same pattern.
wdyt?

ittaiz:
Thanks Johan!
Are you familiar with @katre’s work on allowing multiple toolchains to be
used in the same build? I think that will enable cross compilation which
you refer to and that will land sometime in the next year (which sounds
good to me).
Additionally having the customization venue be macros sounds like an anti
pattern is Bazel so I’d rather avoid it.
One last point is what happens if I depend on something in a different repo
which doesn’t use my macro? I’ll have a mixup.

Re providers- good question. If that will be the case we’ll have multiple
toolchains.

ittaiz:
@johnynek I think we’re ready to merge. Only change which is potential is renaming default_repl_classpath so that it will indicate it’s also for the actual compiler classpath.
If you have an idea tell me otherwise I’d like us to proceed

ittaiz:
Thanks! I've verified and indeed the few repos we have using bazel on master are now running with this and green.
@katre I'd still appreciate your feedback on this "trick" of exposing a toolchain value to a java_binary
@iirina I'd still appreciate your feedback on the java_common/JavaInfo

@johnynek please don't delete the branch just yet, thanks!

PR #611, opened 2018-09-12
VSCO adopts rules_scala

promiseofcake:
Thanks for all of the hard work, [VSCO](https://vsco.co) uses rules_scala in production.


johnynek:
Thanks so much for the PR!

PR #613, opened 2018-09-18
pending test: import-only deps are not reported as used

gkossakowski:
this is a known limitation of the classpath-shrinker approach. the
underlying reason is that imports are lazily resolved upon first access
for a simple identifier (e.g. ArrayUtils in the test case)

### DIFF
--- a/third_party/unused_dependency_checker/src/test/io/bazel/rulesscala/dependencyanalyzer/UnusedDependencyCheckerTest.scala
+++ b/third_party/unused_dependency_checker/src/test/io/bazel/rulesscala/dependencyanalyzer/UnusedDependencyCheckerTest.scala
@@ -59,4 +59,19 @@ class UnusedDependencyCheckerTest extends FunSuite {
     val errorMessages = compileWithUnusedDependencyChecker(testCode, withDirect = direct)
     assert(errorMessages.isEmpty)
   }
+
+  test("do not error on used direct import-only dependency") {
+    val testCode =
+      """object Foo {
+        |  import org.apache.commons.lang3.ArrayUtils
+        |}
+      """.stripMargin
+
+    val commonsTarget = "commonsTarget"
+
+    val direct = List(apacheCommonsClasspath -> commonsTarget)
+
+    val errorMessages = compileWithUnusedDependencyChecker(testCode, withDirect = direct)
+    assert(errorMessages.isEmpty)
+  }
 }


johnynek:
good to see you back here!

travisbrown:
Thanks @gkossakowski! As I noted in the internal issue, though, we're seeing many false positives from the checker that aren't in this category.

johnynek:
Is this kind a problem? I mean with unused imports warning and fatal warnings this is already an error.

I think it would be nice to fix, but I think in some ways this is an unused dependency.

gkossakowski:
Yes, it's not high priority issue but I dropped it for our shared understanding.
AFAIR, 2.11 had bugs in unused imports detection but i can't find the ticket. It works reliably in 2.12+.

ittaiz:
@Jamie5 do we support this use-case? If so do we have a test for it?
If the answer to both is yes I'd be happy to close this PR

Jamie5:
For high-level: there are no tests for this AFAIK. IIRC this definitely was a "problem" in 2.11 (as much as an error pointing out something you shouldn't be doing anyways is a problem), seemed to recall it being a problem in 2.12 as well but don't recall.

For AST mode, it does look like we do detect imports as being used generally speaking (but did not check for all versions, just the default 2.11.x, and only checked a basic import at package level), but don't have a test for it.

ittaiz:
@Jamie5 I’d love a test for this in AST mode if you get around to it.
@gkossakowski I’m closing this PR for staleness.
Feel free to rebase, add a test and reopen.

Jamie5:
@ittaiz ok was wrong, I tested wrong. My test was `object A { val i: Int = 0 }` and then doing `import A.i`. (This was because tests didn't seem to support packages, actually it is fairly easy to do packages but didn't think of it at the time)

That worked. But `package foo ; object A` and `import foo.A` does not work.

The reason being that an import is represented as `Import(expr, selectors)` and `expr` is all the parts of the import except the last part, so in `import foo.bar.{a, b}` expr would be `foo.bar` and selectors would be `[a, b]` roughly.

So in the first case expr=`A` and yay we look at `A` and it works. While in the second case expr=`foo` which is not a class.

This is almost certainly fixable as the compiler itself has to do these sorts of lookups. I wonder if it is worth it though (the fix may be complex).

For one if someone did `import foo.bar` and it was an unused import, not much we can do about that. Because even if someone had `foo.bar.A` and `foo.bar.B` as deps, how do we pick which one (assuming both A and B are not used) is the strict one and which one isn't, in a reasonable manner?

So not sure how worthwhile it is to handle the case of `import foo.A` and would not want to wade into the case of `import foo.bar`, but will await orders.

Can still add a test for the `import A.i` case if we want.

ittaiz:
1. I think it's worthwhile to also add a test even for things we don't support to document the current behavior with a comment saying we can consider to change this.
Mainly because I'd love for us to link to the tests from the readme and say that this is what we know that works and what we know that doesn't work right now.

2. Concretely- I'm not sure why you say "how do we pick which one... is the strict one and which one isn't"? The discussion here is around unused-deps since what happens is that one applies the unused-deps fix and then their build breaks and they're blocked (unless they remove the import or add a local ignore). This specific case of unused might be one that we can live with but to me this is another symptom of the different nuances between strict-deps and unused-deps.

WDYT?

Jamie5:
1. Ok am fine with having tests to reflect current behavior so will submit something.

2. In this case unused feels very much the opposite of strict. From my POV both A and B are unused (and hence not strict) because you see no usages of any types or members of A, and the same for B. And the only reason you have to keep one of them marked as not unused/is strict is that we have this random import statement. And unlike the other case we were discussing about A->B->C->D subclassing, there is no single correct set of jars to import when using direct deps. As a side note I do wonder which people would turn on unused_deps_checker but not unused_imports_checker.

ittaiz:
I think we're using unused and strict in multiple and sometimes different terms.

1. unused-deps: refers to Bazel dependencies that can be removed without breaking the build of the current target.
2. strict-deps: refers to Bazel dependencies that are needed for compilation to pass, are needed by the current source code (an import statement IMHO qualifies since the dev explicitly typed it in) but are not directly declared by the current target.
3. unused imports: imports which are not needed for the current code to compile.

IMHO when we put it like this then A is 3 and can be an infraction of 2 if the current target doesn't declare it. It's not 1 in any case.

I think we should agree on the above (or continue the discussion) and afterwards we can decide if indeed we want to solve it (I think so) or willing to not support it given the cost to the users should be small (not sure about that).

`unused_imports_checker`- This isn't a thing, right? Do you mean using scalafix? I'm trying to think how we (Wix) can migrate to these capabilities. I can't ask everyone to run remove unused imports because asking doesn't scale, we don't have scalafix on bazel yet and currently we won't support unused-deps as is.
I looked into scalafmt but it doesn't appear to support removing unused imports.

Would appreciate your thoughts on both counts.

Jamie5:
Ok by those definitions then yes agreed that `import foo.bar.A` violates 2 and 3. But this case works fine under the existing unused-deps/strict-deps paradigm: `A` is not an unused dep because if it is removed, the build breaks. `A` is a strict dep, because it is referenced in the current source code. The only issue with that case is that we do not return `A` in `findUsedJars` even though we technically should. (maybe we should refer to this as case 1 to contrast with the following, which we can call case 2).

(case 2) In the case of `import foo.bar` where we have different bazel packages with `foo.bar.A` and `foo.bar.B`, then what happens? If we examine A, it is an unused-dep because removing it will not break the build. It is also not a strict dep because it is not mentioned in the code. We can do the same analysis for B. However we can't remove both A and B or the compile will fail. So not quite sure how this fits into 1/2/3.

We just use `-Ywarn-unused-import` and `-Xfatal-warnings`. Which isn't automatic but is not really bothersome.

PR #620, opened 2018-09-27
Change scala_import rule neverlink attribute from int to bool

anchlovi:
Currently `scala_maven_import_external` declares `neverlink` as `bool` while `scala_import` expects `int`


ittaiz:
Important to say it’s the underlying *shared* jvm import external which declares it as bool

PR #622, opened 2018-10-01
Forward kwargs in scala_library_suite and scala_test_suite

### LINKED ISSUE #621: deprecated `jvm_flags` not reflected in `scala_library_suite` or `scala_test_suite`
The README marks the `jvm_flags` attribute to `scala_library` and `scala_test` as deprecated. But, the corresponding `scala_library_suite` and `scala_test_suite` macros don't reflect this deprecation, as they only accept the `jvm_flags` attribute.

aherrmann:
closes #621


aherrmann:
@johnynek I'm afraid I don't understand the Travis failure. Looks like the linting stages are failing. But, it looks like it's complaining about code that this PR isn't touching.

ittaiz:
I think the listing is broken :(
@johnynek,
Am I correct?

johnynek:
I think yeah, we should disable linting until we have it fixed.

ittaiz:
Ok, this seems to break remote execution. I submitted #634 to make sure RBE still passes without this change and it indeed passes.
Can you take a look at the build log? I’m AFK so have limited ability to help with the troubleshooting but if you’ll post specific questions I’ll try to answer ASAP

aherrmann:
I've updated this PR as suggested here: https://github.com/bazelbuild/rules_scala/pull/627#issuecomment-432085769

ittaiz:
Thanks!

johnynek:
thank you!

PR #623, opened 2018-10-02
Add source jar support to Maven import external

anchlovi:
cc @ittaiz

PR #625, opened 2018-10-03
Better way to fetch sources

anchlovi:
This is a better approach to fetch sources following: #623. This also includes a usage example for `scala_maven_import_external` (which fails since `scala_import` has no `srcjar` attribute)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -71,6 +71,7 @@ scala_maven_import_external(
     artifact = "com.google.guava:guava:21.0",
     jar_sha256 = "972139718abc8a4893fa78cba8cf7b2c903f35c97aaf44fa3031b0669948b480",
     fetch_sources = True,
+    srcjar_sha256 = "b186965c9af0a714632fe49b33378c9670f8f074797ab466f49a67e918e116ea",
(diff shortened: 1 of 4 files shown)


anchlovi:
Currently tests will fail with:
```
/private/var/tmp/_bazel_shacharan/87128639a71862f9745ebfe165e3dfdb/external/com_google_guava_guava_21_0_external/BUILD:9:1: @com_google_guava_guava_21_0_external//:com_google_guava_guava_21_0_external: no such attribute 'srcjar' in 'scala_import' rule
```

anchlovi:
Current status: 

I added the `srcjar` to intellij metadata but for some reason intellij is not aware that there is source jar attached

johnynek:
Are you manually linting? The script should be able to fix things up for you. I don’t remember the invocation (I’m on a phone now) but I thought we noted it somewhere in the README.

That said, I also had trouble with the script recently.

anchlovi:
I'm using bazel 0.17.2. The linter doesn't work with this version.
Eventually I gave up and installed also 0.16.1 :-/

PR #626, opened 2018-10-08
Populate the correct classifier when decoding maven artifact coordinates

anchlovi:


WORKSPACE
```diff
@@ -67,15 +67,15 @@ maven_jar(
 
 # test of import external
 scala_maven_import_external(
-    name = "com_google_guava_guava_21_0_external",
-    artifact = "com.google.guava:guava:21.0",
+    name = "com_github_jnr_jffi_native",
+    artifact = "com.github.jnr:jffi:jar:native:1.2.17",
```
ittaiz: can you add a comment saying why the classifier/packaging/both are important? it's unclear from here and someone might change it

PR #627, opened 2018-10-16
Forward the data attribute in scala_test_suite

aherrmann:
The attribute was previously accepted in the scala_test_suite macro but not passed on to the scala_test rule. However, this attribute is imported for example in order to provide runfiles.


johnynek:
Would you be willing to add a test so we don’t accidentally revert this later?

aherrmann:
@johnynek I've added a small test-case that is similar to the use-case that triggered this issue.

ittaiz:
Sorry for the delay. We’ll merge when green

ittaiz:
Sorry, now that I’m trying to catch up. Doesn’t #622 solve it as well in a more generic way with kwargs?

aherrmann:
@ittaiz 
> Sorry, now that I’m trying to catch up. Doesn’t #622 solve it as well in a more generic way with kwargs?

It did not initially, as `data` was already an explicit keyword argument, and just silently ignored. However, the explicitly defined keyword arguments in the `*_suite` macros don't change the default values defined in the corresponding non-suite rules. I've modified #622 to just eliminate the unnecessarily explicitly defined keyword arguments.

This makes this PR redundant.

PR #629, opened 2018-10-17
Add full_jars to default outputs of several rules

exchgr:
- `scala_library`
- `scala_library_for_plugin_bootstrapping_impl`
- `scala_macro_library_impl`

This allows downstream dependents of these rules to access both the scala and java jars for mixed scala/java builds.


exchgr:
Working on getting the CLA for Meetup working correctly with my github

johnynek:
Happy to merge this after we add a test.

See https://github.com/bazelbuild/rules_scala/blob/master/test_rules_scala.sh

for how we are testing things similar to this now.

I think we want to make a mixed target, build the target and see that the java jar exists.

exchgr:
Sorry, my mistake. Going to reopen this when I restore these commits

ittaiz:
I’m not sure an e2e is the right approach since the jars are created. Their problem is with the outputs being propagated.
Two options:
1. Have a test with a dummy rule that echoes the artifacts and assert on that. You can find examples in our codebase or ask here and I’ll send a few links. This is preferable since it demonstrates your need.
2. Have a test with a dummy aspect which asserts about the property as well. Main con is that to me this feels more white box (not 100% clear why the aspect does this).

exchgr:
@ittaiz Sorry for the late response, just getting back around to this. I'd greatly appreciate links to some of those examples! Thank you!

exchgr:
Added a test, not sure if it'll pass in CI just yet, but the one test file I added passes on my machine.

ittaiz:
Sorry for not replying earlier, I’ll try to send the links later today.
I think the test doesn’t help us since it’s irrelevant to the bug (missing info in providers) and I’m fairly certain it already passes without your change. Can you please try without your change?

exchgr:
Ah yes, you're right about that. I'll change it today.

test_scala_library_outputs_mixed_java_scala_jars.sh
```diff
@@ -0,0 +1,36 @@
...
+  grep '{"id":{"targetCompleted' $1 | grep $2 > /dev/null
+}
+
+test_scala_library_outputs_mixed_java_scala_jars() {
+  set -e
+
+  TARGET="MixJavaScalaLib"
+  BUILD_EVENT_FILE="build_event.json"
+
+  bazel build test:"$TARGET" --build_event_json_file=$BUILD_EVENT_FILE
```
exchgr: @ittaiz this was the only way I could find to get the outputs from the command line. If you know of a better way, I'd love to hear about it.

ittaiz:
ok,
Are you sure the new test failed? I can see why but not 100% sure (Which is also why I'm thinking maybe it's best to replace it).
Writing a custom rule:
Basically you add a bzl which depends on a scala_library (which has a scala and a java input) and then the rule creates a fixed file with the names of the files which it takes from the default output provider.
the assertion can be either in a scala_test target which depends on the above target and asserts the contents of the file in scala or probably a genrule.
So: scala_library with java/scala sources is the `setup`, a target with your custom rule is the `act` and you have a third target for `assert`. This might sound complicated but is really straightforward.
https://github.com/bazelbuild/rules_scala/blob/44438c6205f5c4a9bf5b9b762715b88530356ea9/test_expect_failure/missing_direct_deps/internal_deps/custom-jvm-rule.bzl
Using aspects:
You can look at the link below and the area around it. I think it's a bit implicit and usually takes me a few reads to remember what's going on but maybe it's a matter of taste. If this flavor is more understandable and easier to you I'm ok with it. 
https://github.com/bazelbuild/rules_scala/blob/44438c6205f5c4a9bf5b9b762715b88530356ea9/test/aspect/aspect.bzl

Again sorry for the late reply

exchgr:
@ittaiz Thanks for the explanation. I just pushed up a revised test that I think is more along the lines of what you're looking for.

test/mix_java_scala_lib.bzl
```diff
@@ -0,0 +1,11 @@
+def _mix_java_scala_lib_impl(ctx):
+    paths = [f.path for f in ctx.files.deps]
```
ittaiz: Is this what you’re using internally which fails?

test/mix_java_scala_lib.bzl
```diff
@@ -0,0 +1,11 @@
...
+    for path in paths:
+        print(path)
+
+mix_java_scala_lib_rule = rule(
```
ittaiz: Wdyt about mix_java_scala_lib_test_rulr? Or maybe dummy_rule?
Can live with the current one but it surprised me a bit when I first saw it in the call site

test/mix_java_scala_lib.bzl
```diff
@@ -0,0 +1,11 @@
+def _mix_java_scala_lib_impl(ctx):
+    paths = [f.path for f in ctx.files.deps]
+    for path in paths:
+        print(path)
```
ittaiz: I think having an output file with this content is cleaner than using print but it also does the work so leaving it up to you (we’ve had quite a few iterations on this change unfortunately)

test/mix_java_scala_lib.bzl
exchgr: Yeah, this is what we're using in our rules that failed before this change. This maps to the `ctx.outputs` of the `scala_library` rule on which this rule depends:

> In other words, use files to access the default outputs of a dependency.

-- https://docs.bazel.build/versions/master/skylark/lib/ctx.html#files

test/mix_java_scala_lib.bzl
exchgr: I'm alright with changing this name. It was kind of arbitrary

test/mix_java_scala_lib.bzl
exchgr: If I have an output file from here, how do I test that it contains the right jars? And if I do that, why not test the `scala_library` `MixJavaScalaLib`'s outputs directly?

exchgr:
I saw that one of the tests failed before the latest changes, but I didn't really dig into it too deeply since I ended up replacing the change that made it fail anyway, per your suggestion.

test/mix_java_scala_lib.bzl
ittaiz: Two ways:
1. Have a scala_test target that depends on this output as data and asserts its contents (pro- no bash, incremental bazel run)
2. Have your current e2e assert against the content of the file (pro- feels less brittle than asserting on debug prints).
In any case I'm ok with the current implementation. I think it can be better but I don't feel strongly about it. Your call.

ittaiz:
@exchgr let me know if you want to make any changes or you'd rather this be merged as is

test/mix_java_scala_lib.bzl
exchgr: @ittaiz Now I understand what you're saying. Option 1 seems like a good path forward, I'll make the change. Thanks!

exchgr:
@ittaiz I've taken a bunch of your suggestions and mashed them together to form some kind of reasonable test that doesn't rely on a `DEBUG` to stdout. Let me know what you think. I'm alright with merging this if you are.

test_rules_scala.sh
```diff
@@ -883,3 +887,4 @@ $runner bazel build //test_expect_failure/proto_source_root/... --strict_proto_d
 $runner scala_binary_jar_is_exposed_in_build_event_protocol
 $runner scala_test_jar_is_exposed_in_build_event_protocol
 $runner scala_junit_test_jar_is_exposed_in_build_event_protocol
+$runner test_mix_java_scala_lib
```
johnynek: I don’t think we need this. Your test is already a bazel test. This script is for the tests we can’t easily make normal bazel tests.

test/test_scala_library_outputs_mixed_java_scala_jars.sh
```diff
@@ -0,0 +1,20 @@
...
+  $@ || (echo "FAILED: $@"; exit 1)
+}
+
+contains() {
+  grep $@
+}
+
+set -e
+
+TARGET="MixJavaScalaLib"
```
ittaiz: Really small comment: can this also move to parameter? Symmetry wise it feels like it should really sit in the sh_test

test_rules_scala.sh
exchgr: Ah nice, I didn't see the call to `bazel test //...` at first. I'll get rid of this.

test/test_scala_library_outputs_mixed_java_scala_jars.sh
exchgr: Done!

ittaiz:
Thank you!

exchgr:
:D Thanks for your patience through all these changes!

PR #633, opened 2018-10-21
[WIP] Producing a source jar for build targets

hmemcpy:
This adds building a `-src.jar` in addition to the target jar, so that IntelliJ supports navigating to sources.

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -841,10 +841,10 @@ def _pack_source_jars(ctx):
       f for f in ctx.files.srcs if f.basename.endswith(_java_extension)
   ]
   if len(java_sources) > 0:
-      pseudo_java_jara_file = ctx.actions.declare_file(ctx.label.name + "_java.jar")
+      pseudo_java_jar_file = ctx.actions.declare_file(ctx.label.name + "_java.jar")
       java_source_jar = java_common.pack_sources(
          ctx.actions,
-         output_jar = pseudo_java_jara_file,
+         output_jar = pseudo_java_jar_file,
          sources = java_sources,
          java_toolchain = ctx.attr._java_toolchain,
          host_javabase = ctx.attr._host_javabase
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -708,7 +708,7 @@ def _lib(
         transitive = [transitive_rjars, exports_jars.transitive_runtime_jars],
     )
 
-    source_jar = _pack_sources(ctx)
+    source_jars = _pack_source_jars(ctx)
 
     scalaattr = create_scala_provider(
         ijar = outputs.ijar,
@@ -721,7 +721,7 @@ def _lib(
         deploy_jar = ctx.outputs.deploy_jar,
         full_jars = outputs.full_jars,
         statsfile = ctx.outputs.statsfile,
-        source_jar = source_jar,
+        source_jars = source_jars,
     )
 
     java_provider = create_java_provider(scalaattr, jars.transitive_compile_jars)
@@ -809,7 +809,7 @@ def _scala_binary_common(
         collect_data = True,
     )
 
-    source_jar = _pack_sources(ctx)
+    source_jars = _pack_source_jars(ctx)
 
     scalaattr = create_scala_provider(
         ijar = outputs.class_jar,  # we aren't using ijar here
@@ -819,7 +819,7 @@ def _scala_binary_common(
         deploy_jar = ctx.outputs.deploy_jar,
         full_jars = outputs.full_jars,
         statsfile = ctx.outputs.statsfile,
-        source_jar = source_jar,
+        source_jars = source_jars,
     )
 
     java_provider = create_java_provider(scalaattr, transitive_compile_time_jars)
@@ -833,22 +833,49 @@ def _scala_binary_common(
         runfiles = runfiles,
     )
 
-def _pack_sources(ctx):
-  java_srcs = [
+def _pack_source_jars(ctx):
+  source_jars = []
+
+  # collect .java sources and pack a source jar for Java
+  java_sources = [
       f for f in ctx.files.srcs if f.basename.endswith(_java_extension)
   ]
-  sources = [
-      f for f in ctx.files.srcs if f.basename.endswith(_scala_extension)
-  ] + java_srcs
+  if len(java_sources) > 0:
+      pseudo_java_jara_file = ctx.actions.declare_file(ctx.label.name + "_java.jar")
+      java_source_jar = java_common.pack_sources(
+         ctx.actions,
+         output_jar = pseudo_java_jara_file,
+         sources = java_sources,
+         java_toolchain = ctx.attr._java_toolchain,
+         host_javabase = ctx.attr._host_javabase
+      )
+      if java_source_jar:
+          source_jars.append(java_source_jar)
 
-  source_jar = java_common.pack_sources(
+
+  # collect .scala sources and pack a source jar for Scala
+  scala_sources = [
+      f for f in ctx.files.srcs
+      if f.basename.endswith(_scala_extension)
+  ]
+
+  # collect .srcjar files and pack them with the scala sources
+  bundled_source_jars = [
+      f for f in ctx.files.srcs
+      if f.basename.endswith(_srcjar_extension)
+  ]
+  scala_source_jar = java_common.pack_sources(
       ctx.actions,
       output_jar = ctx.outputs.jar,
-      sources = sources,
+      sources = scala_sources,
+      source_jars = bundled_source_jars,
       java_toolchain = ctx.attr._java_toolchain,
       host_javabase = ctx.attr._host_javabase
   )
-  return source_jar
+  if scala_source_jar:
+    source_jars.append(scala_source_jar)
+
+  return source_jars
 
 def scala_binary_impl(ctx):
     scalac_provider = _scalac_provider(ctx)
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -375,6 +375,7 @@ def try_to_compile_java_jar(
     return struct(
         jar = full_java_jar,
         ijar = provider.compile_jars.to_list().pop(),
+        source_jars = provider.source_jars
     )
 
 def collect_java_providers_of(deps):
@@ -406,6 +407,7 @@ def _compile_or_empty(
             java_jar = False,
             full_jars = [ctx.outputs.jar],
             ijars = [ctx.outputs.jar],
+            source_jars = [],
         )
     else:
         in_srcjars = [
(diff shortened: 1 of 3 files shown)


johnynek:
Looks good to me. Did you verify sources are available?

Is there a way we can test this?

johnynek:
Cc @andyjscott might be very nice for IntelliJ users at stripe.

scala/private/rule_impls.bzl
```diff
@@ -739,6 +745,23 @@ def _scala_binary_common(
       rjars,  #calling rules need this for the classpath in the launcher
       runfiles = runfiles)
 
+def _pack_sources(ctx):
+  java_srcs = [
```
ittaiz: @hmemcpy why do we need this? Can’t we just use ctx.files.srcs?

scala/providers.bzl
```diff
@@ -1,11 +1,11 @@
 # TODO: this should really be a bazel provider, but we are using old-style rule outputs
 # we need to document better what the intellij dependencies on this code actually are
 def create_scala_provider(ijar, class_jar, compile_jars,
-                          transitive_runtime_jars, deploy_jar, full_jars,
+                          transitive_runtime_jars, deploy_jar, full_jars, source_jar,
                           statsfile):
 
   formatted_for_intellij = [
-      struct(class_jar = jar, ijar = None, source_jar = None, source_jars = [])
+      struct(class_jar = jar, ijar = None, source_jar = None, source_jars = [source_jar])
```
ittaiz: Isn’t it better to pass it instead of the source_jar=None?

scala/private/rule_impls.bzl
hmemcpy: I tried, `ctx.files.srcs` holds a `list` of files, and that, for some reason, wasn't acceptable for `pack_sources`. But this may have been before I made it work, I'll try again to see if I can use it!

hmemcpy: I tried, `ctx.files.srcs` holds a `list` of files, and that, for some reason, wasn't acceptable for `pack_sources`. But this may have been before I made it work, I'll try again to see if I can use it!

hmemcpy: I tried, `ctx.files.srcs` holds a `list` of files, and that, for some reason, wasn't acceptable for `pack_sources`. But this may have been before I made it work, I'll try again to see if I can use it!

hmemcpy:
Sorry for the delays, I'm trying to add a few tests to this first.

andyscott:
Nice! I'll slot some time to test this with our work repos.

ittaiz:
Important to say @hmemcpy tested it manually and saw it works for bazel
source dependencies.
We’re also working on streamlining the 3rd party sources with scala import
external

johnynek:
any update here folks? Would be nice to get this merged.

hmemcpy:
Hi Oscar, sorry for the delays... we're on it, I'll give a proper update.

sha1n:
I fixed the merge conflicts and tested it again manually, this time with the intellij model source_jar fix (#649) and an intellij plugin fix that takes care of auto-source attachment for Scala (not yet merged into the plugin repo. See https://github.com/wix-playground/intellij/pull/4) and everything seems to be working as expected. Moreover, the intellij aspect tests for `scala_library` and `scala_binary` now fail, because they don't expect the source jars, which is great (I will fix them :) ). 

I have a few questions regarding how to proceed from here:
1. Since source jars are not really written to disk when running a build without the intellij aspect, how would you test the rule functionality? Does anyone here have any experience with similar scenarios and can give me a pointer?
2. Once we merge this PR, the tests I mentioned above will start failing. How do we usually role such multi-repo changes?

Thanks!

@ittaiz - FYI

sha1n:
The changes required in the intellij aspect tests are read for PR here: https://github.com/wix-playground/intellij/tree/adapting-to-rules_scala-src-jar-gen-change

johnynek:
Is there a way we can make a change to the intellij plugin that we can merge first that detects whether we are in the old version or new?

Strong coupling between releases usually creates a bad user experience.

Thanks for continuing to work on this!

sha1n:
@johnynek, thanks for your response! 
The plugin repo depends on a specific commit hash of `rules_scala`, so if we merge the PR I linked to above, technically, it will not break, but that's a bit nasty I guess, especially given that we can't guarantee that the changes are merged very close to each other (@ittaiz if we can do that, it will simplify the process, so if you think you can help here..).. 
If I get you right, you want the tests to behave differently based on the version of `rules_scala`. Well, that's possible for sure, probably in a hacky way, but that should be ok as a temp solution. However, we will have to discuss that with the maintainers of the plug I guess.

Thanks!

scala/private/rule_impls.bzl
ittaiz: @sha1n any chance you can take a look at this comment?
Additionally do we need two source jars? We generate a scala bytecode jar and a java one. Do we need two source jars (for intellij I mean)?
If not then why do we do these two iterations. Sounds like `ctx.files.srcs` should work. Also according to https://docs.bazel.build/versions/master/skylark/lib/ctx.html#files and https://docs.bazel.build/versions/master/skylark/depsets.html

sha1n: Sure! I missed that..

ittaiz:
@sha1n indeed many thanks!
1. Do you know what is the performance overhead of generating the sources? Is there any overhead given one doesn't explicitly ask for the source jars (like intellij does)?
2. `src_jars`- I think this has a bug since it doesn't pack sources which exist in `src_jars`. I don't think we (Wix) care about this use-case but should be easy to test and support once you have a few tests in place.
3.Java outputs. Does this work? It packs them in but will intellij work? Asking since I think intellij will look for `foo-java-src.jar` file but we will only generate `foo-src.jar` which correlates to the scala output. 
4. Is there any coupling between releases? To me this just sounds tests will break but users won't feel this. Am I mistaken?
5. Re your question how to test this-`bazel build //test/foo:bar-src.jar` will trigger the source jar generation of `//test/foo:bar` target

sha1n:
@ittaiz 

1. I couldn't measure any performance impact when running builds from CLI. The source jars are not written to disk in that case, they are only written when the intellij aspect is attached to the build. `pack_sources` Bazel API returns a `FileApi` object, that is a representation of a file that might or might not be read/written. I haven't looked into the implementation, but from the documentation I understand that it is an optimization, which seems to work pretty well in our case.
2. Thanks - I will look into it!
3. Yes. `rules_scala` includes a [test script](https://github.com/bazelbuild/rules_scala/blob/master/test_intellij_aspect.sh) that actually clones the plugin repo and runs the aspect tests. 
4. Yeasaaa! That looks like what I was looking for!

Thanks!

ittaiz:
Re performance that sounds great! Those Bazel devs have definitely thought
this build business through 😉

ittaiz:
@johnynek I’m ok with this coupling of tests. Wdyt?
@sha1n I think we should wait until after the holidays (Jan 6th?) and merge
given we have a PR waiting over there.

scala/private/rule_impls.bzl
sha1n: Yes, that seems to be working just fine. However, I'm trying to get libraries with mixed scala/java sources to work smoothly, so if I end up creating two separate source jars, I might keep the current approach.

sha1n:
So, we are still waiting for https://github.com/bazelbuild/intellij/pull/478 and https://github.com/wix-playground/intellij/pull/4 to be merged on the intellij side. One these are merged and this PR is merged we have:
1. Auto-source attachment when navigating to 3rd party classes
2. Same for second party (scala only at this point)
3. Same for proto generated scala classes

Next, I'm going to try and fix the navigation to Java sources. This is relevant to packages with mixed Java/Scala code.

scala/private/rule_impls.bzl
```diff
@@ -827,6 +833,50 @@ def _scala_binary_common(
...
+def _pack_source_jars(ctx):
+  source_jars = []
+
+  # collect .java sources and pack a source jar for Java
+  java_sources = [
+      f for f in ctx.files.srcs if f.basename.endswith(_java_extension)
+  ]
+  if len(java_sources) > 0:
+      pseudo_java_jar_file = ctx.actions.declare_file(ctx.label.name + "_java.jar")
+      java_source_jar = java_common.pack_sources(
```
ittaiz: Hmm, maybe we actually get this already from the java_common.compile?

scala_proto/scala_proto.bzl
```diff
@@ -542,6 +542,7 @@ def scalapb_proto_library(
 
     scala_library(
         name = name,
+        srcs = [srcjar],
```
ittaiz: Awesome!

scala/private/rule_impls.bzl
sha1n: This is not working yet BTW. Meaning, the java sources are not auto-attached like the Scala sources do. I think it's because of the jar root dir issue, but I haven't proved that yet.
What do you mean by get it from the java_common.compile? The Java sources jar? I don't think so. It does generate a separate `_java.jar` file, but not the sources. Do you mean that we can get that by configuring it differently? Does it have a flag for source gen or something?

ittaiz:
https://docs.bazel.build/versions/master/skylark/lib/java_common.html#compile
https://docs.bazel.build/versions/master/skylark/lib/JavaInfo.html#source_jars
I think we have all we need already but we need to propagate it out

ittaiz:
Strange. Do you mean the java code is packed correctly (without scala folder) and matching to bytecode but IntelliJ doesn’t auto add it? When you click on attach does it automatically work or do you need something else? Any idea why?

sha1n:
The Java sources are packed exactly the same as they were when my code packed them (with the scala directory at the bottom) and intellij obviously behaves the same - requires me to select the scala directory inside the source jar in order to show sources. I have no idea why that happens.. it smells like some kind of convention, but I haven't been able to trace it down yet.

ittaiz:
Does this reproduce also in one of our internal projects? What is the
behavior you’re seeing?
I’m leaning towards leaving this as is but I’d like the extra data point.

sha1n:
@ittaiz I still need to test the source gen code and if you remember, I couldn't make Bazel produce source jars without the intellij aspect attached. So after I read the docs a few times to convince myself that `output_groups` should  work, I tried it on a Java package and guess what.. it works! So my conclusion is that `rules_scala` simply doesn't support it. I want it, so I'll try to see if I can get it working for us.. if you have any tips that can shorten my journey, I'll be more than happy ;)

ittaiz:
given there are tests on the plugin side I'll merge as is

ittaiz:
for future reference @sha1n and @hmemcpy both are part of wix

johnynek:
Small note. Can you remove WIP from titles when they are ready to review?

ittaiz:
Indeed, will do.

beala-stripe:
@johnynek This already works for scrooge generated thrift actually. The scrooge jar is set as the source jar, so when you jump to def, it jumps to the generated code. Jumping to the thrift def itself would be great, but I'm not sure how to get IJ to do that.

johnynek:
oh yeah, since the thrift rule is a bit more complex, I guess it is already sharing the generated source.

Sorry for the noise.

I really need to find the time to make the proto rules work like the thrift rules (e.g. using aspects so we don't transitively build all the protos in each target).

PR #634, opened 2018-10-23
Check if RBE still passes

ittaiz:


johnynek:
Was this just a minor diff to trigger a build?

ittaiz:
Indeed. RBE (remote execution) build failed in the kwargs suite PR which
sounded strange so I wanted to make sure master build works properly
(thought maybe CI configuration changed behind the scenes).
I’ll keep it open for the author of the PR to be able to compare the runs

aherrmann:
@ittaiz Thanks for providing this PR for comparison. Unfortunately, I cannot see the build logs on buildkite. (As I understand, that's where the RBE tests are happening.)
It says "You must log in to see this page"

Issue #598, opened 2018-08-28
Force compressed output jar


hmemcpy:
Seems that the SingleJar utility supports the `--compression` flag.
We have an internal rule that is using the SingleJar utility to produce an uber-jar, and without the `--compression` flag, it is about twice the size it ought to be.

Any reason not to do this by default in the `_build_deployable` rule?

johnynek:
I don't think there is a reason not to do this unless the compression slows us way down, which I assume it won't.

It may really improve perf since we do less IO, especially with caching.

ittaiz:
@hmemcpy,
Can you send the PR?

PR #643, opened 2018-11-20
Use --compression in _build_deployable

prebeta:
For the feature request in: https://github.com/bazelbuild/rules_scala/issues/598

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -495,7 +495,8 @@ def _build_deployable(ctx, jars_list):
     # This calls bazels singlejar utility.
     # For a full list of available command line options see:
     # https://github.com/bazelbuild/bazel/blob/master/src/java_tools/singlejar/java/com/google/devtools/build/singlejar/SingleJar.java#L311
-    args = ["--normalize", "--sources"]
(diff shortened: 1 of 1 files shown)


ittaiz:
For our internal rule (which does an Uber jar but with some other stuff
like filtering and shading) it was a substantial change.
Given this doesn’t have a test should we add a small comment saying why and
that it’s not covered?
Not a must but might be helpful if someone touches the area

prebeta:
Using compression reduced the size of our fat jar from 1.5GB to 500MB. We have not seen any significant regressions in behavior during compilation or in production. It also looks like we use compression by default in our internal JarCreator: https://github.com/bazelbuild/rules_scala/blob/master/src/java/io/bazel/rulesscala/jar/JarCreator.java#L281

@ittaiz We already have a comment with a pointer to the flags available in SingleJar, so I'm not sure what additional information would be useful in this case. If you have a specific suggestion for content, I can add that into the PR.

ittaiz:
I’m really sorry @prebeta but you’ll have to rebase and push

prebeta:
@ittaiz sounds good. I've rebased and added a comment as well.

Issue #644, opened 2018-11-22
Fix Bazel incompatible changes


laurentlb:
I'd like to be able to build rules_scala with `--all_incompatible_changes`

e.g. `cd rules_scala && bazel test scala:all --all_incompatible_changes`

It may not be possible to fix every issue now due to transitive dependencies, but we should try to fix the issues from this repository.

xingao267:
This is blocking turn on --all_incompatible_changes for rules_docker.

Traceback (most recent call last):
	File "/usr/local/google/home/xingao/.cache/bazel/_bazel_xingao/459658d023448dae1285a534aa3be713/external/io_bazel_rules_scala_scala_compiler/BUILD", line 9
		scala_import(name = 'io_bazel_rules_scala_scala_compiler')
	File "/usr/local/google/home/xingao/.cache/bazel/_bazel_xingao/459658d023448dae1285a534aa3be713/external/io_bazel_rules_scala/scala/scala_import.bzl", line 27, in _scala_import_impl
		struct(scala = struct(outputs = struct(...)), ...)])
	File "/usr/local/google/home/xingao/.cache/bazel/_bazel_xingao/459658d023448dae1285a534aa3be713/external/io_bazel_rules_scala/scala/scala_import.bzl", line 32, in struct
		_create_provider(current_jars, transitive_runtime_j..., <5 more arguments>)
	File "/usr/local/google/home/xingao/.cache/bazel/_bazel_xingao/459658d023448dae1285a534aa3be713/external/io_bazel_rules_scala/scala/scala_import.bzl", line 73, in _create_provider
		java_common.create_provider(use_ijar = False, compile_time_jar...]), <3 more arguments>)
create_provider is deprecated and cannot be used when --incompatible_disallow_legacy_javainfo is set. Please migrate to the JavaInfo constructor.
ERROR: Analysis of target '//container:image_test' failed; build aborted: Analysis of target '@io_bazel_rules_scala_scala_compiler//:io_bazel_rules_scala_scala_compiler' failed; build aborted

ittaiz:
Hey,
I’d also like to get there.
The JavaInfo move is complicated since we have a big ball of mud in that
aspect. I hope to tackle this in the next few weeks but I estimate a
correct fix would take me 1-2 full days.

The change is not drop in at all.

laurentlb:
cc @iirina

johnynek:
[...]

PR #645, opened 2018-11-22
Update files for future Bazel changes

laurentlb:
Progress towards #644

This diff is large, because I reformatted all files with Buildifier. This will allow you to use Buildifier automated fixes in the future.


ittaiz:
Do you know if there are plans on running buildifier on buildkite as part of the default CI templates?

laurentlb:
We should look into that.

PR #646, opened 2018-11-27
Update bazel-toolchains pin

mhlopko:
This fixed Bazel CI failures (bazelbuild/bazel#6742) such as   https://buildkite.com/bazel/bazel-with-downstream-projects-bazel/builds/613#6483c40c-c77d-4f29-bcf8-a8dba45b1df0

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -153,10 +153,10 @@ format_repositories()
 
 http_archive(
     name = "bazel_toolchains",
-    sha256 = "4ab012a06e80172b1d2cc68a69f12237ba2c4eb47ba34cb8099830d3b8c43dbc",
-    strip_prefix = "bazel-toolchains-646207624ed58c9dc658a135e40e578f8bbabf64",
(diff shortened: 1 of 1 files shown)


mhlopko:
Travis is red because it couldn't download stuff. I cannot imagine how this PR could break it, so i'll go ahead and submit. Please shout at me if I'm not following the process.

johnynek:
Since you invited this: Please don’t self merge a red PR. We never self merge in this repo. Someone else can merge if they agree. We almost never merge red CI (even when it is only a README change).

mhlopko:
Yeah I was unsure about the policy, and I apologize.

So merging a green PR without review is acceptable, and merging red is not? That explains why the review requirement is not enforced by the github.

Thank you for bearing with me.

ittaiz:
We rarely merge a PR without review. It should have a really good reason
and then it should be green, yes.
Don't know if we want to add the review requirement on github though.
Feels bureaucratic to me but maybe it's a good signal to other bazel people.
Thank you for the fix btw.

PR #649, opened 2018-12-10
Fixed intellij metadata to include correct source_jars info for scala_import libraries

sha1n:
Fixed intellij metadata generated for `scala_import` to include the correct source jars information for intellij to be able properly show sources.

Since the attribute `source_jar` is marked deprecated in [intellij_ide_info.proto](https://github.com/bazelbuild/intellij/blob/02b148713798c71d86121e3c22ca5ef5470cb08b/proto/intellij_ide_info.proto#L36) I removed it from the metadata struct and used `source_jars` for all cases.

Tested manually on intellij 2018.3 and hopefully after this is merged, I will develop a test in the intellij bazel plugin project, to cover that.

We should probably do the corresponding cleanups in `aspect/intellij_info_impl.bzl`, just to keep it clean and avoid confusion.

### DIFF
--- a/scala/scala_import.bzl
+++ b/scala/scala_import.bzl
@@ -93,14 +93,13 @@ def _code_jars_and_intellij_metadata_from(jars, srcjar):
     intellij_metadata = []
     for jar in jars:
         current_jar_code_jars = _filter_out_non_code_jars(jar.files)
-        (current_jar_source_jars, src_jar) = _source_jars(jar, srcjar)
+        current_jar_source_jars = _source_jars(jar, srcjar)
         code_jars += current_jar_code_jars
         for current_class_jar in current_jar_code_jars:  #intellij, untested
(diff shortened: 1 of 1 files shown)


sha1n:
I signed it!

scala/scala_import.bzl
```diff
@@ -93,29 +93,28 @@ def _code_jars_and_intellij_metadata_from(jars, srcjar):
...
-        return ([], srcjar)
+        return [srcjar]
     else:
         jar_source_jars = [
             file
             for file in jar.files.to_list()
             if _is_source_jar(file)
         ]
-        return (jar_source_jars, None)
+        return jar_source_jars
```
johnynek: Why don’t we merge srcjar into the result list? I think this is why thrift sources at Stripe don’t show up with sources: they use code generation which works via source jars.

PR #650, opened 2018-12-11
scala_proto should require JavaInfo provider and not specific rules

ittaiz:
subj.
drive-by: removed old conditional.
@laurentlb @dslomov I really wanted to use `ProtoSourcesProvider` and not `"proto"` like the [docs say](https://docs.bazel.build/versions/master/skylark/rules.html#migrating-from-legacy-providers) but as far as I can see `ProtoSourcesProvider` isn't defined as a symbol. Also see https://github.com/bazelbuild/bazel/issues/3701 and https://github.com/bazelbuild/bazel/issues/3527
Can anyone help us move to the correct pattern?

### DIFF
--- a/scala_proto/scala_proto.bzl
+++ b/scala_proto/scala_proto.bzl
@@ -450,12 +450,7 @@ scala_proto_srcjar = rule(
     attrs = {
         "deps": attr.label_list(
             mandatory = True,
-            allow_rules = [
-                "proto_library",
-                "java_proto_library",
-                "java_library",
-                "scala_library",
-            ],
+            providers = [["proto"],[JavaInfo]],
         ),
         "flags": attr.string_list(default = []),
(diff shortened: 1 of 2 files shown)


ittaiz:
by the way, my motivation came from the need to depend on `scala_import` and so that is the test I added. I decided to implement the generic best practice solution of `providers` since `allow_rules` is specific and deprecated. https://docs.bazel.build/versions/master/skylark/lib/attr.html

ittaiz:
@johnynek any thoughts? I'd like to merge this since this is blocking us

ittaiz:
Thank you!
@laurentlb @dslomov if we can still get your help we'd appreciate it since we'd like to avoid the usage of the legacy providers

laurentlb:
@c-parsons 
@lberki

PR #656, opened 2019-01-06
Allow alternatively discovering junit tests from a separate target

ittaiz:
if `alternatively_discover_tests_from` is set then test discovery is attempted from a different target.
This is in preparation for the introduction of `trim_test_configuration` in bazel which will mean that `scala_library` (even with `testonly = 1`) can't depend on a `scala_junit_test`.
For us (Wix) this will be a problem since often `scala_junit_test` targets contain tests and a bit of test support code which other test support targets (`scala_library`) sometimes depend on.
Additionally such separation can serve us since we can segment compilation rules to test rules and have different worker pools (in remote execution) for the different usage

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -1049,13 +1049,13 @@ def _serialize_archives_short_path(archives):
 def _get_test_archive_jars(ctx, test_archives):
     flattened_list = []
     for archive in test_archives:
-        #1. fix scala_library full_compile_jars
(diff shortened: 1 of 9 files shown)


ittaiz:
@anchlovi I'd appreciate your review here

ittaiz:
Good question. I deliberated a lot about it.
Wdyt about “tests_from”?

ittaiz:
@johnynek wdyt about `tests_from`?

johnynek:
tests_from seems good to me.

PR #658, opened 2019-01-08
Explicitly re-export symbols

laurentlb:
This is needed for a future change in Bazel
(`--incompatible_no_transitive_loads`).

Tested:
  `bazel test //scala:all //test/... --all_incompatible_changes --incompatible_disallow_legacy_javainfo=false --incompatible_disable_deprecated_attr_params=false`


laurentlb:
No, that was not a good idea.
I realize there was only one call site, so I've fixed it.

johnynek:
yeah, I think that was not intentional.

Issue #564, opened 2018-07-18
scrooge_scala_library: Remove `scala` field on provider and instead return jars in JavaInfo provider


beala-stripe:
In order to support the IntelliJ Bazel plugin, #562 returns an [old style provider](https://github.com/beala-stripe/rules_scala/blob/57661fa1ebbdc1359147d573e0ba4306dc6b9b98/twitter_scrooge/twitter_scrooge.bzl#L302-L308) from `scrooge_scala_library`. This is to work around an issue where a `JavaInfo` provider cannot have multiple output jars, so we return a scala provider instead (see PR #562 for an in depth description of the issue and workaround).

Setting multiple output jars on an `JavaInfo` will be addressed by [this fix](https://github.com/bazelbuild/bazel/commit/4ff32da2f79f36fbb1e608830835a31f39772b8a), which is expected to [arrive in 0.17.0](https://groups.google.com/d/msg/bazel-sig-jvm/THl2DRqyYW0/Za5_xZscBgAJ). Once this lands, it will allow us to set multiple output jars by merging multiple `JavaInfo`s and we can then get rid of the `scala` field and set the output jars on `JavaInfo` instead.

johnynek:
relates to #466

Issue #466, opened 2018-03-31
document how intellij bazel rules depend on scala rules


johnynek:
in this file there are two references to intellij:

https://github.com/bazelbuild/rules_scala/blob/master/scala/scala.bzl#L774

It seems that intellij seems to be wanting some information put in the struct accessible at `.scala` on a target. However, we want to move to modern providers, so that seems to mean that the intellij rules will need to be updated.

Can we document:

1. what dependency do the intellij rules have on us? What API exactly do they require?
2. what we need to do to change the API. I guess we will need to make changes in almost lockstep: so we make the breaking change here then send a PR to some other repo.

@ittaiz @gkk-stripe @andyscott do you all know anything about this?

andyscott:
I don't have any context on this-- but I am also extremely curious to learn more.

johnynek:
cc @jjudd who also has a "blame" on one of the lines that mentions intellij (although ittai seems to have the most).

gkossakowski:
i believe the dependency comes from intellij collecting java provider and expecting certain attributes on it, see: https://github.com/bazelbuild/intellij/blob/master/aspect/intellij_info_impl.bzl#L408 and calls to that method

johnynek:
yeah, I wonder why they do that rather than just getting the java provider? I guess we would need a PR to them.

@ittaiz you have followed this: does the native java provider have enough information to support just using the java provider and not having any special rules for scala?

ittaiz:
The important part is a few lines below what @gkossakowski linked to:
```
  if hasattr(target, "scala"):
    return target.scala
  if hasattr(target, "kt") and hasattr(target.kt,"outputs"):
    return target.kt
  # java_common.provider is a work in progress. It will soon expose the information
  # we require (e.g. outputs jars, jdeps), but does not yet do so.
  if java_common.provider in target:
    return target[java_common.provider]
```
Note:

>   java_common.provider is a work in progress. It will soon expose the information
>   we require (e.g. outputs jars, jdeps), but does not yet do so.

I think the work @dbabkin did on the [bazel side](https://github.com/bazelbuild/bazel/issues/3769) took us a lot farther but unfortunately due to some ongoing [discussion](https://github.com/bazelbuild/bazel/issues/4584) with respect to strict-deps labels I'm not sure we can switch to using the new JavaInfo API (I think, not sure, it hides information we currently need for the labels accumulation).

[...]

PR #660, opened 2019-01-10
Fix transitive thrift symbol resolution in IntelliJ

beala-stripe:
Currently, given the following build file structure:

scrooge_scala_library -> thrift_library -> thrift_library -> thrift_library -> ...

where "A -> B" means "A depends on B."

Only symbols in the leftmost `thrift_library` will resolve in IJ. This is because when the compiled scrooge jars are gathered up, transitive jars are not gathered. This fixes this behavior by adding transitive jars to the `ScroogeAspectInfo.output_files` depset.

I'm not sure the best way to test this. I've manually run this over some internal Stripe repos. Are there unit tests you can point me to? The last time I tried to add tests to the IJ plugin project I didn't get a response: https://github.com/bazelbuild/intellij/pull/366

### DIFF
--- a/twitter_scrooge/twitter_scrooge.bzl
+++ b/twitter_scrooge/twitter_scrooge.bzl
@@ -312,7 +312,7 @@ def _scrooge_aspect_impl(target, ctx):
 
         src_jars = depset([scrooge_file])
         output = _compiled_jar_file(ctx.actions, scrooge_file)
-        outs = depset([output])
+        outs = depset([output], transitive = [d[ScroogeAspectInfo].output_files for d in ctx.rule.attr.deps])
         java_info = _compile_scala(
             ctx,
             target.label,
@@ -325,7 +325,7 @@ def _scrooge_aspect_impl(target, ctx):
     else:
         # this target is only an aggregation target
         src_jars = depset()
-        outs = depset()
+        outs = depset(transitive = [d[ScroogeAspectInfo].output_files for d in ctx.rule.attr.deps])
         java_info = _empty_java_info(deps, imps)
 
     return [
@@ -412,10 +412,10 @@ def _create_scala_struct(ctx):
     output_jars = []
 
     for dep in ctx.attr.deps:
-        for j in dep[ScroogeAspectInfo].java_info.outputs.jars:
+        for j in dep[ScroogeAspectInfo].output_files:
             output_jars.append(
                 struct(
-                    class_jar = j.class_jar,
+                    class_jar = j,
                     ijar = None,
                     source_jar = None,
                     source_jars = [],


twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -312,7 +312,7 @@ def _scrooge_aspect_impl(target, ctx):
 
         src_jars = depset([scrooge_file])
         output = _compiled_jar_file(ctx.actions, scrooge_file)
-        outs = depset([output])
+        outs = depset([output], transitive = [d[ScroogeAspectInfo].output_files for d in ctx.rule.attr.deps])
```
johnynek: I don't think this is the right way. Often, these providers explicitly propagate a transitive output as well. I think we should add that and include it there. Then when we create the scala struct, we would add it in there.

That said, the scala struct thing is a hack, and we are supposed to only be able to use JavaInfo eventually. I don't know what the current state of the art is there though since bazel was constantly churning on JavaInfo and intellij plugin was behind on it.

johnynek:
see #466 for an issue on this. It would be really good to kill this custom hack and be clearer about how to get info into intellij.

beala-stripe:
Yup I remember this: https://github.com/bazelbuild/rules_scala/issues/564

According to that, the functionality we need should exist in bazel 0.17, but we'll have to drop support for releases before that iiuc. Unfortunately, not all Stripe repos are on that (yes, terrible I know).

How about I make an explicit transitive output field, and then address https://github.com/bazelbuild/rules_scala/issues/564 in a follow up?

It also looks like I broke some tests, so I'll fix that too.

beala-stripe:
Travis CI is confusing me. It has build two different SHAs that, when clicked, go to the same commit on github, and neither of them are my latest commit 😕 

Could someone with access clear the cache and rebuild? It looks like buildkite is passing.

ittaiz:
I’ll just add that we, rules_scala, need to move to using only JavaInfo as
bazel want to remove the deprecated java_common.create_provider and have
asked me to actively work on it.
Provider handling in our code is unfortunately a big ball of mud and so I
need to clear some time to make the change intelligently.
I expect to get to it in the couple of weeks

beala-stripe:
The most recent commit removes the old-style providers. Note that this change is *not* compatible with bazel < 0.17.

If you want to merge this, it'd be nice to do a normal merge so stripe repos that are behind on upgrades can potentially rely on the earlier commit that is compatible with earlier versions.

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -450,6 +432,7 @@ scrooge_scala_import = rule(
     attrs = {
         "thrift_jars": attr.label_list(allow_files = [".jar"]),
         "scala_jars": attr.label_list(allow_files = [".jar"]),
+        "deps": attr.label_list(),
```
ittaiz: Is this a new attribute? If so do we want a test to show how to use it? Also will help us not delete it by mistake

ittaiz:
@beala-stripe the previous commit seems to be broken? Also we don't merge regularly (for some reason it's even disabled here) but we can just have two PRs which we merge sequentially.
I know @johnynek has allergies from this ;) but I'm ok with it (if both pass the build)

beala-stripe:
Ok let's keep it simple then and just squash merge. We're planning on upgrading bazel soon, so breaking < 0.17 compat won't be an issue for long (as long as yall are ok with it)

twitter_scrooge/twitter_scrooge.bzl
beala-stripe: Good point. I'll see what I can wire up.

beala-stripe: Err wait a minute. I actually don't think this attribute is necessary. I think I got confused between `scrooge_scala_import` and `scrooge_scala_library`. I'll get back to you.

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -387,16 +387,11 @@ def _scrooge_scala_library_impl(ctx):
     else:
         all_java = aspect_info.java_info
 
-    # TODO: Remove `scala` field once JavaInfo supports multiple
-    # output jars. https://github.com/bazelbuild/rules_scala/issues/564
```
johnynek: really, for the TODO we just need to delete this? can we have any test that the output jars are being plumbed through correctly when we have multiple jars?

beala-stripe:
I've pushed up a greatly simplified diff. 

Now that `java_common.merge` merges the `java_info.output` field (since bazel 0.17), all of the extra code that was gathering up outputs has been made obsolete (see the now removed `_create_scala_struct` function). With the change to `java_common.merge`, the existing code now gathers up the outputs in the correct way.

johnynek:
@beala-stripe can we remove the scala provider entirely now, not just from this call site?

ittaiz:
AFAIK relying on java_common.create_provider is problematic since the new APIs (like the merge) assume that the provider was built with JavaInfo. This is important because I think of how they assume the object is populated.
This is why I want to move the code to use the new constructor (complicated since it's a new idiom and we have a big ball of mud in the java/scala area). After that (maybe same PR) I want to move to `java_common.merge` and after extensive manual testing send another PR to remove the scala attribute. Please don't break by mistake the intellij integration. It's super crucial for us.

beala-stripe:
@ittaiz Roger that. What version of bazel are you running? The current diff should work with bazel >= 0.17. If you're on an older version, we need to move back to the previous strategy involving the Scala provider.

I'll try to find some time to write a few tests and increase confidence this doesn't break anything (for compatible bazel versions)

ittaiz:
Thanks!
We're on 21.

beala-stripe:
I've pushed up some unit tests that assert transitive thrift jars are collected up properly.

beala-stripe:
The tests is broken because it requires bazel >= 0.17, and the test script runs with 0.16.1. 

Is it possible to bump the version used by the test script? Or do we want to maintain compat with 0.16.1?

beala-stripe:
I should add that by "breaking compat" this only breaks compat with IJ, in the sense that thrift objects won't resolve correctly. It should still build with bazel versions before 0.17

johnynek:
I think it is fine to bump the lower bound up to 0.17. It would be good also test with a more recent version. We used to try to do two: an upper bound and a lower bound known to work.

But the script may or may not "just work" with a later version of bazel.

johnynek:
screen shot 2019-01-17 at 13 10 26

I hope that is some kind of blip. That time difference is quite concerning. Maybe we are using a travis cache and the 0.17 somehow benefited from that?

beala-stripe:
Thanks for approving! Is there anything else we need to do to get this merged?

johnynek:
since this impacts wix so heavily, I'd like for @ittaiz to also approve. Will you have time soon @ittaiz ?

johnynek:
actually, thinking on it more, wix does not use thrift, I don't think. So, maybe this does not hit them at all.

Is that correct, if so, I think we should just go ahead and merge.

PR #663, opened 2019-01-11
Fix building Finagle Services

### LINKED ISSUE #592: Cannot compile a Scrooge Finagle service
Hello,

Since I upgraded my Scala rules (last version known to work was 21c8913a2cf35130870df9f07a781afc77f33693, from last March), I cannot compile anymore my Thrift services with Scrooge. Creating `scrooge_scala_library` when the IDL file contains only structs is ok, but I cannot create a library when the IDL file also contains a service.

When trying to compile the target `scrooge_scala_library`, I get a lot of compilation errors looking like:
```error: object finagle is not a member of package com.twitter```

Previously, I used to add "//3rdparty/jvm/com/twitter:finagle_thrift" to my `deps` to bring in the Finagle dependency, but if I do so, I now get another error: 
```<target //3rdparty/jvm/com/twitter:finagle_thrift> (rule 'scala_import') doesn't contain declared provider 'ThriftInfo'```

I couldn't spot any documentation or tests about compiling Finagle Scrooge services, so I am a bit stuck. I can contribute tests/documentation if given a hint about how things are supposed to work. Thanks!

### PR DESCRIPTION
**Pull request by samschlegel**
This is a WIP PR to attempt to fix #592

### CODE REVIEW

**Comment by samschlegel** [comment]
CLA should be fixed?

**Comment by samschlegel** [comment]
This appears to require adding `finagle_thrift`, `finagle_core`, and `util_stats` to `_implicit_compile_deps`. 

This fixes the build, except for the following exception:
```
scalarules/test/twitter_scrooge/thrift/Service$FinagleClient.scala:87: error: value readFrom is not a memb
er of object org.apache.thrift.TApplicationException
        val exception = TApplicationException.readFrom(iprot) match {
```

This was added in Thrift 0.10, and Scrooge >18.4.0 supports only Thrift 0.10+, so a version bump there is also required.

I agree with the comments that @johnynek had in #592, that this should probably be solved via some kind of toolchain work, but for now this appears to be the path AFAICT.

File: twitter_scrooge/twitter_scrooge.bzl [hunk 1]
```diff
@@ -48,12 +48,18 @@ def twitter_scrooge(
             "scrooge_core": "00351f73b555d61cfe7320ef3b1367a9641e694cfb8dfa8a733cfcf49df872e8",
             "scrooge_generator": "0f0027e815e67985895a6f3caa137f02366ceeea4966498f34fb82cabb11dee6",
             "util_core": "5336da4846dfc3db8ffe5ae076be1021828cfee35aa17bda9af461e203cf265c",
+            "util_stats": "",
+            "finagle_thrift": "",
+            "finagle_core": "",
```
samschlegel: FIXME: Need to set these

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -28,13 +28,13 @@ def twitter_scrooge(
 
     native.maven_server(
         name = "twitter_scrooge_maven_server",
-        url = "http://mirror.bazel.build/repo1.maven.org/maven2/",
+        url = "http://central.maven.org/maven2/",
```
samschlegel: I'm not sure how to get this on the bazel mirror, so I changed it for now

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -462,6 +524,18 @@ scrooge_scala_import = rule(
                 Label(
                     "//external:io_bazel_rules_scala/dependency/thrift/scrooge_core",
                 ),
+                Label(
```
johnynek: I actually don't think these are used, are they? On the import target there is no compilation.

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -367,6 +418,15 @@ scrooge_aspect = aspect(
                 Label(
                     "//external:io_bazel_rules_scala/dependency/thrift/util_core",
                 ),
+                Label(
```
johnynek: I'm a little concerned that everyone now needs to depend on these jars even though many won't need them. I wish we could find some way to only put these jars on the classpath if the thrifts have services.

Maybe we could add `has_services` to the thrift rules, and then only for thrifts that claim to have services to we include these on the classpath. What do you think?

Increasing the classpath also slows down compilation and invalidates caches in some cases, so minimizing is a good idea.

samschlegel: Yeah, I was concerned about that too. I like the idea of `has_services`, but I'm not sure how to cleanly handle the failure case. If `has_services` isn't set, then the user will get dependency errors and it's not quite obvious why.

Would it be possible to somehow catch that specific failure case? We could also just print out a message saying "If you have services, this might be due to `has_services = False`"

johnynek: I think if the transitive dependencies don't include services, we just wouldn't generate the service output, so they wouldn't get dependency errors. If they do transitively have services, then bazel would error since scalac could not find required code when compiling the services.

I think it should be okay since the scrooge compiler can be instructed not to generate the services. Does that sound like it would work?

samschlegel: @johnynek Finally getting back to this. I see in the comments in `_compile_to_scala` there was mention of putting `include_services` on `thrift_info`. I don't fully grok providers and aspects yet, but it seems like this would make more sense to put on the `scrooge_scala_library` rule, since if it's put on `ThriftInfo`, a single `thrift_library` that wants services would cause all `scrooge_scala_library` targets to generate services instead of letting the user of `scrooge_scala_library` decide.

Will need to spend some time reading up on providers and aspects to understand this code better.

twitter_scrooge/twitter_scrooge.bzl
samschlegel: Will remove these. I'm guessing that if we're importing external libraries they should have the `runtime_deps` set properly

ittaiz:
@samschlegel @pvcnt what is the status here?

liucijus:

Issue #665, opened 2019-01-14
Support blacklisted_protos as in proto_lang_toolchain


simuons:
Hi.

**Motivation**
We generate scala code with `flat_package`. We also use `@google_api_protos//google/api:annotations_proto` in our services. Generated code for `annotations.proto` using `flat_package` clashes with google's java clients libraries. So we would like to have `annotations.proto` pre-generated and include it as runtime jar and prevent others to generate clashing code (it gets generated as `proto_library` dep). It would be also useful to stop generating code for well known types because it is already provided by scalapb.

**Solution**
Initially I considered to add support for [proto_lang_toolchain](https://docs.bazel.build/versions/master/be/protocol-buffer.html#proto_lang_toolchain) (which already have `blacklisted_protos`). Unfortunately according to this https://groups.google.com/forum/#!searchin/bazel-discuss/proto_lang_toolchain%7Csort:date/bazel-discuss/xoy98a4yhAA/hD9eDus9BAAJ it cannot be used from skylark.
Currently I'm thinking about adding `blacklisted_protos` parameter to `scala_proto_srcjar` and remove `blacklisted_protos` from inputs to protoc.

Does it make sense? If yes I'm going to open a PR for this.

ittaiz:
@johnynek did you see this?

PR #666, opened 2019-01-14
Add blacklisted_protos to scala_proto_srcjar

simuons:
Resolves https://github.com/bazelbuild/rules_scala/issues/665

Had to change `com_google_protobuf` and `com_google_protobuf_java` targets in a WORKSPACE otherwise I couldn't run `bazel test //test/...`

### DIFF
--- a/scala_proto/scala_proto.bzl
+++ b/scala_proto/scala_proto.bzl
@@ -386,6 +386,11 @@ def _colon_paths(data):
         for f in sorted(data)
     ])
 
+def _retained_protos(inputs, blacklisted_proto_targets):
+    blacklisted_protos = [dep for target in blacklisted_proto_targets for dep in target.proto.transitive_sources]
+    blacklisted_protos_dict = dict(zip(blacklisted_protos, blacklisted_protos))
+    return [f for f in inputs if blacklisted_protos_dict.get(f, None) == None]
+
 def _gen_proto_srcjar_impl(ctx):
     acc_imports = []
     transitive_proto_paths = []
@@ -406,10 +411,6 @@ def _gen_proto_srcjar_impl(ctx):
 
     deps_jars = collect_jars(jvm_deps)
 
-    blacklisted_protos = [dep for target in ctx.attr.blacklisted_protos for dep in target.proto.transitive_sources]
-    blacklisted_protos_dict = dict(zip(blacklisted_protos, blacklisted_protos))
-    inputs = [f.path for f in sorted(acc_imports) if blacklisted_protos_dict.get(f, None) == None]
-
     worker_content = "{output}\n{paths}\n{flags_arg}\n{packages}\n{inputs}".format(
         output = ctx.outputs.srcjar.path,
         paths = _colon_paths(acc_imports.to_list()),
@@ -418,7 +419,8 @@ def _gen_proto_srcjar_impl(ctx):
         # Command line args to worker cannot be empty so using padding
         packages = "-" +
                    ":".join(depset(transitive = transitive_proto_paths).to_list()),
-        inputs = ":".join(inputs)
+        # Pass inputs seprately because they doesn't always match to imports (ie blacklisted protos are excluded)
+        inputs = ":".join(sorted([f.path for f in _retained_protos(acc_imports, ctx.attr.blacklisted_protos)]))
     )
     argfile = ctx.actions.declare_file(
         "%s_worker_input" % ctx.label.name,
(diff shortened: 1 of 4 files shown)


simuons:
I signed it!

ittaiz:
Thanks. Will reply later or tomorrow. Have a few questions.

test/proto/BUILD
```diff
@@ -60,3 +66,24 @@ scalapb_proto_library(
...
+
+scala_proto_srcjar(
+    name = "test1_proto_scala",
+    deps = ["//test/proto2:test"],
+    generator = "@io_bazel_rules_scala//src/scala/scripts:scalapb_generator")
+
+scala_proto_srcjar(
+    name = "test2_proto_scala",
+    deps = [":test2"],
+    blacklisted_protos = ["//test/proto2:test"],
```
johnynek: this feels like the wrong approach to me. The thrift rules don't work this way. It seems wrong to me that we are transitively building all the protos in each target. Instead, I think each target should only build what they depend on (as the thrift rules do).

If we do this, there is no need for blacklisting, since we only ever build one target.

In the thrift rules, we used aspects to achieve this:
https://github.com/bazelbuild/rules_scala/blob/master/twitter_scrooge/twitter_scrooge.bzl#L404

simuons: @johnynek thanks for the reply. 

I just want to make sure I understand you correctly. Let's say we have targets like this
```
proto_library(
    name = "a_proto",
    srcs = ["a.proto"]
)

proto_library(
    name = "b_proto",
    srcs = ["b.proto"],
    deps = [":a_proto"]
)
```

and if we have following target
```
scalapb_proto_library(
    name = "scala_proto",
    deps = [":b_proto"],
)
```

scala code should be generated only for `b.proto`?

scala_proto/scala_proto.bzl
```diff
@@ -406,14 +406,19 @@ def _gen_proto_srcjar_impl(ctx):
 
     deps_jars = collect_jars(jvm_deps)
 
-    worker_content = "{output}\n{paths}\n{flags_arg}\n{packages}".format(
+    blacklisted_protos = [dep for target in ctx.attr.blacklisted_protos for dep in target.proto.transitive_sources]
+    blacklisted_protos_dict = dict(zip(blacklisted_protos, blacklisted_protos))
+    inputs = [f.path for f in sorted(acc_imports) if blacklisted_protos_dict.get(f, None) == None]
```
ittaiz: I think this `sorted` can be expensive. WDYT about only doing this whole line if `blacklisted_protos` is non empty?
also wdyt about using `if blacklisted_protos_dict.has_key(f)`?
lastly maybe it's best to just do:
` inputs = [f.path for f in retained_protos(acc_imports, blacklisted_protos_dict)`?
then the if and the details can be kept out of this context?

simuons:
Indeed `scalapb_proto_library` behaves differently than `java_proto_library`. I'm referring to https://blog.bazel.build/2017/02/27/protocol-buffers.html (looks like this is the only source which explains how protobuf in bazel works). Indeed it's implemented with aspects as @johnynek pointed and mentioned in blog post. I did small [project](https://github.com/simuons/bazel-proto-test/tree/java-proto-library-outs) to check `java_proto_library` outputs and described them [here](https://github.com/simuons/bazel-proto-test/blob/java-proto-library-outs/notes.md). Looks like `java_proto_library` generates src.jar and speed.jar (not sure what this means) for direct dependency and speed.jar for transitive dependencies.
Anyway I'm not trying to align `scalapb_proto_library` behaviour to `java_proto_library` with this PR. I just need a feature that `java_proto_library` has via [proto_lang_toolchain](https://docs.bazel.build/versions/master/be/protocol-buffer.html#proto_lang_toolchain). Even after behaviour will be aligned I would still need to prevent code generation/compilation for well known protos as this causes trouble for us.

ittaiz:
@johnynek my understanding currently is:
1. Our current implementation is significantly different than java proto one.
2. We have a use-case that's blocking our adoption in some of our services due to the issue above and it is different enough from the issue you're touching,

I'd like us to merge this and have @simuons try to rewrite this using aspects (I guess in ~ a month).
WDYT?
Note that even with aspects will need this feature.

scala_proto/scala_proto.bzl
simuons: type 'dict' has no method has_key(File).
I'll extract to separate function. 
Not sure I understood your comment about sorting. If blacklisted_protos would be empty I would still need to sort that list. I don't think sorting is needed at all. I just followed pattern from `_colon_paths`

ittaiz:
Thanks!

PR #668, opened 2019-01-16
Adding a load rule for new `http_archive` in the code snippet

hmemcpy:
This prevents getting an error ("The native http_archive rule is deprecated.")


johnynek:
Thanks!

PR #671, opened 2019-01-22
Prevent test runner hang in case of exception thrown from specs2

hmemcpy:
This gracefully shuts down specs2 in case of an exception thrown from initialization. This fixes the bazel runner hanging indefinitely.

### DIFF
--- a/src/java/io/bazel/rulesscala/specs2/Specs2RunnerBuilder.scala
+++ b/src/java/io/bazel/rulesscala/specs2/Specs2RunnerBuilder.scala
@@ -3,8 +3,8 @@ package io.bazel.rulesscala.specs2
 import java.util
 import java.util.regex.Pattern
 
-import io.bazel.rulesscala.test_discovery._
 import io.bazel.rulesscala.test_discovery.FilteredRunnerBuilder.FilteringRunnerBuilder
+import io.bazel.rulesscala.test_discovery._
 import org.junit.runner.notification.RunNotifier
 import org.junit.runner.{Description, RunWith, Runner}
 import org.junit.runners.Suite
@@ -12,11 +12,12 @@ import org.junit.runners.model.RunnerBuilder
 import org.specs2.concurrent.ExecutionEnv
 import org.specs2.control.Action
 import org.specs2.main.{Arguments, CommandLine, Select}
+import org.specs2.runner.JUnitRunner
 import org.specs2.specification.core.Env
 import org.specs2.specification.process.Stats
 
-import scala.language.reflectiveCalls
 import scala.collection.JavaConverters._
+import scala.language.reflectiveCalls
 import scala.util.Try
 
 @RunWith(classOf[Specs2PrefixSuffixTestDiscoveringSuite])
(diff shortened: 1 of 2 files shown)


hmemcpy:
Handles both cases with test_filter and without!

src/java/io/bazel/rulesscala/specs2/Specs2RunnerBuilder.scala
```diff
@@ -34,22 +35,14 @@ class Specs2PrefixSuffixTestDiscoveringSuite(suite: Class[Any], runnerBuilder: R
...
-
-  override def getChildren: util.List[Runner] =
+  override def getChildren: util.List[Runner] = {
     super.getChildren.asScala
       .collect {
-        case r: FilteredSpecs2ClassRunner if r.matchesFilter => Some(r)
+        case r: FilteredSpecs2ClassRunner if r.matchesFilter => Some(new SafeSpecs2Runner(r))
         case _: FilteredSpecs2ClassRunner => None
-        case other => Some(other)
+        case other => Some(new SafeSpecs2Runner(other).asInstanceOf[Runner])
```
johnynek: Instead of a cast can we do:
```scala
val runner: Runner = new SafeSpecs2Runner(other)
Some(runner)
```
Casts are very brittle with respect to refactoring. Even when safe now, they may not be safe after a refactor.

hmemcpy:
Yeah, I added a couple of test cases, unfortunately they don't reproduce with the bundled v3.8.8 of specs2 - this bug only repros on v4.x of the library. I would have to either upgrade the bundled specs2 to v4.x, or produce a separate rule that loads the newer jars. Either way, I'll update this PR with a few test cases, since this was a pain in the ass to track :)
Thanks!

src/java/io/bazel/rulesscala/specs2/Specs2RunnerBuilder.scala
hmemcpy: Isn't variance fun!
Good point, I'll fix this.

ittaiz:
Thanks!
I think upgrading to 4.3.3 is better. Synthesizing the problem might look unnatural.
Also should we open an issue for Eric so he’ll be aware?

hmemcpy:
Great, now that I have v4.x inside rules_scala, I can write a test for this. I've rebased this PR on top of #673, so once it's merged, we'll have the fix.

ittaiz:
@hmemcpy given Eric is favorable to your [PR](https://github.com/etorreborre/specs2/pull/727) can we close this?

hmemcpy:
I suggest we keep it. The bug affects all of v4.x of specs2, if it's fixed, it's only going to be fixed in 4.3.x or whatever the latest version. This way we can ensure older clients don't hang too.

hmemcpy:
Added a test. Should be good now!

ittaiz:
OTOH like we discussed in newer versions the test won't reproduce and so we'll have extra code which is defensive without clear reason. I'd rather we close this one

ittaiz:
```
Still running: "scala_specs2_exception_in_initializer_without_filter"
Test "scala_specs2_exception_in_initializer_without_filter" successful (2 sec) 
running test scalac_jvm_flags_are_configured
```
From build log of [Fixing wrong junit hash](https://travis-ci.org/bazelbuild/rules_scala/jobs/485011182)
Given its two seconds and given we should just update the version in a few days when the new version is out I'm closing to have less code...

PR #674, opened 2019-01-26
Fix duplicate key error with ignored unused dependencies on scala junit tests

laurynasl-wix:
If you add `unused_dependency_checker_ignored_targets ` to your scala junit test target, analysis fails with message: 
```
duplicate keyword 'unused_dependency_checker_ignored_targets' in call to scala_junit_test
```
This PR fixes that by merging the ignored targets provided on the target with `_specs2_junit_dependencies`.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -569,7 +569,8 @@ def scala_specs2_junit_test(name, **kwargs):
     scala_junit_test(
         name = name,
         deps = _specs2_junit_dependencies() + kwargs.pop("deps", []),
-        unused_dependency_checker_ignored_targets = _specs2_junit_dependencies(),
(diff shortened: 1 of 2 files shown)


ittaiz:
@laurynasl-wix can you please also sign a private CLA? It’s easier than the jumping through hoops which is needed to add you to the company one (I don’t think anyone is registered there)

laurynasl-wix:
I signed the CLA. Only locally. Just starting out and seeing what sticks. I'm writing a dumb script that parses bazel output and removes all unused dependencies. I need it in order to split up a bigish target I have that is on the critical path for tests

ittaiz:
Cool. Sounds interesting. Good luck and if it works out maybe you can share it (blog post, spotlight, script, all of the above 😉)

Issue #680, opened 2019-01-28
Add scalac_jvm_flags attribute to scalapb_proto_library


akakitani:
I'm working with some pretty large proto definitions, and scalac ends up choking on the generated classes with a StackOverflowError. I've been able to get this to compile by bumping up `-Xss` with `scalac_jvm_flags` on my modified version of `scalapb_proto_library`. Would you accept a PR with this change?

johnynek:
I think this could be related to #548 

Currently the rules have the flaw that every proto target fully transitively rebuilds all dependencies. In a big repo, I fear that could get big, and it is wasteful.

This is just a flaw that should be fixed (to follow the aspect based approach thrift did) but it has not been done yet.

akakitani:
@johnynek ah I see, do you think this attribute would still be useful regardless? e.g. if I need to run `bazel clean`, it seems like we could run into this again even if #548 is fixed

johnynek:
I'm happy to accept the PR if it fixes an issue for you. It may be unrelated to what I'm suggesting. I'm not sure why we should need to have deep recursion depth. Are these really pathological protos?

akakitani:
A little bit yeah - these have a lot of nested `message` definitions and `oneof` fields with 200+ entries, and their dependencies are kind of complicated too. Here's a scalac stack trace in case it rings any bells:
```
Exception in thread "main" java.lang.StackOverflowError
        at scala.tools.nsc.typechecker.SuperAccessors$SuperAccTransformer.transform(SuperAccessors.scala:178)
        at scala.tools.nsc.typechecker.SuperAccessors$SuperAccTransformer.transform(SuperAccessors.scala:71)
        at scala.reflect.internal.Trees$class.itransform(Trees.scala:1345)
        at scala.reflect.internal.SymbolTable.itransform(SymbolTable.scala:16)
        at scala.reflect.internal.SymbolTable.itransform(SymbolTable.scala:16)
        at scala.reflect.api.Trees$Transformer.transform(Trees.scala:2555)
        at scala.tools.nsc.transform.TypingTransformers$TypingTransformer.transform(TypingTransformers.scala:44)
        at scala.tools.nsc.typechecker.SuperAccessors$SuperAccTransformer.mayNeedProtectedAccessor$1(SuperAccessors.scala:186)
        at scala.tools.nsc.typechecker.SuperAccessors$SuperAccTransformer.transformSelect$1(SuperAccessors.scala:332)
[...]

PR #681, opened 2019-01-30
Add scalac_jvm_flags attribute to scalapb_proto_library

akakitani:
Addresses https://github.com/bazelbuild/rules_scala/issues/680


scala_proto/scala_proto.bzl
```diff
@@ -502,7 +502,8 @@ Example:
     scalapb_proto_library(
         name = "exampla_proto_scala",
         with_grpc = True,
-        deps = ["//src/proto:example_service"]
+        deps = ["//src/proto:example_service"],
+        scalac_jvm_fags = ["-Xss8M"],
```
johnynek: 1. you have misspelled `flags` here.

2. I don't think we should set the default in the example, people will copy this example and setting flags should be uncommon.

johnynek:
Thanks for the PR!

Issue #548, opened 2018-06-28
aspect based proto rules


johnynek:
similar to #524 do the same for proto rules to fix the same issues.

ittaiz:
What problems exactly? Because proto doesn't use the `extra_information` it just uses the transitive proto sources from the proto provider

johnynek:
I think that currently the proto rules rebuild the transitive closure of all protos on each invocation.

That is not what you want. It means unless you make sure you have no overlaps (your proto_library forms a tree, not a DAG), you will duplicate class files in the jars across targets.

johnynek:
actually, it does not look like it has that bug, but you do need to manually build the graph correctly.

The aspect approach automatically generates a 1 for 1 proto_library to action DAG.

simuons:
@ittaiz I think this issue can be closed

ittaiz:
Thanks!

PR #685, opened 2019-02-08
Add targets that reproduce proto issues in 0.22.0 (and maybe earlier)

beala-stripe:
This reproduces an issue in bazel 0.22.0 that breaks `scalapb_proto_library`. Issue to follow.

It's expected that the build will fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,10 @@ env:
   # we want to test the last release
   #- V=0.16.1 TEST_SCRIPT=test_lint.sh
   - V=0.17.1 TEST_SCRIPT=test_rules_scala.sh
-  - V=0.21.0 TEST_SCRIPT=test_rules_scala.sh
+  - V=0.22.0 TEST_SCRIPT=test_rules_scala.sh
     #- V=0.14.1 TEST_SCRIPT=test_intellij_aspect.sh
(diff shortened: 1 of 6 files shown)


johnynek:
interesting this passes on 0.17... huh...

I have brought up many times that I think the proto rules need to improved #548 although I don't know if that will help here....

It seems like something at the bazel java skylark API may have changed.

ittaiz:
@simouns can you take a look? Cc @natansil @anchlovi who might have more
context
(I’ll also try to take a look in the next few days)

beala-stripe:
Thanks for the quick responses! I just wanted to call attention to the companion issue I created, which I believe explains the bazel API change that broke this: https://github.com/bazelbuild/rules_scala/issues/687

The question in my mind is does this needs to be addressed in Bazel or in rules_scala? I don't have much context on the proto rules, but I'll keep digging on Monday.

beala-stripe:
I pushed up a fix which mirrors the fix @lberki made here for the java rules: https://github.com/bazelbuild/bazel/commit/af36058

This simply strips out the problematic proto_paths entry. This builds successfully in an internal repo, but none of our repos (afaik) exercise the proto rules very hard.

 I've also added a test to targets that use the `proto_source_root` attribute, which was previously untested.

PR #688, opened 2019-02-13
Fix root path for generated proto files

gseitz:
The owner's workspace_root for proto files that are generated by a
genrule is empty. This means that the root path ends with a trailing
slash. PBGenerateRequest expects the root path to not end in a slash
and removes an additional character (see line 17) which is the expected
path separator after the workspace_root component of the path.
For generated files, this means that the first character of the proto
file path is cut off, and thus the scalapb compiler cannot read the
file.

### DIFF
--- a/src/scala/scripts/PBGenerateRequest.scala
+++ b/src/scala/scripts/PBGenerateRequest.scala
@@ -14,7 +14,7 @@ object PBGenerateRequest {
       val file = if (root.isEmpty) {
         parsed(1)
       } else {
-        parsed(1).substring(root.length + 1)
+        parsed(1).substring(root.length).stripPrefix("/")
       }
       (file, Paths.get(root, file).toString)
     }
--- a/test/proto3/BUILD
+++ b/test/proto3/BUILD
@@ -1,0 +1,23 @@
+load(
+    "//scala_proto:scala_proto.bzl",
+    "scalapb_proto_library",
+)
+
+genrule(
+    name = "generated",
+    srcs = ["test.proto"],
+    outs = ["generated.proto"],
+    cmd = "cp $(SRCS) \"$@\""
+)
+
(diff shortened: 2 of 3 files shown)


gseitz:
An alternative fix could be to use `file.stripPrefix("/")` in [PBGenerateRequest](https://github.com/bazelbuild/rules_scala/blob/master/src/scala/scripts/PBGenerateRequest.scala#L19).

ittaiz:
@gseitz thanks! the alternative fix sounds more coherent to me since it adds more semantics (prefix stripping) instead of adding another edge case in a detached place.
Are there cons with the alternative fix?

gseitz:
@ittaiz: I don't think so, and I agree it's better to not have another edge case. I'll push the change shortly.

gseitz:
@ittaiz: not sure why the builds failed. The tests ran successfully for me locally.

johnynek:
@gseitz if you click on the Travis link you should be able to see the error:
```
ERROR: /home/travis/build/bazelbuild/rules_scala/test/proto/BUILD:35:1: scala //test/proto:test_external_dep failed 
```

Does that target build for you locally? It failed on all the Travis builds.

gseitz:
@johnynek: Sorry, I wasn't clear. I had looked at the failure on Travis, but there's no hint for the reason of the failure. `bazel clean --expunge && bazel build test/proto/...` runs successfully for me locally.

johnynek:
I can restart, but given that all of them failed, it seems unlikely to help, I wonder if the rules are not hermetic in some way?

gseitz:
I pushed the correct fix now. Looks like always stripping the leading `/` is wrong for external dependencies.

gseitz:
Thanks for the quick feedback and merging the PR!

PR #694, opened 2019-02-19
use the JAVA_BIN environmental variable to allow overriding the jdk at runtime

ianoc-stripe:
Similar to the java_binary script we should allow users to override the version of java based on the environment rather than using the runfiles version.

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -553,8 +553,8 @@ def _write_java_wrapper(ctx, args = "", wrapper_preamble = ""):
         content = """#!/usr/bin/env bash
 {preamble}
 DEFAULT_JAVABIN={javabin}
-JAVABIN=${{JAVABIN:-$DEFAULT_JAVABIN}}
-{exec_str}$JAVABIN "$@" {args}
+JAVA_EXEC_TO_USE=${{REAL_EXTERNAL_JAVA_BIN:-$DEFAULT_JAVABIN}}
+{exec_str}$JAVA_EXEC_TO_USE "$@" {args}
 """.format(
             preamble = wrapper_preamble,
             exec_str = exec_str,
@@ -582,7 +582,7 @@ def _write_executable(ctx, rjars, main_class, jvm_flags, wrapper):
         substitutions = {
             "%classpath%": classpath,
             "%java_start_class%": main_class,
-            "%javabin%": "JAVABIN=%s/%s" % (
+            "%javabin%": "export REAL_EXTERNAL_JAVA_BIN=${JAVABIN};JAVABIN=%s/%s" % (
                 _runfiles_root(ctx),
                 wrapper.short_path,
             ),
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -552,8 +552,9 @@ def _write_java_wrapper(ctx, args = "", wrapper_preamble = ""):
         output = wrapper,
         content = """#!/usr/bin/env bash
 {preamble}
-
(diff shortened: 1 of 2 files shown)


scala/private/rule_impls.bzl
```diff
@@ -581,7 +582,7 @@ def _write_executable(ctx, rjars, main_class, jvm_flags, wrapper):
         substitutions = {
             "%classpath%": classpath,
             "%java_start_class%": main_class,
-            "%javabin%": "JAVABIN=%s/%s" % (
+            "%javabin%": "JAVABIN=${JAVABIN:-%s/%s}" % (
```
johnynek: Aren’t we double checking the default? Here we set DEFAULT_JAVA_BIN to either JAVABIN or the old default, then above we set JAVA_BIN to either JAVA_BIN or the default.

Seems like this line shouldn’t need to change.

Also: we seem to have both JAVABIN here and JAVA_BIN above. Should we only have JAVABIN?

ianoc-stripe: Should just be JAVABIN your right. I believe these are somewhat independent call sites for the same issue. this is the scala_binary path, the other is some simple wrapper script we use in places. Both of them seemed to hard code the path to the JDK to be the runfiles location. 

I'm not sure of the history of the java wrapper, but it looks like originally from the comment it was about setting up a java path. (Though the java launcher preamble will have that too).  it injects a preamble to control the tty for the repl, and hard codes args for scala_test by the looks. Doesn't do anything for scala_binary i can see?

So both would need to be updated if we can launch java with and without the wrapper i believe.

ianoc-stripe:
In regards tests i'll try figure that out -- i think we can refer to the extra outputs/override it to be broken as you said and assert that it fails to run with nonsense in the env. The non-set case i think should be covered in the current tests.

scala/private/rule_impls.bzl
ianoc-stripe: Ok so i wasn't following the 2 levels here properly. your right I was setting it twice... which was just plain broken.

So what is happening, now in my understanding(so correct me if i'm wrong). We use the java_binary wrapper, but we always call the inner script. That means however that `JAVABIN` exports to us from the outside world are actually a little tough to obey nicely, since effectively we always want to re-write `JAVABIN` onto  the wrapper. What i've done now is added another new variable and exported it as part of the `java_binary` injection. Then we pick up that variable to control the java path inside the wrapper script. With this I have (finally) proper success with the remote tests! 

(as always, thanks for the good review, led down the path to making it all actually work nicely)

ianoc:
@johnynek I think you'll need to do the merge since i don't have the bits. But i've pushed another change you should look at. The `action_should_fail` tests were actually all broken from the tests i just did locally. the bazel subprocess was invoked as `$(bazel ...)` not assigning the output of the subprocess to a variable, so bash evaluates it. Since the stdout isn't really a command it always fails with exit code 127 on my mac anyway

ianoc:
Happily i think those tests still worked -- the error in the broken build here looks like a transient download failure from what i can see.

ittaiz:
I'm now looking into these lines of code and I'm not sure this was the best choice. This sounds like you get less hemeticity.
Why not use different java_toolchain when running remote builds?

ianoc:
@ittaiz the remote execution bit is special to some experiments(happy to elaborate on these if your interested, but the content of the PR is more general). This change in general is aligning the scala rules to the java rules. The hermetic nature of the build in general won't be effected since that mostly comes via the sandbox we are running in. (And given the java rules obey these env vars if your proxying them one would expect them to effect everything the same?)

Issue #696, opened 2019-02-20
Be able to generate service provider configuration for targets that host service implementations


ity:
### Description of the problem / feature request:

Be able to generate service provider configuration for targets that host service implementations. This automates maintenance of jvm service provider configuration files as described here:
  https://docs.oracle.com/javase/6/docs/api/java/util/ServiceLoader.html

### Feature requests: what underlying problem are you trying to solve with this feature?

Currently, there is no way to add this information to the `scala_library` target so that the generation of this configuration is automated. 

Proposal: to add a `services` attribute to `scala_library` for this additional information such that if its specified like:
```
scala_library(
    ...
    services = {
        "com.a.b": [
            "com.a.b.ServiceInterface1",
            "com.a.b.ServiceInterface2",
        ],
    },
)
```
then this will result in a directory being created in the output jar: `META-INF/services`, and an additional text file per service interface listing the implementing classes like:
```
$ cat $output_jar/META-INF/services/com.a.b
com.a.b.ServiceInterface1
com.a.b.ServiceInterface2
```

### Any other information, logs, or outputs that you want to share?

Attaching a draft wip PR with this issue to kickstart the work and happy to hear thoughts/ideas.

pierrechevalier83:
I'm interested in picking up this work. I have read @ity's code on https://github.com/bazelbuild/rules_scala/pull/697/files and the discussions triggered by it.
Based on these, I wrote a commit that applies cleanly on recent master and I think follows the advice given in the initial draft PR's discussion.

I opened https://github.com/bazelbuild/rules_scala/pull/1063 to pick up the work and the discussion. I would be very grateful for feedback on the approach.

PR #697, opened 2019-02-20
wip - generate service provider configuration for `scala_library` targets

ity:
Add a `services` attribute to `scala_library` for this additional information. See #696 for details.

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -55,6 +55,10 @@ def _adjust_resources_path_by_default_prefixes(path):
     if rel_path:
         return dir_1 + dir_2, rel_path
 
+    (dir1, meta_inf, rel_path) = path.partition("META-INF/")
+    if rel_path:
+        return dir1, meta_inf + rel_path
+
     return "", path
 
 def _adjust_resources_path(path, resource_strip_prefix):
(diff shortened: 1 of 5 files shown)


ittaiz:
Can you please rebase on top of master? You have conflicts and the diff with test_rules_scala.sh is confusing

ity:
> Can you please rebase on top of master? You have conflicts and the diff with test_rules_scala.sh is confusing

just did, thanks for taking a look!

johnynek:
@ity good to see you!

I wonder, would it be possible for this to be a separate rule that just builds a jar that has this file? Then it could be added as a runtime dependency or any java or scala library?

In that way, it wouldn't be coupled to the scala library at all.

How does that sound?

ity:
> @ity good to see you!
> 
> I wonder, would it be possible for this to be a separate rule that just builds a jar that has this file? Then it could be added as a runtime dependency or any java or scala library?
> 
> In that way, it wouldn't be coupled to the scala library at all.
> 
> How does that sound?

@johnynek thanks :)

And yes, making it a separate rule makes sense to me! I can edit and send an update in the same PR.

ity:
Realistically, I might not get to this in the next 2 weeks or so - so if anyone wants to take a stab at this in the meantime, feel free. Otherwise I will get to this as soon as I can.

ittaiz:
@ity do you think you'll get to it or I'll close this?

liucijus:

PR #695, opened 2019-02-20
Support data files in jmh benchmarks

tferi-da:

PR #698, opened 2019-02-21
add plus one deps transitive mode for compilation

ittaiz:
Tries to balance between sternness of strict-deps off and strict-deps on cache issues and false positives.
We, Wix, have been using `strict-deps=error` for a long time and have been enjoying the relative ease a developer has when needing to add a new dependency.
However the current strict-deps compiler plugin has a serious drawback in generating too many false positives (causing errors about dependencies which aren't really used because of the compiler eagerness). This is a much bigger issue for us then the cache hits (which we also care about) since this means that:
1. Builds on master fail from seemingly innocent changes. This is because we have many repositories in a logical head dependency.
2. Additionally this causes developer fatigue which just try to "please the beast".

OTOH using strict-deps off has similar issues of easily breaking builds. For examples you can see the tests I added and also this [scala-center](https://github.com/scalacenter/advisoryboard/blob/master/proposals/009-improve-direct-dependency-experience.md) discussion.

This PR adds a new *heuristic* mode which I hope will strike a better balance between the two. It passes to the compiler all direct dependencies and *their* direct dependencies since very often this is needed by the compiler. This was also confirmed by offline discussions with the google java people.

We also hope in the future to add some kind of post processing action (inside the bazel build) which might work over the byte code or source code to find strict-deps infringements of these plus one deps.
Additionally we have an internal tool in beta mode which goes over the compiler output and via various analyzers (mainly regex based) finds the needed FQN and then looks them up in bazel-out, external binary dependencies as well as a global index we have for our own source code.
This tool is run outside of bazel by the user.
We hope to open source this tool as well as the global index server API.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,10 @@ env:
   # See https://github.com/bazelbuild/rules_scala/pull/622
   # we want to test the last release
   #- V=0.16.1 TEST_SCRIPT=test_lint.sh
-  - V=0.17.1 TEST_SCRIPT=test_rules_scala.sh
+  #- V=0.17.1 TEST_SCRIPT=test_rules_scala.sh
   - V=0.22.0 TEST_SCRIPT=test_rules_scala.sh
     #- V=0.14.1 TEST_SCRIPT=test_intellij_aspect.sh
-  - V=0.17.1 TEST_SCRIPT=test_reproducibility.sh
+  #- V=0.17.1 TEST_SCRIPT=test_reproducibility.sh
   - V=0.22.0 TEST_SCRIPT=test_reproducibility.sh
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,10 @@ env:
   # See https://github.com/bazelbuild/rules_scala/pull/622
   # we want to test the last release
   #- V=0.16.1 TEST_SCRIPT=test_lint.sh
-  #- V=0.17.1 TEST_SCRIPT=test_rules_scala.sh
+  - V=0.17.1 TEST_SCRIPT=test_rules_scala.sh
   - V=0.22.0 TEST_SCRIPT=test_rules_scala.sh
     #- V=0.14.1 TEST_SCRIPT=test_intellij_aspect.sh
-  #- V=0.17.1 TEST_SCRIPT=test_reproducibility.sh
+  - V=0.17.1 TEST_SCRIPT=test_reproducibility.sh
   - V=0.22.0 TEST_SCRIPT=test_reproducibility.sh
 
 before_install:
--- a/scala/plusone.bzl
+++ b/scala/plusone.bzl
@@ -1,3 +1,8 @@
+"""
+Keeps direct compile dependencies of targets.
+This enables targets to pass the to compiler the plus one dependencies in addition to the direct ones.
+For motivation of plus one see the e2e tests     9
+"""
 PlusOneDeps = provider(
     fields = {
         'direct_deps' : 'list of direct compile dependencies of a target',
--- a/scala/plusone.bzl
+++ b/scala/plusone.bzl
@@ -1,0 +1,12 @@
+PlusOneDeps = provider(
+    fields = {
+        'direct_deps' : 'list of direct compile dependencies of a target',
+    }
+)
+
+def _collect_plus_one_deps_aspect_impl(target, ctx):
+    return [PlusOneDeps(direct_deps = getattr(ctx.rule.attr,'deps',[]))]
+
+collect_plus_one_deps_aspect = aspect(implementation = _collect_plus_one_deps_aspect_impl,
+    attr_aspects = ['deps'],
+)
--- a/scala/plusone.bzl
+++ b/scala/plusone.bzl
@@ -1,7 +1,7 @@
 """
 Keeps direct compile dependencies of targets.
 This enables targets to pass the to compiler the plus one dependencies in addition to the direct ones.
-For motivation of plus one see the e2e tests     9
+For motivation of plus one see the e2e tests
 """
 PlusOneDeps = provider(
     fields = {
--- a/scala/private/common.bzl
+++ b/scala/private/common.bzl
@@ -1,4 +1,5 @@
 load("@io_bazel_rules_scala//scala:jars_to_labels.bzl", "JarsToLabelsInfo")
+load("@io_bazel_rules_scala//scala:plusone.bzl", "PlusOneDeps")
 
 def write_manifest(ctx):
     main_class = getattr(ctx.attr, "main_class", None)
@@ -157,16 +158,3 @@ def create_java_provider(scalaattr, transitive_compile_time_jars):
         ),
         transitive_runtime_jars = scalaattr.transitive_runtime_jars,
     )
-
-PlusOneDeps = provider(
-    fields = {
-        'direct_deps' : 'list of direct compile dependencies of a target',
-    }
-)
-
-def _collect_plus_one_deps_aspect_impl(target, ctx):
-    return [PlusOneDeps(direct_deps = getattr(ctx.rule.attr,'deps',[]))]
-
-collect_plus_one_deps_aspect = aspect(implementation = _collect_plus_one_deps_aspect_impl,
-    attr_aspects = ['deps'],
-)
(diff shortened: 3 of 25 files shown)


ittaiz:
btw, I really wanted to refrain from adding the aspect to the `deps` attribute but rather run the aspect from command line but unfortunately this behaved very differently.

johnynek:
@ittaiz can you comment more or share any links about discussions with the google folks? Can you give some references: is this how it works at Google internally, do the standard java rules work this way?

I'd really like to be moving in the other way: less configuration. Adding a new compilation mode adds complexity and tends to cut the effectiveness of our tests since we generally are not multiplicatively expanding our tests to check all configurations.

I wouldn't want to keep inviting new heuristics for people to bolt into the rules that each company thinks is a good solution to the problem. I'd rather push the concerns upstream to scalac where possible, and follow the java standard rules as much as possible.

ittaiz:
the discussions were offline with @cushon.
AFAIU the java rules work similarly to this but not exactly. They add all of the dependencies that were used in the previous layer.
Example:
A->B->(C,D) and D isn' used then AFAIU they will compile A with B and C.

In general the java rules are much more robust (and complicated) since they have more time/people/history behind it.

I agree I'd like less configuration but I have to say that the current default is mainly because it's what Stripe uses and Stripe started the project. It doesn't mean direct-only is the right mode and I'll even add that all of the google people I talked to thought it's unreasonable (see this [discussion](https://github.com/bazelbuild/bazel/issues/4821) for a bit info).

Adding things to scalac might be a good long term goal but we can't hold our breath for it.

I'll add that as far as we're concerned we can definitely discuss removing the current implementation of strict-deps

ittaiz:
I’ll also add that e2e indeed aren’t multiplied but the regular test/... are multiplied

Lastly I’d encourage many different approaches from different companies in hopes we’ll find a good one. From my examples you can see the direct mode breaks on simple usecases. People can get used to adding transitive dependencies without understanding but I don’t think it’s something we want to aspire to you.

ittaiz:
I remembered one more thing about java rules- AFAIR they have one action which gets all of the transitive classpath to it but the differences from our current strict deps approach are:
1. They try to compile with the +1 variation I explained above and if fail they compile with all of it (as a single bazel action)
2. Their strict deps analysis mechanism is vastly superior to ours due to many reasons (javac/java simpler

ittaiz:
Closed by mistake

ittaiz:
@johnynek ?

.travis.yml
```diff
@@ -19,10 +19,10 @@ env:
   # See https://github.com/bazelbuild/rules_scala/pull/622
   # we want to test the last release
   #- V=0.16.1 TEST_SCRIPT=test_lint.sh
-  - V=0.17.1 TEST_SCRIPT=test_rules_scala.sh
+  #- V=0.17.1 TEST_SCRIPT=test_rules_scala.sh
```
johnynek: Let's uncomment or increase this to the lowest version we support. If we are bumping that, let's update the README.

scala/private/common.bzl
```diff
@@ -148,3 +157,16 @@ def create_java_provider(scalaattr, transitive_compile_time_jars):
         ),
         transitive_runtime_jars = scalaattr.transitive_runtime_jars,
     )
+
+PlusOneDeps = provider(
```
johnynek: we often put providers in different files to deal with that old bug. Is that unneeded now?

Can we comment this provider a bit more so people can understand a bit more the purpose?

scala/private/rule_impls.bzl
```diff
@@ -1006,36 +1013,6 @@ def scala_test_impl(ctx):
         jars.jars2labels,
     )
 
-    # _scalatest is an http_jar, so its compile jar is run through ijar
```
johnynek: why is this all removed now? Is this because we shouldn't depend on a test anyway?

test_rules_scala.sh
```diff
@@ -904,6 +904,26 @@ test_scala_import_fetch_sources() {
...
+}
+test_compilation_succeeds_with_plus_one_deps_on_for_external_deps() {
+  bazel build --define plus_one_deps=true //test_expect_failure/plus_one_deps/external_deps:a
+}
+test_compilation_succeeds_with_plus_one_deps_on_also_for_exports() {
+  bazel build --define plus_one_deps=true //test_expect_failure/plus_one_deps/exports_deps:a
+}
+test_plus_one_deps_only_works_for_java_info_targets() {
+  #for example doesn't break scala proto which depends on proto_library
+  bazel build --define plus_one_deps=true //test/proto:test_proto
```
johnynek: what do you think about configuring this in the toolchain?

I don't think we have a clear strategy about what is configured where. It seems this is a new direction, no?

scala/private/common.bzl
ittaiz: Don’t know about the old bug but I’ll put it in a different file if that’s the common thing. I also thought about doing it but wasn’t sure.

I’ll also add comments

test_rules_scala.sh
ittaiz: It’s a new direction for rules_scala but this is the current way for starlark rules to have command line configuration.

I’m good with moving it to the toolchain. Will do.

scala/private/rule_impls.bzl
ittaiz: Good question. 
I have an answer which I’m fairly confident about but unfortunately  not 100%.
The answer is that since I added the scalatest label as part of the base classpath it’s jars and labels are collected in the main flow.
I have to say that no tests failed even before I did this so I think this area of scalatest and integration with strict-deps/unused deps isn’t tested well.

Lastly tests can depend on tests, not a problem also in java.

ittaiz:
@johnynek replied to everything. Feel free to reopen if you feel I've missed something.
Also not 100% sure about the documentation I added so open for feedback.

ittaiz:
@johnynek wdyt? Also the breakage is due to jmh reproducibility flakiness

johnynek:
@ittaiz there was a recent change the jmh rules: #695 I wonder if that could be causing the issue.

scala/plusone.bzl
```diff
@@ -0,0 +1,17 @@
+"""
+Keeps direct compile dependencies of targets.
+This enables targets to pass the to compiler the plus one dependencies in addition to the direct ones.
+For motivation of plus one see the e2e tests     9
```
johnynek: trailing `9`.

scala/private/common.bzl
```diff
@@ -22,21 +23,25 @@ def collect_srcjars(targets):
 def collect_jars(
         dep_targets,
         dependency_analyzer_is_off = True,
-        unused_dependency_checker_is_off = True):
+        unused_dependency_checker_is_off = True,
+        plus_one_deps_is_off = True):
```
johnynek: I wonder if we should have so many default values here. especially with the same type... but we can punt that concern I guess.

scala/private/common.bzl
```diff
@@ -58,11 +63,16 @@ def _collect_jars_when_dependency_analyzer_is_off(
...
+        if (not plus_one_deps_is_off) and (PlusOneDeps in dep_target):
+            plus_one_deps_compile_jars.append(
+                depset(transitive = [dep[JavaInfo].compile_jars for dep in dep_target[PlusOneDeps].direct_deps if JavaInfo in dep ])
+            )
+
     return struct(
         compile_jars = depset(transitive = compile_jars),
         transitive_runtime_jars = depset(transitive = runtime_jars),
         jars2labels = JarsToLabelsInfo(jars_to_labels = jars2labels),
-        transitive_compile_jars = depset(),
```
johnynek: huh, this is interesting. This is a change from before. We weren't really tracking transitive compile jars?

scala/private/common.bzl
ittaiz: I definitely agree...
I also thought about it and the worst thing is that they are all negative case.
I think we still need to do a lot of refactoring here but it's hard to keep up, fix bugs, refactor while also doing all of the other d2d stuff at work...
So thank you also for taking a bigger part of the work of this repo. I hope to remove some of my responsibilities in the next few months so I can give some TLC back here (and maybe also get 1-2 people from Wix to work here some more)

scala/private/common.bzl
ittaiz: if strict deps was off we didn't track it

ittaiz:
re recent change in jmh- don't think it's related since my branch predates it

johnynek:
👍

PR #700, opened 2019-02-25
Make the scala_protobuf aspect based

ianoc-stripe:
Working on a scala_protobuf aspect based rules.

These use the toolchain to configure the options instead of being on the rule. Allowing us to build out the shadow graph and share it.

The only part i haven't implemented and am not entirely sure how i should is the java conversions. This would also be a pretty large change for someone to implement. (Though when you'd want to mix/match these options usually in a repo i don't know, so it might not be much of a deal other than deleting some lines).


johnynek:
@ianoc-stripe this is great! thanks.

Did you see if this addresses some concerns from internal uses? Can you comment on them here so people on google can see some of our motivations?

scala_proto/BUILD
```diff
@@ -0,0 +1,39 @@
+load("//scala_proto:scala_proto_toolchain.bzl", "scala_proto_toolchain")
+
+toolchain_type(
+    name = "toolchain_type",
```
johnynek: I freely admit to not understanding this stuff. It seems really weird to me the whole toolchain_type, toolchain, impl of the toolchain.... but I guess maybe they are mapping a typed system onto dynamic values in skylark.

scala_proto/BUILD
```diff
@@ -0,0 +1,39 @@
...
+    visibility = ["//visibility:public"],
+)
+
+scala_proto_toolchain(
+    name = "default_toolchain_impl",
+    with_grpc = False,
+    with_flat_package=False,
+    # with_java=False,
```
johnynek: why the comment? to show the setting?

scala_proto/private/proto_to_scala_src.bzl
```diff
@@ -0,0 +1,49 @@
...
+        output = jar_output.path,
+        included_proto = "-" + ":".join(sorted(["%s,%s" % (f.root.path, f.path) for f in include_proto])),
+        # Command line args to worker cannot be empty so using padding
+        flags_arg = "-" + ",".join(flags),
+        transitive_proto_paths = "-" + ":".join(transitive_proto_paths),
+        # Command line args to worker cannot be empty so using padding
+        # Pass inputs seprately because they doesn't always match to imports (ie blacklisted protos are excluded)
+        inputs = _colon_paths(compile_proto)
+    )
+    print(worker_content)
```
johnynek: can we remove this?

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,269 @@
...
+            d[ScalaPBAspectInfo].proto_info
+            for d in ctx.rule.attr.deps
+        ] + [target_ti],
+    )
+
+    # we sort so the inputs are always the same for caching
+    compile_protos = sorted(target_ti.direct_sources)
+    transitive_protos = sorted(target_ti.transitive_sources)
+
+    print("transitive_descriptor_sets")
```
johnynek: can we remove or make a flag for debugging?

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,269 @@
...
+    manifest = ctx.actions.declare_file(
+        label.name + "_MANIFEST.MF",
+        sibling = scalapb_jar,
+    )
+    write_manifest_file(ctx.actions, manifest, None)
+    statsfile = ctx.actions.declare_file(
+        label.name + "_scalac.statsfile",
+        sibling = scalapb_jar,
+    )
+    print("Compiling %s -> %s" %(label, scalapb_jar))
```
johnynek: remove or make optional.

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,269 @@
...
+    merged_deps = java_common.merge(deps_java_info + implicit_deps)
+
+    # this only compiles scala, not the ijar, but we don't
+    # want the ijar for generated code anyway: any change
+    # in the thrift generally will change the interface and
```
johnynek: change `thrift` to `proto` here.

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,269 @@
...
+        transitive_compile_time_jars = depset(
+            transitive = [merged_deps.transitive_compile_time_jars],
+        ),
+        transitive_runtime_jars = depset(
+            transitive = [merged_deps.transitive_runtime_jars],
+        ),
+    )
+
+####
+# This is applied to the DAG of thrift_librarys reachable from a deps
```
johnynek: `thrift` here.

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,269 @@
...
+# or a scalapb_scala_library. Each thrift_library will be one scalapb
```
johnynek: thrift.

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,269 @@
...
+
+    # if toolchain.with_java:
+    #     flags.append("java_conversions")
+
+    if compile_protos:
+        # we sort so the inputs are always the same for caching
+        compile_proto_map = {}
+        for ct in compile_protos:
+            compile_proto_map[ct] = True
+        include_thrifts = sorted([
```
johnynek: thrift here should be proto.

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,269 @@
...
+            proto_info = transitive_ti,
+            java_info = java_info,
+        ),
+    ]
+
+scalapb_aspect = aspect(
+    implementation = _scalapb_aspect_impl,
+    attr_aspects = ["deps"],
+    attrs = {
+        "_pluck_scalapb_scala": attr.label(
```
johnynek: I don't think this `pluck` name is relevant here.

src/scala/scripts/ScalaPBGenerator.scala
```diff
@@ -31,6 +32,18 @@ object ScalaPBWorker extends GenericWorker(new ScalaPBGenerator) {
...
 
 class ScalaPBGenerator extends Processor {
+  def setupIncludedProto(includedProto: List[(String, String)]): Unit = {
+    includedProto.foreach { case (root, fullPath) =>
+      val rp = Paths.get(root)
+      val op = Paths.get(fullPath)
+      require(op.toFile.exists, s"Path $fullPath does not exist, which it should as a dependency of this rule")
+      val relativePath = rp.relativize(op)
+
+      println(s"Root: rp , op: $op and relativePath: $relativePath")
```
johnynek: comment out?

src/scala/scripts/ScalaPBGenerator.scala
```diff
@@ -31,6 +32,18 @@ object ScalaPBWorker extends GenericWorker(new ScalaPBGenerator) {
...
+      relativePath.toFile.getParentFile.mkdirs
+      java.nio.file.Files.copy(op, relativePath)
```
johnynek: want to import `Files`?

test/proto/BUILD
```diff
@@ -80,39 +79,36 @@ java_proto_library(
 scalapb_proto_library(
     name = "test_proto_java_conversions",
     visibility = ["//visibility:public"],
-    with_flat_package = True,
-    with_java = True,
     deps = [
         ":test2",
-        ":test_proto_java_lib",
+        # ":test_proto_java_lib",
```
johnynek: can we remove the comment.

test/proto/BUILD
```diff
@@ -80,39 +79,36 @@ java_proto_library(
...
     visibility = ["//visibility:public"],
-    with_grpc = True,
     deps = [":test_service"],
 )
 
-scala_proto_srcjar(
-    name = "test1_proto_scala",
-    deps = ["//test/proto2:test"],
-    generator = "@io_bazel_rules_scala//src/scala/scripts:scalapb_generator")
+# scala_proto_srcjar(
```
johnynek: I guess we want to enable these before merging? somehow? I think passing the generator should not be an option, but building the test1 target.

johnynek:
Thanks taking this on.

I took a look, generally it looks like the right direction to me.

The challenge here is that the options are set once for all targets, I think. That may be okay, or may not. Options to aspects didn't have a lot of features when I looked at this with thrift.

@ittaiz can you and your team take a look at this?

scala_proto/BUILD
ianoc-stripe: I don't really know about how much of this is required/needed tbh. I cargo culted this from the scala tool chain

scala_proto/BUILD
ianoc-stripe: Right now this is commented because i haven't figured out how it should be implemented dynamically -- we need to depend on the java aspect only when this is true. And depending on the java aspect itself since its defined in java i haven't found the right magic invokcation for yet

scala_proto/private/proto_to_scala_src.bzl
ianoc-stripe: Yep

ianoc-stripe:
What this is:
Replacing the existing rules with a scala proto library rule similar to the java one. With this we will effectively create synthetic scala protobuf libraries to match every proto_library instance. This means instead of one large target we will have a tree of smaller targets matching the proto_libraries. 

Why/motivation:
Currently if you build a proto file, the transitive set of proto's are all handed to scalac and wind up i the final jar. This has three pretty large downsides:
1) Compiling is super super slow in some of our targets when there is a large/wide dependency tree
2) This means cache hits tend to be pretty poor
3) We have duplicate classes on the classpath from every scala proto library, which makes the jars bigger and is poor hygiene.


Challenges/differences:

Since the scala_proto's now are going to be generated via the aspect we need to share a bunch of the generating options. I've moved these to a toolchain to enable them for a repo. Similar to how this would mostly be used in sbt from scalapb i hope its not a negative impact for folks.

One thing I haven't quite figured out yet is doing the `with_java` option, since effectively for each of our synthetically generated nodes we will need to depend on the corresponding java synthetic node at the same level. If i can get the right provider from the java_proto rule i think i can get it working, though not sure about dynamically enabling aspects like this based on the toolchain configuration.

Also still to be implemented:
proto blacklist, this will move to the toolchain also.

ittaiz:
Indeed thanks a lot!
I’ll take a look tomorrow but I just now skimmed it and I’m concerned about two things:
1. Current rule is decoupled from scalapb and only the macro gives the scalapb wrapping. We’d like to preserve that if possible.
2. Somewhat related to the above is that we need to pass a custom generator. We need to pass it to all targets but one so not sure how we can tackle it.
3. Related to 1, we used to pass the free form string flags which now doesn’t seem like it’s possible

ianoc-stripe:
@ittaiz on (1) you mean you like having the scala_library just referring to the src jar of the scalapb code? or how do you mean decoupled?

ianoc:
@ittaiz did you have a chance to take another look here? If you have a custom generator that needs to get passed to all targets we could do it via a toolchain config setting?

johnynek:
@ianoc-stripe is this still a WIP? Can we merge master?

ianoc:
@johnynek its still WIP just since there's two features not implemented blacklists and java converters, blacklists seem trivial, not entirely sure quite yet how to correctly handle blacklists. But also to understand for folks like @ittaiz that this is directionally usable for him or we need to possibly consider that this will need to be put in as a new set of rules. (Given aspect based stuff implies more global configuration from the tool chain and less local per-target config).

ittaiz:
@simuons is the one leading our internal proto scala work and I asked him to take a look since I think this will severely break us :( 
He took a look but hasn't gotten a chance to respond yet...
I'll try to take a look now as well

johnynek:
what do you think will break you @ittaiz ? Maybe we can focus on that.

Ideally, the majority of old targets can be made to work with no change, and otherwise have a canonical simple change for the rest.

We are really concerned with performance and this was MUCH faster for us (according to what @ianoc-stripe told me offline).

ittaiz:
In general I'm very much in favor of this approach and want to merge it in even if it breaks some APIs and we need to do some work for it but it has to be compatible.
Re generator via toolchain I think that we have one target that is the inception point which needs a vanilla generator but maybe we can solve it via a nested workspace (a bit iffy but might stick)

ianoc-stripe:
Its a little odd seemingly but it might be easy enough to maintain if we aimed for the ability for a global generator + blacklist for said generator. You could blacklist the inception point then?

ittaiz:
I don't remember the blacklist 100% unfortunately.
Let me take a look at the generator first.
Re coupling I mean that the toolchain has predefined flags (with_grpc for example) while we used scala_proto_srcjar which allowed free form flags and a custom generator.
We then had all targets but one use a macro which defined our custom generator and had a single target (don't know why exactly) use the vanilla generator (I think because of a cycle issue).

ittaiz:
We also used the blacklisted_protos obviously :)

ittaiz:
I think nested workspace won't solve the issue since Bazel is source based. 
The code in the nested workspace will use the toolchain of the host workspace when it's run in it...
Sharing here more details need in case you'll have thoughts.
Our inception BUILD.bazel file has the following declarations:
```
load("@io_bazel_rules_scala//scala_proto:scala_proto.bzl", "scala_proto_srcjar")
load("somewhere_from_wix", "GOOGLE_API_WELL_KNOWN_PROTOS", "PROTOBUF_AND_GOOGLE_API_PROTOS", "PROTOBUF_WELL_KNOWN_PROTOS")

proto_library(
    name = "proto",
    srcs = glob(["**/*.proto"]),
    proto_source_root = native.package_name(),
    deps = PROTOBUF_AND_GOOGLE_API_PROTOS,
)

scala_proto_srcjar(
    name = "proto_scala_srcjar",
    blacklisted_protos = PROTOBUF_AND_GOOGLE_API_PROTOS,
    flags = [
        "ascii_format_to_string",
        "flat_package",
        "grpc",
        "single_line_to_proto_string",
    ],
    generator = "@io_bazel_rules_scala//src/scala/scripts:scalapb_generator",
    deps = [":proto"],
)

scala_proto_srcjar(
    name = "google_api_proto_scala_srcjar",
    blacklisted_protos = PROTOBUF_WELL_KNOWN_PROTOS,
    generator = "@io_bazel_rules_scala//src/scala/scripts:scalapb_generator",
    deps = GOOGLE_API_WELL_KNOWN_PROTOS,
)

scala_library(
    name = "proto_scala",
    srcs = [
        ":google_api_proto_scala_srcjar",
        ":proto_scala_srcjar",
    ],
    deps = [
        "@com_google_protobuf_protobuf_java",
        "@com_thesamet_scalapb_lenses_2_12",
        "@com_thesamet_scalapb_scalapb_runtime_2_12",
    ],
)
```
The proto glob catches a few of our infra protos with extensions to proto.
Our macro then adds the above `proto_scala` target to all of the scala_library it creates since all of then generated scala needs them (due to our custom generator).
Not sure exactly what will happen if it will try to use the custom generator. I'll play with it a bit.

ittaiz:
The code of the custom generator needs the generated scala from the protos so indeed no-go.
@ianoc-stripe wdyt about the following solution: Have a "default generator" attribute and another attribute of dictionary from "target label" to "generator label" to allow for inception targets.
Also taking another look I see that every target can configure its own flags which somehow relate to custom generators (its not a bazel label). @simuons can you please give more context on our needs?

johnynek:
@ittaiz if you think you can’t solve your problem with a tool chain option, isn’t that a problem with toolchains?

I still don’t understand why putting the custom generator in the tool chain can’t work.

ittaiz:
@johnynek Not sure I understand what you mean (or maybe I wasn't clear).
We need to have a custom generator for almost all of our targets (all except for two). If you move the generator to the toolchain this applies to all targets in the repo.
I thought about solving it with a nested repository but that won't solve anything thanks to source dependencies (also in external repositories).
The way to I think we can solve this via a toolchain is to specify in the toolchain a default generator (we'll specify the label for our custom one) and a map of targets which need a different generator.
The aspect will then consult with the map to see which generator is needed (something like `map.getOrElse(targetLabel,defaultGeneratorLabel)`).
Not 100% sure how to solve the flags but I want to get a bit more understanding from @simuons so I/we can suggest a solution.
What do you (and @ianoc-stripe) think about these two more attributes in the toolchain?

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,267 @@
...
+    "aspect_info",
+])
+
+
+
+def lift_proto_to_struct(tis):
+    return struct(
+        direct_sources = tis.direct_sources,
+        transitive_sources = tis.transitive_sources,
+        # transitive = depset(transitive = [t.transitive for t in tis]),
```
ittaiz: do we need the commented out lines here and below?

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,267 @@
...
+        # deps = depset(transitive = [t.deps for t in tis]),
+    )
+
+
+def merge_scalapb_aspect_info(scalapbs):
+    return ScalaPBAspectInfo(
+        src_jars = depset(transitive = [s.src_jars for s in scalapbs]),
+        output_files = depset(transitive = [s.output_files for s in scalapbs]),
+        proto_info = merge_proto_infos([s.proto_info for s in scalapbs]),
+        java_info = java_common.merge([s.java_info for s in scalapbs]),
```
ittaiz: did you see how this affects, if at all, on intellij? Not saying it should but just battle scarred.

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,267 @@
...
+    )
+
+####
+# This is applied to the DAG of thrift_librarys reachable from a deps
+# or a scalapb_scala_library. Each thrift_library will be one scalapb
+# invocation assuming it has some sources.
+def _scalapb_aspect_impl(target, ctx):
+    deps = [d[ScalaPBAspectInfo].java_info for d in ctx.rule.attr.deps]
+
+    if ProtoInfo not in target:
```
ittaiz: maybe add a comment when this is the case?

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,267 @@
...
+            flags.append("flat_package")
+
+        if toolchain.with_single_line_to_string:
+            flags.append("single_line_to_proto_string")
+
+        # if toolchain.with_java:
+        #     flags.append("java_conversions")
+
+        if compile_protos:
+            # we sort so the inputs are always the same for caching
```
ittaiz: redundant?

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,267 @@
...
+            compile_proto_map = {}
+            for ct in compile_protos:
+                compile_proto_map[ct] = True
+            include_thrifts = sorted([
+                trans
+                for trans in compile_protos
```
ittaiz: please forgive my python ignorance but when will trans not be in the map?
Two lines above we add every ct in compile_protos in the map and trans also comes from compile_protos

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,267 @@
...
+    # in the thrift generally will change the interface and
+    # method bodies
+    compile_scala(
+        ctx,
+        Label("%s-fast" % (label)),
+        output,
+        manifest,
+        statsfile,
+        sources = [],
+        cjars = merged_deps.transitive_compile_time_jars,
```
ittaiz: shouldn't this be compile_time_jars?

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,267 @@
...
+        deps = deps_java_info + implicit_deps,
+        runtime_deps = deps_java_info + implicit_deps,
+        exports = deps_java_info + implicit_deps,
+        output_jar = output,
+        compile_jar = output,
+    )
+
+def _empty_java_info(deps_java_info, implicit_deps):
+    merged_deps = java_common.merge(deps_java_info + implicit_deps)
+    return java_common.create_provider(
```
ittaiz: this API is deprecated. We need to replace it in other places as well but if we can refrain from adding usages it's best.
Shouldn't you just return merged_deps here?

scala_proto/scala_proto.bzl
```diff
@@ -18,579 +18,61 @@ load(
...
-        "{root},{path}".format(root = _root_path(f), path = f.path)
-        for f in sorted(data)
-    ])
-
-def _retained_protos(inputs, blacklisted_proto_targets):
-    blacklisted_protos = [dep for target in blacklisted_proto_targets for dep in target.proto.transitive_sources]
-    blacklisted_protos_dict = dict(zip(blacklisted_protos, blacklisted_protos))
-    return [f for f in inputs if blacklisted_protos_dict.get(f, None) == None]
-
-def _valid_proto_paths(transitive_proto_path):
```
ittaiz: not needed anymore?

johnynek:
@ittaiz in reply to: https://github.com/bazelbuild/rules_scala/pull/700#issuecomment-468945137

So, it seems to me a weakness of aspects, toolchains or both that this is a challenge here. for instance, I don’t see why we can’t have two different toolchains in use in one build. And aspects also make it hard to have per target options because you have hidden nodes in the aspect generated graph that can’t have custom options.

Cc @katre I think we have talked about this design problem before: are how to handle selecting on a per rule basis toolchains specific options (as Wix seems to need to do).

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,267 @@
...
+            cfg = "host",
+            default = Label("@io_bazel_rules_scala//src/scala/scripts:scalapb_generator"),
+            allow_files = True,
+        ),
+        "_scalac": attr.label(
+            default = Label(
+                "@io_bazel_rules_scala//src/java/io/bazel/rulesscala/scalac",
+            ),
+        ),
+        "_grpc_deps": attr.label_list(
```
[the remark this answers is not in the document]
ianoc: It should be, i'll move it to the toolchain

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,267 @@
...
+            proto_info = transitive_ti,
+            java_info = java_info,
+        ),
+    ]
+
+scalapb_aspect = aspect(
+    implementation = _scalapb_aspect_impl,
+    attr_aspects = ["deps"],
+    attrs = {
+        "_pluck_scalapb_scala": attr.label(
```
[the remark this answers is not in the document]
ianoc: It does, i was in flight travelling yesterday. I have the changes to make this configurable. Was just working on an override generator piece of code and tests.

scala_proto/private/proto_to_scala_src.bzl
```diff
@@ -0,0 +1,48 @@
...
+
+def _root_path(f):
+    if f.is_source:
+        return f.owner.workspace_root
+    return "/".join([f.root.path, f.owner.workspace_root])
+
+
+def _colon_paths(data):
+    return ":".join([
+        f.path
```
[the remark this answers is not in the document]
ianoc: I've just handled this differently, the external locations all work right in the tests i've done. Instead of a big chain of imports in the rule i just move things to where they are expected. For me it removed a bunch of warnings about `external/com_google_..` not existing. Seemed like it made it simpler though i'm not wed to it.

scala_proto/private/proto_to_scala_src.bzl
```diff
@@ -0,0 +1,48 @@
...
+def proto_to_scala_src(ctx, label, compile_proto, include_proto, transitive_proto_paths, flags, jar_output):
+    worker_content = "{output}\n{included_proto}\n{flags_arg}\n{transitive_proto_paths}\n{inputs}".format(
+        output = jar_output.path,
+        included_proto = "-" + ":".join(sorted(["%s,%s" % (f.root.path, f.path) for f in include_proto])),
+        # Command line args to worker cannot be empty so using padding
+        flags_arg = "-" + ",".join(flags),
+        transitive_proto_paths = "-" + ":".join(transitive_proto_paths),
+        # Command line args to worker cannot be empty so using padding
+        # Pass inputs seprately because they doesn't always match to imports (ie blacklisted protos are excluded)
+        inputs = _colon_paths(compile_proto)
```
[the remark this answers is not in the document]
ianoc: I pushed up the blacklist code handling, this should all be fine. I was trying to understand some of the more complex alias handling from before and i don't think we need it

scala_proto/private/scalapb_aspect.bzl
ianoc: Thats fair, let me ensure this is the case. This is likely just a degree of cargo culting from the scrooge aspect

scala_proto/private/scalapb_aspect.bzl
ianoc: Thats fair, i'm not sure if we are aiming to have strict handling of the compile time classpath. Let me try flip it and see if it all still works right

scala_proto/scala_proto.bzl
ianoc: nope, this is about simplifying the handling of the externals

src/scala/scripts/ScalaPBGenerator.scala
```diff
@@ -39,20 +51,19 @@ class ScalaPBGenerator extends Processor {
...
+    setupIncludedProto(extractRequestResult.includedProto)
+
     val config = ScalaPBC.processArgs(extractRequestResult.scalaPBArgs.toArray)
     val code = ProtocBridge.runWithGenerators(
       protoc = a => com.github.os72.protocjar.Protoc.runProtoc(a.toArray),
       namedGenerators = Seq("scala" -> ScalaPbCodeGenerator),
       params = config.args)
 
     try {
-      if (!config.throwException) {
```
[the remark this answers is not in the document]
ianoc: The previous if/else if caused issues where i was having failures but it just produced empty outputs rather than quitting properly.

I believe this configuration handling is an odd spill over from using the ScalaPB arg parsing. It looks like the behavior in scalapb is to let the user use the exit code for error handling or throw a java exception. And it defaults to throwException in the config as false. 

https://github.com/scalapb/ScalaPB/blob/master/scalapbc/src/main/scala/scalapb/ScalaPBC.scala#L67-L72

I don't think we have any use/interest in this config variable unless i'm missing something? since the user doesn't have any direct control/utility for changing how the error handling is done. (right now if `config.throwException` is false, the default, we will always build the jar(though its probably empty), and not throw the error from my read of it).

scala_proto/private/scalapb_aspect.bzl
ianoc: No I rarely enough use intellij, i'm happy to try out some steps on intellij if you know of any common failure modes/usual utility that should work here.

scala_proto/scala_proto_toolchain.bzl
```diff
@@ -0,0 +1,67 @@
...
+#     blacklisted_protos: list of protobuf targets to exclude from recursive building
+#     code_generator: what code generator to use, usually you'll want the default
+scala_proto_toolchain = rule(
+    _scala_proto_toolchain_impl,
+    attrs = {
+        "with_grpc": attr.bool(),
+        "with_flat_package": attr.bool(),
+        "with_single_line_to_string": attr.bool(),
+        "blacklisted_protos": attr.label_list(default=[]),
+        "code_generator": attr.label(
```
ianoc-stripe: So after some `bazel clean` action this seems to be a problem -- if i have the code generator here, and its an executable dependency. The runfiles aren't available to the actions generated in the aspect. If i add it as a dep to the aspect it works. Anyone know if there is some syntax i'm missing here to thread a proper dependency onto the target via the tool chain on the generator code?

simuons:
I might be late to the party but I'll try anyway.

First of all thanks for taking this!

I'm a bit worried that proposed toolchain is tightly coupled to scalapb. Reason is that we have our own generators (for example we generate validators for each message based on annotations in proto file or generate tagless-final style services etc) and for those generators scalapb flags are not really relevant or they might have their own flags. With `scala_proto_srcjar` we had that flexibility to specify flags and I think it will be gone with new toolchain.

Have you seen [proto_lang_toolchain](https://docs.bazel.build/versions/master/be/protocol-buffer.html#proto_lang_toolchain)? It's supposed to be the toolchain for all `LANG_proto_library`. Unfortunately it cannot be used from skylark now ([according to this](https://groups.google.com/forum/#!searchin/bazel-discuss/proto_lang_toolchain%7Csort:date/bazel-discuss/xoy98a4yhAA/hD9eDus9BAAJ)). It is more generic so that you can set flags for command line, pass plugin/generator and and pass runtime against which generated code should be implemented. 

Maybe `scala_proto_toolchain` should mimic `proto_lang_toolchain` and eventually switch to it once available in skylark?

johnynek:
What if we follow the arg style we use with scalacprocessor? In that case we pass a key value list which give us named args.

In that case, a custom generator can ignore args it doesn’t care about? Would that work?

Secondly, I wonder if assuming you can replace the generator is a good assumption without specify more how args are to be passed and what the output is. I’m worried if we feel that toolchains and aspects won’t work and so we avoid them just because in fact you want a different backend which I think would work with a toolchain?

Lastly, since we expose the skylark couldn’t you have two differently named rules, one uses the generator you want, one uses the standard generator? Then you could pick two different toolchains potentially?

scala_proto/private/scala_proto_default_repositories.bzl
```diff
@@ -0,0 +1,392 @@
...
+        scala_version = _default_scala_version(),
+        maven_servers = ["http://central.maven.org/maven2"]):
+    major_version = _extract_major_version(scala_version)
+
+    native.maven_server(
+        name = "scala_proto_deps_maven_server",
+        url = "http://central.maven.org/maven2/",
+    )
+
+    native.maven_jar(
```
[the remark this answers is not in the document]
dkelmer: Thanks for pointing it out @ittaiz! I am planning on deprecating it, but it is currently blocked on finding a satisfying alternative to maven_server. 

Though in this case, I think the following would work instead of both the native.maven_server and native.maven_jar rules:

```
jvm_maven_import_external(
  name = "scala_proto_rules_protoc_jar",
  artifact = "com.github.os72:protoc-jar:3.6.0.1",
  server_urls = ["http://central.maven.org/maven2/"], 
  artifact_sha256 = "44d513ece5a7d9b09f2d865653d82c1d1556900b2450f0e7d629c3bc13ee0e13",
  licenses = ["notice"] # Apache 2.0
)
```

ianoc-stripe:
I would like as much as possible to make it easy for folks to use it, so free form flag fields feel like a minefield -- indeed the existing rules are broken in the conversion of flags to the scalapb ones. So the more failures it causes imo the better that the flag doesn't exist sooner. In regards the generic tool chain, once its fully in skylark i think we can look at it, but most of these API's are not stable so i'm pretty slow to want to try emulate something in flight we can't use.

ianoc-stripe:
So we will require bazel 0.23.0 for this, but now the tool can properly be configured on the toolchain.

Needs https://github.com/bazelbuild/bazel/commit/5a20a447514f42e0f071c800f53d99c4286c297d
until this it doesn't seem like it worked.

i've also made the changes to just use the google protoc now too.

With this change we can look closer at what would be required to support a custom generator for some specific labels

ianoc:
ok @ittaiz see what you think now maybe. With this you can specify a second code generator, and then a set of targets to give to that code generator. Would that work for Wix?

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,235 @@
...
+
+scalapb_aspect = aspect(
+    implementation = _scalapb_aspect_impl,
+    attr_aspects = ["deps"],
+    required_aspect_providers = [
+        [ProtoInfo],
+        [ScalaPBImport],
+    ],
+    attrs = {
+        "_protoc": attr.label(executable = True, cfg = "host", default = "@com_google_protobuf//:protoc"),
```
johnynek: @simuons this is the same change you have, do you agree?

scala_proto/scala_proto_toolchain.bzl
johnynek: this is resolved by 0.23, right?

ianoc: Yep

src/scala/scripts/PBGenerateRequest.scala
```diff
@@ -2,48 +2,38 @@ package scripts
 
 import java.nio.file.{Files, Path, Paths}
 
-class PBGenerateRequest(val jarOutput: String, val scalaPBOutput: Path, val scalaPBArgs: List[String], val protoc: Path)
+class PBGenerateRequest(val jarOutput: String, val scalaPBOutput: Path, val scalaPBArgs: List[String], val includedProto: List[(String, String)], val protoc: Path)
```
johnynek: wonder why this isn't a case class and we can remove all the `val`s...

src/scala/scripts/PBGenerateRequest.scala
```diff
@@ -2,48 +2,38 @@ package scripts
...
-    // This is to match the expected behavior of
-    // proto_library and java_proto_library where proto files
-    // can import other proto files using only the relative path
-    val imports = parsedProtoFiles.map { case (relPath, absolutePath) =>
-      s"-I$relPath=$absolutePath"
-    }
     val protoFiles = args.get(4).split(':')
+    val includedProto = args.get(1).drop(1).split(':').distinct.map { e =>
+      val p = e.split(',')
+      (p(0), p(1))
```
johnynek: should we just use `Path` here so we don't have to do the conversion later?

test/proto/BUILD
```diff
@@ -80,39 +146,47 @@ java_proto_library(
...
-    with_grpc = True,
-    deps = [":test_service"],
+    deps = [
+        ":blacklisted_proto",
+        ":test_service",
+        "@io_bazel_rules_scala//test/proto:other_blacklisted_proto",
+    ],
 )
 
-scala_proto_srcjar(
```
johnynek: can we no longer directly exercise srcjar?

scala_proto/private/scalapb_aspect.bzl
simuons: Yes it is.

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,235 @@
...
+            if(lbl.label == target_absolute_label):
+                compile_protos = False
+
+        code_generator = toolchain.code_generator
+        for lbl in toolchain.override_code_generator_targets:
+            if(lbl.label == target_absolute_label):
+                code_generator = toolchain.override_code_generator
+
+
+        for lbl in toolchain.override_code_generator_targets:
```
simuons: This is identical to three lines above. I guess it should be removed.

ittaiz:
@ianoc @johnynek,
The PR currently doesn't work for us.
Previous implementation allowed for very low level free form manipulation that we used.
I think that @ianoc went out of his way to get the PR to work for our needs but because he doesn't have these pains (and because we weren't very good at explaining our needs) I think this is becoming a bit too hard
Because we don't want to block you more what we suggest is that we'll copy over the existing implementation to our own code and we'll work on a solution to our needs based on rules_scala master (with the aspect).
This will allow you to merge, allow Wix to continue to upgrade rules_scala and allow us to formulate a sound solution to our needs (which we think are general when you use proto in a large organization)
WDYT?
In general our needs are around being able to have multiple and custom generators per target and possible allowing these generators to have custom flags.

On a more abstract note- I have to say that I'm not sure how this should have turned out. One the one hand there is a significant advancement (written above) but on the other hand this is breaking existing features that work for us. Don't know...

ianoc-stripe:
@ittaiz instead of an override generator if that doesn't quite fit the bill -- how about we have a `scala_proto_srcjar` target similar to what we had before with just the free form flags for the generator in the args rule. And in that target though we assert that the target is in the toolchain's blacklist?

I'm not entirely sure tbh still of your needs here, but i think then you could construct the core targets anyway you need, followed by adding those targets into the toolchains dependency set so all the toolchain generated targets depend on them?

test/proto/BUILD
ianoc: Nope, we could bring it back if it was useful, though since the aspect generates one internally they would be duplicates of some form

src/scala/scripts/PBGenerateRequest.scala
ianoc: Yep will update

src/scala/scripts/PBGenerateRequest.scala
ianoc: will try make that change to make it shorter

scala_proto/private/scalapb_aspect.bzl
ianoc: 👍

ittaiz:
I hope @simuons will give a more thorough description of our needs but basically I under-represented the complexity. The core targets are a small issue.
The bigger issue is that we allow for many different additional custom generators. You can add tagless-final for FP oriented code, you can add accord validation and others.
The current approach is via a sort of dynamic decorator which takes a list of arguments (hence the need of free form flags in each target), splits them to scalapb args and additional generators list and according to the list runs additional generators.
This means that every target decides on the additional generators.
@simuons and his team are now trying to think and evaluate what is the best approach for this in bazel.
Their current approach, getting rid of the decorator generator in favor of multiple top level generators (multiple --out) or maybe having multiple protoc actions each with its own generator.

One core requirement is to allow the introduction of a custom generator without it needing to be blessed by the infra team. Another desired feature (not must) is to avoid generating all generators if a target doesn't want it. Not 100% sure about the latter.

One question though- I seem to remember that the aspect can read the attributes of the target it's inspecting. Why not have custom flags attribute on the rule?

ianoc-stripe:
@ittaiz there is a tree of targets generated that are shared by multiple calls to the scala proto library rule. So the rule we are building it for is the proto_library not the scala_proto_library. the scala_proto_library expresses then a dependency on one of the generated by aspect targets. So it becomes difficult to say how a shared target should be built based on different args passed for generators

johnynek:
@ittaiz the aspect mechanism does not allow for custom flags per rule. I think you could attach custom flags to the proto and have it propogate, but the whole idea is that each proto rule gets mapped to a scala proto src jar. There is no place to hook custom args, that I know of, onto the proto.

I don't know why you can't use one generator for everyone and control the behavior by adding annotations to the proto itself (or in some comment).

At twitter, which was a pretty large org when I was there, we only used 1 generator for thrift. We did not have some concept that each rule would pass bespoke options. We tried to push any options into the thrift sources, which you could control. I'm a bit skeptical that this must be solved by passing custom generators to proto rules, and I'm concerned that design clashes with aspects for the reason I mention above, so is somewhat ill-suited to a bazel best practice of aspects for code generation.

scala_proto/private/scalapb_aspect.bzl
```diff
@@ -0,0 +1,230 @@
...
+    )
+    merged_deps = java_common.merge(deps_java_info + implicit_deps)
+
+    # this only compiles scala, not the ijar, but we don't
+    # want the ijar for generated code anyway: any change
+    # in the proto generally will change the interface and
+    # method bodies
+    compile_scala(
+        ctx,
+        Label("%s-fast" % (label)),
```
johnynek: starlark really needs some kind of gensym thing that makes sure generated symbols are unique.

scala_proto/scala_proto_toolchain.bzl
```diff
@@ -0,0 +1,68 @@
...
+            default = Label("@io_bazel_rules_scala//src/scala/scripts:scalapb_generator"),
+            allow_files=True
+        ),
+        "override_code_generator": attr.label(
+            executable = True,
+            cfg = "host",
+            default = Label("@io_bazel_rules_scala//src/scala/scripts:scalapb_generator"),
+            allow_files=True
+        ),
+        "override_code_generator_targets": attr.label_list(
```
johnynek: should we remove this override code if @ittaiz says it won't be useful for them anyway? simplify the code?

ittaiz:
@johnynek thanks for your inputs. Given the generic proto toolchain @simuons found I have a feeling we can find a solution in bazel.
We'll see.

scala_proto/private/proto_to_scala_src.bzl
```diff
@@ -0,0 +1,49 @@
...
+        inputs = _colon_paths(compile_proto),
+        protoc = ctx.executable._protoc.path
+    )
+    argfile = ctx.actions.declare_file(
+        "%s_worker_input" % label.name,
+        sibling = jar_output,
+    )
+    ctx.actions.write(output = argfile, content = worker_content)
+    ctx.actions.run(
+        executable = code_generator.files_to_run,
```
tferi-da: @johnynek even after registering the default toolchain in my WORKSPACE, I'm getting the following error:
```
ERROR: /private/var/tmp/_bazel_feri/27c97e79ab5b7e8bcdeb0d679a4bc60c/external/com_google_protobuf/BUILD:248:2: in @io_bazel_rules_scala//scala_proto/private:scalapb_aspect.bzl%scalapb_aspect aspect on proto_library rule @com_google_protobuf//:any_proto:
	File "/private/var/tmp/_bazel_feri/27c97e79ab5b7e8bcdeb0d679a4bc60c/external/com_google_protobuf/BUILD", line 248
		@io_bazel_rules_scala//scala_proto/private:scalapb_aspect.bzl%scalapb_aspect(...)
	File "/private/var/tmp/_bazel_feri/27c97e79ab5b7e8bcdeb0d679a4bc60c/external/io_bazel_rules_scala/scala_proto/private/scalapb_aspect.bzl", line 175, in _scalapb_aspect_impl
		proto_to_scala_src(ctx, target.label, code_generator, com..., <4 more arguments>)
	File "/private/var/tmp/_bazel_feri/27c97e79ab5b7e8bcdeb0d679a4bc60c/external/io_bazel_rules_scala/scala_proto/private/proto_to_scala_src.bzl", line 39, in proto_to_scala_src
		ctx.actions.run(executable = code_generator.file..., <6 more arguments>)
expected value of type 'File or string' for parameter 'executable', in method call run(FilesToRunProvider executable, list inputs, list outputs, string mnemonic, string progress_message, dict execution_requirements, list arguments) of 'actions'
```

tferi-da: @gseitz FYI

ianoc: @tferi-da This needs a fix from a bazel update -- the latest rules require bazel 0.23 or above now i'm afraid

tferi-da: @ianoc thanks for the tip!

PR #702, opened 2019-02-27
Ignore JMH in reproducibility tests

andyscott:
Uses .bazelignore to ignore the JMH paths during the reproducibility test.

This solution was quick and easy. Alternatively, you could perhaps filter targets by tags (and add tags on the jmh targets).

### DIFF
--- a/test_reproducibility.sh
+++ b/test_reproducibility.sh
@@ -16,6 +16,7 @@ non_deploy_jar_md5_sum() {
 }
 
 test_build_is_identical() {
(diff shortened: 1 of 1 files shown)


ittaiz:
Something seems to be wrong there

andyscott:
It's totally not working, yet ;)

johnynek:
we should diagnose what is going on there, but at least we don't depend on jmh in other targets, so I think it shouldn't really be a problem.

johnynek:
seems that #791 handled this.

Issue #690, opened 2019-02-13
Usage of Worker/Processor and protoc-bridge/protoc-jar in proto rules


simuons:
Hi,

what is the reason/rationale for proto rules being implemented with `Worker`/`Processor` and `protoc-bridge`/`protoc-jar`

From my understanding current implementation has few "drawbacks"

1. `@com_google_protobuf//:protoc` is ignored. It's used by `proto_library` but `scalapb_proto_library` uses one of the https://github.com/os72/protoc-jar/tree/master/bin
    a. it requires extra work/care to keep both protoc at same version
2. it requires extra hops to generate scala code
    a. bazel invokes scala via `Worker`/`Processor`
    b. protoc_bridge is invoked which
            i. extracts own `protoc` to temp directory https://github.com/os72/protoc-jar/blob/master/src/main/java/com/github/os72/protocjar/Protoc.java#L197
            ii. starts new `protoc` process https://github.com/os72/protoc-jar/blob/master/src/main/java/com/github/os72/protocjar/Protoc.java#L125
    c. protoc calls scala plugin to generate code
3. to use custom plugin we had to copy `Worker` and `Processor` to supply it

As I know there is an aim to use persistent worker. But I guess because new process is spawned benefits of persistent worker cannot be used (my knowledge is limited on this topic).

I have a POC https://github.com/simuons/rules_scala/blob/scala-proto-gen/scala_proto/scala_proto.bzl#L595 which
1. uses `@com_google_protobuf//:protoc` from bazel (the same as `proto_library` uses)
2. calls `protoc` directly with supplied plugin and is based on descriptor sets to avoid `--proto_path` construction (mentioned here https://blog.bazel.build/2017/02/27/protocol-buffers.html # Descriptor Sets)
3. needs less code to supply own plugin https://github.com/simuons/rules_scala/blob/scala-proto-gen/src/scala/scripts/ScalaPBPlugin.scala to a rule https://github.com/simuons/rules_scala/blob/scala-proto-gen/test/proto/BUILD#L124

What do you think? If all I wrote makes any sense I'd like to proceed with that.

johnynek:
This all makes sense to me.

@azymnis contributed the original version and I think he just wanted to get unblocked. Ideally, I'd love to have only 1 version of the rules, and not break previous builds. So, if we could improve without breakage, that would be great, e.g. not breaking any of the current tests.

[...]

PR #705, opened 2019-03-01
WIP: Generate scala code from protobuf using @com_google_protobuf//:protoc

simuons:
Resolves https://github.com/bazelbuild/rules_scala/issues/690

Introduces new rule `scala_proto_gen` which
1. generates scala code using implicit `@com_google_protobuf//:protoc`. This means single version of `protoc` for `proto_library` and `scalapb_proto_library` (instead of one which comes from `protoc-jar`)
2. accepts plugin as an executable which implements [plugin.proto](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/plugin.proto) contract. This means less code for us to supply own scala plugins

`scala_proto_srcjar` was left intact so it wouldn't break existing users. My suggestion is to deprecate it and remove with subsequent pr.

`scala_proto_gen` generates code in same way as `scala_proto_srcjar` does ie all direct and transitive deps are generated into single srcjar (no aspects in this pr)

 `scalapb_proto_library` is implemented on top of `scala_proto_gen`. One thing I don't like though is that I had to create "private" rule to support JavaInfo in  `scalapb_proto_library` deps (maybe there is a better way to achieve that). I believe `scalapb_proto_library` deps should accept only `proto_library` as `java_proto_library` does.

`--proto_path` were replaced with `--descriptor_set_in` this removes burden of constructing `-I` aliases.

I haven't done benchmarks other than `bazel clean` followed by `bazel test //test/proto/...`. I got same times in this pr and master. I don't think this pr has impact on performance. In current implementation protoc process is launched from worker which in turn calls executable plugin and I think in actual code generation phase workers are not used (I might be wrong here).


simuons:
BTW I'm going to test this on our codebase. I should have done it before opening PR but better later than never I guess.

johnynek:
Looks like we have two racing proto rewrites. Can you and @ianoc-stripe look at each other’s changes and comment on what aspects from each should be used?

#

johnynek:
See

johnynek:
See #700

scala_proto/scala_proto.bzl
```diff
@@ -594,3 +598,109 @@ def scalapb_proto_library(
...
+    roots = depset([f for dep in protos for f in dep.proto.transitive_proto_path]).to_list()
+    inputs = depset([_strip_root(f, roots) for f in _retained_protos(sources, ctx.attr.blacklisted_protos)]).to_list()
+
+    srcdotjar = ctx.actions.declare_file("_" + ctx.label.name + "_src.jar")
+
+    ctx.actions.run(
+        inputs = [ctx.executable._protoc, ctx.executable.plugin] + descriptors,
+        outputs = [srcdotjar],
+        arguments = [
+            "--plugin=protoc-gen-scala=" + ctx.executable.plugin.path,
```
johnynek: This is going to prevent using workers right? So you have to spin up a new JVM for each file, right?

ittaiz:
@johnynek I think @simuons needs to upload a diagram of our (rules_scala) current process map when using proto. 
He explained it to me and AFAIU this will be a net performance improvement.
I do however agree that we need to see how we have both rewrites.
@simuons I think you should consider yielding to @ianoc since he was first unless you think yours greatly simplifies his (which I don't see but might be wrong).

simuons:
@ittaiz @johnynek I didn't mean to race/interfere with aspects PR. I'm trying to solve different issue here (ie use single `@com_google_protobuf//:protoc`) which is orthogonal to aspects. I'd say aspects PR should move forward and ignore this one. I still need to investigate performance impact. 

PS thanks for your time and sorry for delayed response.

scala_proto/scala_proto.bzl
simuons: Yes it would prevent using workers. I thought this is what is happening now as well. Now I think my assumptions might be wrong as `protoc-bridge` is doing some tricks to reuse jvm. I'll dig more and update.

ittaiz:
@lberki @philwo
May I ask if java rules reuse workers for java generation from proto or
does this just work natively without opening a jvm?

simuons:
Apparently my assumptions about worker usage when generating scala code were wrong. Code generation indeed happens in worker by the trick of `protoc-bridge`. `protoc` is supplied with a plugin that pipes it's inputs/outputs to files https://github.com/scalapb/protoc-bridge/blob/master/src/main/scala/protocbridge/frontend/PosixPluginFrontend.scala#L50 and worker reads/writes to them.

Closing this in favour of less aggressive attempt to reuse `@com_google_protobuf//:protoc` https://github.com/bazelbuild/rules_scala/pull/708

lberki:
@ittaiz : iff something goes through `java_common.compile`, it should use workers. AFAIU the Scala plugin is written in Scala, so I suppose then no workers will be used? (modulo, of course, `protoc-bridge`, which I don't really understand...)

PR #708, opened 2019-03-04
Reuse @com_google_protobuf//:protoc when generating scala from proto

simuons:
Remove `protoc-jar` which brings own versions of `protoc`

### DIFF
--- a/src/scala/scripts/PBGenerateRequest.scala
+++ b/src/scala/scripts/PBGenerateRequest.scala
@@ -2,7 +2,7 @@ package scripts
 
 import java.nio.file.{Files, Path, Paths}
 
-class PBGenerateRequest(val jarOutput: String, val scalaPBOutput: Path, val scalaPBArgs: List[String], val protoc: String)
+class PBGenerateRequest(val jarOutput: String, val scalaPBOutput: Path, val scalaPBArgs: List[String], val protoc: Path)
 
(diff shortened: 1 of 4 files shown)


simuons:
Resolves https://github.com/bazelbuild/rules_scala/issues/690

src/scala/scripts/PBGenerateRequest.scala
```diff
@@ -2,7 +2,7 @@ package scripts
 
 import java.nio.file.{Files, Path, Paths}
 
-class PBGenerateRequest(val jarOutput: String, val scalaPBOutput: Path, val scalaPBArgs: List[String])
+class PBGenerateRequest(val jarOutput: String, val scalaPBOutput: Path, val scalaPBArgs: List[String], val protoc: String)
```
johnynek: Should protoc be a Path?

simuons: Changed to `Path`

simuons:
@johnynek yes this is orthogonal to @ianoc-stripe's changes. These changes are not in his PR but he can take them and incorporate to his PR. Your call, I have no objections.

simuons:
One side effect is that messages like

johnynek:
cc @ianoc-stripe this seems like a good thing to copy in the aspect version.

PR #709, opened 2019-03-04
Unify protobuf-java versions

simuons:
Looks like there are two versions of `protobuf-java`. One comes from `@com_google_protobuf//:protobuf_java` version `3.6.1.3` and another one from `_scala_maven_import_external` version `3.1.0`. There are warnings which comes from `protobuf-java 3.1.0` although `3.6.1.3` should be used.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -97,13 +97,6 @@ http_archive(
     strip_prefix = "protobuf-3.6.1.3",
 )
 
-http_archive(
-    name = "com_google_protobuf_java",
-    sha256 = "9510dd2afc29e7245e9e884336f848c8a6600a14ae726adb6befdb4f786f0be2",
-    urls = ["https://github.com/protocolbuffers/protobuf/archive/v3.6.1.3.zip"],
-    strip_prefix = "protobuf-3.6.1.3",
-)
-
 new_local_repository(
     name = "test_new_local_repo",
     build_file_content =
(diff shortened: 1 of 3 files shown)


johnynek:
It is a real shame we can’t use a tool here. The fact that bazel and the community hasn’t settled on one tool for doing maven resolutions is a shame...

johnynek:
I didn’t look at the Travis failure yet.

johnynek:
seems like a really good and simple change. I'm going to merge this one.

andrew-stripe:
I'm seeing the following error in one of our builds that goes away when I revert this change:
```ERROR: /mnt/tmpfs/bazel/bazel/external/io_bazel_rules_scala/scala/scala.bzl:471:5: no such package '@com_google_protobuf//': The repository could not be resolved and referenced by '//external:io_bazel_rules_scala/dependency/com_google_protobuf/protobuf_java'```

Maybe the `_scala_maven_import_external` should be retained and only the `http_archive` removed?

johnynek:
you need to declare `@com_google_protobuf` my guess is the internal build in question is not doing that.

johnynek:
as you can see, we do so in the workspace of these rules:

https://github.com/bazelbuild/rules_scala/pull/709/files#diff-09498dbadf45966909850dc8a47ebb13R93

PR #710, opened 2019-03-05
Update bazel toolchains pin and add toolchain config target

nlopezgi:
New target will be used by BuildKite CI rbe_ubuntu1604 config


nlopezgi:
fyi @ittaiz

johnynek:
thanks for the PR.

PR #714, opened 2019-03-08
use grpc 1.19

tferi-da:
grpc 1.18 causes a warrning to displayed on JVM versions above 8:
`WARNING: An illegal reflective access operation has occurred`
This is fixed with 1.19.


johnynek:
Thank you for the contribution!

PR #715, opened 2019-03-12
Replace maven_jar with scala_maven_import_external

andrew-stripe:
`native.maven_jar` does not properly support HTTP(S)_PROXY environment variable. This prevents us from using `rules_scala` in certain network-restricted environments. This PR replaces `native.maven_jar` with `scala_maven_import_external` which does respect proxy environment variables.

I had a few questions:
* How confident are you that test would catch issues with this change? I'll need to manually validate any part of this without clear testing.
* There are a few `maven_jar` references in the WORKSPACE file that I didn't replace because comments describe them as specifically for testing. Should I swap out those too?
* There are some expectation failures in `//test/aspect`. I can't tell if these are real failures or just a consequence of changing how deps are pulled. The log output is:
```
Expected these deps from scala_junit_test:
//test/aspect:scala_junit_test, @io_bazel_rules_scala_junit_junit//jar:jar, @io_bazel_rules_scala_org_hamcrest_hamcrest_core//jar:jar, @io_bazel_rules_scala_scala_library//:io_bazel_rules_scala_scala_library,
but got these instead:
//test/aspect:scala_junit_test, @io_bazel_rules_scala_junit_junit//:io_bazel_rules_scala_junit_junit, @io_bazel_rules_scala_org_hamcrest_hamcrest_core//:io_bazel_rules_scala_org_hamcrest_hamcrest_core, @io_bazel_rules_scala_scala_library//:io_bazel_rules_scala_scala_library
Expected these deps from scala_specs2_junit_test:
//test/aspect:scala_specs2_junit_test, @io_bazel_rules_scala//specs2:specs2, @io_bazel_rules_scala_junit_junit//jar:jar, @io_bazel_rules_scala_org_hamcrest_hamcrest_core//jar:jar, @io_bazel_rules_scala_org_specs2_specs2_junit//:io_bazel_rules_scala_org_specs2_specs2_junit, @io_bazel_rules_scala_scala_library//:io_bazel_rules_scala_scala_library, @io_bazel_rules_scala_scala_library//:io_bazel_rules_scala_scala_library, @io_bazel_rules_scala_scala_parser_combinators//:io_bazel_rules_scala_scala_parser_combinators, @io_bazel_rules_scala_scala_reflect//:io_bazel_rules_scala_scala_reflect, @io_bazel_rules_scala_scala_xml//:io_bazel_rules_scala_scala_xml,
but got these instead:
//test/aspect:scala_specs2_junit_test, @io_bazel_rules_scala//specs2:specs2, @io_bazel_rules_scala_junit_junit//:io_bazel_rules_scala_junit_junit, @io_bazel_rules_scala_org_hamcrest_hamcrest_core//:io_bazel_rules_scala_org_hamcrest_hamcrest_core, @io_bazel_rules_scala_org_specs2_specs2_junit//:io_bazel_rules_scala_org_specs2_specs2_junit, @io_bazel_rules_scala_scala_library//:io_bazel_rules_scala_scala_library, @io_bazel_rules_scala_scala_library//:io_bazel_rules_scala_scala_library, @io_bazel_rules_scala_scala_parser_combinators//:io_bazel_rules_scala_scala_parser_combinators, @io_bazel_rules_scala_scala_reflect//:io_bazel_rules_scala_scala_reflect, @io_bazel_rules_scala_scala_xml//:io_bazel_rules_scala_scala_xml
```

### DIFF
--- a/jmh/jmh.bzl
+++ b/jmh/jmh.bzl
@@ -1,57 +1,73 @@
 load("//scala:scala.bzl", "scala_binary", "scala_library")
+load(
+    "@io_bazel_rules_scala//scala:scala_maven_import_external.bzl",
+    _scala_maven_import_external = "scala_maven_import_external",
+)
 
-def jmh_repositories():
-    native.maven_jar(
+def jmh_repositories(maven_servers = ["http://central.maven.org/maven2"]):
+    _scala_maven_import_external(
         name = "io_bazel_rules_scala_org_openjdk_jmh_jmh_core",
         artifact = "org.openjdk.jmh:jmh-core:1.20",
-        sha1 = "5f9f9839bda2332e9acd06ce31ad94afa7d6d447",
+        jar_sha256 = "1688db5110ea6413bf63662113ed38084106ab1149e020c58c5ac22b91b842ca",
+        licenses = ["notice"],
+        server_urls = maven_servers,
     )
     native.bind(
         name = "io_bazel_rules_scala/dependency/jmh/jmh_core",
         actual = "@io_bazel_rules_scala_org_openjdk_jmh_jmh_core//jar",
     )
-    native.maven_jar(
+    _scala_maven_import_external(
         name = "io_bazel_rules_scala_org_openjdk_jmh_jmh_generator_asm",
         artifact = "org.openjdk.jmh:jmh-generator-asm:1.20",
-        sha1 = "3c43040e08ae68905657a375e669f11a7352f9db",
+        jar_sha256 = "2dd4798b0c9120326310cda3864cc2e0035b8476346713d54a28d1adab1414a5",
+        licenses = ["notice"],
+        server_urls = maven_servers,
     )
     native.bind(
         name = "io_bazel_rules_scala/dependency/jmh/jmh_generator_asm",
         actual = "@io_bazel_rules_scala_org_openjdk_jmh_jmh_generator_asm//jar",
     )
-    native.maven_jar(
+    _scala_maven_import_external(
         name = "io_bazel_rules_scala_org_openjdk_jmh_jmh_generator_reflection",
         artifact = "org.openjdk.jmh:jmh-generator-reflection:1.20",
-        sha1 = "f2154437b42426a48d5dac0b3df59002f86aed26",
+        jar_sha256 = "57706f7c8278272594a9afc42753aaf9ba0ba05980bae0673b8195908d21204e",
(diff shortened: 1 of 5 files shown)


andrew-stripe:
It also looks like I have a signficant rebase in scala_proto.bzl to work on, tackling that now.

andrew-stripe:
Rebased, should be ready for initial review with the caveat that I had a question about the failing tests.

johnynek:
looks like the `//test/aspect:aspect_test` test fails. Does it pass for you locally?

I'm fine with this change, but note, the intent of those function was to make it easy for repos to make their own choices. Those dependency functions are intended to be customized. We could make any version of that function we want internally.

I think this is worth doing, but it is unlikely we can satisfy all users (in terms of versions and config), so that's why the functions exist: clearly demark a customization point.

andrew-stripe:
@johnynek: That test fails for me locally as well but I'm not quite sure how to interpret the results. Can you tell from the logs I pasted in the description whether it's a change you'd expect from importing the dependencies in a different way or likely something I broke?

johnynek:
So, it looks like the target names have changed. The maven_jar makes `jar:jar` dependencies, but the scala import seems to make a different name.

So, we have to consider:
1. do we care that these targets changed names, or are they only used in tests. I don't recall, but bazel query could answer some of the question potentially.
2. do we care about these names changing for downstream users who may be broken next time they bump if the were referring to these jars. Put another way, are these names part of the API? I guess I would tend to say no.

I still consider the most expedient path to fixing stripe's concerns is to make a custom version of these dependency files.

@ittaiz do you have any comments?

johnynek:
I guess since we have used the binds explicitly to make those who we depend on things, so my inclination would be to change the test to depend on the new strings.

andrew-stripe:
I'm not familiar with the IntelliJ integration here unfortunately (I don't do much Scala dev at all, just trying to fix some infra). If you think it's necessary to merge, is there anywhere I can look for guidance?

ittaiz:
It's mainly about opening Intellij and seeing you can edit. @johnynek any chance you can ask one of your colleagues who uses IJ to try this branch?
We can't check it since we're blocked on upgrade due to the proto issue

andrew-stripe:
I was able to load and run a test in IntelliJ in one of our repos that invokes scala_proto_repositories. Is that sufficient validation or do you need me to check something else?

ittaiz:
Sounds good. Thanks!

PR #716, opened 2019-03-18
Get jar label from manifest stamp when present

ceason:
This PR gets a jar's label from the 'Target-Label' MANIFEST.MF attribute when available (this is what `java_common.{run_ijar,stamp_jar}` both populate).

### DIFF
--- a/third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/DependencyAnalyzer.scala
+++ b/third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/DependencyAnalyzer.scala
@@ -1,8 +1,11 @@
 package third_party.dependency_analyzer.src.main.io.bazel.rulesscala.dependencyanalyzer
 
+import java.util.jar.JarFile
+
 import scala.reflect.io.AbstractFile
 import scala.tools.nsc.plugins.{Plugin, PluginComponent}
 import scala.tools.nsc.{Global, Phase}
+import scala.util.Try
 
 class DependencyAnalyzer(val global: Global) extends Plugin {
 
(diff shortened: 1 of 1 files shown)


third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/DependencyAnalyzer.scala
```diff
@@ -77,6 +80,19 @@ class DependencyAnalyzer(val global: Global) extends Plugin {
 
   }
 
+  private def getJarLabel(jarPath: String): Option[String] = {
+    // First check the jar's manifest for a stamped label
+    Try(new JarFile(jarPath)).toOption.flatMap { jf ⇒
+      try {
```
johnynek: can we change to:
```scala
try {
  val jf = new JarFile(jarPath)
  ...
```
rather than `Try().toOption` since we have a try following anyway? Seems a bit clearer

johnynek:
looks like this breaks some tests we have around unused dependencies.

Can you inspect those errors and see about fixing those tests? It may be that the test is over strict, or it may be that something really broke.

ittaiz:
I’ll take a look at the impl but in general this is definitely something I
want to do (switch to label lookup from manifest rather than passing
labels).
Note that we need to use stamp_jar in scala_import and scala macro library.
Also this will break some tests which talk about exports but maybe that’s
the price we need to pay

johnynek:
This is still interesting but has seemed to get a bit stale.

@ittaiz can you take a look, and @ceason can you merge master?

johnynek:
@ittaiz is there someone on your end that can look at this? I know the skylark label propagation is something you all wanted. It would be great to switch to the same approach that java is using.

cc @andyscott @long-stripe if you are looking for clean up tasks. :)

ittaiz:
We wanted it but due to disagreements with upstream java people we've dropped the usage. Essentially we allow label re-mapping (via export) whereas they determine the label at point of creation. @or-shachar maybe this is a relevant task for Shay?
It will take him through a few of the interesting places.

ittaiz:
I think I’ll have someone in a month. If Long or Andy want to give it a crack I’ll be happy to guide them

liucijus:

Issue #642, opened 2018-11-19
'Unused dependency checking' is causing failures in rules_scala/test build


majcherm-da:
![scala_rules_unused_deps](https://user-images.githubusercontent.com/43614029/48721645-a66cee00-ec22-11e8-9777-8e509a8c09e8.PNG)

To reproduce run:

    bazel test //test:HelloLibTest

in `rules_scala` / `test`s

johnynek:
@majcherm-da thanks for the bug report. That plugin is not totally perfect. We have seen some cases where it errors saying something is unused, but if removed, it does not compile.

For those targets, we disable the checker on the target currently.

cc @andyscott @beala-stripe maybe Stripe can prioritize taking a stab at improving this, but so far we have been working around.

An idea outcome would be getting this a supported flow in scalac, but so far, we have not had a ton of luck advocating for that. @stuhood had a proposal for scalacenter related to this, but I can't find it now, and I don't think it has moved forward. Unfortunately, as far as I can tell, lightbend, scala center and EPFL are putting close to zero effort on the large build problem (the domain of bazel, pants, buck).

majcherm-da:
I created a PR to fix this problem. It's specifically a Windows related issue: https://github.com/bazelbuild/rules_scala/pull/717

liucijus:
Solved in #717

Issue #641, opened 2018-11-19
'No such file or directory' on Windows


majcherm-da:
`rules_scala` doesn't work properly on Windows.

This can be reproduced by building:

    bazel test //test:HelloLibTest

on `rules_scala` itself (`test` folder).

After switching to a commit before: https://github.com/bazelbuild/rules_scala/pull/568/files/bd8ff42e35d8f8bb0181aea5cf0439512c6df32e
which introduced another issue, this is what I'm getting:
![scala_rules_runfiles](https://user-images.githubusercontent.com/43614029/48721186-a91b1380-ec21-11e8-978b-9238918ae85d.PNG)

It tries to lookup a `JAVABIN` wrapper in `<name>.runfiles` folder, but this one is almost empty on Windows.

(Similar thing happens for `rules_java` - this folder is also almost empty, but in case of `java` a different laucher is used for which this doesn't cause a problem:
https://github.com/bazelbuild/bazel/blob/556839466dd49958a6b058a3a4c2ad341c8d06f1/src/main/java/com/google/devtools/build/lib/bazel/rules/java/BazelJavaSemantics.java#L380)

Wrapper script is properly generated, but it resides in the same location as runfiles directory only. On unix boxes the file is present in both locations and thus it works.

ittaiz:
I don’t understand, what is rules_java? Is this in bazel-core?
If so are you saying bazel itself doesn’t work on windows?

majcherm-da:
By `rules_java` I meant: https://docs.bazel.build/versions/master/be/java.html. I've seen that `rules_scala` is based on it and I've also noticed similarities in artifacts which are getting produced for java and scala on different OSes. By this I meen that on OSX `<name>.runfiles` folder is fully populated with jdk, built jars, wrappers, etc. This happens for both java and scala. On Windows the runfiles folder is almost empty and it's like this for both java and scala, altough **Bazel's Java rules work correctly on Windows)**. In java rules (from bazel) launcher is dfferent on Windows than on other OSes, which does not seem to be the case for `rules_scala`. `rules_scala` is trying to access files from `<name>.runfiles` folder, which is not populated on Windows.

I could paste several screenshot showing what's happening, but probably the easiest way to understand what's happening is running it on Windows. This can be very easily reproduced by following the steps from above.

majcherm-da:
I created a PR to fix this problem. In short, use exe launcher instead of bash script on Windows:  https://github.com/bazelbuild/rules_scala/pull/717

PR #717, opened 2019-03-19
Windows support

majcherm-da:
This resolves 2 issues I raised for running `rules_scala` on Windows, that is:
- https://github.com/bazelbuild/rules_scala/issues/641 (`%1 is not a valid Win32 application`)
- https://github.com/bazelbuild/rules_scala/issues/642 (valid dependencies reported as usused on Windows)

Main change is a port of Bazel's exe launcher from java rules to scala rules. Without it Bazel attempts to execute a bash script (generated from `java_stub_template`) as Bazel's executable, which fails on Windows. 

See https://github.com/bazelbuild/bazel/blob/8ac46c157c5da63e97a5825316ce6f0c3290e189/src/main/java/com/google/devtools/build/lib/bazel/rules/java/BazelJavaSemantics.java#L411 for how it's done in java rules.

The `UnusedDependencyChecker` is fixed by normalizing Windows paths, replacing backslashes with slashes. Without it all targets report unused deps on Windows.

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -590,74 +590,97 @@ def _jar_path_based_on_java_bin(ctx):
     jar_path = java_bin.rpartition("/")[0] + "/jar"
     return jar_path
 
-def _write_executable(ctx, rjars, main_class, jvm_flags, wrapper, use_jacoco):
-    template = ctx.attr._java_stub_template.files.to_list()[0]
+def _write_executable(ctx, executable, rjars, main_class, jvm_flags, wrapper, use_jacoco):
+    if (_is_windows(ctx)):
+        classpath = ";".join(
+            [("external/%s" % (j.short_path[3:]) if j.short_path.startswith("../") else j.short_path) for j in rjars.to_list()],
+        )
+        jvm_flags_str = ";".join(jvm_flags)
+        java_for_exe = "%s/%s" % (ctx.workspace_name, str(ctx.attr._java_runtime[java_common.JavaRuntimeInfo].java_executable_exec_path))
 
-    jvm_flags = " ".join(
-        [ctx.expand_location(f, ctx.attr.data) for f in jvm_flags],
-    )
-
-    javabin = "export REAL_EXTERNAL_JAVA_BIN=${JAVABIN};JAVABIN=%s/%s" % (
-        _runfiles_root(ctx),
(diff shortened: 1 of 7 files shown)


majcherm-da:
I signed it!

ittaiz:
@majcherm-da Thank you and sorry for the long silence. I'll get to this very soon.
I've skimmed and just want to verify something- the "exe" launcher which is a java_binary is used only in windows, right?  
Also this starts a jvm which then starts a windows executable from core-bazel?
Is this indirection mandatory? Sounds like it's added complexity and performance but I don't know the windows challenges well enough.

ittaiz:
Also,
@johnynek I think this relates to the discussion we had about dropping scala_binary and its friends in favor of java_binary and friends.
We are making some headway at Wix towards being able to move from scala_junit_test to using java_test (still not there but I think we're close).
If we can run scala_test on java_test/java_binary then we can remove a lot of code/concerns from this repo.

majcherm-da:
@ittaiz Yes, the "exe" java_binary is used only on Windows. It's java based, since it was the simpliest way I found to create the resulting exe file based on the Bazel's `bazel_tools/tools/launcher/launcher.exe` (porting the solution from Bazel's `JavaBinary`). The `launcher.exe` (binary file) needs to be appended with target specific arguments, which together constitues a final exe file. Bazel's `JavaBinary` is written in Java itself, but here the entrypoint is a Skylark rule that's why it feels a bit unnatrual. The "exe" java_binary is only used to create the final exe file, which is a replacement for the sh start script for Linux/Mac to start the target app/binary, so the statement that "this starts a jvm which then starts a windows executable from core-bazel" is not fully correct. As an alternative, to get rid of the need of starting JVM to create the exe file, we could re-implement the [Bazel's LaunchInfo](https://github.com/bazelbuild/bazel/blob/59f17d6e0550bf63a0b6ef182e2d63474e058ede/src/main/java/com/google/devtools/build/lib/analysis/actions/LauncherFileWriteAction.java#L102) and a merge of it's output with the laucher.exe binary with sth. more lightweight, like a `run_shell` call, but, not mentioning the rewrite itself, it might be a bit tricky to do in bash as it would need to support the runfiles.rlocation to obtain the path of the launcher.exe.

johnynek:
can we close this since #718 includes everything here?

PR #720, opened 2019-03-21
Use jar from java runtime and not local_jdk

ittaiz:
Work around and somewhat prepare for resolution of
https://github.com/bazelbuild/bazel/issues/6955


ittaiz:
@anchlovi would love your thoughts here

ittaiz:
@beala-stripe if you're using scala_binary/repl/test (as opposed to java_binary) you can take a look at this solution. It requires copying the stub to your workspace.
I think I'll improve this and expose it as an http_archive.
This is better than the workaround of local_jdk IMHO

beala-stripe:
Thanks! I'll try to find some time to test this out later.

ittaiz:
@johnynek this is ready apart from the fact that I'd like to serve the stub from a public place and use http_archive

ittaiz:
@johnynek any thoughts or objections?

johnynek:
Thanks @ittaiz

ittaiz:
Thank you!

PR #723, opened 2019-03-27
Updated java_stub_template in scala_repositories

ittaiz:
adding it to the WORKSPACE was wrong since now each user needs to do the same action (and we have two contradicting declarations)


ittaiz:
@johnynek I think I was mistaken in adding the java_stub_template directly to the workspace.
This approach seems wiser.

ittaiz:
thanks! I'll merge it when green. Want to customize the commit message

PR #724, opened 2019-03-27
Replace `native.package_name()` with `package_name` in BUILD file.

laurentlb:
The notation `native.` will not be available in BUILD files in the future.


laurentlb:
cc @ittaiz 

The Travis test seems to fail randomly.

johnynek:
@laurentlb sorry for the latency. This looks good to me.

The reproducibility test is flakey on our jmh benchmark rules for reasons we haven’t run down yet. It only started in the last couple of months. Anyway.

PR #728, opened 2019-04-17
Improve persistent Scalac worker docs in README.md

thundergolfer:
### 1st Commit

`test` inherits from `build`, so the 2nd line is redundant. See: https://docs.bazel.build/versions/master/guide.html#bazelrc-syntax-and-semantics

### 2nd Commit

I'd say that, _generally_, sandboxing in Bazel is preferred. By default you get non-sandboxed workers in Bazel, so I've added the option that enables the sandboxing as an 'easy win'.


johnynek:
thanks!

PR #731, opened 2019-04-23
Handle experimental java coverage configuration

steinar:
Experimental java coverage was [added in Bazel 0.24.0](https://github.com/bazelbuild/bazel/commit/68c7e5a3c679be51f750d44aae146007f0f04b4d), which introduced a new template variable `%set_java_coverage_new_implementation%` to `java_stub_template.txt`. Not handling that substitution results in the following error on build:

```bash
/private/var/tmp/[...]: line 262: fg: no job control
```

This PR resolves the error by enabling the experimental coverage functionality if coverage is enabled, but disables it otherwise. This corresponds to the behaviour of the Java rules. I'm not complete sure if that is applicable for Scala as well.


steinar:
I signed it!

ittaiz:
Thanks!

PR #733, opened 2019-04-23
Add support and test for custom generator

ianoc-stripe:
This allows adding more generators to scalapb to run within the toolchain. 

Use case here is that i want to generate service definitions that are friendly to twitter's util/futures rather than scala futures. So want to make it extensible to add more scala code generators into the output.

Thoughts @johnynek and @ittaiz ?

### DIFF
--- a/scala_proto/private/proto_to_scala_src.bzl
+++ b/scala_proto/private/proto_to_scala_src.bzl
@@ -19,8 +19,12 @@ def _colon_paths(data):
     ])
 
 
-def proto_to_scala_src(ctx, label, code_generator, compile_proto, include_proto, transitive_proto_paths, flags, jar_output):
-    worker_content = "{output}\n{included_proto}\n{flags_arg}\n{transitive_proto_paths}\n{inputs}\n{protoc}".format(
+def encode_named_generators(named_generators):
+    return ",".join([k + "=" + v for (k, v) in named_generators.items()])
+
+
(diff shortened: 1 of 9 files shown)


ittaiz:
Super interesting, thanks!
I skimmed and mainly looked at the usage example.
I'll try to take a closer look soon. @simuons any chance you can also take a look?

ianoc-stripe:
This is relatively narrow right now in an additive generator, but depending on folks thoughts similar things would be:
-> Ensure factoring to instead specify the list of generators(i.e. it would mean you could use these rails to replace scala generation at all i guess you could directly use it for clojure/kotlin if it was applicable)
-> Attempt to factor the code to make it 'easy' to build multiple independent ones (so practice for a consumer should be add a new toolchain/rule easily than adding more generators to existing rule?)

scala_proto/private/proto_to_scala_src.bzl
```diff
@@ -19,8 +19,12 @@ def _colon_paths(data):
     ])
 
 
-def proto_to_scala_src(ctx, label, code_generator, compile_proto, include_proto, transitive_proto_paths, flags, jar_output):
-    worker_content = "{output}\n{included_proto}\n{flags_arg}\n{transitive_proto_paths}\n{inputs}\n{protoc}".format(
+def encode_named_generators(named_generators):
+    return ",".join([k + "=" + v for (k, v) in named_generators.items()])
```
johnynek: I wonder if we need to sort this?

ianoc: when in doubt, sort. Will put in a PR

PR #735, opened 2019-04-29
[DOCS] README updates 

sholokhov:
Updated Scala and scala libs versions, highlight the possibility to specify scala_version for `scala_proto_repositories` rule (because default is `2.11.8` and it obviously doesn't work with default scala version).


sholokhov:
I signed it!

johnynek:
Thanks!

Issue #699, opened 2019-02-22
jmh code might be unreproducible


ittaiz:
One of the travis builds failed on the reproducibility test without apparent relation to the change (I think). Makes me think maybe it just surfaced there:
https://travis-ci.org/bazelbuild/rules_scala/jobs/497218553
```
< 47c3d415c9e5dbe424c573d08f40da1e  bazel-bin/test/jmh/test_benchmark_compiled_benchmark_lib.jar
< 47c3d415c9e5dbe424c573d08f40da1e  bazel-bin/test/jmh/test_benchmark.runfiles/io_bazel_rules_scala/test/jmh/test_benchmark_compiled_benchmark_lib.jar
1501a1500
> 4f5fa74aa94da701176f042a1e47c841  bazel-bin/test/jmh/test_benchmark_codegen_resources.jar
5393a5393,5394
> d9a26e038170d9e26c593555af311b79  bazel-bin/test/jmh/test_benchmark_compiled_benchmark_lib.jar
> d9a26e038170d9e26c593555af311b79  bazel-bin/test/jmh/test_benchmark.runfiles/io_bazel_rules_scala/test/jmh/test_benchmark_compiled_benchmark_lib.jar
7596d7596
< ef6ee9a44b47c199f47ee962eaf89e7f  bazel-bin/test/jmh/test_benchmark_codegen_resources.jar
```
@johnynek wdyt?

ittaiz:
mainly since the PR passed and then I pushed two commits and it broke:
https://github.com/bazelbuild/rules_scala/pull/698/commits/b7979bac2d9f2809849ef81085e8b24d1a753d49
https://github.com/bazelbuild/rules_scala/pull/698/commits/a530f8b29b201561624127c7b0bea9c9b658dc61
They seem really unrelated

ittaiz:
indeed passed after rerun without code changing

johnynek:
yeah, it looks like it is unreproducible. Fortunately, we never depend on a benchmark, we only ever run them, so I guess it is not totally critical to fix, but it would be nice to.

Surprising this pops up now years later.

ittaiz:
yeah, very surprising.
maybe we should exclude it from the reproducibility test for now?

johnynek:
I just noticed that the shas are always the same:

```
< 47c3d415c9e5dbe424c573d08f40da1e  bazel-bin/test/jmh/test_benchmark_compiled_benchmark_lib.jar
[...]

PR #736, opened 2019-04-30
Port bazelbuild/bazel#8196 : improve java_stub_template MANIFEST.MF construction speed

joshrosen-stripe:
This PR ports [bazelbuild/bazel#8196: "Speed up construction of MANIFEST.MF lines in java_stub_template"](https://github.com/bazelbuild/bazel/pull/8196) to `rules_scala`.

This port is necessary because `rules_scala` currently uses its own version of `java_stub_template` (see https://github.com/bazelbuild/rules_scala/pull/720 and https://github.com/bazelbuild/rules_scala/pull/723; /cc @ittaiz)


johnynek:
I'm kind of confused how this PR is working. It is linking to a self reference to this repo. Is it just advancing the pointer?

joshrosen-stripe:
Yeah, I'm a bit confused about that, too: the existing master branch fetches the `java_stub_template` via a `http_archive` downloading a file from an archive hosted on GItHub and created from an unreachable ref that was force-pushed over: https://github.com/bazelbuild/rules_scala/pull/720#issuecomment-476472506

I tried to mimic that series of steps here. This was pretty painful and hard to reason about.

Do you know if there's a reason that we have to use `http_archive` instead of just including this as a regular file here?

johnynek:
So, the history was we used to copy a file, then it would get really out of sync, then we pointed to the bazel repo to make it easier to stay in sync, but then as you see above, the wheels came off somehow.

I *thought* we used the self pointer during the tests, but looking back now, that doesn't seem to be the case. It's not well documented and since it is unusual it should be.

@ittaiz can you shed some light here?

johnynek:
But how is this a fix @joshrosen-stripe ? It is just pointing to a previous sha in this repo. That fixes things?

joshrosen-stripe:
@johnynek,

What I did here was:

1. Download the old version of `java_stub_template` from https://github.com/bazelbuild/rules_scala/archive/8b8271e3ee5709e1340b19790d0b396a0ff3dd0f.tar.gz, the URL referenced in the `http_archive` rule.
2. Commit ⬆️ as c287aef0d90c50625cc43bead116b6d46622866d
3. Apply the changes from my bazelbuild PR in 75217c13f73aefb9e9e90fc6346f9c7e1ceed426.
4. _Delete_ all of the files added in the previous three commits in 6e3989c659dead4fc1aa5542475d00e8cfacdd65
5. Update the `http_archive` to point to an URL reflecting the state in step 3.

This is rather convoluted, but I was trying to follow / mirror the existing approach (since I assumed it was presumably done for a good reason).

Maybe we can just include `java_stub_template` as a regular file and skip this whole `http_archive` indirection?

beala-stripe:
I think we should just include it as a regular file for now. Once the following issue is resolved, we can change it to a pointer to bazelbuild/bazel: https://github.com/bazelbuild/rules_scala/blob/8092d5f6165a8d9c4797d5f089c1ba4eee3326b1/scala/scala.bzl#L508-L509

Since this is affecting a target in Zoolander, I can take on making these changes tomorrow if you don't have the bandwidth @joshrosen-stripe .

fwiw, I got this working as [a pointer to bazelbuild/bazel](https://github.com/beala-stripe/rules_scala/blob/63ff39b4a1b1d036ad3bb8926e24e2552efa0b26/scala/scala.bzl#L511-L522), which is what I think was originally intended, and then saw that comment above that the upstream version was missing a fix we need.

joshrosen-stripe:
@beala-stripe, I've invited you as a collaborator to my fork, so you should now have push access if you want to make the changes in this PR / my branch.

beala-stripe:
@joshrosen-stripe Thanks! I've made the suggested change.

@johnynek Does this seem like a reasonable solution for now? I'm happy to change it back to a pointer to bazelbuild/bazel once https://github.com/bazelbuild/bazel/issues/6955 is fixed. I can probably carve out some time to try and upstream our fix for https://github.com/bazelbuild/bazel/issues/6955

johnynek:
Sorry this took so long to merge. This is a good solution. Thanks @joshrosen-stripe and @beala-stripe — it would be great to upstream this and include some example timing before and after this change.

joshrosen-stripe:
The Bazel fix is already upstreamed in https://github.com/bazelbuild/bazel/pull/8196; there's some performance measurements in that PR.

johnynek:
wait....

if it is already upstreamed, why can't we just point the java_stub remote repo to a version of bazel that has it and not copy the file here? I'm confused.

joshrosen-stripe:
> why can't we just point the java_stub remote repo to a version of bazel that has it and not copy the file here?

Because `rules_scala`'s fork of this file includes additional changes which are working around https://github.com/bazelbuild/bazel/issues/6955 and those other changes aren't upstreamed yet.

johnynek:
Ok. I understood https://github.com/bazelbuild/rules_scala/pull/736#issuecomment-510575712 to mean the full fix is already upstreamed. I guess only a partial fix has been upstreamed.

beala-stripe:
There are two different issues:
1. https://github.com/bazelbuild/bazel/issues/6955: This one has _not_ been upstreamed, and is the reason we need to fork `java_stub_template.txt`
2. https://github.com/bazelbuild/bazel/pull/8196: This one has been upstreamed by @joshrosen-stripe. He opened this PR and manually copied pasta'd it here because of (1).

The scala-team has a burndown sprint coming up (starting July 22nd), and I think I could fit upstreaming (1) into that sprint, but for now I think keeping the forked `java_stub_template.txt` is the quickest path to resolving the Zoolander issue. Here's the internal ticket for upstreaming (1) https://jira.corp.stripe.com/browse/SCALA-1034

beala-stripe:
@johnynek Anything else we should do to get this across the finish line? It looks like the windows tests are flaking out. Should I disable those?

johnynek:
sorry, I restarted the build and forgot to come back to it. everything was green except reproducibility of jmh which is a known issue #699

Issue #737, opened 2019-05-04
different argument name for sha256 in jvm_import_external rule in bazel repo and this repo


natansil:
`jvm_import_external` has duplicate definitions with different arg name for the same field:
in [bazel repo](https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/repo/jvm.bzl#L211) the arg name is `artifact_sha256`
in [rules_scala](https://github.com/bazelbuild/rules_scala/blob/6a9f81aa29563a07cc69a2555e54ac3cdfd396ed/scala/scala_maven_import_external.bzl#L211) the arg name is `jar_sha256`

I think rules_scala should not have an implementation of jvm_import_external at all...
so this will probably mean adding an alias to bazel repos's definition...

PR #738, opened 2019-05-04
add arg  to _jvm_import_external to align with bazel repo version

natansil:
bazel repo has [artifact_sha256](https://github.com/bazelbuild/bazel/blob/d8ffad56129e2c358c5b208edb62918c8ed1220c/tools/build_defs/repo/jvm.bzl#L56) arg:
This change aligns rules_scala version of `jvm_import_external` to also have this arg.
See https://github.com/bazelbuild/rules_scala/issues/737

### DIFF
--- a/scala/scala_maven_import_external.bzl
+++ b/scala/scala_maven_import_external.bzl
@@ -210,7 +210,7 @@ jvm_import_external = repository_rule(
         "rule_name": attr.string(mandatory = True),
         "licenses": attr.string_list(mandatory = True, allow_empty = False),
         "jar_urls": attr.string_list(mandatory = True, allow_empty = False),
-        "jar_sha256": attr.string(default = None, doc = "'jar_sha256' is deprecated. Please use 'artifact_sha256'"),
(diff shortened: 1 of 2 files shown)


scala/scala_maven_import_external.bzl
```diff
@@ -208,7 +208,7 @@ jvm_import_external = repository_rule(
         "rule_name": attr.string(mandatory = True),
         "licenses": attr.string_list(mandatory = True, allow_empty = False),
         "jar_urls": attr.string_list(mandatory = True, allow_empty = False),
-        "jar_sha256": attr.string(),
+        "jar_sha256": attr.string(doc = "'jar_sha256' is deprecated. Please use 'artifact_sha256'"),
```
natansil: @ittaiz is this what you meant by deprecation?

ittaiz: I meant https://github.com/bazelbuild/bazel/issues/5654
TLDR:
```
set the default value of the attribute to None and then in the rule implementation check if the attribute is set to anything other than None and print a warning to the user
```

natansil: done. string type can't get None

PR #739, opened 2019-05-05
Portable classpath in stub template

gmishkin:
Change to use the platform-specific path separator. The stub template script already sets `IFS=';'` on Windows so having colons in `CLASSPATH` means it doesn't get split properly.

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -618,7 +618,7 @@ def _write_executable(ctx, rjars, main_class, jvm_flags, wrapper, use_jacoco):
             template = template,
             output = ctx.outputs.executable,
             substitutions = {
-                "%classpath%": classpath,
+                "%classpath%": "\"%s\"" % classpath,
                 "%javabin%": javabin,
                 "%jarbin%": _jar_path_based_on_java_bin(ctx),
                 "%jvm_flags%": jvm_flags,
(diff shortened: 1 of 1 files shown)


johnynek:
Thanks for the catch.

It would be great to follow this up with a PR that would exercise this in a test so that we don't regress on this since it may be important to you.

gmishkin:
There's another thing I need to do to get the whole thing working (bazel-deps), which I thought I would get back to sooner (but didn't) and then think about how to test it.

For some reason `templates/jar_artifact_resource.bzl` is ending up at the root of the C: drive instead of in the jar or wherever, so the `parse` target doesn't run. I'm guessing it has something to do with how the `resources` attribute is handled in `scala_binary` targets with Windows paths. I need to look into it more.

The other thought I had which I'm not sure is a direction you'd consider going in but would apply to this class path issue too: would it be possible to share more with the native Java rules? Those create a native .exe on Windows instead of using the stub template.

johnynek:
you did see this PR, right: #718 will this be helpful?

gmishkin:
Ah, I did see it a while ago before I even ran into this issue, but it was just a curiosity at the time so it slipped my mind.

I think the launcher improvement was not sufficient to get bazel-deps specifically working. Some issue either there, or here in the way these rules handle resources paths on Windows, is still causing `jar_artifact_backend.bzl` to get dropped in `C:\templates` instead of where (I suppose) it should be.

```
Exception in thread "main" java.lang.NullPointerException
        at com.github.johnynek.bazel_deps.Writer$.jarArtifactBackend$lzycompute(Writer.scala:13)
        at com.github.johnynek.bazel_deps.Writer$.jarArtifactBackend(Writer.scala:12)
        at com.github.johnynek.bazel_deps.Writer$.workspace(Writer.scala:149)
        at com.github.johnynek.bazel_deps.MakeDeps$.apply(MakeDeps.scala:79)
        at com.github.johnynek.bazel_deps.ParseProject$.main(ParseProject.scala:17)
        at com.github.johnynek.bazel_deps.ParseProject.main(ParseProject.scala)
```

Issue #721, opened 2019-03-22
scala_import is not documented


kojustin:
The scala import rule is not documented in the README.

long-stripe:
I think we can close this with https://github.com/bazelbuild/rules_scala/pull/746 being merged - it's start, at least.

PR #746, opened 2019-05-10
Document scala_import

long-stripe:
Related issue: #721

## Summary

This PR adds an initial doc for `scala_import`.

I chose to do it as a Markdown table written by hand. The rest of the docs are HTML tables that I think were generated long ago by Skydoc when it still worked. As soon as Stardoc works, generated files should replace these hand maintained files!

I do need a second look with these attributes though, since I don't really know all of them really mean and cribbed the descriptions from:

- https://docs.bazel.build/versions/master/skylark/lib/JavaInfo.html
- https://github.com/bazelbuild/rules_scala/blob/master/docs/scala_library.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ This project defines core build rules for [Scala](https://www.scala-lang.org/) t
 * [thrift_library](docs/thrift_library.md)
 * [scalapb_proto_library](docs/scalapb_proto_library.md)
 * [scala_toolchain](docs/scala_toolchain.md)
+* [scala_import](docs/scala_import.md)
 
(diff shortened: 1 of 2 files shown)


long-stripe:
It's probably easiest to look at the GitHub rendered Markdown: 
https://github.com/long-stripe/rules_scala/blob/document-scala-import/docs/scala_import.md

johnynek:
Can we add the motivation here: we are simply reimplementing java_import except we aren’t using ijar which breaks scala macros and generally doesn’t help much for external dependencies which rarely change.

PR #748, opened 2019-05-13
Fix for Bazel upcoming change --incompatible_string_join_requires_strings

laurentlb:
The argument of string.join should be a list of strings. In the future,
Bazel won't perform this implicit conversion.


johnynek:
filed #749 as it looks like our windows CI is flaky.

PR #750, opened 2019-05-14
Fix compilation for --incompatible_depset_is_not_iterable

laurentlb:
Soon, Bazel will forbid iterating directly on a depset.


laurentlb:
(the diff is slightly longer than it should because the file was not properly formatted with Buildifier)

johnynek:
Thanks!

PR #756, opened 2019-05-20
Update protobuf dependency

laurentlb:
With this commit, this repository works with Bazel flag `--incompatible_disable_deprecated_attr_params`.

Progress towards #644

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -91,18 +91,6 @@ maven_jar(
     sha1 = "6c6c702c89bfff3cd9e80b04d668c5e190d588c6",
 )
 
-http_archive(
-    name = "com_google_protobuf",
-    sha256 = "d82eb0141ad18e98de47ed7ed415daabead6d5d1bef1b8cccb6aa4d108a9008f",
-    strip_prefix = "protobuf-b4f193788c9f0f05d7e0879ea96cd738630e5d51",
-    # Commit from 2019-05-15, update to protobuf 3.8 when available.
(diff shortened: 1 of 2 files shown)


WORKSPACE
```diff
@@ -93,11 +93,16 @@ maven_jar(
...
-    strip_prefix = "protobuf-3.6.1.3",
+    sha256 = "d82eb0141ad18e98de47ed7ed415daabead6d5d1bef1b8cccb6aa4d108a9008f",
+    strip_prefix = "protobuf-b4f193788c9f0f05d7e0879ea96cd738630e5d51",
+    # Commit from 2019-05-15, update to protobuf 3.8 when available.
+    url = "https://github.com/protocolbuffers/protobuf/archive/b4f193788c9f0f05d7e0879ea96cd738630e5d51.tar.gz",
 )
 
+load("@com_google_protobuf//:protobuf_deps.bzl", "protobuf_deps")
+
+protobuf_deps()
```
johnynek: Can you update the getting started documentation since I assume if we merge this this will break all out downstream users when they update.

At least explain they need to add that to their build but also if possible link to any documentation that might exist on configuring that function (some such dependency functions have optional args).

laurentlb: Good point.

protobuf_deps contains only zlib. I can inline its definition into `scala_repositories` so that it's transparent for end users.

laurentlb: and the protobuf http_archive can be moved there, to simplify the getting started code.

PR #759, opened 2019-05-27
Prevent showing filtered-out specs2 tests in the test.xml

hmemcpy:
This change fixes a bug in test.xml generation produced by the bazel JUnit runner. Before this fix, the test.xml produced by the test runner will contain all test cases, regardless whether they ran or not, and incorrectly display them in IntelliJ as successful.

</img>
(shows `foo` as having successfully ran even though it never actually ran, due to test filter) 

When providing a `test_filter`, the string is passed to the underlying specs2 runner, so it runs just the filtered tests, but the test model contains all the tests, regardless of the filter. In Java, the underlying model is modified, so the test children which don't match the filter are removed from the model. In Scala the model is immutable, so the solution is to generate a test model that is already filtered. This works by first flattening the tree structure of a model, filtering the tests according to the `test_filter` string, and reconstructing the tree model from filtered results.

An additional benefit of this fix is the resulting test.xml - it will no longer contain redundant "should" blocks!
<details>
<summary>
  <strong>XML differences</strong>
</summary>
<p>
Given a test like the screenshot above:

```scala
class MyTest extends SpecWithJunit {
  "my test" should {
    "foo" in ...
    "bar" in ...
  }
}
```
When running with `--test_filter=MyTest#my test should::foo$`,

**Before:**
```xml
<?xml version='1.0' encoding='UTF-8'?>
<testsuites>
  <testsuite name='Aggregate Specs2 Test Suite' timestamp='1970-01-01T17:14:20.074Z' hostname='localhost' tests='2' failures='0' errors='0' time='0.002' package='' id='0'>
    <properties />
    <system-out />
    <system-err /></testsuite>
  <testsuite name='MyTest' id='1'>
    <properties />
  </testsuite>
  <testsuite name='my test should' id='2'>  // <--- 1
    <properties />
    <testcase name='my test should::foo' classname='MyTest' time='0.002' />
    <testcase name='my test should::bar' classname='MyTest' time='0.0' />    // <-- 2
  </testsuite>
</testsuites>
```

Notice the **1** and **2** noise.

**After:**
```xml
<?xml version='1.0' encoding='UTF-8'?>
<testsuites>
  <testsuite name='Aggregate Specs2 Test Suite' ... package='' id='0'>
    <properties />
  <testsuite name='MyTest' id='1'>
    <properties />
    <testcase name='my test should::foo' classname='MyTest' time='0.002' />
  </testsuite>
</testsuites>
```
</p>

### DIFF
--- a/src/java/io/bazel/rulesscala/specs2/Specs2RunnerBuilder.scala
+++ b/src/java/io/bazel/rulesscala/specs2/Specs2RunnerBuilder.scala
@@ -53,6 +53,15 @@ object Specs2FilteringRunnerBuilder {
 class FilteredSpecs2ClassRunner(testClass: Class[_], testFilter: Pattern)
   extends org.specs2.runner.JUnitRunner(testClass) {
 
+  override def getDescription(env: Env): Description = {
+    val root = super.getDescription(env)
+    val filtered = flatten(root).filter(matchingFilter)
+
+    val flattenedRoot = root.childlessCopy()
+    filtered.foreach(flattenedRoot.addChild)
+    flattenedRoot
+  }
+
   def matchesFilter: Boolean = {
     val fqn = testClass.getName + "#"
     val matcher = testFilter.matcher(fqn)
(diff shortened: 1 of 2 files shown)


johnynek:
Looks good to me. Thanks for the PR!

Issue #230, opened 2017-06-15
support building scaladoc


johnynek:
this might be an optional output in the default scala rules (like the deploy jar) or we could use an another approach.

ittaiz:
What do the java rules do with javadoc?

pauldraper:
Community implementations, but nothing in core java rules.

https://github.com/bazelbuild/bazel/issues/446

PR #760, opened 2019-05-29
Add scala_doc rule

long-stripe:
Closes https://github.com/bazelbuild/rules_scala/issues/230.

### Summary

This PR adds a basic implementation of `scala_doc`.

I hesitate to recommend using it widely yet since I haven't figured out if there are weird edge cases. Would it be better to somehow designate this as experimental or something?

I did _not_ automatically compress outputted scaladoc, instead I think you could use `pkg_tar`.

### Performance

For an internal target, scaladoc takes ~17sec to scaladoc for "324 documentable templates".

Comparatively, cats-core takes ~32sec to scaladoc for "771 documentable templates":
```
sbt:cats> coreJVM/doc
[info] Main Scala API documentation to /Users/long/oss/cats/core/.jvm/target/scala-2.12/api...
model contains 771 documentable templates
[info] Main Scala API documentation successful.
[success] Total time: 32 s, completed May 29, 2019 2:01:16 PM
```

... which seems to makes sense to me.

To wit, I've added a note in the markdown documentation that users may want to mark `scala_doc` targets as manually built but YMMV.

### Testing

- Added a simple test target that also pulls in kind-projector and passes that along via aspect as a compiler plugin to `scaladoc`.
- Ran this against an internal `scala_doc` target.

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -32,6 +32,10 @@ load(
     "@io_bazel_rules_scala//scala:plusone.bzl",
     _collect_plus_one_deps_aspect = "collect_plus_one_deps_aspect",
 )
+load(
+    "@io_bazel_rules_scala//scala:scala_doc.bzl",
+    _scala_doc = "scala_doc",
+)
 
 _launcher_template = {
     "_java_stub_template": attr.label(
@@ -700,3 +704,5 @@ def scala_specs2_junit_test(name, **kwargs):
         suite_class = "io.bazel.rulesscala.specs2.Specs2DiscoveredTestSuite",
         **kwargs
     )
+
+scala_doc = _scala_doc
--- a/scala/scala_doc.bzl
+++ b/scala/scala_doc.bzl
@@ -5,6 +5,7 @@ load("@io_bazel_rules_scala//scala/private:common.bzl", "collect_plugin_paths")
 ScaladocAspectInfo = provider(fields = [
     "src_files",
     "compile_jars",
+    "plugins",
 ])
 
 def _scaladoc_aspect_impl(target, ctx):
@@ -18,9 +19,14 @@ def _scaladoc_aspect_impl(target, ctx):
         # Collect compile_jars from visited targets' deps.
         compile_jars = depset(transitive = [dep[JavaInfo].compile_jars for dep in ctx.rule.attr.deps if JavaInfo in dep])
 
+        plugins = depset()
+        if hasattr(ctx.rule.attr, "plugins"):
+            plugins = depset(direct = ctx.rule.attr.plugins)
+
         return [ScaladocAspectInfo(
             src_files = src_files,
             compile_jars = compile_jars,
+            plugins = plugins,
         )]
     else:
         return []
@@ -41,7 +47,7 @@ def _scala_doc_impl(ctx):
     compile_jars = depset(transitive = [dep[ScaladocAspectInfo].compile_jars for dep in ctx.attr.deps])
 
     # Get the 'real' paths to the plugin jars.
-    plugins = collect_plugin_paths(ctx.attr.plugins)
+    plugins = collect_plugin_paths(depset(transitive = [dep[ScaladocAspectInfo].plugins for dep in ctx.attr.deps]))
 
     # Construct the full classpath depset since we need to add compiler plugins too.
     classpath = depset(transitive = [plugins, compile_jars])
@@ -73,7 +79,6 @@ scala_doc = rule(
             aspects = [scaladoc_aspect],
             providers = [JavaInfo],
         ),
-        "plugins": attr.label_list(allow_files = [".jar"]),
         "_scaladoc": attr.label(
             cfg = "host",
             executable = True,
--- a/scala/scala_doc.bzl
+++ b/scala/scala_doc.bzl
@@ -40,7 +40,8 @@ scaladoc_aspect = aspect(
 )
 
 def _scala_doc_impl(ctx):
-    output_path = ctx.outputs.html
+    # scaladoc warns if you don't have the output directory already created, which is annoying.
+    output_path = ctx.actions.declare_directory("{}.html".format(ctx.attr.name))
 
     # Collect all source files and compile_jars to pass to scaladoc by way of an aspect.
     src_files = depset(transitive = [dep[ScaladocAspectInfo].src_files for dep in ctx.attr.deps])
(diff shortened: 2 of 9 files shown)


src/scala/io/bazel/rules_scala/scaladoc_support/BUILD
```diff
@@ -0,0 +1,17 @@
+load("//scala:scala.bzl", "scala_binary")
+
+# A simple scala_binary to run scaladoc.
+# `bazel run` this target with "-help" as a param for usage text:
+# bazel run -- "//src/scala/io/bazel/rules_scala/scaladoc_support:scaladoc_generator" -help
+scala_binary(
```
long-stripe: I put this target in `src/scala/io/bazel/rules_scala/scaladoc_support` because I initially had an actual Scala program do heavy lifting, but it turned out to be unnecessary and I could pass in everything through scaladoc/scalac args.

Let me know if this should be moved to the `//scala` package or something.

scala/scala.bzl
```diff
@@ -700,3 +704,5 @@ def scala_specs2_junit_test(name, **kwargs):
         suite_class = "io.bazel.rulesscala.specs2.Specs2DiscoveredTestSuite",
         **kwargs
     )
+
+scala_doc = _scala_doc
```
long-stripe: I'm following a pattern I see in [other rules](https://github.com/tweag/rules_haskell/blob/1375cc4ac8fb30474113a62f0daa9200dc42e08a/haskell/haskell.bzl#L354-L386), in which you basically re-export rules that are loaded from other .bzl files, makes it such that a user can load multiple rules from the same .bzl file.

long-stripe:
For your convenience, a link to view [scala_doc.md](https://github.com/long-stripe/rules_scala/blob/add-scaladoc-rule/docs/scala_doc.md).

scala/scala_doc.bzl
```diff
@@ -0,0 +1,91 @@
+"""Scaladoc support"""
+
+load("@io_bazel_rules_scala//scala/private:common.bzl", "collect_plugin_paths")
+
+ScaladocAspectInfo = provider(fields = [
```
andyscott: If this can be private, add a `_` suffix.

johnynek:
This is great, thanks @long-stripe !

One question: is there an easy way to combine this with a bazel query to find all the scala_libraries in a repo and build one giant documentation build? It might be impractical, but many folks might want to build one build of the docs per repo.

I guess some `bazel query` with `kind` of `scala_library` might do it. If we could add that to the documentation, it might help people.

long-stripe:
@johnynek I do seem to remember running a query that found all targets of a kind of rule, so that certainly sounds plausible! Would be nice to throw up in the docs as a way to build a big scala_doc target.

long-stripe:
@johnynek is scaladoc-ing external/3rd party dependencies (e.g. cats-core) something users would want to do?

As far as I can tell most of the time it's about the user's library code so `srcs` is what we're looking for here. If we pulled in source jars indiscriminately I could imagine this exploding into a truly gigantic ball of documentation... or am I misunderstanding?

johnynek:
@long-stripe if you return a type in your code, it is a bit of a bummer if you can't see anything about that type in the generated docs. We can certainly merge without that feature, but I think the ability to build the full documentation fully linked would be nice.

src/scala/io/bazel/rules_scala/scaladoc_support/BUILD
johnynek: I think this is fine.

johnynek:
@long-stripe question: have you tried a build of the internal target pointing at this rule defined remotely? sometimes we shake out bugs with the remote repositories.

src/scala/io/bazel/rules_scala/scaladoc_support/BUILD
long-stripe: Cool - happy to keep this here in case we do need to introduce a real Scala program.

long-stripe:
@johnynek 

1. Yeah that would be nice - I am not sure how it works out in the greater Scala ecosystem. How does http://circe.github.io/circe/api/io/circe/Decoder$.html#Result[A]=Either[io.circe.DecodingFailure,A], for example, link back to `scala.util.Either`? And it doesn't seem to be consistent for other projects... hmm.

2. I have not tried that. I assume when you say remotely you mean on CI? I've been testing by pointing to a local rules_scala with `native.local_repository`. I wouldn't expect there to be problems given that I'm reusing some already defined targets like `scala_compiler` and I'm not doing anything _that_ weird...

johnynek:
I'm in favor of merging and iterating rather than trying to imagine all outcomes. Let's get using this from the main repo, and see if we hit any issues.

long-stripe:
Same - would rather land this, possibly marked as experimental, and iterate.

long-stripe:
Oof- this doesn't work on Windows, probably because of platform differences like classpaths being defined with `;` or something.

https://travis-ci.org/bazelbuild/rules_scala/jobs/538958622#L326

johnynek:
the windows build failed, but it often does and the original contributor hasn't had time to diagnose.

johnynek:
actually, I was wrong, you are right. This is a new failure related to scaladoc.

@long-stripe do you think you could take a stab at making it work? There are a few places where we have to do things like that now.

long-stripe:
@johnynek yeah I think that should be fixed prior to merging. Do you have an example of switching per-platform behavior?

long-stripe:
Pushed a few more changes, again testing against an internal target:

- We now properly collect all `src_files` and `compile_jars` for deps up the propagation graph and not just one level in from the rule's deps.
- Turned off warnings since my bet is that `scala_doc` targets will tend to be larger collations, warnings will surely obscure actual showstopper errors and don't get in the way of scaladoc generation. It's better to have the warnings generated by regular `bazel build`s on the `scala_library` calls anyhow.
- Turns out we needed to propagate `scalacopts` for more complex source files! Duh.

long-stripe:
(I'll look into fixing Windows builds tomorrow)

long-stripe:
@johnynek looks like the classpath thing was it for Windows builds... would appreciate a re-review since I added new behavior in https://github.com/bazelbuild/rules_scala/pull/760/commits/48fc3d4f63219b90fe30afaaf1eea2bb41dc5c50, its intended purpose is to collect all src_files and compile_jars along the dependency graph.

johnynek:
looks great to me. Let's merge and iterate.

I'm a bit dubious of hardwiring the `-nowarn` vs making it the default `scalacopts`, but I do see your argument that you should be dealing with code warnings in your scala_library targets, not here.

PR #761, opened 2019-05-31
Explicitly convert depset to list

laurentlb:
This will be required by Bazel 0.27, where the flag
`--incompatible_no_support_tools_in_action_inputs` will be on by
default.

The function `collect_plugin_paths` iterates over its argument, so we
need to flatten the depset.


laurentlb:
cc @vladmos, @long-stripe

johnynek:
Thanks!

PR #762, opened 2019-06-03
fix grpc opencensus stats integration

mackenziestarr:
- upgrade `opencensus-java` packages to the latest (`0.22.1`)
- add `opencensus-impl` and `opencensus-impl-core` 

The opencensus stats integration support for gRPC when using a `scala_proto_toolchain` with `with_grpc = True` seems to be missing some transitive dependencies. I'm exporting these metrics with `opencensus-exporter-stats-prometheus` and without this patch nothing is put into the metrics registry. These are [runtime dependencies](https://github.com/census-instrumentation/opencensus-java#add-the-dependencies-to-your-project-1) so it's possible I've declared them in the wrong place in this patch. I'm getting around this right now in my build with this target configuration:

```
load("@io_bazel_rules_scala//scala_proto:default_dep_sets.bzl", "DEFAULT_SCALAPB_GRPC_DEPS")

scala_proto_toolchain(
    name = "scala_proto_toolchain_configuration",
    grpc_deps = DEFAULT_SCALAPB_GRPC_DEPS + [
        "//third_party/jvm/io/opencensus:opencensus_impl",
        "//third_party/jvm/io/opencensus:opencensus_impl_core",
    ],
    with_flat_package = True,
    with_grpc = True,
)
```
but I feel they should they should be included in `DEFAULT_SCALAPB_GRPC_DEPS`

In addition to adding these dependencies, I've upgraded all the OpenCensus deps to the latest  `0.22.1` since they are compatible with grpc 1.19.0 (see [opencensus-java gradle build](https://github.com/census-instrumentation/opencensus-java/blob/595cb11bcffca5bf9b861bdb81cf941a4f551764/build.gradle#L163))

thanks for taking a look!

### DIFF
--- a/scala_proto/default_dep_sets.bzl
+++ b/scala_proto/default_dep_sets.bzl
@@ -25,6 +25,7 @@ DEFAULT_SCALAPB_GRPC_DEPS = [
     "//external:io_bazel_rules_scala/dependency/proto/guava",
     "//external:io_bazel_rules_scala/dependency/proto/opencensus_api",
     "//external:io_bazel_rules_scala/dependency/proto/opencensus_impl",
+    "//external:io_bazel_rules_scala/dependency/proto/disruptor",
     "//external:io_bazel_rules_scala/dependency/proto/opencensus_impl_core",
     "//external:io_bazel_rules_scala/dependency/proto/opencensus_contrib_grpc_metrics",
     "//external:io_bazel_rules_scala/dependency/proto/google_instrumentation",
--- a/scala_proto/default_dep_sets.bzl
+++ b/scala_proto/default_dep_sets.bzl
@@ -24,6 +24,8 @@ DEFAULT_SCALAPB_GRPC_DEPS = [
     "//external:io_bazel_rules_scala/dependency/proto/grpc_context",
     "//external:io_bazel_rules_scala/dependency/proto/guava",
     "//external:io_bazel_rules_scala/dependency/proto/opencensus_api",
+    "//external:io_bazel_rules_scala/dependency/proto/opencensus_impl",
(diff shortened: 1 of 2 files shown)


johnynek:
Thanks for the PR!

Looks like CI failed.

@ianoc can you also take a look as someone who has worked at lot in this space.

mackenziestarr:
oh yeah that is a legitimate test failure, judging from the stack trace it seems like `opencensus-impl` is expecting another transitive dependency to exist `Caused by: java.lang.NoClassDefFoundError: com/lmax/disruptor/WaitStrategy`. I checked my project's build and i had already pulled in that dependency from another package which makes sense why I hadn't seen this before. I will try to update the patch later today.

mackenziestarr:
looks like one of the configurations in CI failed on a different test with the latest commit but I'm having trouble deciphering if it is an error or CI is just flaky https://travis-ci.org/bazelbuild/rules_scala/jobs/541568563. Given some conversations going on in [grpc-java](https://github.com/grpc/grpc-java/pull/5520) I'm wondering if it would be better to allow users to opt into these runtime dependencies, possibly with a separate argument (e.g. `with_opencensus=True`). I should also point out that Android seems to use a different artifact for the stats implementation, (impl_lite)[https://github.com/census-instrumentation/opencensus-java/tree/master/impl_lite] and I'm not sure what including `impl_core` on the classpath in Android would do, maybe @ianoc has some more context on this.

ianoc:
Alas https://github.com/bazelbuild/rules_scala/issues/699 looks to still be open. Which explains your failure. Proper ci targets all look green.

mackenziestarr:
@johnynek closed and reopened the PR to trigger CI again and it is passing now

Issue #753, opened 2019-05-16
scalapb_proto_library additional scala sources


holinov:
Now it is impossible to add some custom scala sources into scalapb library. So i could not do something like this:
```scala
trait TagRecordCompanionTrait {
    implicit val ordering: Ordering[TagRecord] = (x: TagRecord, y: TagRecord) => {
     if (x.count < y.count) 1
     else if (x.count > y.count) -1
     else x.name compareTo y.name
   }
}
```
```proto
message TagRecord {
    option (scalapb.message).companion_extends = "TagRecordCompanionTrait";

    required string name = 1;
    required int64 count = 2;
}
```
The only way to do this is to use preamble in proto file

ianoc:
This appears to be a duplicate of https://github.com/bazelbuild/rules_scala/issues/751 

If you want to weigh in there around the discussion that would be great

holinov:
No. That issue is about adding some external dep. In my case i need to add some scala sources into scalapb_library. I need to create base trait for message/message companion and set sef-type to message (so i could not compile my scala files and flies from protobuf in same scala library because of cyclic dependency)

ianoc:
Interesting, scalapb and its many options around this really make the aspect like stuff quite tricky. 
OOC have you managed to use any of the scalapb rules since/if so any change/improvement in performance? (it was huge for us, but we probably have more proto's than most).


On this, if you have any thoughts on how/what an API should look like here or how generic/often you need to do this would love to hear it. Some thoughts of ways we could try do it:

1) Use the experimental (new) api around adding another printer and emit the scala to do ordering or other things dynamically from that. (this is on master today so it would be usable -- just needs more docs written, and as more people use it we may update the API's to bake it in more too).
2) Use a manual scalapb rule and not use the aspects at all. We could just share some of the generation steps but have a non-aspect set of rules here too. Mixing the two would be odd in terms of duplicate classes. But if folks aren't getting a win and it makes stuff too complex then thats maybe not worth forcing everything down one path here.
3) It feels maybe a bit iffy of a plot. But we could specify in the toolchain a per target filegroup to inject similar to the dependencies discussed in #751. The two could probably be designed to compliment each other reasonably well

holinov:
I use this option all most every time i use scalapb. (it allows to extend generated messages/companions with methods and implicits)
When i'm thinking about api i think of it as 
```bazel

proto_library(
    name = "common_proto",
    srcs = [ "common.proto" ],
[...]

Issue #751, opened 2019-05-14
scalapb_proto_library doesn't support scala_library as a dependency anymore?


sholokhov:
Hi!

After switching on aspect-based protobuf compilation (#700) it seems like it's no longer possible to use custom `TypeMapper` from ScalaPB. At the first glance the reason is that bazel tries to generate scala code from `proto_library` (`%name-fast` rule) but there is no way to specify scala libraries for `proto_library` (at least I haven't found). 

It feels like it used to work before because previously bazel was using `scalapb_proto_library` with its `deps` to generate final scala code, where you can specify additional custom classes. 

I've also prepared [small example which reproduces this issue](https://github.com/sholokhov/bazel-scalapb-example).

Bazel version - `0.25.2`
rules_scala - `b2e12827240fc488bbe5ebbbbcbc7a4f12ca02f2`

johnynek:
cc @ianoc-stripe who may have some ideas.

Would you be willing to make PR with a unit test that you would like to pass and we can see how we can get it to pass. My guess is that we will need to add the ability of adding extra dependencies to the scalapb toolchain since we can't customize the individual aspect generated targets in any way that I know of.

ianoc:
@sholokhov you can add the dependencies to the scalapb toolchain which lets you add more deps. It will get added to all of the scalapb targets in one swoop in the repo.

You can define
https://github.com/bazelbuild/rules_scala/blob/master/scala_proto/scala_proto_toolchain.bzl#L46
part of your toolchain to contain the reference to your scala libraries.

sholokhov:
@ianoc thanks for helping, but now it fails with following error (during well-known proto files compilation):
```
sholokhov$ bazel build //scala-app:messages_scala_proto
INFO: Analyzed target //scala-app:messages_scala_proto (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: /private/var/tmp/_bazel_sholokhov/557977177af35e2d2806c45eff908f1d/external/com_google_protobuf/BUILD:248:2: scala @com_google_protobuf//:descriptor_proto-fast failed (Exit 1) scalac failed: error executing command bazel-out/host/bin/external/io_bazel_rules_scala/src/java/io/bazel/rulesscala/scalac/scalac @bazel-out/darwin-fastbuild/bin/external/com_google_protobuf/descriptor_proto-fast_scalac_worker_input

Use --sandbox_debug to see verbose messages from the sandbox
[...]

PR #763, opened 2019-06-04
[Api Discussion via docs] Adding a more manual/advanced scalapb rule

ianoc:
In the change i made to make these rules aspect based previously it has blocked/made difficult  some of scalapb's options that don't translate well into aspects (and aren't supported in the java protobuf so don't come up there). This proposes adding another rule which would give full flexibility in invoking the scalapb to cover these use cases. 

Using both rules in a repo does pose the risk of classpath conflicts of multiple implementations that are different of the same class name. Though I don't see a way around this issue in general. We could have the direct rule check to ensure none of its dependencies also contain the same class name it is emitting, to ensure it doesn't depend on an aspect for the same protobuf? 

Thoughts @ittaiz @simuons @johnynek ?


ittaiz:
@ianoc thanks for taking the time!
We are interested in the aspect approach for performance and since the current approach generates multiple copies of dependent protos (causes issues in IJ as well).
We are planning towards end of next week to prototype a suggested API on how to support multiple custom generators using toolchains and aspects. We might fail but we want to give it a try :)

ittaiz:
In general I'm not against this if you have concrete use-cases internally you're aware of.
I think that the conflict check can be valuable depending on cost (both build time and dev time)

ianoc:
We don't have them internally, the use cases that were coming up are:

https://github.com/bazelbuild/rules_scala/issues/765
https://github.com/bazelbuild/rules_scala/issues/753
https://github.com/bazelbuild/rules_scala/issues/751

so if your work managed to cover these and we didn't need it at all that would be better again

ittaiz:
Very interesting. #751 and #753  indeed sound very very similar to our needs and concerns.
This is what we'll try to solve (don't know if we'll be able to though).
Can you or @johnynek refresh my memory about why aspects are better?
My idea before the aspect approach was to separate the "to-be-generated" protos from the "classpath" protos (which I'm fairly certain is an option in protoc). This should solve the multiple copies of a proto appearing in the scala classpath later on and will also greatly improve performance (less generation and less compilation in the next step).

Am I missing something? Are these indeed the benefits but you think I'm mistaken about protoc support for this?

ianoc:
@ittaiz You can with protoc for sure separate out the proto's from having multiple copies. (using the internal function of src to jar thats in there now would do this). The aspects make the whole chain fully automatic which is the big win(to add the aspects to our largest scala repo i had to add the toolchain then strip the per-target options -- so a pretty simple PR, it immediately dropped 6-8 minutes off average our CI times for the repo which was super nice). Aspects make a few things better here though:

- You don't need to define the scalapb targets at every step of the chain. In our case this is huge since we have cross language bazel dependencies/generated proto's and similar, so you can define the scalapb in the repo you need it/where you need it and all of the intermediate targets are automatically supplied
- You don't have folks globbing protobuf files/proto_library targets out of laziness into one scalapb (well they can, but internally via the aspect thats not what will happen).
- With the former two issues you can wind up with 2 scalapb's referring to the same proto_library without a bunch of tooling/convention to stop it. So the aspects ensure you don't get duplicated classes from this. 
- Not the aspects so much, but the fact the options are configured on the toolchain while a pain for the extra options in scalapb makes the whole repo consistent in how the scala is generated which is a plus. (Mostly here the `flat_package` option has caused us problems in the past)


I think mostly what I was suggesting here is pretty close to what you were thinking about doing pre-aspects here, just having a rule that wouldn't produce all the classes all the time.

jsbrucker:
@ianoc any plans to pick up work on this PR again?

ianoc:
No I'm afraid there didn't seem to be a consensus on design to drive it. So I've long since moved onto other things.

liucijus:

PR #764, opened 2019-06-04
Make sure that plus-one-deps from exports of direct deps also propagate

ittaiz:
According to [java_library.export](https://docs.bazel.build/versions/master/be/java.html#java_library.exports):
```
Exported libraries.
Listing rules here will make them available to parent rules, as if the parents explicitly depended on these rules. This is not true for regular (non-exported) deps.
```
Currently we break this by not propagating +1 for exported targets


ittaiz:
@anchlovi @or-shachar would appreciate your review

ittaiz:
@johnynek any comments?

johnynek:
looks good to me.

Sorry for the latency.

ittaiz:
no worries. Mine are worse unfortunately :( 
Thanks! [ (and if you need me you know where to ping me ;) ]

ittaiz:
Btw, I hope you noticed the toolchain toggle in the aspect. I added it especially so that y'all wouldn't need to pay the iteration cost and memory when you're not using +1.

johnynek:
I did notice that! Thank you.

I think we should reconsider strictness possibly when the input pruning change to bazel lands. I can imagine deps + 1 along with input pruning being a very good default for many applications. It would be interesting to benchmark that at Stripe and see how performance looks.

https://github.com/bazelbuild/proposals/pull/102

cc @andyscott @long-stripe

long-stripe:
I'm not exactly sure what deps + 1 is, but input pruning sounds pretty neat. Is it applied by Bazel ex post facto on re-execution only? It sounds like rule implementors would need to emit a diff list of unused inputs (like unused classpath dependencies) but I'm a little confused as to how that unused list file is used.

ittaiz:
+1 is where we pass the compiler the direct dependencies + their direct
dependencies since often (not always) it needs that extra layer of
indirection (for example inheritance).
This is purely a heuristic designed at lowering the number of WTF moments a
developer has when compilation breaks and they need to add a dependency
they don't think they're using.
Btw re unused-deps: I've tried using it internally and several times it
asked me to remove dependencies that the compiler actually needs.
@johnynek how satisfied are y'all with it? How many exclusions and hacks do
you need to do to make it happy?

johnynek:
It works well enough that we removed a ton of dependencies which improved build caching.

We keep it turned on in general, but it isn’t perfect. I think we seem more biased towards minimizing rebuilds at some expense to usability. That may not be the trade you want to make.

We very rarely see cases where it can’t be satisfied but it happens. In those cases we disable it.

johnynek:
It would be really nice if the plugin were more precise and even better if scalac included this functionality built in.

ittaiz:
+1 for more precise and +100 for scalac including it.
I wanted to turn it on after we moved to +1 but saw enough issue  to keep
it off.
I should really investigate them more to see if there's any improvement we
can make to the plugin.
What people here now use is an internal tool (which we're considering to
OSS) called DepFixer.
DepFixer basically scans the compiler output to find FQN scalac needs, then
looks up in a few indices (one of them is a global Wix codebase index) to
find the matching labels and then according to scoring adds the label it
thinks is most relevant.
Then it runs bazel again to see that it indeed fixed the dependencies. It
can iterate for a long time and has many features like trackback and giving
up after no changes.
This allows us to remove all dependencies at all (other than proto) and let
it go wild :)

For now it's giving us some value in the unused deps front (it was created
for other purposes) so it's not that burning to turn on unused-deps itself.

PR #766, opened 2019-06-12
Specs2 filtering of JUnit descriptions

hmemcpy:
This PR augments the fix introduced in #759. The previous PR introduced a bug with incorrectly flattening JUnit descriptions - their original hashCode value was lost, and this caused failed tests to appear as succeeded in the bazel runner.

This PR replaces the flattening logic by having Specs2 itself create the Descriptions, keeping their original hashCode values.

Adding some tests to make sure only valid results appear in the XML.

@ittaiz

### DIFF
--- a/src/java/io/bazel/rulesscala/specs2/Specs2RunnerBuilder.scala
+++ b/src/java/io/bazel/rulesscala/specs2/Specs2RunnerBuilder.scala
@@ -3,8 +3,8 @@ package io.bazel.rulesscala.specs2
 import java.util
 import java.util.regex.Pattern
 
-import io.bazel.rulesscala.test_discovery._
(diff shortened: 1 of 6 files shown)


hmemcpy:
Ugh, this needs a bit more work.

hmemcpy:
Hmm, I don't know why two of these seem to fail... They pass on my machine too, any ideas @ittaiz @johnynek?

ittaiz:
Not sure exactly what's going on:
```
Test "scala_specs2_only_failed_test_shows_in_the_xml" failed  (3 sec) 
The command "if [[ "${TRAVIS_OS_NAME}" == "windows" ]]; then
```

hmemcpy:
OMG finally! Turns out you need to set `set +e` in your test if you're expecting test failures. Everything should pass now. @ittaiz please take a look.

PR #770, opened 2019-06-16
Add Canva to the adopters list

thundergolfer:
### Description

We're using Bazel in our Data Engineering + Data Science monorepo to build Scala (and Python) 🙂


ittaiz:
Happy to hear! Thanks for updating :)

Issue #771, opened 2019-06-18
README instructions seem to be failing me:


Parth:
When I add the following to my `WORKSPACE` file (only thing in that file), 
```
rules_scala_version="a89d44f7ef67d93dedfc9888630f48d7723516f7" # I have tried a handful of versions here, same result

load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "io_bazel_rules_scala",
    strip_prefix = "rules_scala-%s" % rules_scala_version,
    type = "zip",
    url = "https://github.com/bazelbuild/rules_scala/archive/%s.zip" % rules_scala_version,
)

load("@io_bazel_rules_scala//scala:toolchains.bzl", "scala_register_toolchains")
scala_register_toolchains()

load("@io_bazel_rules_scala//scala:scala.bzl", "scala_repositories")
scala_repositories()

protobuf_version="66dc42d891a4fc8e9190c524fd67961688a37bbe"
protobuf_version_sha256="983975ab66113cbaabea4b8ec9f3a73406d89ed74db9ae75c74888e685f956f8"

http_archive(
    name = "com_google_protobuf",
    url = "https://github.com/protocolbuffers/protobuf/archive/%s.tar.gz" % protobuf_version,
    strip_prefix = "protobuf-%s" % protobuf_version,
    sha256 = protobuf_version_sha256,
)
```

I see:

```DEBUG: Rule 'io_bazel_rules_scala' indicated that a canonical reproducible form can be obtained by modifying arguments sha256 = "e341dc30a397c77e6723a953d4a84539c050b49f163b4c50cf865c698ea598fe"
ERROR: /private/var/tmp/_bazel_parth.mehrotra/f3d049f4cd59e2f8a0b016862811f8be/external/io_bazel_rules_scala/scala/scala.bzl:111:13: cfg must be either 'host' or 'target'.
ERROR: /private/var/tmp/_bazel_parth.mehrotra/f3d049f4cd59e2f8a0b016862811f8be/external/io_bazel_rules_scala/scala/scala.bzl:125:22: Traceback (most recent call last):
        File "/private/var/tmp/_bazel_parth.mehrotra/f3d049f4cd59e2f8a0b016862811f8be/external/io_bazel_rules_scala/scala/scala.bzl", line 125
                _common_attrs.update(_common_attrs_for_plugin_bootstr...)
[...]

PR #772, opened 2019-06-19
Specify which version of bazel is required.

Parth:
#771


Parth:
Done! Let me know your thoughts

PR #773, opened 2019-06-19
Depset is not iterable

ianoc-stripe:
In 0.27 by default depset's aren't iterable. So just need a simple to_list to do this


ianoc-stripe:
using rules_scala in bazel 0.27 -- so its possible there are some other locations that haven't come up yet

ittaiz:
I think we went over the other places a while back. This probably has no
tests so didn't show up on CI before.
Note that Bazel folk build our rules with latest bazel with incompatible
flags so they surface such issues beforehand (if they surface in tests)

PR #774, opened 2019-06-19
JavaInfo provider should be given for deps

long-stripe:
It seems like `deps` should behave like `runtime_deps` in that a `JavaInfo` provider must be given.

Opening a drive-by PR now to see if CI will build this and confirm my hunch...

### DIFF
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -131,10 +131,10 @@ _common_attrs_for_plugin_bootstrapping = {
             _collect_plus_one_deps_aspect,
             _coverage_replacements_provider.aspect,
         ],
-        providers = [JavaInfo],
+        providers = [[JavaInfo]],
     ),
     "plugins": attr.label_list(allow_files = [".jar"]),
-    "runtime_deps": attr.label_list(providers = [JavaInfo]),
+    "runtime_deps": attr.label_list(providers = [[JavaInfo]]),
     "data": attr.label_list(allow_files = True),
     "resources": attr.label_list(allow_files = True),
     "resource_strip_prefix": attr.string(),
@@ -673,7 +673,7 @@ _scala_junit_test_attrs.update(_common_attrs)
 _scala_junit_test_attrs.update(_junit_resolve_deps)
 
 _scala_junit_test_attrs.update({
-    "tests_from": attr.label_list(providers = [JavaInfo]),
+    "tests_from": attr.label_list(providers = [[JavaInfo]]),
 })
 
 scala_junit_test = rule(
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -126,10 +126,13 @@ _common_attrs_for_plugin_bootstrapping = {
         ".srcjar",
         ".java",
     ]),
-    "deps": attr.label_list(aspects = [
-        _collect_plus_one_deps_aspect,
-        _coverage_replacements_provider.aspect,
-    ]),
+    "deps": attr.label_list(
+        aspects = [
+            _collect_plus_one_deps_aspect,
+            _coverage_replacements_provider.aspect,
+        ],
+        providers = [[JavaInfo]],
+    ),
     "plugins": attr.label_list(allow_files = [".jar"]),
     "runtime_deps": attr.label_list(providers = [[JavaInfo]]),
     "data": attr.label_list(allow_files = True),
--- a/scala/scala.bzl
+++ b/scala/scala.bzl
@@ -131,10 +131,10 @@ _common_attrs_for_plugin_bootstrapping = {
             _collect_plus_one_deps_aspect,
             _coverage_replacements_provider.aspect,
         ],
-        providers = [[JavaInfo]],
+        providers = [JavaInfo],
     ),
     "plugins": attr.label_list(allow_files = [".jar"]),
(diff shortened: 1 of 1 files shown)


long-stripe:
cc: @andyscott

andyscott:
It should just be a singly nested array instead of a doubly nested array. The doubly nested syntax is disjunctions of conjunctions of provider requirements whereas the singly nested is just a conjunction of requirements.

long-stripe:
> disjunctions of conjunctions

SCAS2002-2__52557 1390369862 1280 1280

... anyhow, [Bazel docs](https://docs.bazel.build/versions/master/skylark/lib/attr.html#label_list) say:

> The format of this argument is a list of lists of providers -- *Info objects returned by provider() (or in the case of a legacy provider, its string name). The dependency must return ALL providers mentioned in at least ONE of the inner lists. As a convenience, this argument may also be a single-level list of providers, in which case it is wrapped in an outer list with one element.

I'm fine writing it either way, but it does sound like a list of lists of providers is the norm, I only very slightly worry that Bazel would deprecate the singly nested usage as Bazel may do from time to time. Bazel, why are you the way that you are?

ittaiz:
We had a very good reason to not add JavaInfo provider to deps. IIRC it was mainly for Stripe's needs. I'll try to take a look today or tomorrow at why we needed it. Would appreciate if we don't merge too fast.

ittaiz:
It was @johnynek's [decision](https://github.com/bazelbuild/rules_scala/pull/523) at the time. didn't have the time to dive into the PR discussion on why not to add to `deps` (maybe it's irrelevant now)

long-stripe:
Ah well, I checked with an internal Stripe build running rules_scala off of a464f61f80c9dd70cb0f719fa226d3364fab8bd4 and it definitely fails:

```
in deps attribute of scala_test rule //[snip]: '//[snip]' does not have mandatory providers: 'JavaInfo'
```

It does seem to make sense that `deps` must return `JavaInfo`, but reality does not seem to reflect that.

(also, the test failure for a464f61f80c9dd70cb0f719fa226d3364fab8bd4 seems to go away locally if I revert it, let's see if CI says the same).

johnynek:
Sorry I had forgotten that.

Long before providers we were really loose in what we accepted, even raw files we would just put on the class path.

I guess that is what is being exploited on a few targets at Stripe. I think we should keep this open, investigate why our targets are failing and I think it will probably be the case that they could be changed slightly and use this change.

I think using providers to the utmost is a good idea to get reach. They are the closest things starlark has to types.

long-stripe:
> I think using providers to the utmost is a good idea to get reach. They are the closest things starlark has to types.

Agreed 100%

ittaiz:
+1
Just need to make this transition.
I think the best option for now would be to add a check in the code that
fails if no JavaInfo exists *unless* someone explicitly enables it via
—define.
This means we start a migration period (and the fail message can say that
and point to the workaround for now).
Wdyt?

long-stripe:
@ittaiz makes sense to have a migration period since non-`JavaInfo` `deps` were allowed in the past, but I assume that would mean any rule implementation that uses `_common_attrs_for_plugin_bootstrapping` would need to check `deps`? You could stuff that checking logic into a util function, but you'd still need to call it from every rule impl right?

ittaiz:
Indeed.
Due to lack of visibility on bzl files we indeed can't be sure who needs it.
Still if we cover the ~10 (20?) rules we have with this call we should be ok

johnynek:
@long-stripe any update here? How many targets were red internally? Could those have actually been buggy targets (depending on something that was actually meaningless) or are they doing something sane?

long-stripe:
@johnynek I haven't looked at it again in depth. Let me give it another poke when I have some time, either today or early this week.

long-stripe:
Quick update: got a green build after some whack a mole. 16 targets that had non-JavaInfo deps that weren't necessary, and 3 `scala_repl` targets that were somewhat unprincipled in their passing in deps. The latter would need a little more massaging to get working correctly, or be disabled/commented out altogether.

I'll poke again later but it isn't _that_ bad after all.

johnynek:
seems like we should fix those, and consider merging this PR since it seems we have no legitimate uses of it?

long-stripe:
Yeah I agree, not too hard to fix. At the very least I can merge my internal fixes and this can be landed.

ittaiz:
Oscar,
I'd love to test this internally to also verify.
Long Cao,
Any chance you can rebase onto master? This way I'll take your commit and
test it.

long-stripe:
@ittaiz Sure! Btw my recent test was against a rebased branch, so things are looking good.

long-stripe:
Rebased and now TravisCI is failing now; also reproduced the build failure locally. I don't know enough to know what's wrong here, what do y'all think?

johnynek:
the only build failure I see is #699 

Is that what you reproduced locally? We should fix that....

johnynek:
@ittaiz let us know when you run this. This found bugs internally we are fixing. I think it is worth a merge to help others find similar bugs.

ittaiz:
Ran it, fixed some things.
We’re good to merge as far as I’m concerned

PR #775, opened 2019-06-20
Migrate rules_scala for https://github.com/bazelbuild/bazel/issues/7153

hlopko:
This PR simply follows the migration instructions in
https://github.com/bazelbuild/bazel/issues/7153, that is renames
`proto_source_root` to `strip_import_prefix` and prepends the attribute
value with `/`


hlopko:
It seems the travis is failing, but Bazel CI not, and I cannot make it fail for me locally (even with 0.23.1), do you have any advice how to reproduce please?

Thank you!

johnynek:
There are negative tests that the bazel CI does not test. These are tests where we expect a failure to compile or failure of a test:

https://travis-ci.org/bazelbuild/rules_scala/jobs/548557592#L506

Can you run the target it is complaining about with the options shown there?

johnynek:
PS: if you have any suggestions for clean ways to do negative tests in bazel, that would be great. We want to test that building some targets errors, or running some tests fail. It hasn't been clear to us if it is safe to run bazel recursively: e.g. a bazel rule that runs bazel and expects failure on a test. It seems most rule authors should need this functionality because we absolutely need negative tests in rules.

hlopko:
cc @iirina 

Sorry for the radio silence, I hope to come back to this soon.

johnynek:
Am I right that #793 supersedes this?

johnynek:
Please reopen if this is still needed.

PR #778, opened 2019-06-29
Update "Getting started" WORKSPACE block

tenorviol:
I'm a longtime user of SBT, but completely new to Bazel. The documentation on getting started compiling Scala projects with Bazel is pretty sparse, so thank you for this project! I had to make the following changes to get my first tiny Scala project to build with Bazel. It's unclear to me why I needed to add the Skylib reference, but the build command by default complains about its lack.


tenorviol:
I signed it!

PR #780, opened 2019-07-08
Allow for code coverage to be run on 0.27.1

d-haxton:
I spent some time investigating code coverage and I ran into a few issues on the latest version of bazel.

- 0.27.0 had a regression and is entirely borked. Thankfully they released a patchfor that: https://github.com/bazelbuild/bazel/issues/8670

- We had pinned our Jacoco and asm versions to what bazel previously had. Caused significant conflicts when attempting to instrument. 
- We were incorrectly passing `_lcov_merger` into the runpath. This caused the infamous zip error seen on the other issue tracking code coverage.

The latter 2 have been addressed in this PR and you can now run coverage on 0.27.1 *

* ...if you don't reference a local java_library. I've lost the needle here and don't have much to go on at this point. Hopefully, someone here may know something or at the very least can point me in a direction they think I might find an error message because thus far I have none left to debug.

Combined reports do indeed work. You can test this out with the following command:

```
 bazel coverage //test/coverage/... --incompatible_no_support_tools_in_action_inputs=false --incompatible_depset_is_not_iterable=false --extra_toolchains="//test/coverage:enable_code_coverage_aspect" --coverage_report_generator="@bazel_tools//tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator:Main" --combined_report=lcov
```


johnynek:
looks like some of our tests are red. Can you take a look and see if we can get them green?

(also perhaps merge master).

johnynek:
Also, thank you very much for the PR!

d-haxton:
Everything should be kosher now.

johnynek:
Thank you!

PR #781, opened 2019-07-08
Migrate from java_common.create_provider to JavaInfo()

iirina:
Progress on https://github.com/bazelbuild/bazel/issues/5821.

### DIFF
--- a/scala/Dummy.java
+++ b/scala/Dummy.java
@@ -1,0 +1,5 @@
+public class Dummy {
+	public static void main(String[] args) {
+
+	}
+}
--- a/scala/Dummy.java
+++ b/scala/Dummy.java
@@ -1,5 +1,0 @@
-public class Dummy {
-	public static void main(String[] args) {
-
-	}
-}
--- a/scala/private/common.bzl
+++ b/scala/private/common.bzl
@@ -62,11 +62,14 @@ def _collect_jars_when_dependency_analyzer_is_off(
     runtime_jars = []
     jars2labels = {}
 
+    deps_providers = []
+
     for dep_target in dep_targets:
         # we require a JavaInfo for dependencies
         # must use java_import or scala_import if you have raw files
         if JavaInfo in dep_target:
             java_provider = dep_target[JavaInfo]
+            deps_providers.append(java_provider)
             compile_jars.append(java_provider.compile_jars)
             runtime_jars.append(java_provider.transitive_runtime_jars)
 
@@ -84,12 +87,14 @@ def _collect_jars_when_dependency_analyzer_is_off(
             plus_one_deps_compile_jars.append(
                 depset(transitive = [dep[JavaInfo].compile_jars for dep in dep_target[PlusOneDeps].direct_deps if JavaInfo in dep ])
             )
+            deps_providers.append([dep[JavaInfo] for dep in dep_target[PlusOneDeps].direct_deps if JavaInfo in dep ])
 
     return struct(
         compile_jars = depset(transitive = compile_jars),
         transitive_runtime_jars = depset(transitive = runtime_jars),
         jars2labels = JarsToLabelsInfo(jars_to_labels = jars2labels),
         transitive_compile_jars = depset(transitive = compile_jars + plus_one_deps_compile_jars),
+        deps_providers = deps_providers,
     )
 
 def _collect_jars_when_dependency_analyzer_is_on(dep_targets):
@@ -97,12 +102,14 @@ def _collect_jars_when_dependency_analyzer_is_on(dep_targets):
     jars2labels = {}
     compile_jars = []
     runtime_jars = []
+    deps_providers = []
 
     for dep_target in dep_targets:
         # we require a JavaInfo for dependencies
         # must use java_import or scala_import if you have raw files
         if JavaInfo in dep_target:
             java_provider = dep_target[JavaInfo]
+            deps_providers.append(java_provider)
             current_dep_compile_jars = java_provider.compile_jars
             current_dep_transitive_compile_jars = java_provider.transitive_compile_time_jars
             runtime_jars.append(java_provider.transitive_runtime_jars)
@@ -124,6 +131,7 @@ def _collect_jars_when_dependency_analyzer_is_on(dep_targets):
         transitive_runtime_jars = depset(transitive = runtime_jars),
         jars2labels = JarsToLabelsInfo(jars_to_labels = jars2labels),
         transitive_compile_jars = depset(transitive = transitive_compile_jars),
+        deps_providers = deps_providers,
     )
 
 # When import mavan_jar's for scala macros we have to use the jar:file requirement
@@ -163,36 +171,3 @@ def _provider_of_dependency_label_of(dependency, path):
         return dependency[JarsToLabelsInfo].jars_to_labels.get(path)
     else:
         return None
-
-# TODO this seems to have limited value now that JavaInfo has everything
-def create_java_provider(scalaattr, transitive_compile_time_jars):
-    compile_time_deps = [
-        JavaInfo(
-            output_jar = jar,
-            compile_jar = jar,
-        )
-        #TODO: Avoid flattening the depset here.
-        for jar in transitive_compile_time_jars.to_list()
-    ]
-
-    compile_providers_neverlink = [
-        JavaInfo(
-            output_jar = jar,
-            compile_jar = jar,
-            # these are only header jars and they should not be used as runtime
-            neverlink = True)
-        for jar in scalaattr.compile_jars.to_list()
-    ]
-
-    runtime_providers = [
-        JavaInfo(
-            output_jar = jar,
-            compile_jar = jar,
-            deps = compile_time_deps
-        )
-        #TODO: Avoid flattening the depset here.
-        for jar in scalaattr.transitive_runtime_jars.to_list()
-    ]
-
-    merged_provider = java_common.merge(runtime_providers + compile_providers_neverlink)
-    return merged_provider
--- a/scala/private/common.bzl
+++ b/scala/private/common.bzl
@@ -166,31 +166,33 @@ def _provider_of_dependency_label_of(dependency, path):
 
 # TODO this seems to have limited value now that JavaInfo has everything
 def create_java_provider(scalaattr, transitive_compile_time_jars):
-
     compile_time_deps = [
         JavaInfo(
             output_jar = jar,
-            compile_jar = jar
+            compile_jar = jar,
         )
         #TODO: Avoid flattening the depset here.
         for jar in transitive_compile_time_jars.to_list()
     ]
 
-    compile_providers = [
+    compile_providers_neverlink = [
         JavaInfo(
             output_jar = jar,
             compile_jar = jar,
-            deps = compile_time_deps)
-        #TODO: Avoid flattening the depset here.
+            # these are only header jars and they should not be used as runtime
+            neverlink = True)
         for jar in scalaattr.compile_jars.to_list()
     ]
+
     runtime_providers = [
         JavaInfo(
             output_jar = jar,
-            compile_jar = jar
+            compile_jar = jar,
+            deps = compile_time_deps
         )
         #TODO: Avoid flattening the depset here.
         for jar in scalaattr.transitive_runtime_jars.to_list()
     ]
 
-    return java_common.merge(compile_providers + runtime_providers)
+    merged_provider = java_common.merge(runtime_providers + compile_providers_neverlink)
+    return merged_provider
--- a/scala/private/common.bzl
+++ b/scala/private/common.bzl
@@ -166,12 +166,27 @@ def _provider_of_dependency_label_of(dependency, path):
 
 # TODO this seems to have limited value now that JavaInfo has everything
 def create_java_provider(scalaattr, transitive_compile_time_jars):
-    return java_common.create_provider(
-        use_ijar = False,
-        compile_time_jars = scalaattr.compile_jars,
-        runtime_jars = scalaattr.transitive_runtime_jars,
-        transitive_compile_time_jars = depset(
-            transitive = [transitive_compile_time_jars, scalaattr.compile_jars],
-        ),
-        transitive_runtime_jars = scalaattr.transitive_runtime_jars,
-    )
+
+    compile_time_deps = [
+        JavaInfo(
+            output_jar = jar,
+            compile_jar = output_jar
+        )
+        for jar in transitive_compile_time_jars
+    ]
+
+    compile_providers = [
+        JavaInfo(
+            output_jar = compile_jar,
+            compile_jar = compile_jar,
+            deps = compile_time_deps)
+        for compile_jar in scalaattr.compile_jars
+    ]
+    runtime_providers = [
+        JavaInfo(
+            output_jar = runtime_jar,
+            compile_jar = compile_jar)
+        for runtime_jar in scalaattr.transitive_runtime_jars
+    ]
+
+    return java_common.merge(compile_providers, runtime_providers)
--- a/scala/private/common.bzl
+++ b/scala/private/common.bzl
@@ -172,7 +172,8 @@ def create_java_provider(scalaattr, transitive_compile_time_jars):
             output_jar = jar,
             compile_jar = jar
         )
-        for jar in transitive_compile_time_jars
+        #TODO: Avoid flattening the depset here.
+        for jar in transitive_compile_time_jars.to_list()
     ]
 
     compile_providers = [
@@ -180,14 +181,16 @@ def create_java_provider(scalaattr, transitive_compile_time_jars):
             output_jar = jar,
             compile_jar = jar,
             deps = compile_time_deps)
-        for jar in scalaattr.compile_jars
+        #TODO: Avoid flattening the depset here.
+        for jar in scalaattr.compile_jars.to_list()
     ]
     runtime_providers = [
         JavaInfo(
             output_jar = jar,
             compile_jar = jar
         )
-        for jar in scalaattr.transitive_runtime_jars
+        #TODO: Avoid flattening the depset here.
+        for jar in scalaattr.transitive_runtime_jars.to_list()
     ]
 
-    return java_common.merge(compile_providers, runtime_providers)
+    return java_common.merge(compile_providers + runtime_providers)
--- a/scala/private/common.bzl
+++ b/scala/private/common.bzl
@@ -170,23 +170,24 @@ def create_java_provider(scalaattr, transitive_compile_time_jars):
     compile_time_deps = [
         JavaInfo(
             output_jar = jar,
-            compile_jar = output_jar
+            compile_jar = jar
         )
         for jar in transitive_compile_time_jars
     ]
 
     compile_providers = [
         JavaInfo(
-            output_jar = compile_jar,
-            compile_jar = compile_jar,
+            output_jar = jar,
+            compile_jar = jar,
             deps = compile_time_deps)
-        for compile_jar in scalaattr.compile_jars
+        for jar in scalaattr.compile_jars
     ]
     runtime_providers = [
         JavaInfo(
-            output_jar = runtime_jar,
-            compile_jar = compile_jar)
-        for runtime_jar in scalaattr.transitive_runtime_jars
+            output_jar = jar,
+            compile_jar = jar
+        )
+        for jar in scalaattr.transitive_runtime_jars
     ]
 
     return java_common.merge(compile_providers, runtime_providers)
--- a/scala/private/common.bzl
+++ b/scala/private/common.bzl
@@ -87,7 +87,6 @@ def _collect_jars_when_dependency_analyzer_is_off(
             plus_one_deps_compile_jars.append(
                 depset(transitive = [dep[JavaInfo].compile_jars for dep in dep_target[PlusOneDeps].direct_deps if JavaInfo in dep ])
             )
-            deps_providers.append([dep[JavaInfo] for dep in dep_target[PlusOneDeps].direct_deps if JavaInfo in dep ])
 
     return struct(
         compile_jars = depset(transitive = compile_jars),
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -318,10 +318,10 @@ StatsfileOutput: {statsfile_output}
     )
 
 def _interim_java_provider_for_java_compilation(scala_output):
-    return java_common.create_provider(
-        use_ijar = False,
-        compile_time_jars = [scala_output],
-        runtime_jars = [],
+    return JavaInfo(
+        output_jar = scala_output,
+        compile_jar = scala_output,
+        neverlink = True
     )
 
 def _scalac_provider(ctx):
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -396,12 +396,12 @@ def _compile_or_empty(
         implicit_junit_deps_needed_for_java_compilation,
         unused_dependency_checker_mode,
         unused_dependency_checker_ignored_targets,
-        deps_providers = []):
+        deps_providers):
     # We assume that if a srcjar is present, it is not empty
     if len(ctx.files.srcs) + len(srcjars.to_list()) == 0:
         _build_nosrc_jar(ctx)
 
-        scala_compilation_provider = _get_scala_compilation_provider(ctx, ctx.outputs.jar, deps_providers)
+        scala_compilation_provider = _create_scala_compilation_provider(ctx, ctx.outputs.jar, deps_providers)
 
         #  no need to build ijar when empty
         return struct(
@@ -476,7 +476,7 @@ def _compile_or_empty(
             #  so set ijar == jar
             ijar = ctx.outputs.jar
 
-        scala_compilation_provider = _get_scala_compilation_provider(ctx, ijar, deps_providers)
+        scala_compilation_provider = _create_scala_compilation_provider(ctx, ijar, deps_providers)
 
         # compile the java now
         java_jar = try_to_compile_java_jar(
@@ -513,7 +513,7 @@ def _compile_or_empty(
             merged_provider = merged_provider
         )
 
-def _get_scala_compilation_provider(ctx, ijar, deps_providers):
+def _create_scala_compilation_provider(ctx, ijar, deps_providers):
     exports = []
     if hasattr(ctx.attr, "exports"):
         exports = [dep[JavaInfo] for dep in ctx.attr.exports]
@@ -748,8 +748,6 @@ def _collect_jars_from_common_ctx(
         _is_plus_one_deps_off(ctx),
     )
 
-#    deps_jars.deps_providers
-
     (
         cjars,
         transitive_rjars,
@@ -909,7 +907,7 @@ def _scala_binary_common(
         java_wrapper,
         unused_dependency_checker_mode,
         unused_dependency_checker_ignored_targets,
-        deps_providers = [],
+        deps_providers,
         implicit_junit_deps_needed_for_java_compilation = [],
         runfiles_ext = []):
     write_manifest(ctx)
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -28,7 +28,6 @@ load(
     "collect_jars",
     "collect_plugin_paths",
     "collect_srcjars",
-    "create_java_provider",
     "not_sources_jar",
     "write_manifest",
 )
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -401,20 +401,7 @@ def _compile_or_empty(
     if len(ctx.files.srcs) + len(srcjars.to_list()) == 0:
         _build_nosrc_jar(ctx)
 
-        exports = []
-        if hasattr(ctx.attr, "exports"):
-            exports = [dep[JavaInfo] for dep in ctx.attr.exports]
-        runtime_deps = []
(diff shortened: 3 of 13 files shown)


scala/scala_import.bzl
```diff
@@ -53,14 +57,14 @@ def _create_provider(
         neverlink,
         source_jar,
         intellij_metadata):
-    transitive_runtime_jars = [
+    new_transitive_runtime_jars = [
```
johnynek: You are only changing the name of this where you create it not where it is used. That seems like a bug to me.

scala/private/common.bzl
```diff
@@ -172,22 +172,25 @@ def create_java_provider(scalaattr, transitive_compile_time_jars):
...
             output_jar = jar,
             compile_jar = jar
         )
-        for jar in scalaattr.transitive_runtime_jars
+        #TODO: Avoid flattening the depset here.
+        for jar in scalaattr.transitive_runtime_jars.to_list()
     ]
 
-    return java_common.merge(compile_providers, runtime_providers)
+    return java_common.merge(compile_providers + runtime_providers)
```
johnynek: Rather than using + here can we make a list like all_providers and extend into it. I think this should be better performance since skylark does not have sublinear list concat AFAIK.

iirina:
@johnynek thanks for the comments! The PR is still a draft and I'm still figuring out some things. I'll let you know when it's ready for review.

iirina:
I'm struggling now with re-implementing `create_java_provider` from `scala/private/common.bzl` (surprize surprize). I'm not familiar with the code and I don't understand why the old provider was created that way. If someone could explain that would help me a lot.

I'm stuck now because the following command fails with the message:

```
$ bazel build test:HelloLib
INFO: Analyzed target //test:HelloLib (6 packages loaded, 608 targets configured).
INFO: Found 1 target...
ERROR: /usr/local/google/home/elenairina/rules_scala/test/BUILD:63:1: scala //test:HelloLib failed (Exit 1)
test/HelloLib.scala:20: warning: Remove the if expression and use the condition directly.
  def dumb(x: Int) = if (x == 3) true else false
                           ^
error: Target '//test:Exported' is specified as a dependency to //test:HelloLib but isn't used, please remove it from the deps.
You can use the following buildozer command:
buildozer 'remove deps //test:Exported' //test:HelloLib
one warning found
one error found
one warning found
one error found
java.lang.RuntimeException: Build failed
	at io.bazel.rulesscala.scalac.ScalacProcessor.compileScalaSources(ScalacProcessor.java:244)
	at io.bazel.rulesscala.scalac.ScalacProcessor.processRequest(ScalacProcessor.java:69)
Target //test:HelloLib failed to build
INFO: Elapsed time: 1.314s, Critical Path: 0.38s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
```

I'm wondering how it's determined that `Exported` is not used. What process looks where?

@johnynek @ittaiz

johnynek:
@iirina we have a scala compiler plugin that errors (when enabled) if you have dependencies you don't use. This is pretty important in scala because the compiler is so much slower and otherwise we would build up a lot of useless dependencies that slow down the build.

My guess, without yet looking at the latest code, is that you have changed what makes it onto the classpath, which is breaking some tests that are looking for strict classpaths.

scala/private/common.bzl
```diff
@@ -166,12 +166,33 @@ def _provider_of_dependency_label_of(dependency, path):
...
-        ),
-        transitive_runtime_jars = scalaattr.transitive_runtime_jars,
-    )
+    compile_time_deps = [
+        JavaInfo(
+            output_jar = jar,
+            compile_jar = jar,
+        )
+        #TODO: Avoid flattening the depset here.
+        for jar in transitive_compile_time_jars.to_list()
```
johnynek: isn't this putting all transitive dependencies as direct dependencies? Note previously there was a distinction between `compile_time_jars` and `transitive_compile_time_jars`. It seems we are losing that distinction here.

scala/scala_import.bzl
```diff
@@ -24,20 +24,27 @@ def _scala_import_impl(ctx):
...
+        compile_jar = jar,
+        exports = [target[JavaInfo] for target in ctx.attr.exports],
+        deps = [target[JavaInfo] for target in ctx.attr.deps],
+        runtime_deps = [target[JavaInfo] for target in ctx.attr.runtime_deps],
+        source_jar = ctx.file.srcjar,
+        neverlink = ctx.attr.neverlink
+    )
+    for jar in current_target_compile_jars]
+
+    if not current_target_providers:
```
ittaiz: this is for cases where there are no current jars but only exports?

ittaiz: I wonder if the below shouldn't happen in any case. Not sure.

scala/scala_import.bzl
```diff
@@ -70,18 +77,29 @@ def _create_provider(
...
-            exports.transitive_compile_jars,
-        ]),
-        transitive_runtime_jars = depset(transitive = transitive_runtime_jars),
-        source_jars = source_jars,
+    deps = [
+        JavaInfo(
+            output_jar = jar,
+            compile_jar = jar
+        )
+        for jar in jars.transitive_compile_jars.to_list()
```
ittaiz: should transitive compile_time_jars be added as direct compile time deps?

scala/scala_import.bzl
```diff
@@ -70,18 +77,29 @@ def _create_provider(
...
+    ]
+
+    deps.append([
+        JavaInfo(
+            output_jar = jar,
+            compile_jar = jar
+        )
+        for jar in transitive_runtime_jars
```
ittaiz: should transitive *runtime* jars be added as direct compile time deps?

test_expect_failure/missing_direct_deps/internal_deps/custom-jvm-rule.bzl
```diff
@@ -1,13 +1,8 @@
...
     transitive_compile_jars = _collect(ctx.attr.deps)
-    return struct(
-        providers = [
-            java_common.create_provider(
-                transitive_compile_time_jars = transitive_compile_jars,
-            ),
-        ],
-    )
+    providers = [JavaInfo(output_jar = jar, compile_jar = jar) for jar in transitive_compile_jars]
+    return [java_common.merge(providers)]
```
ittaiz: doesn't this mean that the new merged JavaInfo will have `transitive_compile_jars` as top level dependencies? Currently they are transitive and not direct.

src/java/io/bazel/rulesscala/scalac/jvm_export_toolchain.bzl
```diff
@@ -14,12 +14,8 @@ def _export_scalac_repositories_from_toolchain_to_jvm_impl(ctx):
     default_repl_classpath_files = _files_of(
         default_repl_classpath_deps,
     ).to_list()
-    java_common_provider = java_common.create_provider(
-        use_ijar = False,
-        compile_time_jars = default_repl_classpath_files,
-        runtime_jars = default_repl_classpath_files,
-    )
-    return [java_common_provider]
+    providers = [JavaInfo(output_jar = jar, compile_jar = jar) for jar in default_repl_classpath_files]
```
ittaiz: cool!

scala/private/rule_impls.bzl
```diff
@@ -318,10 +318,10 @@ StatsfileOutput: {statsfile_output}
     )
 
 def _interim_java_provider_for_java_compilation(scala_output):
```
ittaiz: I think I need to read the code a bit more in depth but from what I remember this won't work because this way we won't have the java code present at runtime. I think the whole "interim" approach is now wrong and was needed before.
I might be mistaken but my hunch is that a bit more of a refactor is needed to have the scala compilation propagate out the JavaInfo of the java (or a different refactoring)

scala/private/common.bzl
```diff
@@ -166,12 +166,33 @@ def _provider_of_dependency_label_of(dependency, path):
...
+    ]
+
+    compile_providers_neverlink = [
+        JavaInfo(
+            output_jar = jar,
+            compile_jar = jar,
+            # these are only header jars and they should not be used as runtime
+            neverlink = True)
```
ittaiz: I think this is wrong. These aren't "compile only" but are derived from `deps` which means "compile" and "runtime". Now that you mention it I'm thinking we might have a bug in ignoring neverlink for some cases but I need to dive more into the code

scala/private/common.bzl
ittaiz: @johnynek I don't think so since @iirina later uses `compile_time_deps` as dependencies of `runtime_providers` which means you have another layer of indirection.
If I understand correctly this preserves semantics for you but breaks it for us (who use +1).
Maybe what's needed is to create another indirection provider and put that one as a dependency on the final provider. This is hacky but should be simple to do and maintain the current semantics.

scala/private/common.bzl
```diff
@@ -166,12 +166,33 @@ def _provider_of_dependency_label_of(dependency, path):
...
+        for jar in scalaattr.compile_jars.to_list()
+    ]
+
+    runtime_providers = [
+        JavaInfo(
+            output_jar = jar,
+            compile_jar = jar,
+            deps = compile_time_deps
```
ittaiz: should every runtime_provider have all `compile_time_deps`?

scala/scala_import.bzl
iirina: Yes, for the case with only exports. I added a TODO to figure out if merging the exports providers is always needed. IMO it's not, because the `JavaInfo`s created above specify the exports dependencies.

scala/scala_import.bzl
iirina: It was just part of trying to understand the variables used here. This method is not used anymore so I removed it.

scala/scala_import.bzl
iirina: I removed this method.

scala/scala_import.bzl
iirina: I removed this method.

src/java/io/bazel/rulesscala/scalac/jvm_export_toolchain.bzl
iirina: Finally something straightforward :smiley:

test_expect_failure/missing_direct_deps/internal_deps/custom-jvm-rule.bzl
iirina: That's true. I'm not sure how to create a provider that only has transitive jars and also I'm not sure we should. What is the purpose of `custom_jvm` rule? Why does it only have `deps`?

scala/private/rule_impls.bzl
iirina: What is the reason why `runtime_jars` was previously empty? 

IIUC this provider is needed to compile Java sources with scala dependencies. Does scala also need this provider?

iirina:
Thank you @johnynek and @ittaiz for taking time to review this PR! As I mentioned in one of the comments I refactored some logic for scala_library and scala_binary and created the providers there. 

I'm investigating now the failed test.

scala/private/rule_impls.bzl
```diff
@@ -394,11 +395,27 @@ def _compile_or_empty(
         jars2labels,
         implicit_junit_deps_needed_for_java_compilation,
         unused_dependency_checker_mode,
-        unused_dependency_checker_ignored_targets):
+        unused_dependency_checker_ignored_targets,
+        deps_providers = []):
```
ittaiz: why aren't we passing it in to the scala_binary? targets can depend on the binary (I think it's also supported in java_binary)

scala/private/rule_impls.bzl
```diff
@@ -471,6 +489,22 @@ def _compile_or_empty(
...
+            runtime_deps = [dep[JavaInfo] for dep in ctx.attr.runtime_deps]
+        scala_compilation_provider = JavaInfo(
+            output_jar = ctx.outputs.jar,
+            compile_jar = ijar,
+            deps = deps_providers,
+            exports = exports,
+            runtime_deps = runtime_deps,
+        )
+
+        deps_provider_merged = java_common.merge(deps_providers)
```
ittaiz: is this used?

scala/private/rule_impls.bzl
```diff
@@ -394,11 +395,27 @@ def _compile_or_empty(
...
         implicit_junit_deps_needed_for_java_compilation,
         unused_dependency_checker_mode,
-        unused_dependency_checker_ignored_targets):
+        unused_dependency_checker_ignored_targets,
+        deps_providers = []):
     # We assume that if a srcjar is present, it is not empty
     if len(ctx.files.srcs) + len(srcjars.to_list()) == 0:
         _build_nosrc_jar(ctx)
 
+        exports = []
```
ittaiz: should this be in a method? It's almost the same code below (apart from the ijar thing which actually confused me)

scala/private/rule_impls.bzl
iirina: Yes, exactly. This only works because it's neverlink, otherwise we need to pass in the full jar as well (output_jar=full_jar, compile_jar=ijar) to avoid having an ijar at runtime.

scala/private/rule_impls.bzl
iirina: Yup you're right, I forgot to pass deps_providers from scala_binary. I hope that was what you meant.

scala/private/rule_impls.bzl
iirina: Not used anymore. I removed it, thanks!

scala/private/rule_impls.bzl
iirina: Yup, done.

scala/private/common.bzl
```diff
@@ -84,25 +87,29 @@ def _collect_jars_when_dependency_analyzer_is_off(
             plus_one_deps_compile_jars.append(
                 depset(transitive = [dep[JavaInfo].compile_jars for dep in dep_target[PlusOneDeps].direct_deps if JavaInfo in dep ])
             )
+            deps_providers.append([dep[JavaInfo] for dep in dep_target[PlusOneDeps].direct_deps if JavaInfo in dep ])
```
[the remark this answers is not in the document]
iirina: OK, I removed it.

iirina:
@ittaiz Thansk for the review! If you don't mind taking a look into the [travis CI](https://travis-ci.org/bazelbuild/rules_scala/builds/556756307?utm_source=github_status&utm_medium=notification), that would be helpful.

scala/private/rule_impls.bzl
ittaiz: can we remove the default argument? Currently it might cause bugs (like the one we found)

ittaiz:
The test related to custom-jvm-rule passes (and so is my remark about it since I made a small hack for it). There is a new test that fails. I'll see if I can lend a hand there. Not sure.

scala/private/rule_impls.bzl
iirina: Yes, you're right. Removed the default value. Initially I needed it to test parts of the code before migrating the other usages.

scala/private/rule_impls.bzl
```diff
@@ -498,8 +510,24 @@ def _compile_or_empty(
             ijars = ijars,
             java_jar = java_jar,
             source_jars = source_jars,
+            merged_provider = merged_provider
         )
 
+def _get_scala_compilation_provider(ctx, ijar, deps_providers):
```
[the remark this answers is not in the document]
iirina: Done.

scala/private/rule_impls.bzl
```diff
@@ -877,6 +909,7 @@ def _scala_binary_common(
         java_wrapper,
         unused_dependency_checker_mode,
         unused_dependency_checker_ignored_targets,
+        deps_providers = [],
```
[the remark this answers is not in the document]
iirina: No, it doesn't. The same reasoning as above. I removed the default value.

scala/private/rule_impls.bzl
```diff
@@ -720,16 +748,20 @@ def _collect_jars_from_common_ctx(
         _is_plus_one_deps_off(ctx),
     )
 
+#    deps_jars.deps_providers
```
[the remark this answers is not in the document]
iirina: Done.

iirina:
Thanks @ittaiz for your contribution! Please also reply to @googlebot to say it's ok to be co-author of this PR (I don't know the magic words though).

@googlebot I'm ok with it ?

I'm investigating now why `test_scala_import_expect_failure_on_missing_direct_deps_warn_mode` fails. I'll come back with more details.

ittaiz:
I'm ok with it.

ittaiz:
@iirina Note to project maintainer: This is a terminal state, meaning the cla/google commit status will not change from this state. It's up to you to confirm consent of all the commit author(s), set the cla label to yes (if enabled on your project), and then merge this pull request when appropriate.

iirina:
List of failing test cases to investigate:

- [ ] test_scala_import_expect_failure_on_missing_direct_deps_warn_mode
- [ ] test_scala_import_library_passes_labels_of_direct_deps
- [ ] test_coverage_on

iirina:
@ittaiz Thanks for confirming. This googlebot story is a bit confusing :confused:

ittaiz:
indeed. I've spoken with one of the googlers responsible for this IIRC but
to no avail...
You got the list of test cases by commenting out the tests until you got a
green one? Because the bash script fails fast

ittaiz:
re the scala import e2e tests which fail I'd guess that scala_import exposes something different in the graph

iirina:
> You got the list of test cases by commenting out the tests until you got a green one? Because the bash script fails fast.

Yes, that's what I did. Fortunately there aren't many failing tests, otherwise it would have taken me a long time. :smile: 

> re the scala import e2e tests which fail I'd guess that scala_import exposes something different in the graph

Yup. It's the case with no jars, but with exports/deps/runtime_deps. Because there is no way to make express them as transitive dependencies, the targets depending on a scala_import with no jars sees its deps as direct deps as opposed to indirect. Thinking now of a workaround...

iirina:
A dirty workaround would be to create dummy top-level jars for these scala_imports, but that pollutes the classpaths and creates additional actions affecting performance :slightly_frowning_face:

ittaiz:
In java_import I can do it, right? What’s the difference?

iirina:
> In java_import I can do it, right? What’s the difference?

In java_import you can have empty jars. However my understanding is that it's a side effect, not an intended feature. java_import doesn't check for empty jars and when it creates the providers it simply adds empty lists wherever the jars are added.

In `JavaInfo()` we made the choice to forbid such behavior. IMO this constructor should not allow an empty `output_jar`. I understand the need for having this layer of indirection, so bazel shouldn't just forbid it. Having a new API that's more meaningful and special cased for the specific use case is IMO the best way forward (e.g. `java_common.export_deps(deps, runtime_deps, exports)`).

I was trying to migrate `rules_scala` without any bazel changes so the flag can be flipped in bazel 1.0. If we don't find any workarounds, the plan would be:
- add the new API in bazel 0.29 (august release)
- use the new API in scala_import and merge this PR in August
- flip the flag in bazel 1.0 (sept release)

iirina:
In my latest commit I added a temporary hack:

* a dummy java_library `//scala:dummy`
* `scala_import` always depends on `@rules_scala//scala:libdummy.jar`
* `scala_import` adds the dummy jar as top-level `output_jar` in `JavaInfo` when no jars are specified.

This approach preserves the same direct and indirect dependencies. We should remove this once we have a new API to properly express the layer of indirection.

Let me know what you think.

iirina:
Now all tests pass except test_reproducibility on linux. I wonder why it only fails on linux. Will investigate.

scala/scala_import.bzl
```diff
@@ -25,24 +25,33 @@ def _scala_import_impl(ctx):
...
+            JavaInfo(
+                output_jar = ctx.file._dummy,
+                compile_jar = ctx.file._dummy,
+                exports = [target[JavaInfo] for target in ctx.attr.exports],
+                deps = [target[JavaInfo] for target in ctx.attr.deps],
+                runtime_deps = [target[JavaInfo] for target in ctx.attr.runtime_deps],
+            ),
+        ]
+
+    #        current_target_providers = exports + deps + runtime_deps
```
ittaiz: commented out line

scala/scala_import.bzl
```diff
@@ -25,24 +25,33 @@ def _scala_import_impl(ctx):
...
+    ]
 
     # Handle the case with no jars.
     # TODO: Figure out if these providers should always be merged.
     if not current_target_providers:
-        exports = [java_common.merge([target[JavaInfo] for target in ctx.attr.exports])]
-        deps = [java_common.merge([target[JavaInfo] for target in ctx.attr.deps])]
-        runtime_deps = [java_common.merge([target[JavaInfo] for target in ctx.attr.runtime_deps])]
-        current_target_providers = exports + deps + runtime_deps
+        current_target_providers = [
```
ittaiz: extract again? not 100% sure. Just because of the repetition of details. wdyt?

scala/scala_import.bzl
```diff
@@ -25,24 +25,33 @@ def _scala_import_impl(ctx):
...
+            deps = [target[JavaInfo] for target in ctx.attr.deps],
+            runtime_deps = [target[JavaInfo] for target in ctx.attr.runtime_deps],
+            source_jar = ctx.file.srcjar,
+            neverlink = ctx.attr.neverlink,
+        )
+        for jar in current_target_compile_jars
+    ]
 
     # Handle the case with no jars.
     # TODO: Figure out if these providers should always be merged.
```
ittaiz: what should we do about this TODO?

ittaiz:
Sadly reproducibility test is flaky due to a non reproducible benchmarking utility rule (likely has no dependencies so reproducibility issues aren't dramatic)

ittaiz:
I'd really appreciate the new java_common API. Am willing to live with this hack for now.

ittaiz:
I tried running this on Wix's code and it fails miserably since we have some rules that depend on `create_java_provider` which is removed. I'll try to get them up to speed and report again.

ittaiz:
@iirina btw, if I don't pass anything to `compile_jar` will `JavaInfo` constructor take `compile_jar` as is? Docs are a bit unclear about it (they say ` If you do not wish to use either, you can simply pass output_jar.`)

iirina:
> Sadly reproducibility test is flaky due to a non reproducible benchmarking utility rule (likely has no dependencies so reproducibility issues aren't dramatic)

Thanks for retriggering the tests! It's nice to see everything green :smiley: 

> I'd really appreciate the new java_common API. Am willing to live with this hack for now.

Thank you! I'll file a new issue for it.

> I tried running this on Wix's code and it fails miserably since we have some rules that depend on create_java_provider which is removed. I'll try to get them up to speed and report again.

Ah, ok. I'm curious why you need to create this provider outside of the rules implementation.

> @iirina btw, if I don't pass anything to compile_jar will JavaInfo constructor take compile_jar as is? Docs are a bit unclear about it (they say If you do not wish to use either, you can simply pass output_jar.)

Sorry for the documentation. Both `output_jar` and `compile_jar` are mandatory, the build fails if either is unspecified.

iirina:
I opened https://github.com/bazelbuild/bazel/issues/8867.

iirina:
@johnynek Can you take another look and let me know what you think of this new approach? Thanks!

ittaiz:
> Ah, ok. I'm curious why you need to create this provider outside of the rules implementation.

One instance is that we needed to fork the scala proto implementation for a while because there were disagreements on priorities and we didn't want to block stripe. Essentially we could have just used `java_common.create_provider` ourselves but the code we copied used that method so we kept it. Another instance is for documentation generation.

> Sorry for the documentation. Both output_jar and compile_jar are mandatory, the build fails if either is unspecified.

That's even worse. The documentation says that `compile_jar` has a default value of `None`.

ittaiz:
> I opened bazelbuild/bazel#8867.

Thanks!

ittaiz:
Please don't forget though that I'm still trying to validate this internally so I'd appreciate you not merging just yet

johnynek:
@beala-stripe this is a significant change. We should test this internally like Wix is doing.

scala/Dummy.java
```diff
@@ -0,0 +1,5 @@
+public class Dummy {
```
johnynek: I'd like to put this into some namespace we control on the off chance it causes strange errors for users who make a Dummy class in an example.

scala/Dummy.java
```diff
@@ -0,0 +1,5 @@
...
+	public static void main(String[] args) {
```
johnynek: why add this? Just seems like more code in the class file?

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -251,16 +251,7 @@ def _compile_scala(
 
 def _empty_java_info(deps_java_info, implicit_deps):
```
johnynek: can we remove this entire method? it seems like it is just `java_common.merge(deps_java_info + implicit_deps)` now, and I'd rather that be inlined.

long-stripe:
For posterity: I tried https://github.com/bazelbuild/rules_scala/pull/781/commits/86a229efc766f75166d93f9fb70669a019b09806 on our largest internal repo at Stripe and with ~4 lines of changes (whatever changed tripped the unused deps checker, strange how it wasn't picking those up before) I was able to get a green build.

scala/Dummy.java
ittaiz: I also thought about this from a different perspective of WTF in 6 months. Maybe something like `PlaceHolderClassToCreateEmptyJarForScalaImport` with a comment above pointing to the bazel issue

ittaiz:
Wix is green as well

scala/Dummy.java
iirina: I renamed it to `PlaceHolderClassToCreateEmptyJarForScalaImport`, thanks for the suggestions!

scala/Dummy.java
iirina: You're right, I removed the method.

scala/scala_import.bzl
iirina: Obsolete now. I replaced it with a new TODO pointing at the new API issue.

scala/scala_import.bzl
iirina: Sounds good, done.

scala/scala_import.bzl
iirina: Done.

twitter_scrooge/twitter_scrooge.bzl
iirina: Sure, done.

scala/private/common.bzl
iirina: Sorry I missed this comment before. I replaced all `+` usages for `java_common.merge` with list extensions.

iirina:
Thank you everyone for your helpful contributions and for your understanding! I'm happy with the result thanks to the productive discussions!

I addressed the final comments and if everyone agrees this can be merged. You can try the final version on your codebase just to be extra safe.

ittaiz:
I’m good. Not concerned with the latest changes.
Thank you for picking up this ball! Not handling it myself, like we agreed in the past,  was definitely weighing on me for a few good months...

scala/private/rule_impls.bzl
```diff
@@ -498,8 +510,24 @@ def _compile_or_empty(
...
         )
 
+def _create_scala_compilation_provider(ctx, ijar, deps_providers):
+    exports = []
+    if hasattr(ctx.attr, "exports"):
+        exports = [dep[JavaInfo] for dep in ctx.attr.exports]
+    runtime_deps = []
+    if hasattr(ctx.attr, "runtime_deps"):
+        runtime_deps = [dep[JavaInfo] for dep in ctx.attr.runtime_deps]
+    return JavaInfo(
```
ittaiz: I just realized that this doesn’t pass the source_jar.
Any chance you can add that? If not I’ll add it in a separate PR

iirina: I'm not sure which is the source jar here. Is it `_pack_source_jars(ctx)`? Note that `JavaInfo()` takes only one source jar, so probably there needs to be some more packing before.

iirina:
@ittaiz Can you re-trigger the reproducibility test again? Thanks!

ittaiz:
AFK and Travis doesn’t let me retrigger from mobile :(
I’ll do it as soon as I get to my laptop

ittaiz:
@iirina re-triggered

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -379,7 +372,7 @@ def _scrooge_scala_library_impl(ctx):
     )
     if ctx.attr.exports:
         exports = [exp[JavaInfo] for exp in ctx.attr.exports]
-        all_java = java_common.merge(exports + [aspect_info.java_info])
+        all_java = java_common.merge(_concat_lists(exports, [aspect_info.java_info]))
```
johnynek: nit: why not do `exports.append(aspect_info.java_info)`? that is O(1) in python so I hope skylark, whereas this is O(N) full copy.

johnynek:
looks like we are passing except for #699 

@ittaiz are we good to merge? Seems like we both have green internal CIs on this change.

ittaiz:
We’re good.
There is a small gap of the source_jar but I think existing solution didn’t pass it as well so it’s not a regression but more of alignment and using the new API.
I’ll send a separate PR for it.

twitter_scrooge/twitter_scrooge.bzl
ittaiz: @iirina any thoughts? I’ll do a follow up OR and can fix this if you agree

PR #782, opened 2019-07-12
check e2e with bazel 0.28.0

ittaiz:


johnynek:
looks like some skylark is rejected in 0.28 that we need to fix:

```
File "/home/travis/.cache/bazel/_bazel_travis/7f7e07d167ded9af0372bc305ceff99e/external/io_bazel_rules_scala/scala/private/rule_impls.bzl", line 181, in compile_scala
		",".join([labels[j.path] for j in transit...])
sequence element must be a string (got 'Label'). See https://github.com/bazelbuild/bazel/issues/7802 for information about --incompatible_string_join_requires_strings.
```

ittaiz:
I’ll let Ian take the lead here

PR #783, opened 2019-07-13
Passing source-jar into JavaInfo

ittaiz:
also did a few nits from the review of #781

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -991,6 +991,7 @@ def _pack_source_jar(ctx):
 
 def _pack_source_jars(ctx):
     source_jar = _pack_source_jar(ctx)
+    #_pack_source_jar may return None if java_common.pack_sources returned None (and it can)
     return [source_jar] if source_jar else []
 
 def scala_binary_impl(ctx):
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -401,7 +401,7 @@ def _compile_or_empty(
     if len(ctx.files.srcs) + len(srcjars.to_list()) == 0:
         _build_nosrc_jar(ctx)
 
-        scala_compilation_provider = _create_scala_compilation_provider(ctx, ctx.outputs.jar, deps_providers)
+        scala_compilation_provider = _create_scala_compilation_provider(ctx, ctx.outputs.jar, None, deps_providers)
 
         #  no need to build ijar when empty
         return struct(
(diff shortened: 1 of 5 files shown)


scala/private/rule_impls.bzl
```diff
@@ -986,10 +986,12 @@ def _pack_source_jars(ctx):
...
     )
-    if scala_source_jar:
-        source_jars.append(scala_source_jar)
 
-    return source_jars
+    return scala_source_jar
+
+def _pack_source_jars(ctx):
+    source_jar = _pack_source_jar(ctx)
+    return [source_jar] if source_jar else []
```
johnynek: isn't `source_jar` always defined now? I'm confused how this can ever return `[]`. Can we add a comment if I am missing something?

johnynek: PS: if I am right, can we remove `_pack_source_jars` and and just use `_pack_source_jar`?

ittaiz: the `java_common.pack_sources` may return None. @iirina maybe you know in which cases?

ittaiz: @johnynek just add a comment saying that `java_common.pack_sources` may return `None`? Honestly I want to do some refactoring here as I'd like us to drop the `scala` attribute or at least significantly trim down what we're passing there to rely on JavaInfo.

johnynek: yeah, adding that comment would be fine to me, but I wonder if we really need two methods, maybe just put this logic at the end of the first call?

I would also love to remove the `scala` attribute and just use `JavaInfo`. If it doesn't break intellij, I think we should be able to do that. We should have an issue (maybe already do) to track that.

johnynek:
thanks for the cleanup!

johnynek:
(merge when green)

scala/private/rule_impls.bzl
ittaiz: Just now saw your comment about removing `_pack_source_jars` completely. I preferred to focus on this change since (like I wrote above) I want to consider greatly limiting the scala attribute and so probably also will be able to remove `_pack_source_jars`.

ittaiz:
Sure thing :) I hate leaving a mess and I also didn't want to bother Irina with all the small nits.
Btw,
Inside Wix we have a service that allows us to comment #automerge on PRs and on commits and it squash merges the PRs when they are green (notifies you via slack on success/failure). I wish we had that here...

scala/private/rule_impls.bzl
johnynek: no problem.

iirina: `java_common.pack_sources` returns `None` if there's nothing to pack, meaning if both `source_jars` and `sources` are empty.

iirina:
Thanks for taking care of this!

PR #784, opened 2019-07-13
remove usage of attr scala with JavaInfo outputs

ittaiz:
@iirina I used to think that because we use `java_common.create_provider` then our `JavaInfo` instances are faulty and return empty sequence for `archive[JavaInfo].runtime_output_jars`.
I've now tried to remove this hack thanks to your work but it still returns an empty sequence.
1. Are we constructing the JavaInfo incorrectly?
2. Is the below usage of `outputs.jars` correct?
3. When is `runtime_output_jars` non empty?

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -1236,13 +1236,8 @@ def _serialize_archives_short_path(archives):
 def _get_test_archive_jars(ctx, test_archives):
     flattened_list = []
     for archive in test_archives:
-        # because we (rules_scala) use the legacy JavaInfo (java_common.create_provider)
-        # runtime_output_jars contains more jars than needed
-        if hasattr(archive, "scala"):
-            jars = [jar.class_jar for jar in archive.scala.outputs.jars]
-        else:
-            jars = archive[JavaInfo].runtime_output_jars
-        flattened_list.extend(jars)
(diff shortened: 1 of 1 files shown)


ittaiz:
Thanks!
I didn’t remove the scala attr but removed one usage of it where we took
something from the custom attribute instead from JavaInfo due to a bug.
I want to remove internal usages first before touching the external side of
IntelliJ (which should just work with JavaInfo but we’ll need to manually
test it).
In any case I’d like Irina’s feedback on my change before continuing since
I don’t understand why my other solution didn’t work.

ittaiz:
@iirina
I’d still appreciate your inputs here as I’m not sure this is the *best*
solution and I don’t understand why the runtime_output_jars didn’t work

iirina:
@ittaiz It's probably because the direct runtime jars aren't merged in `java_common.merge`. I have a WIP PR https://github.com/bazelbuild/bazel/pull/8847 aiming to solve this. Do you know if there's a simple way to test that my PR solves the problem on your codebase?

ittaiz:
You can run the test_rules_scala.sh locally.
If you change the file to just use the outputs.
This would be exciting to have!

iirina:
> If you change the file to just use the outputs.

What do you mean? Should I run the tests with your PR as well?

ittaiz:
I’ll open a new failing PR and mention you there. Ok?

PR #790, opened 2019-07-19
Simplify _jacoco_offline_instrument.

benjaminp:
resolve_command shouldn't be required here.

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -1333,10 +1333,6 @@ def _jacoco_offline_instrument(ctx, input_jar):
     if not ctx.configuration.coverage_enabled or not hasattr(ctx.attr, "_code_coverage_instrumentation_worker"):
         return _empty_coverage_struct
 
-    worker_inputs, _, worker_input_manifests = ctx.resolve_command(
-        tools = [ctx.attr._code_coverage_instrumentation_worker],
-    )
-
     output_jar = ctx.actions.declare_file(
         "{}-offline.jar".format(input_jar.basename.split(".")[0]),
     )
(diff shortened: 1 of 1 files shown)


johnynek:
Thanks for the PR.

I have to admit, I don't understand when this should or should not be required. Can you add any links to documentation as to why you think it *shouldn't* be required?

Tests passing are a strong hint, but I'd like to find some documentation that could also guide us.

benjaminp:
I think `resolve_command` is only useful if you pass the first string argument. Then, it can parse out of the `$(location //some/tool)` directives for you and return the requisite tools. But if you already know the executable you want, `resolve_command` is killing a fly with a cannon.

scala/private/rule_impls.bzl
```diff
@@ -1351,10 +1347,9 @@ def _jacoco_offline_instrument(ctx, input_jar):
 
     ctx.actions.run(
         mnemonic = "JacocoInstrumenter",
-        inputs = [in_out_pair[0] for in_out_pair in in_out_pairs] + worker_inputs,
+        inputs = [in_out_pair[0] for in_out_pair in in_out_pairs],
         outputs = [in_out_pair[1] for in_out_pair in in_out_pairs],
-        executable = ctx.attr._code_coverage_instrumentation_worker.files_to_run.executable,
-        input_manifests = worker_input_manifests,
+        executable = ctx.attr._code_coverage_instrumentation_worker.files_to_run,
```
andyscott: This seems fine to me. Do you (@johnynek?) know if we've re-enabled the coverage tests in this repo? They were disabled, but I recall seeing a PR to fix/reenable them for current Bazel versions.

benjaminp:
Travis is printing:
```
running test test_coverage_on
Test "test_coverage_on" successful (20 sec) 
```

johnynek:
Thanks!

PR #791, opened 2019-07-19
Exclude jmh from reproducibility test, since the code generator is non-deterministic.

benjaminp:
https://github.com/bazelbuild/rules_scala/issues/699


ittaiz:
The Travis failure is about Windows rules_scala test failing because of resources (can’t start java).
Merging.

Issue #797, opened 2019-07-24
Dependencies for host are used for target when set in grpc_deps


ignasl:
I have defined and registered my own `scala_proto_toolchain`:
```python
scala_proto_toolchain(
    name = "my_shiny_toolchain",
    grpc_deps = [
        "//important/dependency"
    ],
    code_generator = "//important/generator:my_generator"
)
```
Naturally, I am declaring the rule for proto and another scala library that depends on it:
```python
scalapb_proto_library(
   name = "my_proto",
   deps = [...]
)

scala_library(
   name = "my_library",
   deps = [
      ":my_proto",
      "//important/dependency"
   ]
)
``` 
So dependencies are eventually set-up like this:
1. `my_proto` -> `//important/dependency`
2. `my_library` -> `//important/dependency`
3. `my_library` -> `my_proto`

Now, the problem arises that when `"my_library"` gets built, the classpath contains two copies of the `//important/dependency` jars:
1. `/bazel-out/host/bin/important/dependency/dependency.jar`
2. `/bazel-out/darwin-fastbuild/bin/important/dependency/dependency.jar`

This is a problem, because `jar`s meant for the host find their way into dependency list meant for the target. I cannot remove the `//important/dependency' from the `grpc_deps` list, because generated code requires it and this causes compile failures.

I believe (correct me if I am wrong) that bazel or rules_scala incorrectly resolves libraries' paths to `host`. With JVM languages compilers can happily accept them because they are essentially different. However, if this was, for example, C++ cross-platform build (host is x86 and target ARM) compiler would not allow it. Since bazel supports cross-compiling, the issue must be with scala_rules. Otherwise, please, let me know what am doing wrong.

ignasl:
@ianoc-stripe, could you provide some insights?

ianoc-stripe:
Hi @ignasl it would be great if you could add a test in a PR for this? it should be possible to write one that fails i believe? it would make it much easier to debug/fix.

ianoc-stripe:
@ignasl merged a changed yesterday i think should act as a workaround for you, unfortunately the api's changed slightly due to limitations in bazel in making this nicer at the moment. But it should stop the host deps coming for target

ignasl:
Thanks a lot for solving this one. API change is not a problem at all we are still in POC mode. I have already started experimenting.

Issue #749, opened 2019-05-13
windows CI is flaky


johnynek:
see: https://travis-ci.org/bazelbuild/rules_scala/jobs/531959514#L436

cc @majcherm-da 

this has shown up in: #748 and #745

johnynek:
I've seen this several times...

I wonder if it is flaky locally or if travis is flaky for windows.

ittaiz:
I've seen another different issue of windows flakiness attached here (I actually saw two different ones but don't have the log of the other)
[failing_windows_travis_log.txt](https://github.com/bazelbuild/rules_scala/files/3414497/failing_windows_travis_log.txt)
@majcherm-da do you think you'll have some capacity to take a look?

johnynek:
I think proto tests should just be disabled on windows until someone interested wants to invest the time to get them passing.

ittaiz:
+1.
We should probably decide that this is the policy for most tests that flake
out after some time.
The unification Andy sent will probably make this a bit more complicated
bash setup wise

andyscott:
I'd prefer that we wait-- I intend to split apart the test scripts so it's easy to tag which tests run for each configuration.

ittaiz:
👍🏽
One note though- long term vision IMHO should be to move the bash tests
inside Bazel using bazel-integration-testing library.
If you’re doing a small medium effort that’s ok but if you’re going to be
doing a large one then maybe we should reconsider investing in moving over
to it.

PR #801, opened 2019-07-29
Due to limitations in the toolchains implementation this is required …

ianoc-stripe:
…to stop these being host deps rather than target deps

First commit here should add a test, second one (once travis starts) will add a refactor to fix the test.

The bad part here is that this is due to https://groups.google.com/d/msg/bazel-discuss/vXrG6uFL5V8/RjK8_chaEgAJ

so a limitation in toolchains. The only real way i could see around this is to remove the deps portion from the toolchain and move them to be binds.

r? @johnynek 
cc @ignasl

relates to https://github.com/bazelbuild/rules_scala/issues/797

### DIFF
--- a/scala_proto/scala_proto.bzl
+++ b/scala_proto/scala_proto.bzl
@@ -15,7 +15,6 @@ load(
 )
 
 def register_default_proto_dependencies():
-    print("here")
     if native.existing_rule("io_bazel_rules_scala/dependency/proto/grpc_deps") == None:
         native.bind(
             name = "io_bazel_rules_scala/dependency/proto/grpc_deps",
--- a/scala_proto/scala_proto.bzl
+++ b/scala_proto/scala_proto.bzl
@@ -14,10 +14,30 @@ load(
     "scalapb_aspect",
 )
 
+def register_default_proto_dependencies():
+    print("here")
+    if native.existing_rule("io_bazel_rules_scala/dependency/proto/grpc_deps") == None:
+        native.bind(
+            name = "io_bazel_rules_scala/dependency/proto/grpc_deps",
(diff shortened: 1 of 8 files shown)


johnynek:
This seemed to fail on windows CI the same way this did: https://travis-ci.org/bazelbuild/rules_scala/builds/563804940?utm_source=github_status&utm_medium=notification I think there is still a lurking issue in proto somehow that seems much more likely to be hit on windows.

johnynek:
cc @katre

scala_proto/scala_proto.bzl
```diff
@@ -14,10 +14,30 @@ load(
     "scalapb_aspect",
 )
 
+def register_default_proto_dependencies():
+    print("here")
```
johnynek: can we remove this?

ianoc: oops, of course

johnynek:
Looks good. Thanks!

ianoc-stripe:
@johnynek i'm trying out 0.28.1 for travis just to try debug see if this windows thing is unstable or whats happened

johnynek:
looks like the windows CI is set up with chocolatey which only has 0.28.0: https://chocolatey.org/packages/bazel

ianoc-stripe:
Ahh good thanks

ianoc-stripe:
failure looks like some sort of windows travis issue this time
```
Error occurred during initialization of VM
Could not reserve enough sp
```

johnynek:
failure related to flakey windows CI #749

ittaiz:
Thank you so much for tackling this!
As you can understand this is part of our attempt to move to the aspect
based implementation

PR #803, opened 2019-08-02
Add scalac_jvm_flags option to toolchain

beala-stripe:
This is useful for globally setting the max heap usage of Scalac workers. At Stripe, we've seen Scalac workers get killed by the kernel OOM killer because the default JVM heap size is too large.

### DIFF
--- a/docs/scala_toolchain.md
+++ b/docs/scala_toolchain.md
@@ -63,6 +63,9 @@ scala_register_toolchains()
         <p>
           Extra compiler options for this binary to be passed to scalac. 
         </p>
+        <p>
+          This is overridden by the `scalac_jvm_flags` attribute on individual targets.
+        </p>
       </td>
     </tr>
     <tr>
--- a/docs/scala_toolchain.md
+++ b/docs/scala_toolchain.md
@@ -2,8 +2,6 @@
 
 `scala_toolchain` allows you to define global configuration to all Scala targets.
 
-Currently, the only option that can be set is `scalacopts` but the plan is to expand it to other options as well.
-
 **Some scala_toolchain must be registered!**
 
 ### Several options to configure `scala_toolchain`:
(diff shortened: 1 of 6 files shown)


docs/scala_toolchain.md
```diff
@@ -46,3 +44,54 @@ scala_register_toolchains()
     # WORKSPACE
     register_toolchains("//toolchains:my_scala_toolchain")
     ```
+
+<table class="table table-condensed table-bordered table-params">
```
ptarjan: I think you can do a table in github flavored markdown https://help.github.com/en/articles/organizing-information-with-tables

test_expect_failure/scalac_jvm_opts/BUILD
```diff
@@ -0,0 +1,35 @@
...
+    name = "failing_toolchain_impl",
+    # This will fail because 1M isn't enough
+    scalac_jvm_flags = ["-Xmx1M"],
+    visibility = ["//visibility:public"],
+)
+
+scala_toolchain(
+    name = "passing_toolchain_impl",
+    # This will pass because 5G is enough
+    scalac_jvm_flags = ["-Xmx1G"],
```
ptarjan: `s/1/5` ?

beala-stripe: Good catch.

docs/scala_toolchain.md
beala-stripe: I tried converting this to MD, but it's not clear I got any readability wins. In order to insert newlines into cells, I needed to mix MD and HTML. I also cannot indent the HTML properly because a newline starts a new cell.

johnynek:
happy to merge when you fix the comment pt found.

beala-stripe:
Thanks all! Comment fixed.

I also made a slight behavior change. It occurred to me that flags on the target should override flags passed on the toolchain (rather than combining them). WDYT?

johnynek:
@beala-stripe +1 to making the target override. The toolchain should be the default, but if you write it on a target, we get exactly what is there.

That's a good approach.

PR #804, opened 2019-08-05
Add `scala_test_jvm_flags` to toolchain

beala-stripe:
This gives `jvm_flags` for `scala_test` the same treatment as in https://github.com/bazelbuild/rules_scala/pull/803. This allows a global config for the maximum amount of heap a `scala_test` may use. We've seen scala_test processes getting OOM-killed in CI. 

Setting an `Xmx` on test runs makes memory-related test failures more deterministic. If a test fails due to excessive memory usage, it should be more deterministic than the whims of the kernel OOM-killer.

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -304,7 +304,7 @@ StatsfileOutput: {statsfile_output}
     # scalac_jvm_flags passed in on the target override scalac_jvm_flags passed in on the
     # toolchain
     final_scalac_jvm_flags = _first_non_empty(
-        _expand_location(ctx, scalac_jvm_flags),
+        scalac_jvm_flags,
         ctx.toolchains["@io_bazel_rules_scala//scala:toolchain_type"].scalac_jvm_flags
     )
 
@@ -326,7 +326,7 @@ StatsfileOutput: {statsfile_output}
         # consume the flags on startup.
         arguments = [
             "--jvm_flag=%s" % f
-            for f in final_scalac_jvm_flags
+            for f in _expand_location(ctx, final_scalac_jvm_flags)
         ] + ["@" + argfile.path],
     )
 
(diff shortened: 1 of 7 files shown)


test_expect_failure/scala_test_jvm_flags/Empty.scala
```diff
@@ -0,0 +1,3 @@
+package test_expect_failure.scala_test_jvm_flags
+
+class Empty
```
johnynek: shouldn't this be some kind of no-op test or something?

beala-stripe: 👍 I'll expand this out into a trivial test.

beala-stripe: Fixed.

scala/private/rule_impls.bzl
```diff
@@ -1202,13 +1209,20 @@ def scala_test_impl(ctx):
         ])
         coverage_runfiles = ctx.files._jacocorunner + ctx.files._lcov_merger + coverage_replacements.values()
 
+    # jvm_flags passed in on the target override scala_test_jvm_flags passed in on the
+    # toolchain
+    final_jvm_flags = _first_non_empty(
+        ctx.attr.jvm_flags,
```
johnynek: why don't we need `_expand_location` here?

beala-stripe: It wasn't wrapped before, so I left it unwrapped, but thinking about this more, we probably want `final_jvm_flags` itself wrapped.

johnynek:
looks like then windows build is broken in a new way.

beala-stripe:
Thanks for the quick turnaround!

johnynek:
@beala-stripe thank you!

johnynek:
PS: I added this issue #805 about the windows CI

PR #806, opened 2019-08-07
Fix a regression that breaks scalapb when jvm_flags are set on the toolchain

beala-stripe:
It looks like my change in https://github.com/bazelbuild/rules_scala/pull/804 caused a regression that breaks `scalapb_proto_library` targets when the `scalac_jvm_flags` attribute is set on the toolchain. Here's the error failing in our CI:

```
 ERROR: /src/tests/workspaces/scala_with_proto_and_byo_3rdparty/src/BUILD:3:1: in @io_bazel_rules_scala//scala_proto/private:scalapb_aspect.bzl%scalapb_aspect aspect on proto_library rule //src:test_proto: 
  	File "/src/tests/workspaces/scala_with_proto_and_byo_3rdparty/src/BUILD", line 3
  		@io_bazel_rules_scala//scala_proto/private:scalapb_aspect.bzl%scalapb_aspect(...)
  	File "/mnt/tmpfs/bazel/bazel/external/io_bazel_rules_scala/scala_proto/private/scalapb_aspect.bzl", line 188, in _scalapb_aspect_impl
  		_compile_scala(ctx, toolchain.scalac, target.label, o..., <3 more arguments>)
  	File "/mnt/tmpfs/bazel/bazel/external/io_bazel_rules_scala/scala_proto/private/scalapb_aspect.bzl", line 70, in _compile_scala
  		compile_scala(ctx, Label(("%s-fast" % label)), out..., <16 more arguments>)
  	File "/mnt/tmpfs/bazel/bazel/external/io_bazel_rules_scala/scala/private/rule_impls.bzl", line 311, in compile_scala
  		ctx.actions.run(inputs = ins, outputs = outs, execut..., <5 more arguments>)
  	File "/mnt/tmpfs/bazel/bazel/external/io_bazel_rules_scala/scala/private/rule_impls.bzl", line 329, in ctx.actions.run
  		_expand_location(ctx, final_scalac_jvm_flags)
  	File "/mnt/tmpfs/bazel/bazel/external/io_bazel_rules_scala/scala/private/rule_impls.bzl", line 131, in _expand_location
  		ctx.expand_location(f, ctx.attr.data)
  	File "/mnt/tmpfs/bazel/bazel/external/io_bazel_rules_scala/scala/private/rule_impls.bzl", line 131, in ctx.expand_location
  		ctx.attr.data
  No attribute 'data' in attr. Make sure you declared a rule attribute with this name.
```

The issue is that the `_expand_location` function expects the rule to have a `data` attribute, but this isn't true inside the scalapb aspect.

I'm not sure what the best practice is in dyn-langs when I don't know what shape my data is in. For now, I've added a `hasattr` guard.

I've also added a regression test that shows this fixes the issue.


beala-stripe:
Actually, before merging, let me test this in Zoolander and make sure if fixes everything.

test_expect_failure/scalac_jvm_opts/BUILD
```diff
@@ -41,3 +45,17 @@ scala_library(
     # the `failing_scala_toolchain` is used.
     scalac_jvm_flags = ["-Xmx1G"],
 )
+
+proto_library(
```
johnynek: why is this in `expect_failure` We expect this to work don't we?

We should only put tests in here that won't build or won't pass with `bazel test //...`

This seems like it is a positive, not a negative test. It should go in `test/...`

beala-stripe: I can move it there. I noticed some tests that require extra arguments (eg, `bazel build --extra_toolchains ...`) were being put in `expect_failure` because targets in `test/` are all invoked with `bazel build //test/...` and there's no way to specify extra arguments.

Maybe we need another test directory called `test_manual` or something for this use case?

johnynek: yeah, I'd rather have a separate directory. This should be for tests or builds that we are expecting to fail. If we've accidentally allowed others to get in here, we should address that (but not necessarily in this PR).

beala-stripe: Sounds good. I've moved the passing manual tests to a `manual_test` directory.

johnynek:
Thanks!

PR #807, opened 2019-08-09
Migrate from old-style .java provider to JavaInfo.

iirina:
Migrating in preparation of flipping the flag [--incompatible_disallow_legacy_java_provider](https://github.com/bazelbuild/bazel/issues/7598) in bazel.

### DIFF
--- a/jmh/jmh.bzl
+++ b/jmh/jmh.bzl
@@ -77,7 +77,7 @@ def jmh_repositories(maven_servers = ["http://central.maven.org/maven2"]):
 
 def _scala_construct_runtime_classpath(deps):
     scala_targets = [d.scala for d in deps if hasattr(d, "scala")]
-    java_targets = [d.java for d in deps if hasattr(d, "java")]
+    java_targets = [d[JavaInfo] for d in deps if hasattr(d, "JavaInfo")]
     files = []
     for scala in scala_targets:
         files.append(scala.transitive_runtime_jars)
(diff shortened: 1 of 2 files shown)


jmh/jmh.bzl
```diff
@@ -77,7 +77,7 @@ def jmh_repositories(maven_servers = ["http://central.maven.org/maven2"]):
 
 def _scala_construct_runtime_classpath(deps):
     scala_targets = [d.scala for d in deps if hasattr(d, "scala")]
-    java_targets = [d.java for d in deps if hasattr(d, "java")]
+    java_targets = [d[JavaInfo] for d in deps if hasattr(d, "JavaInfo")]
```
johnynek: This was written before scala returned a java provider. I think we can remove the scala targets entirely and just use `targets` in fact we can probably combine this with the loop below that constructs the depset.

scala/private/common.bzl
```diff
@@ -44,8 +44,8 @@ def collect_plugin_paths(plugins):
             paths.append(p)
         elif hasattr(p, "scala"):
             paths.extend([j.class_jar for j in p.scala.outputs.jars])
-        elif hasattr(p, "java"):
-            paths.extend([j.class_jar for j in p.java.outputs.jars])
+        elif hasattr(p, "JavaInfo"):
```
johnynek: Can we change this to `elif JavaInfo in d`? Similar comment about removing the scala branch.

johnynek:
Thanks for the PR!

jmh/jmh.bzl
iirina: Great, done!

scala/private/common.bzl
iirina: Done.

johnynek:
Great! Thank you for this improvement!

iirina:
Thanks for the quick code reviews! \o/

PR #808, opened 2019-08-10
Move scala_repositories macro out to its own file + move scala_doc.bzl

long-stripe:
## Summary

Moves the `scala_repositories` macro out to its own file in and publicly expose the same macro in `scala.bzl` for compatibility and as the intended public interface:

```python
load("@io_bazel_rules_scala//scala:scala.bzl", "scala_repositories")
```

Also moves `scala_doc.bzl` to private as a precursor to some other re-organization I'd like to do, described below.

## Motivation

Loosely inspired by the organization of [bazelbuild/rules_go](https://github.com/bazelbuild/rules_go/tree/9b614efc502964e60de3f8e5d0f0eb740a2a6930/go/private/rules) and [tweag/rules_haskell](https://github.com/tweag/rules_haskell/tree/54eb35afe0645c82c627af6e5ca8378ada70b3ae/haskell), my intention is to eventually break up the very large [scala.bzl](https://github.com/bazelbuild/rules_scala/blob/96176ae9e0be8582918a5212a7d4b44d885db8f6/scala/scala.bzl) and [rules_impl.bzl](https://github.com/bazelbuild/rules_scala/blob/96176ae9e0be8582918a5212a7d4b44d885db8f6/scala/private/rule_impls.bzl) files into per-rule files, with common code factored out elsewhere. Ultimately, a user should still really only care about loading rules from `scala.bzl` and their build definitions should continue to work as-is if they have been loading from `scala.bzl`.

My _real_ personal motivation here is that I would like to contribute more materially, but looking at giant files sometimes makes my eyeballs glaze over 😁. Plus, I like to organize stuff.

## Future work

If this seems alright and gets merged, I propose an codebase organization ending up roughly like (bunch of stuff elided):

```
scala/
├── private
│   ├── common.bzl
│   ├── ...
│   ├── macros # per-macro files
│   │   ├── scala_repositories.bzl
│   │   └── ...
│   └── rules # per-rule files
│       ├── scala_doc.bzl
│       └── ...
├── scala.bzl # re-export private rules/macros, public interface
└── ...

```

Amenable to suggestions for the above!


long-stripe:
\o/ thanks! I will follow up in the near future with some more PRs then.

johnynek:
Thank you! will be nice to clean things up. Also, having a small .md doc describing the structure might help people making prs.

long-stripe:
Good point, I'll add a guideline in `CONTRIBUTING.md` or `README.md` (lmk which one you prefer but the former sounds most appropriate) in my next PR.

PR #812, opened 2019-08-14
Pin Bazel versions

andyscott:
**Description:**

Ports over bazel pinning functionality from https://github.com/higherkindness/rules_scala.

With this, rules_scala developers on Linux/OSX will have Bazel pinned to the version rules_scala expects, automatically, regardless of the global bazel install version. Windows developers will still need to match global bazel with the rules_scala bazel version.

The `tools/bazel` script has been tested extensively (via other projects) for OSX/Linux developers.

**Motivation:**

One hurdle for development on bazelbuild/rules_scala is a lack of Bazel version pinning that's consistent for local development and CI.

### DIFF
--- a/tools/bazel
+++ b/tools/bazel
@@ -1,8 +1,6 @@
 #!/usr/bin/env bash
 set -e
 
-workspace="$(cd "$(dirname "$0")"/..; pwd)"
-
 default_bazel_version='0.28.1'
 
 if [ "$BUILDKITE" = true ]; then
--- a/tools/bazel
+++ b/tools/bazel
@@ -5,7 +5,10 @@ workspace="$(cd "$(dirname "$0")"/..; pwd)"
 
 default_bazel_version='0.28.1'
 
-case $(uname -s) in
+if [ "$BUILDKITE" = true ]; then
+    bazel_version='host'
+else
+    case $(uname -s) in
     Darwin|Linux)
 	# shellcheck disable=SC2153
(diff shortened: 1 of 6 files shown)


tools/bazel
```diff
@@ -0,0 +1,85 @@
...
+	;;
+    esac
+fi
+
+case "$bazel_version" in
+    'host')
+        bazel_version=$("$BAZEL_REAL" version | awk '/Build label/ {print $3}' | cut -d '-' -f 1)
+        bazel="$BAZEL_REAL"
+        ;;
+    '0.28.1')
```
johnynek: We used to be able to support several versions of bazel with the matrix. How can we do this here? We like to test on the oldest we can support and the newest (although sometimes that is a single version).

andyscott: I'll add some notes. The TLDR is that a build matrix across bazel versions is very well supported with this script.

andyscott:
Updated with notes covering how to test with multiple bazel versions.

andyscott:
@ittaiz How do you feel about these changes?

johnynek:
Thanks for the clean up! I'll merge now, if @ittaiz has any concerns we can deal with them.

ittaiz:
1. Thanks @andyscott  for asking and thanks @johnynek  even more so for not letting red tape take us down.
2. @andyscott I didn't deep dive into the script code but it sounds very similar to https://github.com/bazelbuild/bazelisk. Shouldn't we use that? This is also what Wix is moving to internally.

PR #813, opened 2019-08-15
Move scala_binary to its own file

long-stripe:
This PR moves `scala_binary` to its own file, per convention established in https://github.com/bazelbuild/rules_scala/pull/808.

In the process, I made quite a few formerly private functions in `rule_impls.bzl` public in order for it to be imported by `scala_binary.bzl`. These functions will eventually be moved somewhere else but I'm leaving them in place until rules are split out to avoid even larger PRs. Even with the newly exposed public functions, I figure that by convention a user isn't supposed to be directly loading something from a `private` package anyhow, so I think we can abdicate responsibility and give no guarantees of compatibility if a user is doing that.

Too bad Starlark doesn't have better access modifiers than a leading underscore, but I think if I ever want to untangle the big files then something has to give here.

Also:

* Moves any common attributes that `scala_binary` depends on to `common_attributes.bzl`
* Moves common outputs that `scala_binary` depends on to `common_outputs.bzl`
* Added a short note about code organization in CONTRIBUTING.md, feel free to add suggestions for wording or more detail.

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -341,7 +341,7 @@ def _interim_java_provider_for_java_compilation(scala_output):
         neverlink = True
     )
 
-def _scalac_provider(ctx):
+def scalac_provider(ctx):
     return ctx.toolchains["@io_bazel_rules_scala//scala:toolchain_type"].scalac_provider_attr[_ScalacProvider]
 
 def try_to_compile_java_jar(
@@ -358,7 +358,7 @@ def try_to_compile_java_jar(
         implicit_junit_deps_needed_for_java_compilation,
     )
     providers_of_dependencies += collect_java_providers_of(
-        _scalac_provider(ctx).default_classpath,
+        scalac_provider(ctx).default_classpath,
     )
     scala_sources_java_provider = _interim_java_provider_for_java_compilation(
         scala_output,
@@ -592,7 +592,7 @@ def _java_bin(ctx):
         javabin = "%s/%s" % (runfiles_root, java_path)
     return javabin
 
-def _write_java_wrapper(ctx, args = "", wrapper_preamble = ""):
+def write_java_wrapper(ctx, args = "", wrapper_preamble = ""):
     """This creates a wrapper that sets up the correct path
          to stand in for the java command."""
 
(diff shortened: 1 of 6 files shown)


long-stripe:
Looks like some more Windows CI build sadness unrelated to my changeset.

If I have spare cycles or as part of the great reunification plan, Someone™️ should take a look at the Windows build. Perhaps even get off of TravisCI to something else one day too, but I don't want to bikeshed too much in my comment :).

scala/private/rule_impls.bzl
```diff
@@ -1014,68 +1014,25 @@ def _pack_source_jars(ctx):
...
-        main_class = ctx.attr.main_class,
-        rjars = out.transitive_rjars,
-        use_jacoco = False,
-        wrapper = wrapper,
-    )
-    return out
-
 def scala_repl_impl(ctx):
-    scalac_provider = _scalac_provider(ctx)
+    _scalac_provider = scalac_provider(ctx)
```
johnynek: why did you add the `_` on the local var? you removed it in other cases? can we just use a different name?

maybe the function should be `get_scalac_provider(ctx)` if it is going to be public?

scala/private/rule_impls.bzl
```diff
@@ -1137,18 +1094,18 @@ def scala_test_impl(ctx):
     if len(ctx.attr.suites) != 0:
         print("suites attribute is deprecated. All scalatest test suites are run")
 
-    scalac_provider = _scalac_provider(ctx)
+    _scalac_provider = scalac_provider(ctx)
```
johnynek: same comment as above.

scala/private/rule_impls.bzl
```diff
@@ -1275,12 +1232,12 @@ def scala_junit_test_impl(ctx):
         fail(
             "Setting at least one of the attributes ('prefixes','suffixes') is required",
         )
-    scalac_provider = _scalac_provider(ctx)
+    _scalac_provider = scalac_provider(ctx)
```
johnynek: same comment as above

scala/private/rules/scala_binary.bzl
```diff
@@ -0,0 +1,85 @@
...
+    "declare_executable",
+    "get_unused_dependency_checker_mode",
+    "scala_binary_common",
+    "scalac_provider",
+    "write_executable",
+    "write_java_wrapper",
+)
+
+def _scala_binary_impl(ctx):
+    _scalac_provider = scalac_provider(ctx)
```
johnynek: same comment.

johnynek:
I do worry a bit that we are foregoing the real access control mechanism we have to split files. I wonder if that trade is a great trade.

scala/private/rule_impls.bzl
andyscott: It looks like he stuck with file reorganization while keeping any function/variable renames to a minimum. A lot of these now public methods should be refactored further-- tackling the name right now won't matter too much.

long-stripe: Agree with you @johnynek - I think `get_scalac_provider` is probably the right move here, I just changed things to get it to build properly. I'll follow up with that instead.

long-stripe:
@johnynek yeah, I noted that in my PR summary. It's rather unfortunate that the only mechanism of access control in Starlark is a leading `_` for per-file private access, but it seems acceptable to put things behind a `private` package that purely by convention shouldn't be touched by users, like how [rules_go](https://github.com/bazelbuild/rules_go/tree/master/go/private) and [rules_haskell](https://github.com/tweag/rules_haskell/tree/master/haskell/private) currently does. It seems to work for them so I'm inclined to follow their example.

johnynek:
Thanks for the cleanups, @long-stripe !

PR #814, opened 2019-08-16
Delurk Powershell for a more unified test setup

andyscott:
This delurks the separate windows powershell scripts and uses the same test scripts regardless of OS.

The goal is to begin simplifying and reworking the test setup so that folks can iterate more quickly on feature development.


andyscott:
This seems to work in the sense that it's consistent with previous known failures:
- the main windows tests fail with protobuf issues at a certain point
- the reproducibility windows tests were previously disabled to make the job green

After this I'll begin work to refactor tests to accommodate a build matrix of OSs and (later Scala versions).

andyscott:
@long-stripe WDYT?

johnynek:
Thanks Andy!

PR #815, opened 2019-08-17
Move scala_test/scala_test_suite to their own file

long-stripe:
This PR moves `scala_test`/`scala_test_suite` to its own file, per #808 

I relocated `sanitize_string_for_usage` from `scala/scala.bzl` to `scala/private/common.bzl` for now and loaded it in `scala/scala.bzl`, but as things move into `private`, these functions will only be used in `private`.

As I get more familiar with the little private library of functions rules_scala has, perhaps organizing them a bit better like how [rules_haskell](https://github.com/tweag/rules_haskell/blob/b3bea97d7cf133bc5896dfa9d02f0258547b43b8/haskell/private/set.bzl) does for its own private lib would make sense, dunno yet.

### DIFF
--- a/scala/private/common_attributes.bzl
+++ b/scala/private/common_attributes.bzl
@@ -135,17 +135,3 @@ resolve_deps = {
         allow_files = False,
     ),
 }
-
-test_resolve_deps = {
-    "_scala_toolchain": attr.label_list(
-        default = [
-            Label(
-                "//external:io_bazel_rules_scala/dependency/scala/scala_library",
-            ),
-            Label(
-                "//external:io_bazel_rules_scala/dependency/scalatest/scalatest",
-            ),
-        ],
-        allow_files = False,
-    ),
-}
--- a/scala/private/common_attributes.bzl
+++ b/scala/private/common_attributes.bzl
@@ -135,3 +135,17 @@ resolve_deps = {
         allow_files = False,
     ),
 }
+
+test_resolve_deps = {
+    "_scala_toolchain": attr.label_list(
+        default = [
+            Label(
+                "//external:io_bazel_rules_scala/dependency/scala/scala_library",
+            ),
+            Label(
+                "//external:io_bazel_rules_scala/dependency/scalatest/scalatest",
+            ),
+        ],
+        allow_files = False,
+    ),
+}
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -127,7 +127,7 @@ touch {statsfile}
         arguments = [],
     )
 
-def _expand_location(ctx, flags):
+def expand_location(ctx, flags):
     if hasattr(ctx.attr, "data"):
         data = ctx.attr.data
     else:
@@ -138,7 +138,7 @@ def _join_path(args, sep = ","):
     return sep.join([f.path for f in args])
 
 # Return the first non-empty arg. If all are empty, return the last.
-def _first_non_empty(*args):
+def first_non_empty(*args):
     for arg in args:
         if arg:
             return arg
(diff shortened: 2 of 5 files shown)


scala/private/rule_impls.bzl
```diff
@@ -127,7 +127,7 @@ touch {statsfile}
         arguments = [],
     )
 
-def _expand_location(ctx, flags):
+def expand_location(ctx, flags):
```
ittaiz: Should we move every public method to common.bzl or maybe interim_common.bzl? 
It's somewhat surprising to find a public method in private/rule_impls.bzl

scala/private/common_attributes.bzl
```diff
@@ -135,3 +135,17 @@ resolve_deps = {
         allow_files = False,
     ),
 }
+
+test_resolve_deps = {
```
ittaiz: Can't/Shouldn't this be confined to scala_test.bzl?

long-stripe: Good point - I moved it blindly but you are right that it's only used by `scala_test`

scala/private/rule_impls.bzl
long-stripe: I plan to eventually, basically the common functions in rule_impls.bzl will get moved to common.bzl or some other sort of private lib organization. I didn't in these PRs to reduce the size of the incidental diffs during the refactoring process.

long-stripe:
@ittaiz Nope, I haven't touched any implementation thus far, only moves, and one of my motivations for doing such refactoring is to make it less intimidating to look at large monolithic files while trying to make real meaty improvements!

Checking the test suites, the results are the same as before and ✅.

ittaiz:
Will merge on green

scala/private/rules/scala_test.bzl
```diff
@@ -0,0 +1,228 @@
...
+    if ctx.attr.full_stacktraces:
+        flags += "F"
+    else:
+        flags += "S"
+    if not ctx.attr.colors:
+        flags += "W"
+    return flags
+
+def _scala_test_impl(ctx):
+    if len(ctx.attr.suites) != 0:
```
johnynek: we can probably remove this... this has been here for a very long time, and the migration is pretty simple, just remove `ctx.attr.suites`.

Maybe do that in a follow up.

long-stripe:
Don't think the build will be green in this case since the Windows build has some unrelated sadness it seems

scala/private/rules/scala_test.bzl
long-stripe: I went ahead and did this in https://github.com/bazelbuild/rules_scala/pull/815/commits/9a69e3823376f74a01c52b2542c65476b79f3071 to also try a CI rebuild.

long-stripe:
I took a 10 minute glance at the Windows build and couldn't figure out what might be wrong since I don't know enough there. Windows support is unfortunately harder without a local env too.

johnynek:
I think we need to disable windows CI unfortunately. Between the beta status of travis CI on windows, and less support from bazel, it is making our CI *very* noisy.

long-stripe:
@johnynek I stopped short of suggesting so but I think that's probably the right move for now, the resources for supporting Windows are scarce.

scala/private/rules/scala_test.bzl
```diff
@@ -0,0 +1,225 @@
...
+        providers = out.providers,
+        runfiles = ctx.runfiles(coverage_runfiles, transitive_files = out.runfiles.files),
+        scala = out.scala,
+    )
+
+_scala_test_attrs = {
+    "main_class": attr.string(
+        default = "io.bazel.rulesscala.scala_test.Runner",
+    ),
+    "suites": attr.string_list(),
```
johnynek: can we remove this also? We had it deprecated for years now with a print nag warning. I think it should be safe to remove.

johnynek:
we can remove `suites` attribute on a follow up.

PR #816, opened 2019-08-17
Remove all trailing newlines

andyscott:
**Description**
Removes trailing newlines by running all files through a noop sed replacement.

**Motivation**
As part of cleanup, I need the ability to rewrite files using sed without creating an excessive amount of diffs. This adjusts current files so that those diffs won't be present in future work.


johnynek:
seems to fail buildkite. I didn't investigate.

andyscott:
It's probably for the docs, if they're checked in CI. I'll look later.

johnynek:
I'm confused, are we adding a removing trailing newlines? Some files, e.g.:  https://github.com/bazelbuild/rules_scala/pull/816/files#diff-4832b5628d76dd1afbd2bc4f53292883L9 are reporting they went from no trailing new line to a single trailing newline.

Are you making sure there is exactly 1 trailing new line?

Can you share the sed command you ran in the comments/description?

PR #820, opened 2019-08-18
Move scala_repl rule to its own file

long-stripe:
### Description
Moves `scala_repl` to its own file.

### Motivation
Another in the series of splitting out rules to their own files as started in #808.


johnynek:
Thanks!

scala/private/rules/scala_repl.bzl
```diff
@@ -0,0 +1,104 @@
...
+_scala_repl_attrs.update(implicit_deps)
+
+_scala_repl_attrs.update(common_attrs)
+
+_scala_repl_attrs.update(resolve_deps)
+
+scala_repl = rule(
+    attrs = _scala_repl_attrs,
+    executable = True,
+    fragments = ["java"],
```
johnynek: We should double check this. I think this may be a deprecated or even removed Java sandwich API. Cc @iirina

iirina: Do you mean requesting the java fragment? It's required when using something from the java configuration, like java_toolchain.

PR #826, opened 2019-08-20
Buildifier as the only lint

andyscott:
### Description
Redoes our lint scripts so that we're running a simplified buildifier setup (and checking it in CI).

### Motivation
The previous lints were disabled. Additionally, buildifier seems to support the full breadth of starlark syntax now.


.travis.yml
```diff
@@ -43,7 +43,7 @@ env:
   # Linting is broken. Disable until fixed.
   # See https://github.com/bazelbuild/rules_scala/pull/622
   # we want to test the last release
-  #- TEST_SCRIPT=test_lint.sh
+  - TEST_SCRIPT=test_lint.sh
```
johnynek: can we avoid running the lint on all platforms? Is that easy?

andyscott: Yep! I'm going to rework travis so we use the more configurable job pipelines instead of the strict square matrix.

andyscott:
This is green now, sans Windows. I'll take care of windows in another PR.

johnynek:
@ittaiz merge if you think this is okay. I’m very pleased.

ittaiz:
I’ll take a look tonight or tomorrow morning at the latest.
Does this add lint running on CI? Are we using the bazel CI built in
mechanism for this (I’ve asked them to support it a while back and they
did, thanks @philwo!)?
If you need I can find the relevant links

andyscott:
This removes the most all of the disabled lint scripts in favor of just running buildifier in CI.

.travis.yml
ittaiz: I think we should test linting via buildkite. See this readme for more [info] (https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md#running-buildifier-on-ci)

ittaiz:
Actually, one thing I'm a bit concerned about is the "noise" building buildifier from source adds.
It's not a blocker to this change but if we get buildifier checks free from buildkite then maybe to run buildifier people should just get it via `go get`? https://github.com/bazelbuild/buildtools/tree/master/buildifier#setup
I'm ok with merging as is though I think less code is better here

ittaiz:
Any,
Wdyt about moving to buildkite buildifier?

PR #827, opened 2019-08-20
Move library rules

long-stripe:
### Description
Moves `scala_macro_library` to its own file, and `library_attrs` to `common_attributes.bzl` for now.

### Motivation
More moves per direction laid out in #808.


ittaiz:
I have to say I’m not sure I like this PR. I feel scala_macro_library is very closely related to scala_library and this change also exposes “lib” which is a very private part. This hints to me we shouldn’t separate the two.

long-stripe:
@ittaiz Fair enough, I was also considering moving all the lib rules to one place and keep `_lib` private. This confirms it, I'll follow up with your vision in a bit.

scala/private/rule_impls.bzl
```diff
@@ -548,7 +548,7 @@ def _create_scala_compilation_provider(ctx, ijar, source_jar, deps_providers):
         runtime_deps = runtime_deps,
     )
 
-def _build_deployable(ctx, jars_list):
+def build_deployable(ctx, jars_list):
```
johnynek: if we are going to make this public, we could actually comment it and improve the arguments. One api could be:
```
def merge_jars(single_jar, actions, deploy_jar, main_class, jars):
```
Without taking the full `ctx` we prevent limit what the function can do and make it clearer what the dependencies are.

Then call it with:
```
merge_jars(ctx.executable._singlejar, ctx.actions, ctx.outputs.deploy_jar, getattr(ctx.attr, "main_class", None), jars_list)
```

scala/private/rule_impls.bzl
```diff
@@ -949,7 +832,7 @@ def scala_binary_common(
     )  # no need to build an ijar for an executable
     rjars = depset(outputs.full_jars, transitive = [rjars])
 
-    _build_deployable(ctx, rjars.to_list())
+    build_deployable(ctx, rjars.to_list())
```
johnynek: we should probably look at all the linearization of depsets we do in these rules and see if it is avoidable at all. It is a shame you can't lazily linearize a depset into args so it can happen at build time not analysis time... maybe you can.

ittaiz: totally agree. I have a feeling our analysis perf can be greatly improved but I haven't gotten around to profiling it

scala/private/rule_impls.bzl
long-stripe: I'll see if I can follow up with a PR for this next.

PR #831, opened 2019-08-22
Default to usage of scala_proto_library instead of scalapb_proto_libr…

andyscott:
# Description
Updates usages of the protobuf rules to use the direct method `scala_proto_library` instead of the alias `scalapb_proto_library`.

The alias is preserved, so folks aren't forced into a migration.

### Motivation
In https://github.com/bazelbuild/rules_scala/pull/821, we renamed `scalapb_proto_library` to `scala_proto_library` (with an alias, for migration). This is follow up.


ittaiz:
LGTM

PR #832, opened 2019-08-22
Refactor build_deployable

long-stripe:
### Description
Refactors `build_deployable` and documents it a bit per suggestion in https://github.com/bazelbuild/rules_scala/pull/827#discussion_r315972566.

### Motivation
Restrict the API of this function rather than taking a `ctx` object wholesale.

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -548,7 +548,7 @@ def _create_scala_compilation_provider(ctx, ijar, source_jar, deps_providers):
         runtime_deps = runtime_deps,
     )
 
-def merge_jars(actions, deploy_jar, singlejar_executable, label, jars_list, main_class = ""):
+def merge_jars(actions, deploy_jar, singlejar_executable, jars_list, main_class = "", progress_message = ""):
     """Calls Bazel's singlejar utility.
 
     For a full list of available command line options see:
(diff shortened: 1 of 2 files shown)


scala/private/rule_impls.bzl
```diff
@@ -548,22 +548,32 @@ def _create_scala_compilation_provider(ctx, ijar, source_jar, deps_providers):
...
+    args.extend(["--output", deploy_jar.path])
+    actions.run(
         inputs = jars_list,
-        outputs = [ctx.outputs.deploy_jar],
-        executable = ctx.executable._singlejar,
+        outputs = [deploy_jar],
+        executable = singlejar_executable,
         mnemonic = "ScalaDeployJar",
-        progress_message = "scala deployable %s" % ctx.label,
+        progress_message = "scala deployable %s" % label,
```
johnynek: maybe rather than taking the label, take the full message. So we could do `"scala_binary jar: %s" % ctx.label` or `"scala_library jar: " + ctx.label` or something.

Because, really this function has nothing to do with scala anymore. It is really just a jar merging rule.

long-stripe:
I'm not sure I see the motivation behind making it a proper public rule: what end user would want to use it instead of `scala_binary` or `scala_library` with multiple deps?

johnynek:
You might want to use this without scala at all. You might want to use it to merge some deploy jars together. You might want to use it with the jar_jar rule I wrote.

It is public, so no need to make a change for that. I was just making the comment that the rule is useful on its own.

long-stripe:
@johnynek Ah yeah, makes sense. It's a pretty simple wrapper over an executable that does the heavy lifting anyhow. If someone expresses interest it may be useful to elevate it as a rule in some other package of rules_scala (`util`? Half joking).

johnynek:
It might make sense as a PR to rules_pkg:

https://github.com/bazelbuild/rules_pkg

It is packaging related.

long-stripe:
Personally, I don't know if either of those projects seem like places where a `merge_jars` rule would be useful. `skylib` is full of stdlib-like, language agnostic things like Starlark hash sets, and `rules_pkg` is for package (e.g. deb) distribution?

I'll keep it in mind to find somewhere where this thing could live, but could that imply a dependency on another ruleset if it's upstreamed elsewhere?

ittaiz:
It could but that depends on the maturity of that ruleset. In any case I
think it’s not pressing.
Maybe an issue to upstream bazel to say “hey this is the bike shedding we
needed to do to use singlejar, maybe it can be useful for all things jvm.
Where do you think it should go?”
Maybe they’ll take it in core-bazel for the time being.

PR #833, opened 2019-08-23
rename scrooge rules to scala_thrift prefix

andyscott:
### Description

Renames the thrift rules to follow the naming conventions used across the Bazel community for schema generation functionality. Aliases for the original names are retained to ensure that there are no hard breakages during migrations.

With this change, the rules start to look like this:
- `scala_library`
- `scala_repositories`
- `scala_proto_library`
- `scala_proto_repositories`
- `scala_thrift_library`
- `scala_thrift_repositories`
- `proto_library` (broader bazel API)
- `thrift_library`

There's still a bit of clean up needed for some of the other rules on the fringe (junit/specs2/scalatest/tut). I'll chip away at that when I get the chance.

### Motivation

The thrift rule names are inconsistent with the proto rule names, which I find somewhat confusing. Additionally, the naming is inconsistent with the rest of the Bazel community.


twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -1,437 +1,15 @@
...
-    "src_jars",
-    "output_files",
-    "java_info",
-])
-
-ScroogeInfo = provider(fields = [
-    "aspect_info",
-])
-
-ScroogeImport = provider(fields = [
```
johnynek: this don't exist now right? Can we import `ScroogeImport = ScalaThriftImport` or something so we don't break the provider API? Similarly for the others.

ianoc:
i'm kinda -1 on this, limited upside. lots more code/backwards compat stuff. Do many people other than stripe actually use this?

andyscott:
>thrift is apache, and this is not the apache thrift generator for scala (which could exist, but doesn't AFAIK).
>there are two scala thrift implementations: scrooge and foursquare's spindle: https://github.com/foursquare/spindle I think scrooge is more popular, but I don't know for sure.

Ideally `thrift_scala_library` is the public API and implementation agnostic. The backend can be configured by a toolchain or a provider (via an attribute). I'll admit, this might not be possible if the two different libraries are dramatically different-- I haven't looked at spindle.

twitter_scrooge/twitter_scrooge.bzl
andyscott: 👍can do

jjudd:
It'd be nice to have a consistent naming paradigm throughout the rules, which this moves towards. If there is a way to ease the transition for current consumers, including Stripe, then that seems like a good idea.

I like Andy's idea of having `thrift_scala_library` be the api the the implementation be configurable. Unless, of course, the implementations are so wildly different that this is not possible.

johnynek:
The problem is, you can't really change the backend and get compatible source code out the other end. So, I feel that kind of plugability is bad. For instance, if I depend on repo A, which uses scala_foo_library with toolchain google-foo but in my repo I use twitter-foo which generates different code, things won't build.

I don't really see it as workable to use the same name for a rule that generates incompatible outputs.

So, a spindle rule will need a different name for that rule: scala_spindle_thrift_library or something like that.

Like I said, I don't really mind scrooge being the preferred implementation, but I want us to be clear: this isn't about configuring different backends, it is about privileging our chosen implementation.

johnynek:
@jjudd do you all use scrooge?

I'd like to have an idea of the stakeholders. Like, I *think* etsy does, but I'm not sure.

This seems to be a bit contentious compared to other things. Maybe our energy is better spent on other issues.

long-stripe:
@johnynek Ok, I think I can buy that. For illustrative purposes and to present a simple example of what I think you're talking about for my own understanding:

Let's say toolchain `google-foo` takes `bar.thrift` and produces:

```scala
case class Bar(
  theFirstField: String,
  theSecondField: Option[Int],
  theThirdField: Long
)
```

Toolchain `twitter-foo` takes `bar.thrift`, has different opinions, and produces something that looks similar but isn't exactly interchangeable, as consumers wouldn't compile against this because field names are actually different:

```scala
case class Bar(
  the_first_field: String,
  the_second_field: Option[Int],
  this_third_field: Long
)
```

Is that about right?

ianoc:
@long depending on support for options the classes could have different names or even be in different package spaces! :) scrooge has some non-thrift-standard package naming support iirc to support scala and java being around together

johnynek:
It's much more complex than that @long-stripe. These code generators generated subtypes of fixed classes. Users often have reflective code that expects that. Scrooge, for reasons, generates traits and immutable case class implementations behind those traits. Scalapb, for instance, does not do that, and generates case classes, but they extend certain types.

I would turn it around another way are there any examples of generators that emit fully source compatible outputs? I don't know of any. Maybe different versions of a given library.

long-stripe:
fwiw I think if this retains backward compat I _think_ it's ok, taking a stance on scrooge sounds fine to me. My read is this PR isn't even important, but it seems nice to make names look consistent. Obviously I have my own opinion about what looks nice. :)

@ianoc @johnynek Yup that makes sense. Thanks for the insight!

tdyas-stripe:
Pants might offer some prior art to consider. Pants seems to make the choice of "compiler" for its `java_thrift_library` target an attribute on the target. See, for example, https://github.com/pantsbuild/pants/blob/a68153447adf427cda4e3f2a09a3bf0c8bc78a7a/contrib/scrooge/examples/src/thrift/org/pantsbuild/contrib/scrooge/android_generator/BUILD#L3. Then again, Foursquare's custom Pants plugin for Spindle has a `spindle_thrift_library` target. (https://github.com/foursquare/fsqio/blob/master/src/python/fsqio/pants/spindle/targets/spindle_thrift_library.py). So maybe no clear answer but their choice seems to make it an attribute of the target.

johnynek:
Yes, but scrooge’s Java code it emits is designed to be Apache compatible IIRC. So there it makes sense because the output should match.

ittaiz:
Not a Scrooge user here so I’ll try to limit my response to the meta part
of this PR:
I think consistent naming has value but I feel the value here is limited
since this area isn’t widely used.
I feel we have very big fish to fry with respect to technical debt in these
rules and I’d be very appreciative if we could work on those first.
I understand the need to do some janitorial/non-semantic tasks first to get
the lay of the land but I’m not sure changing APIs is the best way to start.
Would it help if Oscar and me go over the (many) open issues and tag them
as relevant for onboarding?
I’m loving the motivation both you and Long are coming with and I’d love
for it to go towards the warts that constrain development here (I think the
naming here is more minor to that effect).

P.S. depreciations aren’t a magic cure, they cost in maintenance

ittaiz:
To summarize (since I’m not sure it’s clear after all):
1. Not against this PR.
2. Super appreciate the motivation and capacity to improve.
3. Not sure this is where we should put energy right now
4. Would gladly help with ideas on where we need help the most

long-stripe:
@ittaiz Thanks Ittai, I really appreciate it!

Absolutely, since you have the most context, I think a brief sweep of the issues and tagging them as "good first issue" or "low hanging fruit" would be great for aspiring contributors. [typelevel/cats](https://github.com/typelevel/cats/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) does something like this and it seems nice. I know for myself that I want to contribute more substantial stuff in the future but can't always tell if something will be a quick fix, a meaty issue, or a multiday slog. Closing obviously irrelevant issues too would be nice, but that usually involves a little more investigation than you may have time for (perhaps also worthy of an "investigation needed" label?)

As for myself, sending these little PRs was a way to build momentum contributing _something_ while I get more context on the lay of the land.

jjudd:
Was out climbing this weekend and am back at a computer now.

@johnynek Lucid does not use Thrift. I'm just a fan of the consistent names.

I also did not realize that different thrift implementations for Scala produced such different outputs. In that case, I suppose the options are either be opinionated on a default with the current naming or having a more specific name like `scala_thirft_scrooge_library`? 

Again, I have no stake in this other than the naming convention.

andyscott:
The feedback/notes here seems very valid. In particular, scrooge is just one thrift implementation, so it does make sense to have the the implementation specific naming for the rules.

With that in mind, I'm generally indifferent to this change. I'm going to close the PR, and we can reopen if someone would like to argue for merging this.

PR #838, opened 2019-09-01
Use java_common.merge instead of manual _collect util functions in scala_import

long-stripe:
### Description

Reworks some code in `scala_import` to use `java_common.merge` instead of manually creating depsets in utils.

Tests should still pass.

### Motivation

I had remembered seeing this `TODO` before and it seems easy enough to fix. Deleting unneeded code!

7a8

### DIFF
--- a/scala/scala_import.bzl
+++ b/scala/scala_import.bzl
@@ -13,9 +13,10 @@ def _scala_import_impl(ctx):
         intellij_metadata,
     ) = (target_data.code_jars, target_data.intellij_metadata)
     current_jars = depset(current_target_compile_jars)
-    exports = _collect(ctx.attr.exports)
-    transitive_runtime_jars = _collect_runtime(ctx.attr.runtime_deps)
-    jars = _collect(ctx.attr.deps)
+    exports = java_common.merge([export[JavaInfo] for export in ctx.attr.exports])
+    transitive_runtime_jars = \
+        java_common.merge([dep[JavaInfo] for dep in ctx.attr.runtime_deps]) \
+            .transitive_runtime_jars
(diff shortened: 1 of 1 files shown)


scala/scala_import.bzl
```diff
@@ -13,9 +13,10 @@ def _scala_import_impl(ctx):
         intellij_metadata,
     ) = (target_data.code_jars, target_data.intellij_metadata)
     current_jars = depset(current_target_compile_jars)
-    exports = _collect(ctx.attr.exports)
-    transitive_runtime_jars = _collect_runtime(ctx.attr.runtime_deps)
-    jars = _collect(ctx.attr.deps)
```
long-stripe: I wanted to be sure to call this line out: as it stands in master, this `jars` is unused and is deleted here.

johnynek:
This looks great to me! Thanks for the simplification.

@iirina there were bugs around merge in the past (something with empty lists, or not merging everything about the provider... are those fixed? Do you recall what I am referencing?)

iirina:
This is such a nice change! :) Most bugs in `java_common.merge` are fixed. The only remaining issue is that the _direct_ runtime jars are not merged, but all the transitive closure is. If you have provider A with direct_jars=A.jar and B with direct_jars=B.jar, `java_common.merge(A,B).runtime_output_jars` is an empty list (probably the culprit of https://github.com/bazelbuild/bazel/issues/8916). This PR is only dealing with transitive closure, so all should work as expected.

johnynek:
@iirina thank you for looking!

johnynek:
@ittaiz if you have time will you double check and merge?

Issue #669, opened 2019-01-19
scala_library shouldn't have a jvm_flags attribute


ittaiz:
AFAICT scala_library doesn't use `jvm_flags` attribute anymore since we have `scalac_jvm_flags` and `javac_jvm_flags`.
It's only used by binary/repl/test/junit_test which are binary rules.
I'd like to remove it.
On the one hand it's a breaking change OTOH the current attribute is deprecated for many months now.
@johnynek any objections?
cc @aherrmann since I think he was interested in this attribute from #621

johnynek:
I guess it is fine to remove it. It should be easy to migrate since it is a no-op. If people follow the pattern of making their own macro wrapper for rules so they can control default options it would be even easier to migrate: their wrapper can accept the option but just discard it.

ittaiz:
I think bazel people would call this abuse of macros but OTOH I'm not sure
they thoroughly considered the use-case of OSS eco system.
I'll send a PR for this

PR #840, opened 2019-09-03
Remove jvm_flags debug print for scala_library

long-stripe:
### Description
Closes https://github.com/bazelbuild/rules_scala/issues/669.

Removes the warning for `jvm_flags` usage in `scala_library` and instances where they were passed in tests.

### Motivation

Follow up from https://github.com/bazelbuild/rules_scala/pull/795 which advocates for removing the debug completely after some period of time (~a month). Feel free to table this if it's too soon and I'll close for now.

### DIFF
--- a/scala/private/rules/scala_library.bzl
+++ b/scala/private/rules/scala_library.bzl
@@ -135,6 +135,11 @@ def _lib(
 ##
 
 def _scala_library_impl(ctx):
+    if hasattr(ctx.attr, "jvm_flags"):
+        fail(
+            msg = "'jvm_flags' for scala_library is deprecated. It does nothing today and will be removed from scala_library to avoid confusion.",
+            attr = "jvm_flags",
+        )
     scalac_provider = get_scalac_provider(ctx)
     unused_dependency_checker_mode = get_unused_dependency_checker_mode(ctx)
     return _lib(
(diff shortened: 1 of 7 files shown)


scala/private/rules/scala_library.bzl
```diff
@@ -135,8 +135,6 @@ def _lib(
 ##
 
 def _scala_library_impl(ctx):
-    if ctx.attr.jvm_flags:
```
johnynek: Do we not need to change the rule to make using them an error?

long-stripe:
Cool - made using `jvm_flags` in `scala_library` a hard fail, and re-added `jvm_flags` to the executable-type rules.

scala/private/rules/scala_library.bzl
```diff
@@ -135,8 +135,11 @@ def _lib(
...
-        print("'jvm_flags' for scala_library is deprecated. It does nothing today and will be removed from scala_library to avoid confusion.")
+    if hasattr(ctx.attr, "jvm_flags"):
```
johnynek: do we need this? Won't bazel fail if users pass attributes that the rule doesn't support?

long-stripe: Ah, right you are! Didn't think about that.

PR #841, opened 2019-09-03
Fix to error message

mantasak:
Was not properly adding space, as such:

```
java.lang.IllegalStateException: Was not able to discover any classes for archives=testsuite/test/example/example.jarprefixes=Set(), suffixes=Set(Test)
```

Fixing for better readability of errors.


ittaiz:
Thanks!

PR #842, opened 2019-09-05
Improve classpath manifest file construction

anchlovi:
### Description
Improve classpath manifest file construction

### Motivation
See: https://github.com/bazelbuild/bazel/pull/9314


johnynek:
Great work! Thanks for upstreaming this also!

Issue #845, opened 2019-09-16
Scrooge Error when defining Thrift exception


KaoruDev:
I noticed when a Thrift file defines an exception, scrooge throws an exception:

For example adding:

```
exception FooBar {
  1: string code
}
```

to `test/src/main/scala/scalarules/test/twitter_scrooge/thrift/thrift2/thrift3/Thrift3.thrift` results in this:

```
ERROR: /absolute/path/to/test/src/main/scala/scalarules/test/twitter_scrooge/thrift/thrift2/thrift3/BUILD:4:1: Couldn't build file test/src/main/scala/scalarules/test/twitter_scrooge/thrift/thrift2/thrift3/thrift3_scrooge.jar: scala //test/src/main/scala/scalarules/test/twitter_scrooge/thrift/thrift2/thrift3:thrift3 failed (Exit 1)
bazel-out/darwin-fastbuild/bin/test/src/main/scala/scalarules/test/twitter_scrooge/thrift/thrift2/thrift3/tmp4446905357582031766/scalarules/test/twitter_scrooge/thrift/thrift2/thrift3/FooBar.scala:171: error: object finagle is not a member of package com.twitter
  extends ThriftException with com.twitter.finagle.SourcedException with ThriftStruct
                                           ^
one error found
one error found
java.lang.RuntimeException: Build failed
	at io.bazel.rulesscala.scalac.ScalacProcessor.compileScalaSources(ScalacProcessor.java:244)
	at io.bazel.rulesscala.scalac.ScalacProcessor.processRequest(ScalacProcessor.java:69)
	at io.bazel.rulesscala.worker.GenericWorker.runPersistentWorker(GenericWorker.java:45)
	at io.bazel.rulesscala.worker.GenericWorker.run(GenericWorker.java:111)
	at io.bazel.rulesscala.scalac.ScalaCInvoker.main(ScalaCInvoker.java:41)
```

johnynek:
so, not everyone uses finagle (e.g. data uses of thrift which aren't about services). I'm reluctant to change the classpaths of everyone using the rules.

The alternate way is to configure which jars come in by changing the bind here:
https://github.com/bazelbuild/rules_scala/blob/master/twitter_scrooge/twitter_scrooge.bzl#L69

in your own repo, e.g. make that bind point to `scala_library` that exports both scrooge_core and finagle.

KaoruDev:
[...]

PR #846, opened 2019-09-16
add finagle-core to scrooge compile time to build thrift exceptions

KaoruDev:
### Description
Fixes https://github.com/bazelbuild/rules_scala/issues/845 (Fixes Scrooge compiling exception)

### Motivation
Adding `finagle-core` to the Scrooge compile time allows `Scrooge` to properly translate a Thrift exception definition to a POJO.

### DIFF
--- a/src/scala/io/bazel/rules_scala/scrooge_support/BUILD
+++ b/src/scala/io/bazel/rules_scala/scrooge_support/BUILD
@@ -12,6 +12,7 @@ scala_library(
         "//external:io_bazel_rules_scala/dependency/thrift/scrooge_generator",
         "//external:io_bazel_rules_scala/dependency/thrift/util_core",
         "//external:io_bazel_rules_scala/dependency/thrift/util_logging",
+        "//external:io_bazel_rules_scala/dependency/thrift/finagle_core",
     ],
 )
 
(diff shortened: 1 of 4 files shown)


johnynek:
as I wrote in #845 I'm nervous about adding the finagle dependency for everyone. See here the list of all finagle's dependencies:

https://search.maven.org/artifact/com.twitter/finagle-core_2.12/19.9.0/jar

it's a considerable list. I think the right way is to document how to configure your scrooge builds so you can depend on finagle in your personal build. Ideally, we would set up a scrooge_toolchain that has the jars to put on the classpath when compiling the code. An alternate idea, would be to put finagle on the classpath when compiling, but not add it to the transitive runtime deps, requiring downstream users to add finagle themselves when they need it, but not polluting the classpath with the full transitive dependencies of finagle for everyone.

ittaiz:
+1 for toolchain. More expensive contribution but that should be our
direction.

KaoruDev:
🆗 let me try your suggestions @johnynek, sorry I'm relatively new to the world of Bazel and just trying to get a POC out for my team.

ittaiz:
Thanks Kaoru and good luck!

Issue #843, opened 2019-09-11
Repository gets polluted with _virtual_imports


ignasl:
I have encountered an interesting issue that reproduces in Bazel 0.27.0 and later versions when dependency tree is a bit more complex.

Consider such set-up:
* There are two repositories: [`repo_one`](https://github.com/ignasl/bazel-proto-scala-test-one) and [`repo_two`](https://github.com/ignasl/bazel-proto-scala-test-two) -- links point to actual repositories where this issue reproduces.
* `two:proto` in `repo_two` depends on `one:proto` in `repo_one`
* `one:proto` in `repo_one` depends on `three:proto` in `repo_two`
* Apparently, it is important that `one:proto` must be at certain level within `repo_two`. In this test it is under `module/submodule/one/src/main/proto/`

In `repo_two` when running `bazel build --sandbox_debug //...` two problems occur:
1. `//three/_virtual_imports/proto/three.proto` gets copied (all Bazel versions after and including 0.27.0) -- repository gets polluted.
2. Build fails because several copies of `three.proto` gets passed to the `protoc` at the same time. Reproduced mostly on 0.28.1, but it seems to be not 100% consistent. Last arguments for `protoc` look like this:
```
bazel-out/darwin-fastbuild/bin/three/_virtual_imports/proto/_virtual_imports/proto/_virtual_imports/proto/three.proto
bazel-out/darwin-fastbuild/bin/three/_virtual_imports/proto/_virtual_imports/proto/three.proto
bazel-out/darwin-fastbuild/bin/three/_virtual_imports/proto/three.proto
```

I suspect that [`ScalaPBGenerator.setupIncludedProto`](https://github.com/bazelbuild/rules_scala/blob/f4a24fe2f76e1b1f2b24757f7740a2699692951f/src/scala/scripts/ScalaPBGenerator.scala#L37) is not completely correct way to go. Why is it necessary to copy `proto` files someplace?.. It seems like the working directory has changed in 0.27.0, which caused the copies to go into the working copy of the repository.

ignasl:
@lberki, @ittaiz suggested you might have some insights.
@ianoc-stripe, any insights on the copying business?
Issue https://github.com/bazelbuild/rules_scala/issues/779 seems closely related.

ianoc-stripe:
Imports have been a bit of a pain in bazel. I believe 
https://github.com/bazelbuild/bazel/commit/f65cc726655eb81c6e0aeb04d05b42831640dce9
[...]

PR #848, opened 2019-09-18
Copy the protos we need into a temp folder

ianoc-stripe:
### Description
We have errors around colliding files. The copying is putting things into trees that aren't part of the sandbox per build.

Eventually i believe the virtual_imports work will be a nicer solution to this, but from what i can tell part of it is still behind a flag. I wasn't quite sure exactly how to implement it all using that today. So for now instead of copying some stuff around and maybe colliding. Just copy everything into a temp location.

Also we weren't getting any error messages from protoc when run for scalapb so i made some changes here to redirect the output to a file and reply it back so we now get the error messages

### Motivation

Errors/problems in the copying approach

### DIFF
--- a/src/scala/scripts/PBGenerateRequest.scala
+++ b/src/scala/scripts/PBGenerateRequest.scala
@@ -6,17 +6,17 @@ case class PBGenerateRequest(jarOutput: String, scalaPBOutput: Path, scalaPBArgs
 
 object PBGenerateRequest {
 
-  def from(args: java.util.List[String]): PBGenerateRequest = {
+  def from(tmpDir: Path)(args: java.util.List[String]): PBGenerateRequest = {
     val jarOutput = args.get(0)
     val protoFiles = args.get(4).split(':')
+    val protoFilesToBuild = protoFiles.map { e => s"${tmpDir.toFile.toString}/$e"}
+
     val includedProto = args.get(1).drop(1).split(':').distinct.map { e =>
       val p = e.split(',')
       // If its an empty string then it means we are local to the current repo for the key, no op
-      (Some(p(0)).filter(_.nonEmpty), p(1))
-    }.collect {
-      // if the to compile files contains this absolute path then we are compiling it and shoudln't try move it around(duplicate files.)
-      case (Some(k), v) if !protoFiles.contains(v) => (Paths.get(k), Paths.get(v))
-    }.toList
+      val absolutePath = Some(p(0)).filter(_.nonEmpty).getOrElse(p(1))
+      (Paths.get(absolutePath), Paths.get(p(1)))
+    }.toList ++ protoFiles.map { p => (Paths.get(p), Paths.get(p))}
 
     val flagOpt = args.get(2) match {
       case "-" => None
@@ -27,7 +27,7 @@ object PBGenerateRequest {
       case "-" => Nil
       case s if s.charAt(0) == '-' => s.tail.split(':').toList //drop padding character
       case other => sys.error(s"expected a padding character of - (dash), but found: $other")
-    }) ++ List(".")
+    }) ++ List("")
 
     val tmp = Paths.get(Option(System.getProperty("java.io.tmpdir")).getOrElse("/tmp"))
     val scalaPBOutput = Files.createTempDirectory(tmp, "bazelscalapb")
@@ -43,7 +43,8 @@ object PBGenerateRequest {
     }.toList
 
 
-    val scalaPBArgs = outputSettings ::: (padWithProtoPathPrefix(transitiveProtoPaths) ++ protoFiles)
+    val scalaPBArgs = outputSettings ::: (padWithProtoPathPrefix(tmpDir)(transitiveProtoPaths) ++ protoFilesToBuild)
+
     val protoc = Paths.get(args.get(5))
 
     val extraJars = args.get(7).drop(1).split(':').filter(_.nonEmpty).distinct.map {e => Paths.get(e)}.toList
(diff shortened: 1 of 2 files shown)


ianoc-stripe:
fixes https://github.com/bazelbuild/rules_scala/issues/843

johnynek:
Thanks for tackling!

looks like legit test failures.

src/scala/scripts/ScalaPBGenerator.scala
```diff
@@ -89,10 +92,21 @@ class ScalaPBGenerator extends Processor {
...
   }
 
-  protected def exec(protoc: Path): Seq[String] => Int = (args: Seq[String]) =>
-    new ProcessBuilder(protoc.toString +: args: _*).inheritIO().start().waitFor()
+  protected def exec(protoc: Path): Seq[String] => Int = (args: Seq[String]) => {
+    val tmpFile = Files.createTempFile("stdout", "log")
+    try {
+      val ret = new ProcessBuilder(protoc.toString +: args: _*).redirectErrorStream(true).redirectOutput(ProcessBuilder.Redirect.to(tmpFile.toFile)).start().waitFor()
+      Try {
+        scala.io.Source.fromFile(tmpFile.toFile).getLines.foreach{System.out.println}
```
johnynek: this looks like debug. Can we remove?

ianoc-stripe: This is how we get the stdout/stderr from protoc into the bazel output stream. So can't remove. 

(prior to this PR any failure in protoc would result in just seeing a message that it exited with code 1 and no further info)

src/scala/scripts/PBGenerateRequest.scala
```diff
@@ -6,16 +6,44 @@ case class PBGenerateRequest(jarOutput: String, scalaPBOutput: Path, scalaPBArgs
...
     val jarOutput = args.get(0)
     val protoFiles = args.get(4).split(':')
-    val includedProto = args.get(1).drop(1).split(':').distinct.map { e =>
-      val p = e.split(',')
-      // If its an empty string then it means we are local to the current repo for the key, no op
-      (Some(p(0)).filter(_.nonEmpty), p(1))
-    }.collect {
-      // if the to compile files contains this absolute path then we are compiling it and shoudln't try move it around(duplicate files.)
-      case (Some(k), v) if !protoFiles.contains(v) => (Paths.get(k), Paths.get(v))
+    val protoFilesToBuild = protoFiles.map { e => s"${tmpDir.toFile.toString}/${stripToVirtual(e)}"}
```
johnynek: nit, you don't need `.toString` inside the interpolation. Also, I think just `$tmpDir/${stripToVirtual(e)}` should work. Why do we need to convert to File?

src/scala/scripts/PBGenerateRequest.scala
```diff
@@ -6,16 +6,44 @@ case class PBGenerateRequest(jarOutput: String, scalaPBOutput: Path, scalaPBArgs
...
+
+    val includedProtoSplit: List[(String, String)] = args.get(1).drop(1).split(':').map { e =>
+      val arr = e.split(',')
```
johnynek: want to add a `require(arr.length == 2, s"expected exactly two comma separated items in ${arr.toList}")` just to guard against anything funny changing later?

src/scala/scripts/PBGenerateRequest.scala
```diff
@@ -6,16 +6,44 @@ case class PBGenerateRequest(jarOutput: String, scalaPBOutput: Path, scalaPBArgs
...
+      (arr(0), arr(1))
+    }.toList
+    val includedProto: List[(Path, Path)] = (includedProtoSplit ++ protoFiles.toList.map { e => ("", e)}).distinct.map { case (repoPath, protoPath) =>
```
johnynek: nit: can we make this line not quite so long? Maybe newline after the `)` concatenating the two lists?

src/scala/scripts/PBGenerateRequest.scala
```diff
@@ -6,16 +6,44 @@ case class PBGenerateRequest(jarOutput: String, scalaPBOutput: Path, scalaPBArgs
...
+      } else if (repoPath.nonEmpty) {
+        // We have a repo specified
+        // if the repo path and the file path are the same, no-op
+        // otherwise get a relative path.
+        val relativePath: Path = if(repoPath == protoPath) Paths.get(protoPath) else Paths.get(repoPath).relativize(Paths.get(protoPath))
+        (Paths.get(protoPath), relativePath)
+      } else {
+        (Paths.get(protoPath), Paths.get(protoPath))
+      }
     }.toList
```
johnynek: isn't this already a list?

src/scala/scripts/PBGenerateRequest.scala
```diff
@@ -27,7 +55,7 @@ object PBGenerateRequest {
       case "-" => Nil
       case s if s.charAt(0) == '-' => s.tail.split(':').toList //drop padding character
       case other => sys.error(s"expected a padding character of - (dash), but found: $other")
-    }) ++ List(".")
+    }) ++ List("")
```
johnynek: `::: List("")` I *think* is a bit more efficient on lists than `++` but maybe not...

src/scala/scripts/ScalaPBGenerator.scala
johnynek: oh yeah... I overlooked that. nit: can we do `foreach(System.out.println)` (use parens rather than `{}` when passing a method as a function argument)?

Yannic:
I agree that writing these rules isn't as easy as it could be. We have a [design doc](https://docs.google.com/document/d/1u95vlQ1lWeQNR4bUw5T4cMeHTGJla2_e1dHHx7v4Dvg/edit) in review that proposes some helper functions to improve this situation (cmp. p 3-5), however they assume that the generator is a protoc plugin. From what I've seen, `rules_scala` uses a persistent worker for this (I assume because of legacy reasons and/or startup time?), so there's probably some work required to make it work but I'm confident you'll be able to reuse at least some of the functionality.

ignasl:
Specifically for the issue #843 and our own build problems I might have a smaller fix that actually avoids copying altogether. I will send a PR for consideration.

ignasl:
See: https://github.com/bazelbuild/rules_scala/pull/850

ianoc:
Closing in favor of https://github.com/bazelbuild/rules_scala/pull/850

PR #850, opened 2019-09-23
Fix paths in --proto_path and avoid copying proto files

ignasl:
### Description
Through numerous trial-and-errors it was found that copying `proto` files is only necessary when they are withing well-known `com_google_protobuf` package. For unknown reasons (could be related to the fact that `protoc` also comes from `com_google_protobuf`) Bazel passes incorrect path that is used in `--proto_path`. This proposal shows a way to fix the `transitiveProtoPaths` by analyzing the `includedProtos` list of `proto` files.
 
1. We have a list of all required `.proto` files in `includedProtos` -- this is a tuple list (root -> full path), for example:
```
bazel-out/k8-fastbuild/bin -> bazel-out/k8-fastbuild/bin/external/com_google_protobuf/google/protobuf/source_context.proto
```
2. Convert the full path to relative from the root (as it was done in `ScalaPBGenerator.setupIncludedProto`):
```
bazel-out/k8-fastbuild/bin -> external/com_google_protobuf/google/protobuf/source_context.proto
```
3. In all the `includedProtos` we find the first one that is located within the `transitiveProtoPaths` we are processing -- relative path in `includedProtos` starts with the path we are processing
4. If found -- the include folder is our "orphan" and is not anchored in either host or target. To fix we prepend root of found proto. If not found, return original. This works as long as `external/com_google_protobuf` is available in target root:
```
bazel-out/k8-fastbuild/bin/external/com_google_protobuf
```

Normally, external `proto` paths are passed with the root (target or host):
```
bazel-out/k8-fastbuild/bin/external/com_my_shiny_dependency
```
However, some (namely `com_google_protobuf`) are passed without root:
```
external/com_google_protobuf
```
This PR "anchors" those "orphan" include paths and makes them available.

### Motivation
The `protoc` plugin fails to find some `proto` files because they end-up missing from the paths within given `transitiveProtoPaths`. However, the `includedProtos` list contains all required `proto` files and they are all available to plugin. As a solution plugin implementation copies all of the required files to the relative folder. This is not sustainable as relative folder depends on working folder thus changes in working folder potentially breaks the behavior, as happened with Bazel 0.27. After working directory was changed from `sanbox` to `execroot`, plugin started copying proto files to the repo it is building. Also, this caused issues when working with `proto` files in remote directories -- due to multiple copies same `proto` files a being build over and over again.

This PR attempts to avoid coping.

This fixes https://github.com/bazelbuild/rules_scala/issues/843
Quite possibly, the `fixTransitiveProtoPath` (and copying) function would become obsolete when https://github.com/bazelbuild/bazel/issues/7964 gets fixed in Bazel.
Alternative to https://github.com/bazelbuild/rules_scala/pull/848

### DIFF
--- a/src/scala/scripts/PBGenerateRequest.scala
+++ b/src/scala/scripts/PBGenerateRequest.scala
@@ -19,10 +19,15 @@ object PBGenerateRequest {
   //  (4) If found -- the include folder is "orphan" and is not anchored in either host or target. To fix we prepend
   //      root. If not found, return original. This works as long as "external/com_google_protobuf" is available in
   //      target root.
-  def fixTransitiveProtoPath(includedRelProto: List[(String, String)], orig: String): String =
-    includedRelProto.find { case (_, rel) => rel.startsWith(orig) } match {
-      case Some((root, _)) => s"$root/$orig"
-      case None => orig
+  def fixTransitiveProtoPath(includedProto: List[(Path, Path)]): String => String = {
+    val includedRelProto = includedProto.map { case (root, full) => (root.toString, root.relativize(full).toString) }
+
+    { orig =>
+      includedRelProto.find { case (_, rel) => rel.startsWith(orig) } match {
+        case Some((root, _)) => s"$root/$orig"
+        case None => orig
+      }
+    }
   }
 
   def from(args: java.util.List[String]): PBGenerateRequest = {
@@ -43,13 +48,11 @@ object PBGenerateRequest {
       case other => sys.error(s"expected a padding character of - (dash), but found: $other")
     }
 
-    val includedRelProto = includedProto.map { case (root, full) => (root.toString, root.relativize(full).toString) }
-
     val transitiveProtoPaths: List[String] = (args.get(3) match {
       case "-" => Nil
       case s if s.charAt(0) == '-' => s.tail.split(':').toList //drop padding character
       case other => sys.error(s"expected a padding character of - (dash), but found: $other")
-    }).map(fixTransitiveProtoPath(includedRelProto, _)) ++ List(".")
+    }).map(fixTransitiveProtoPath(includedProto)) ++ List(".")
 
     val tmp = Paths.get(Option(System.getProperty("java.io.tmpdir")).getOrElse("/tmp"))
     val scalaPBOutput = Files.createTempDirectory(tmp, "bazelscalapb")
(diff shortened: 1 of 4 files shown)


johnynek:
sorry for the latency on review.

@ianoc You have done the most work on protobuf. Can you also review? this seems a bit simpler, but also more coding to the bug, so possible brittle with bazel versions changing behavior?

src/scala/scripts/PBGenerateRequest.scala
```diff
@@ -6,6 +6,25 @@ case class PBGenerateRequest(jarOutput: String, scalaPBOutput: Path, scalaPBArgs
...
+  //  (1) We have a list of all required .proto files; this is a tuple list (root -> full path), for example:
+  //        bazel-out/k8-fastbuild/bin -> bazel-out/k8-fastbuild/bin/external/com_google_protobuf/google/protobuf/source_context.proto
+  //  (2) Convert the full path to relative from the root:
+  //        bazel-out/k8-fastbuild/bin -> external/com_google_protobuf/google/protobuf/source_context.proto
+  //  (3) From all the included protos we find the first one that is located within dir we are processing -- relative
+  //      path starts with the dir we are processing
+  //  (4) If found -- the include folder is "orphan" and is not anchored in either host or target. To fix we prepend
+  //      root. If not found, return original. This works as long as "external/com_google_protobuf" is available in
+  //      target root.
+  def fixTransitiveProtoPath(includedProto: List[(Path, Path)])(orig: String): String = includedProto
```
johnynek: nit: can we store the first map through the list:
```scala
def fixTransitiveProtoPath(includedProto: List[(Path, Path)]): String => String = {
  val relPath = includedProto.map { case (root, full) => (root.toString, root.relativize(full).toString) }

  { orig: String =>
    relPath.find { case (_, rel) => rel.startsWith(orig) } match {
       case Some((root, _)) => s"$root/$orig"
       case None => orig
    }
  } 
}
```
This is a bit longer, but it builds the list and converts to strings once.

ignasl: Good point! Optimized.

ianoc:
This lgtm -- i think once they add the rest of the support upstream/tooling we'll be changing all this to use the built in bazel helpers. Possibly we might retire this for the direct plugin support unless folks have a distinct need for the workers (we mostly don't use them for scalapb/disable them since the memory outlay for lots of workers isn't worth it). So i guess long way of saying i think more upheavals from bazel itself probably lie in the future.

One request would just be some scala unit tests directly on `fixTransitiveProtoPath`, the comment looks great, but having a sample or two in code showing how the paths appear, and how they are transformed. (looking like a regression test/code debugging for when bazel changes this stuff in a few months too).

ignasl:
Added a couple of tests showing how paths are fixed. Should I add more?

ianoc:
Nope this looks good to me now. Thank you.

johnynek:
Thanks!

The bazel style checker failed. Can you run buildifier?

johnynek:
Thanks! I really appreciate the help!

Issue #856, opened 2019-10-03
test_scala_import_source_jar_should_not_be_fetched_when_env_bazel_jvm_fetch_sources_is_set_to_non_true is not working


borkaehw:
It actually gets error `Command 'test_scala_import_fetch_sources' not found. Try 'bazel help'.`, but the message was silenced by `$runner`.

`action_should_fail` is designed to take a Bazel command, not a shell function.

PR #858, opened 2019-10-03
Fix test_scala_import_source_jar.sh

borkaehw:
### Description

It fixes the issue that `test_scala_import_source_jar_should_not_be_fetched_when_env_bazel_jvm_fetch_sources_is_set_to_non_true` is not testing correctly.

Fixes https://github.com/bazelbuild/rules_scala/issues/856

### Motivation

Intermediate step for test refactoring.

### DIFF
--- a/test/shell/test_scala_import_source_jar.sh
+++ b/test/shell/test_scala_import_source_jar.sh
@@ -11,7 +11,7 @@ test_scala_import_fetch_sources_with_env_bazel_jvm_fetch_sources_set_to() {
   local expect_failure=$2
 
   if [[ ${expect_failure} ]]; then
-    action_should_fail test_scala_import_fetch_sources
+    test_scala_import_fetch_sources $expect_failure
   else
     test_scala_import_fetch_sources
   fi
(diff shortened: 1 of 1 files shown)


test/shell/test_scala_import_source_jar.sh
```diff
@@ -22,21 +22,32 @@ test_scala_import_fetch_sources_with_env_bazel_jvm_fetch_sources_set_to() {
...
+      exit 0
+    fi
   else
-    echo "File $1 does not exist."
-    exit 1
+    if [[ -f $2 ]]; then
+      echo "File $2 exists."
+      exit 0
+    else
+      echo "File $2 does not exist."
```
johnynek: can we add "but we expect it not to exist" here to be consistent with line 37?

borkaehw: Shouldn't it be "but we expect it to exist" ?

johnynek: yes, I was wrong and made a mistake above.

borkaehw: Good idea. Changed.

johnynek:
Thank you!

Issue #855, opened 2019-10-03
action_should_fail_with_message does not exit with 1


borkaehw:
`action_should_fail_with_message` is not working as expected. Changing the failing message doesn't fail the test since `action_should_fail_with_message` doesn't `exit 1` in the `elif [ $GREP_RES -ne 0 ]; then` statement.

This bug affects 4 following tests:
1. test_javac_jvm_flags.sh: javac_jvm_flags_are_expanded
2. test_javac_jvm_flags.sh: javac_jvm_flags_via_javacopts_are_expanded
3. test_scalac_jvm_flags.sh: test_scalac_jvm_flags_are_expanded
4. test_toolchain.sh: java_toolchain_javacopts_are_used

Once `exit 1` is added, it fixes 1 and 2, but it also reveals 3 and 4 are existing failing tests.

ittaiz:
Do you have an idea on how to fix 3/4? Do you need a hand? Would you rather send a PR where they fail and have me or @johnynek iterate on it to fix 3/4?

borkaehw:
I currently have no idea how to fix 3/4. I appreciate for the help.

This is the PR https://github.com/bazelbuild/rules_scala/pull/860. You may not see 4 failing since our test framework stops when it encounters the first failure. It will be better if CI can run all the tests before stopping.

liucijus:
Closed via #1198

PR #860, opened 2019-10-07
Fix action_should_fail_with_message does not exit 1

borkaehw:
### Description

`action_should_fail_with_message` should exit 1 if the action doesn't fail with the message. 

Track the issue here: https://github.com/bazelbuild/rules_scala/issues/855

### Motivation

Intermediate step for test refactoring.


liucijus:
Solved via #1198

PR #861, opened 2019-10-10
Add a test build for bazel 1.0

ianoc-stripe:
### Description
We might want to move the default over soon. But figure we should just get a sense of the master trunk or new PR's work for 1.0

### Motivation
1.0 got released today


ianoc-stripe:
@johnynek yeah ==, i think if we can at all be nice to keep compatiblity for a ~while at least. 

i think i've gotten it right now

Issue #862, opened 2019-10-12
Bazel 1.0.0 Compatibility


thundergolfer:
### Description

Bumping `rules_scala` to current `HEAD`, `f985e5e0d6364970be8d6f15d262c8b0e0973d1b`, allowed me to build a very minimal amount of Scala code (in [thundergolfer/the-one-true-bazel-monorepo](https://github.com/thundergolfer/the-one-true-bazel-monorepo)) using Bazel 1.0.0. Reassuring, but hardly a proper test of compatibility.

The `README` currently doesn't list `1.0.0` in its compatibility table, so I'm opening this issue to highlight that and track `1.0.0` compatibility. 

Thanks for all the work on these rules. Hopefully I can contribute back someday 🙂.

johnynek:
See #861 — all tests pass.

Why not send your first PR as a note in the README that 1.0 is supported?

ittaiz:
readme is updated. We're using bazel 2.0 and rules_scala on prod

PR #863, opened 2019-10-13
Include Bazel 1.0.0 in compatibility table

thundergolfer:
### Description

[From this it's my understanding](https://github.com/bazelbuild/rules_scala/pull/861#issuecomment-540713730) that `0.28.0` can remain at `HEAD` as CI is still being run for that version.

### Motivation

https://github.com/bazelbuild/rules_scala/issues/862#issuecomment-541296893


johnynek:
Thanks!

PR #869, opened 2019-11-04
[WIP] see what happens if we drop scala attribute

ittaiz:
### Description
Checking impact of dropping scala attribute

### Motivation
Bazel is going to drop the sturct/attribute syntax (https://github.com/bazelbuild/bazel/issues/9014)
We need to understand the impact of dropping the scala attribute.
One thing is that this requires Bazel 1.1.0 due to both junit and IJ import needing this fix https://github.com/bazelbuild/bazel/issues/8916

### DIFF
--- a/scala/private/rules/scala_library.bzl
+++ b/scala/private/rules/scala_library.bzl
@@ -127,7 +127,6 @@ def _lib(
         jars_to_labels = jars.jars2labels,
         providers = [outputs.merged_provider, jars.jars2labels] + outputs.coverage.providers,
         runfiles = runfiles,
-        scala = scalaattr,
     )
 
 ##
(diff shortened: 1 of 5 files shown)


ittaiz:
I've marked this as ready for review to trigger travis (for some reason it doesn't work on draft PRs).
Don't merge this PR.

scala/private/rules/scala_library.bzl
```diff
@@ -127,7 +127,6 @@ def _lib(
         jars_to_labels = jars.jars2labels,
         providers = [outputs.merged_provider, jars.jars2labels] + outputs.coverage.providers,
         runfiles = runfiles,
-        scala = scalaattr,
```
johnynek: Note above where we construct scalaattr: can we remove that? Indeed can we grep for all calls to create_scala_provider and make sure we aren’t calling that function or that it can be refactored to return a JavaInfo?

johnynek:
This is exciting as it removes some of the oldest cruft in the rules. Thank you!

ittaiz:
The failures are in the bazel run pre 1.1.0 so we're good

ittaiz:
:) indeed.
Re the scala attribute creation and such- definitely, you know I don’t
leave garbage behind. I was just trying to understand how much heavy
lifting is required.

PR #881, opened 2019-11-22
Add compatibility test for v1.2.0

### LINKED ISSUE #880: Upgrade to be compatible with Bazel 1.2.0
As Bazel [v1.2.0](https://github.com/bazelbuild/bazel/releases/tag/1.2.0) is out, we should upgrade the build config to check on the compatibility.

chenrui333:
### Description

Fixes #880

Update build config for Bazel 1.1.0 and Bazel 1.2.0

### Motivation

Keep up to date with the latest Bazel and meetup is actually actively using this lib.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,27 +51,11 @@ jobs:
     - <<: *linux
       env: TEST_SCRIPT=test_rules_scala
     - <<: *linux
-      env: TEST_SCRIPT=test_rules_scala BAZEL_VERSION=1.0.0
-    - <<: *linux
-      env: TEST_SCRIPT=test_rules_scala BAZEL_VERSION=1.1.0
-    - <<: *linux
       env: TEST_SCRIPT=test_reproducibility
-    - <<: *linux
-      env: TEST_SCRIPT=test_reproducibility BAZEL_VERSION=1.0.0
-    - <<: *linux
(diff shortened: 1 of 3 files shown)


chenrui333:
Right now we are actually using Bazel 1.1.0 in production, so I am confident with Bazel 1.1.0 upgrade. 

But it would be nice to also introduce the compatibility test for Bazel 1.2.0 as well. :)

chenrui333:
Also relates to https://github.com/meetup/rules_openapi/pull/22

ittaiz:
Thanks!
I’d rather we update 1.0.0 to 1.2.0 since our past experience tells us that having too many versions on Travis kills our ability to merge code in easily.

chenrui333:
@ittaiz Sounds totally reasonable. Let me update it.

ittaiz:
build failed. Can you PTAL?

ittaiz:
We can have two PRs by the way. We can have to upgrade to 1.1.0 and then a followup one to replace it with 1.2.0
If we do that I'd appreciate the diffs to be as small as possible (should be close to replacing 1.0.0 to 1.1.0 and so on)

chenrui333:
Sure. Sounds like a better idea.

chenrui333:
Breakdown the upgrade steps in new PRs, closing this one for now :)

PR #902, opened 2019-12-19
Bump Bazel to v2.0.0

chenrui333:
### Description

Bump Bazel to the latest `v2.0.0`.

relates to bazelbuild/bazel#10131

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,19 +51,19 @@ jobs:
     - <<: *linux
       env: TEST_SCRIPT=test_rules_scala BAZEL_VERSION=0.28.1
     - <<: *linux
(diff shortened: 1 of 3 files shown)


ittaiz:
Thanks! You’ve moved to it already?
Let’s see if the tests pass...

chenrui333:
Yeah, I am trying to get all our dependencies aligned with the new version before pulling the switch, I might do it after the new year though. :)

ittaiz:
I suggest you take a look at #865 ? It’s a big change and I suggest you see meetup works well with it

chenrui333:
Cool, I will take a look!

PR #907, opened 2019-12-26
expand locations in scalacptops, take 2

andyscott:
https://github.com/bazelbuild/rules_scala/pull/890 but with an additional adjustment to support usages of the compile function from rules without the `plugins` attribute.

Specifically:

```python
scalacopts_expansion_targets = getattr(ctx.attr, 'plugins', [])
scalacopts = [ctx.expand_location(v, scalacopts_expansion_targets) for v in toolchain.scalacopts + in_scalacopts]
```

instead of

```python
scalacopts = [ctx.expand_location(v, ctx.attr.plugins) for v in toolchain.scalacopts + in_scalacopts]
```


andyscott:
Formatting error. 🤦‍♂️

PR #909, opened 2019-12-28
Plugin expansion- Use input plugins param instead of ctx

ittaiz:
### Description
Plugin expansion- Use input plugins param instead of ctx

### Motivation
Reduce usage of ctx in inner methods.
This also allows custom logic in the rules to which attributes they use


ittaiz:
@andyscott @johnynek this is where we actually messed up (apart from the proto coverage which I'll check). We should strive to use the inputs wherever possible and not use ctx if we can.

andyscott:
Yep, agreed!

PR #912, opened 2020-01-08
Phase Scalafmt

borkaehw:
### Description

This is the first attempt to use phase extension. The new phase can format Scala source code via [Scalafmt](https://scalameta.org/scalafmt/).

It is inserted at the end of customizable phases. The extension adds an additional boolean attribute `format` to the rule. If `format` is set to true, users can run
```
bazel run TARGET_NAME.format
```
to format the source code, and run
```
bazel run TARGET_NAME.format-test
```
to test the formatting (without modifying source code).

### Motivation

Formatting is one of the features in higherkindness/rules_scala. I believe it will bring value to this repo.

### TODO:
- [x] Write documentation once we agree to move forward.


WORKSPACE
```diff
@@ -40,6 +40,35 @@ load("//specs2:specs2_junit.bzl", "specs2_junit_repositories")
...
 
+load("//scala/scalafmt:scalafmt.bzl", "scalafmt_default_config", "scalafmt_repositories")
+
+scalafmt_repositories()
+
+scalafmt_default_config()
+
+RULES_JVM_EXTERNAL_TAG = "3.1"
+
+http_archive(
```
borkaehw: I don't know why I can't import `org.scalameta:scalafmt-core_2.11:2.0.0` by `scala_maven_import_external`, but `rules_jvm_external` works.

scala/scalafmt/BUILD
```diff
@@ -0,0 +1,37 @@
...
+
+filegroup(
+    name = "testrunner",
+    srcs = ["private/format-test.template.sh"],
+    visibility = ["//visibility:public"],
+)
+
+scala_binary(
+    name = "scalafmt",
+    srcs = glob(["scalafmt/ScalafmtRunner.scala"]),
```
ittaiz: plain srcs? you're not globbing anything here

scala/scalafmt/BUILD
```diff
@@ -0,0 +1,37 @@
...
+    main_class = "io.bazel.rules_scala.scalafmt.ScalafmtRunner",
+    visibility = ["//visibility:public"],
```
ittaiz: do you want users to be able to use this target?

scala/scalafmt/BUILD
```diff
@@ -0,0 +1,37 @@
...
+    ],
+)
+
+load(
+    "//scala/scalafmt:phase_scalafmt_ext.bzl",
+    "scalafmt_singleton",
+)
+
+scalafmt_singleton(
+    name = "phase_scalafmt",
```
ittaiz: what does this name mean?

scala/scalafmt/BUILD
```diff
@@ -0,0 +1,37 @@
...
+        "@scalafmt//:org_scalameta_scalafmt_core_2_11",
+    ],
+)
+
+load(
+    "//scala/scalafmt:phase_scalafmt_ext.bzl",
+    "scalafmt_singleton",
+)
+
+scalafmt_singleton(
```
ittaiz: `scalafmt_standalone`? what does it mean it's singleton?

ittaiz: how is this used?

scala/scalafmt/scalafmt.bzl
```diff
@@ -0,0 +1,100 @@
+load(
```
ittaiz: should this file be named `scalafmt_repositories`?

scala/scalafmt/scalafmt.bzl
```diff
@@ -0,0 +1,100 @@
...
+        server_urls = maven_servers,
+    )
+
+    native.bind(
+        name = "io_bazel_rules_scala/scalafmt/scalafmt_core",
+        actual = "@scalafmt_core",
+    )
+
+    _scala_maven_import_external(
+        name = "argparse4j",
```
ittaiz: don't we have basic arg parsing in our existing workers?
can we use it? should we?

scala/scalafmt/phase_scalafmt_ext.bzl
```diff
@@ -0,0 +1,54 @@
+load(
```
ittaiz: why does this file's name ends with `_ext`?

scala/scalafmt/private/format.template.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash -e
+RUNPATH="${TEST_SRCDIR-$0.runfiles}"/%workspace%
```
ittaiz: Haven't reviewed it or the format-test file since I think a few comments would really make my work easier. Any chance you can document it?

ittaiz: Can you write here where do you use it? I have a feeling it's in phase_scalafmt.bzl but couldn't find it

test/scalafmt/formatted-test.scala
```diff
@@ -0,0 +1,8 @@
+package scalarules.test.scalafmt
+import org.scalatest._
+class FormatTest extends FlatSpec {
+  val message = "We will format this test!"
+  "FormatTest" should "be formatted" in {
+    assert(message.equals("We will format this test!"))
```
ittaiz: the content doesn't matter, right?

test/shell/test_scalafmt.sh
```diff
@@ -0,0 +1,58 @@
...
+    exit 1
+  fi
+
+  diff $FILE_PATH/un${FILENAME}.scala $FILE_PATH/${FILENAME}.scala
+  if [ $? -ne 0 ]; then
+    echo -e "${RED} un${FILENAME}.scala should be the same as ${FILENAME}.scala after formatting. $NC"
+    exit 1
+  fi
+
+  cp $FILE_PATH/un${FILENAME}.template.scala $FILE_PATH/un${FILENAME}.scala
```
ittaiz: should we do this in the beginning of the test and then add the `un${FILENAME}.scala` to .gitignore?
What we'll get is that we don't need to keep the three of those checked in and also if the test fails you have the files to look at

test/shell/test_scalafmt.sh
```diff
@@ -0,0 +1,58 @@
+# shellcheck source=./test_runner.sh
+dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+. "${dir}"/test_runner.sh
+. "${dir}"/test_helper.sh
+runner=$(get_test_runner "${1:-local}")
+
+run_non_default_formatting() {
```
ittaiz: why is it non_default_formatting? I don't see anything customizing the scalafmt conf file.
Speaking of this maybe we need a test to demonstrate passing in a custom scalafmt file?

borkaehw:
Thanks for always giving thorough review. I am still not very used to the PR principle here, I will respond to the comments soon.

> This PR is >600 LOC

I went through the code again, everything is related and it makes sense to be in the same PR, IMO. But I agree it becomes a big one. I would like to know how to make a PR like this (introducing a new feature), which needs a decent chunk of code to work correctly, small enough for reviewing?

I have thought of separating functions and tests, but does it make sense to merge something without tests? Or merging functions step by step, but does it make sense to merge code which is functionally incomplete?

What do you think? Thanks.

ittaiz:
I agree and this is why I wrote that I don't have any concrete feedback here. I wrote it because I wanted to align ourselves about balancing small vs large PRs and also features.
Often we can do small but important features and we should try and prioritize those.
Additionally I'll say that I just opened a PR that doesn't advance us functionally but it's a step in the right direction. The reason I opened it as-is is to help review.
I do agree however it's not exactly the same example and I'm not 100% sure how it translates to this PR (maybe only add format without check, don't know 🤷‍♀ )

WORKSPACE
borkaehw:
```
ERROR: /home/borkaehw/Desktop/code/rules_scala/test/scalafmt/BUILD:9:1: ScalaFmt test/scalafmt/test/scalafmt/formatted.scala failed: Worker process did not return a WorkResponse:

---8<---8<--- Start of log, file at /home/borkaehw/.cache/bazel/_bazel_borkaehw/5e84e9463749ae9195051aff1f87cebb/bazel-workers/worker-1-ScalaFmt.log ---8<---8<---
Exception in thread "main" java.lang.NoClassDefFoundError: pprint/TPrintColors
	at org.scalafmt.config.Config$.fromHoconFile$default$3(Config.scala:46)
	at io.bazel.rules_scala.scalafmt.ScalafmtProcessor.processRequest(ScalafmtRunner.scala:40)
	at io.bazel.rules_scala.scalafmt.ScalafmtRunner$.main(ScalafmtRunner.scala:16)
	at io.bazel.rules_scala.scalafmt.ScalafmtRunner.main(ScalafmtRunner.scala)
Caused by: java.lang.ClassNotFoundException: pprint.TPrintColors
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:583)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
	... 6 more
---8<---8<--- End of log ---8<---8<---
```

ittaiz:
Also build is failing so can you please get it to green before we iterate more?

scala/private/phases/phase_scalafmt.bzl
```diff
@@ -0,0 +1,81 @@
...
+        if src.path.endswith(".scala") and src.is_source:
+            file = ctx.actions.declare_file(src.short_path)
+            files.append(file)
+            args = ctx.actions.args()
+            args.add("--config")
+            args.add(ctx.file.config.path)
+            args.add(src.path)
+            args.add(file.path)
+            args.set_param_file_format("multiline")
+            args.use_param_file("@%s", use_always = True)
```
ittaiz: Small question- why `use_always = True`? Have you seen issues with Bazel's decision of when to spill over?

scala/private/phases/phase_scalafmt.bzl
```diff
@@ -0,0 +1,81 @@
...
+            ctx.file._runner.path,
+            ctx.workspace_name,
+            manifest.short_path,
+            ctx.outputs.scalafmt_runner.path,
+        ],
+        execution_requirements = {},
+    )
+
+def _format_tester(ctx, manifest, files):
+    ctx.actions.run_shell(
```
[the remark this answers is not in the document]
borkaehw: Sounds good. Fixed.

borkaehw:
Sorry, I think sometimes the builds fail because of error downloading third party stuff. It would be good if the contributors have a way to rerun the failing builds, rather than rebasing.

Got busy on other stuff, will address the comments as soon as possible.

scala/scalafmt/BUILD
borkaehw: Good catch, thanks.

scala/scalafmt/BUILD
borkaehw: They can if they want. I can't think of a strong reason to forbid it. What do you think?

scala/scalafmt/BUILD
borkaehw: It's a phase to format Scala source code by [scalafmt](https://scalameta.org/scalafmt/). I believe `fmt` is shortened from `format.

scala/scalafmt/BUILD
ittaiz: I usually try to be explicit about what I expose as a public API. Mainly since if we change this and people depend on it then they'd be surprised the contract changed. 
If we have a good reason then fine, otherwise maybe change to `//...` equivalent (which means only this repo can use this target). If there is a good reason then we should discuss.

WDYT?

scala/scalafmt/BUILD
ittaiz: I'm sorry I wasn't clear. I meant I don't understand the scalafmt_singleton pattern and also what does then name mean (in the bazel concept, not literaly)

WORKSPACE
ittaiz: this is because scalafmt has dependencies.
scala_maven_import_external uses the bazel (and rules_scala) idiom of statically listing the dependencies.
rules_jvm_external does this implicitly (and in build time given you didn't check in a _install.json file).
see this [gist](https://gist.github.com/ittaiz/3655a8b65ec61460a4bbdf38932b8c75) for 90% of what you need

ittaiz: I've given it some more thought and I'd really prefer not adding another "deps" pattern or rules_external. The way to go is toolchains. We need to have a scalafmt toolchain that defines it's dependencies. We need to offer one out of the box (maybe two also for 2.12) and users can always build and configure their own toolchain with their own deps.
Having said that I'm willing to let this be a followup PR if you'd rather make this one more concise (but still not with rules_jvm_external). WDYT?

borkaehw: It fixes it, thanks! I am curious how do you generate these dependencies?

borkaehw:
Force-pushed to get the https fix. I will keep the commits history to insure we have records of PR related changes.

scala/scalafmt/scalafmt.bzl
borkaehw: It makes sense. Fixed.

scala/scalafmt/scalafmt.bzl
borkaehw: I tried to search in the codebase but I couldn't find something related. Could you point me to the right direction?

scala/scalafmt/phase_scalafmt_ext.bzl
borkaehw: It means `extension`. I think we need to decide a naming convention of this kind of files before it gets too many.

test/scalafmt/formatted-test.scala
borkaehw: Right. The content doesn't matter.

scala/private/phases/phase_scalafmt.bzl
```diff
@@ -0,0 +1,81 @@
...
+def phase_scalafmt(ctx, p):
+    if ctx.attr.format:
+        manifest, files = _build_format(ctx)
+        _format_runner(ctx, manifest, files)
+        _format_tester(ctx, manifest, files)
+    else:
+        ctx.actions.write(
+            output = ctx.outputs.scalafmt_runner,
+            content = "",
+            is_executable = True,
```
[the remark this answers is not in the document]
borkaehw: Good catch, it probably shouldn't be. Fixed.

WORKSPACE
ittaiz: We have a tool similar to bazel-deps that does this. It's similar to bazel-deps but different in a few key ways (it has support for a bazel repo with managed dependencies for a many repo org, it lays out the deps in external repositories and not build targets so they only get built if needed and a few other things).
We're actually planning to open-source it since we're open sourcing a few of our tools and they're using this pattern so we'll need this in.
cc @natansil as the one driving the requirement to open source it and cc @or-shachar as the one who'll probably own the open sourcing in some capacity

scala/scalafmt/phase_scalafmt_ext.bzl
ittaiz: How do you define an extension? I'm guessing it depends on how you define the other thing (core?)? Maybe it should just be phase_scalafmt?
I'm not against but I have a hunch there's a methodology/pattern here which is clear to you and which I'm trying to make explicit so we can discuss and agree

test/scalafmt/formatted-test.scala
ittaiz: I'd consider removing it for a "success" case mainly since it can surprise others. Having said that I don't feel strongly about it so your decision

scala/scalafmt/BUILD
ittaiz: ok, now I saw how this is used (and re-read the docs). I agree we need naming convention here.
I think we need it for the bzl filename like you say, for the extension rule and extension target. I'm also not sure if it's a good idea that it will be named like the implementation function (at least this is how the [docs](https://github.com/bazelbuild/rules_scala/blob/master/docs/customizable_phase.md) show it).

WORKSPACE
borkaehw: The tool sounds amazing. I wonder why don't we just use `rules_jvm_external`?

scala/private/phases/phase_scalafmt.bzl
```diff
@@ -0,0 +1,81 @@
...
+#
+# Outputs to format the scala files when it is explicitly specified
+#
+def phase_scalafmt(ctx, p):
+    if ctx.attr.format:
+        manifest, files = _build_format(ctx)
+        _format_runner(ctx, manifest, files)
+        _format_tester(ctx, manifest, files)
+    else:
+        ctx.actions.write(
```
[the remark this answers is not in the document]
borkaehw: Do you mean like
```
ctx.actions.run_shell(
    inputs = [empty_file],
    outputs = [output_file],
    command = "cp $1 $2",
    arguments = [
        empty_file.path,
        output_file.path,
    ],
    execution_requirements = {},
)
```
Will we really see an obvious performance gain? But it's a bit harder to understand at first glance, IMO.

scala/private/phases/phase_scalafmt.bzl
```diff
@@ -0,0 +1,81 @@
...
+    manifest = ctx.actions.declare_file("format/{}/manifest.txt".format(ctx.label.name))
+    ctx.actions.write(manifest, "\n".join(manifest_content) + "\n")
+
+    return manifest, files
+
+def _format_runner(ctx, manifest, files):
+    ctx.actions.run_shell(
+        inputs = [ctx.file._runner, manifest] + files,
+        outputs = [ctx.outputs.scalafmt_runner],
+        command = "cat $1 | sed -e s#%workspace%#$2# -e s#%manifest%#$3# > $4",
```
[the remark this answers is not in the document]
borkaehw: `$1`, `$2`, ... indicate the arguments in order. The command means read the content in `$1`, replace `%workspace%` with `$2` and replace `%manifest%` with `$3`, and write to `$4`.

See `scala/scalafmt/private/format-test.template.sh` and `scala/scalafmt/private/format.template.sh` for where are `%workspace%` and `%manifest%`.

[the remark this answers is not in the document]
borkaehw: Comment added to the code.

scala/private/phases/phase_scalafmt.bzl
```diff
@@ -0,0 +1,81 @@
...
+        )
+        ctx.actions.write(
+            output = ctx.outputs.scalafmt_testrunner,
+            content = "",
+            is_executable = True,
+        )
+
+def _build_format(ctx):
+    files = []
+    runner_inputs, _, runner_manifests = ctx.resolve_command(tools = [ctx.attr._fmt])
```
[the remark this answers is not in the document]
borkaehw: Good question, I am not sure neither. Do you see a benefit to use regular `executable`?

scala/private/phases/phase_scalafmt.bzl
```diff
@@ -0,0 +1,81 @@
...
+                outputs = [file],
+                input_manifests = runner_manifests,
+                inputs = [ctx.file.config, src],
+                tools = runner_inputs,
+                execution_requirements = {"supports-workers": "1"},
+                mnemonic = "ScalaFmt",
+            )
+            manifest_content.append("{} {}".format(src.short_path, file.short_path))
+
+    manifest = ctx.actions.declare_file("format/{}/manifest.txt".format(ctx.label.name))
```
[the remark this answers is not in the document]
borkaehw: You don't understand why we need a manifest file?

scala/private/phases/phase_scalafmt.bzl
```diff
@@ -0,0 +1,81 @@
...
+            is_executable = True,
+        )
+
+def _build_format(ctx):
+    files = []
+    runner_inputs, _, runner_manifests = ctx.resolve_command(tools = [ctx.attr._fmt])
+    manifest_content = []
+    for src in ctx.files.srcs:
+        if src.path.endswith(".scala") and src.is_source:
+            file = ctx.actions.declare_file(src.short_path)
```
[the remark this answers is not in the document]
borkaehw: It seems to work. I will change it if it's preferable.

scala/private/phases/phase_scalafmt.bzl
```diff
@@ -0,0 +1,81 @@
...
+            files.append(file)
+            args = ctx.actions.args()
```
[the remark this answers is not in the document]
borkaehw: Done.

scala/scalafmt/BUILD
borkaehw: Sounds reasonable, but there is a challenge. Every target that uses format extension will need access to this binary, it could also lives in users' private workspace. Could it be not public?

scala/scalafmt/BUILD
borkaehw: To be honest, I don't have answer. It comes from the convention of `higherkindness/rules_scala`. I personally interpret it as a dummy rule just to pass in `_ScalaRulePhase` provider.

We don't have to follow any convention. I lean toward setting up principle for future reference. Unless @andyscott can provide more insight here.

scala/scalafmt/BUILD
borkaehw: Agree on setting up a naming convention before more PRs are coming. If it's not affecting performance, we could address it soon after this is merged.

scala/scalafmt/private/format.template.sh
borkaehw: In `scala/scalafmt/BUILD`, `format.template.sh` and `format-test.template.sh` are included by `filegroup`. The `filegroup`s are used in extra attributes `_runner` and `_testrunner`. So it is indeed used in `phase_scalafmt.bzl`.

test/shell/test_scalafmt.sh
borkaehw: Since it's an extension, we don't format code by default.

Agree on having a test to demonstrate passing a custom scalafmt conf file.

scala/private/phases/phase_scalafmt.bzl
borkaehw: Never tried `False`. Do you suggest to change?

test/scalafmt/formatted-test.scala
borkaehw: > removing it for a "success" case

Not sure what exactly you mean here. But I can see the dummy string could be confusing. I change it to `assert(true)`, it should be more obvious.

Let me know if I misunderstand.

test/shell/test_scalafmt.sh
borkaehw: Test added.

test/shell/test_scalafmt.sh
borkaehw: Let me think about a better way to test. Get back to you soon.

borkaehw:
I have responded to all the comments, sorry for the delay again. I now feel the tests for scalafmt are so hard to maintain, I will think of a better way soon.

Please take another round of review, thanks.

WORKSPACE
ittaiz: 1. It’s a long discussion which I’d rather not do on this PR.
2. For that matter you can also ask why not bazel-deps.
3. I think this challenge hasn’t hit us so far because we strive to have minimum dependencies and so making it easier to have large transitive graphs is actually not that great. 
4. I think that given that everything works via toolchains then it will be much easier for users to override whatever rules scala suggests out of the box and so at that point maybe introducing an easier entry level tool while still allowing more advanced users to do whatever they want is a better trade off.

scala/private/phases/phase_scalafmt.bzl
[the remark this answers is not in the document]
ittaiz: Let’s ignore

scala/private/phases/phase_scalafmt.bzl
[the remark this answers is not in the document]
ittaiz: Yes, it’s simpler. This is a more advanced mechanism and so makes one think why is it used and not the simpler more common mechanism.

scala/private/phases/phase_scalafmt.bzl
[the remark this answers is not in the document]
ittaiz: I think it’s preferable because often we ls the bazel-bin folders and then it helps to have tool namespaces on the outputs. Wdyt?

scala/scalafmt/BUILD
ittaiz: Sure thing. Let’s do that

scala/scalafmt/phase_scalafmt_ext.bzl
ittaiz: Let’s close this one for now and defer it to the convention discussion.
I think this question of whether we should separate to two files should also be there

test/scalafmt/formatted-test.scala
ittaiz: Sounds great (In specs2 you can have the body contain just `success`)

scala/private/phases/phase_scalafmt.bzl
ittaiz: Yes, it will save IO for small actions (I’m guessing the majority)

ittaiz:
I’ll go over the changes soon.
Can you share what you feel is hard about the tests? Maybe I can assist?
This is an area that’s very close to my heart

scala/scalafmt/BUILD
ittaiz: I'm pretty sure this isn't a problem. Visibility doesn't effect transitive dependencies but only who can directly depend on it.

scala/scalafmt/scalafmt.bzl
ittaiz: https://github.com/bazelbuild/rules_scala/blob/master/src/java/io/bazel/rulesscala/scalac/CompileOptions.java
The idea is to have minimum dependencies for the core rules operations. 
If we can't use the existing infra then maybe we should use argparse4j but I'd prefer to minimize if possible.

scala/scalafmt/private/format.template.sh
ittaiz: 👍 wdyt about adding a few comments inside? Skimming it wasn't easy (bash...)

test/shell/test_scalafmt.sh
ittaiz: WDYT about renaming to `run_formatting()`? What's the value of saying it's non-default? Are you concerned someone will think this happens by default? Why? Not saying you're wrong but I don't see the issue

ittaiz: Also like on the custom conf test

scala/scalafmt/scalafmt_repositories.bzl
```diff
@@ -10,7 +10,7 @@ def scalafmt_default_config(path = ".scalafmt.conf"):
     build.append("    srcs = [\"{}\"],".format(path))
     build.append("    visibility = [\"//visibility:public\"],")
     build.append(")")
-    native.new_local_repository(name = "scalafmt_default", build_file_content = "\n".join(build), path = "")
+    native.new_local_repository(name = "scalafmt_default", build_file_content = "\n".join(build), path = "scala/scalafmt/conf")
```
ittaiz: Is this location documented anywhere?

scala/private/phases/phase_scalafmt.bzl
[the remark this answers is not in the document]
borkaehw: Sounds reasonable, fixed.

test/shell/test_scalafmt.sh
borkaehw: I agree on your comment. Renamed.

scala/private/phases/phase_scalafmt.bzl
borkaehw: When I change it to `False`, I got this
```
ERROR: /home/borkaehw/Desktop/code/rules_scala/test/scalafmt/BUILD:15:1: ScalaFmt test/scalafmt/test/scalafmt/formatted/formatted-encoding.scala.fmt.output failed: Worker strategy cannot execute this ScalaFmt action, because the command-line arguments do not contain at least one @flagfile or --flagfile=
```
Any thoughts?

ittaiz: Did you change it to False or just drop the argument (and still keep the method call obviously)? Asking because I used the args without this in a PR a couple of days ago and it worked

test/shell/test_scalafmt.sh
borkaehw: I like your suggestion. But we have to keep `un${FILENAME}.scala` checked in otherwise `bazel build test/...` in `test_rules_scala.sh` will complain no such source files. Instead, we don't have to check in `un${FILENAME}.template.scala`.

Re-organize the test a bit. Please take a look again.

scala/private/phases/phase_scalafmt.bzl
borkaehw: > Did you change it to False or just drop the argument

I tried both, same issue.

scala/private/phases/phase_scalafmt.bzl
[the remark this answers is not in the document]
borkaehw: This is a file to record the source path and the formatted file path. So that the `format.template.sh` and `format-test.template.sh` know where to find the formatted file to copy from (to replace the source file if it's not formatted).

scala/scalafmt/BUILD
borkaehw: Sorry, it might be obvious to you but I can't get it to work if not public.

I got the message
```
ERROR: in scalafmt_scala_binary rule //test/scalafmt:formatted-binary: target '//scala/scalafmt:scalafmt' is not visible from target '//test/scalafmt:formatted-binary'. Check the visibility declaration of the former target if you think the dependency is legitimate
```
I appreciate if you can point the problem out.

ittaiz: It seems I was wrong. I'm not really sure why but let's ignore for now so we can get this in

scala/private/phases/phase_scalafmt.bzl
ittaiz: I want to suggest the following and LMK what you think- I suggest we merge it with the use_always and then open a separate PR just with this argument dropped. Once we have a failure there I can tag bazel people to try and help us with this pinpointed issue. WDYT?

scala/private/phases/phase_scalafmt.bzl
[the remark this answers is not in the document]
ittaiz: I see. Thanks for explaining. I'd love for this explanation to be documented in the code. WDYT?

scala/private/phases/phase_scalafmt.bzl
borkaehw: Sounds good.

scala/scalafmt/scalafmt.bzl
borkaehw: It turns out we may not need a parser here, so I dropped it. Let me know if you can see any problem without a parser.

scala/private/phases/phase_scalafmt.bzl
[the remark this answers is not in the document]
borkaehw: I honestly don't know why we did it this way at the first place, but dropping it doesn't seem to break anything. Fixed.

scala/private/phases/phase_scalafmt.bzl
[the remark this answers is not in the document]
borkaehw: Sure thing.

scala/scalafmt/scalafmt_repositories.bzl
borkaehw: I will document it.

borkaehw:
It seems like we have addressed most of the concerns. I will start writing documentation. Do we want it to be in a separate file or in an existing file?

I will also write more comments in the code. Feel free to point out the must-have places.

I figure there is some unnecessary code which is based on the assumption in previous repository. After cleaning, `format.template.sh` and `format-test.template.sh` share quite a bit of duplicate code. I will make some changes later.

I am now thinking is there a better way to toggle `.format` and `.format-test`? The only difference is the former formats the source and the latter just reports the difference. 99% of the code are the same except the last part in `.sh`. Or the other question is, how valuable is `.format-test`? Is there a case that people just want to check but don't want to change the source? WDYT?

borkaehw:
> Can you share what you feel is hard about the tests? Maybe I can assist?

The challenge is I don't feel confident that I am setting up the tests in the most maintainable and straightforward way (hopefully it already is). I appreciate if you can take a look. Thanks.

scala/scalafmt/private/format.template.sh
borkaehw: I will make it just `RUNPATH`.

test/shell/test_scalafmt.sh
```diff
@@ -4,31 +4,26 @@ dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 . "${dir}"/test_helper.sh
 runner=$(get_test_runner "${1:-local}")
 
-backup_unformatted_file() {
+backup_unformatted() {
   FILE_PATH=$1
   FILENAME=$2
-  cp $FILE_PATH/unformatted/un${FILENAME}.scala $FILE_PATH/unformatted/un${FILENAME}.backup.scala
+  cp $FILE_PATH/unformatted/unformatted-$FILENAME.scala $FILE_PATH/unformatted/unformatted-$FILENAME.backup.scala
```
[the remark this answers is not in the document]
borkaehw: Good suggestion!

scala/private/phases/phase_scalafmt.bzl
```diff
@@ -50,6 +53,7 @@ def _formatter(ctx, manifest, files, input_runner, output_runner):
     )
 
 def _write_empty_content(ctx, output_runner):
+    # generate empty output if format=false
```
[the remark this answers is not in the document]
borkaehw: Make sense. Fixed.

borkaehw:
It seems like the only thing left is documentation? I will update it when I have it ready.

Also, I will rebase later.

ittaiz:
Indeed!
Two thoughts (which definitely can land in follow-up PRs):
1. See if we can somehow offer users (given a flag) to fail the build on formatting test failure (maybe just make it an sh_test and have this action be an optional one in a phase that will only do it if parameter is on?)
2. Move scalafmt_repositories to a toolchain. I'd like to get rid of the repositories pattern

borkaehw:
Just a heads-up that I probably don't have time for documentation until early next week. I think it's also a good idea to list all the follow-ups in a Google doc. Will do it later.

> fail the build on formatting test failure

I remember you mentioned this before but not exactly sure what it will look like. Could you elaborate more?

ittaiz:
Re documentation- sure, thanks for letting me know
Re follow up doc- SGTM (if there are just a few issues we can open GH issues for them)
Re fail build- I’ll try to send a PR to this PR
So I won’t forget- did you change the naming here according to the new convention?

ittaiz:
Thought about "failing the build" some more and I see a few options:
1. Have an independent macro that users set up in addition to their `scalafmt_scala_library` targets which looks something like:
```
def scala_fmt_test(name,target):
    native.sh_test(
        name = name + "_scala_fmt_test",
        srcs = [target + ".format-test"],
    )
```
This currently doesn't work as is because the phase doesn't add the outputs to the runfiles of the scalafmt_scala_library rule (hopefully we'll change that in #958) and I had a bit of a hard time hacking around it because the phase puts the output files in weird places (`format/{}/manifest.txt".format(ctx.label.name)`)
2. Have `scalafmt_scala_library` actually be a macro that delegates to `_scalafmt_scala_library_rule` and `scala_fmt_test`.
3. If you want this to be toggled via a toolchain then you can do something like 2 but instead of using a regular `sh_test` you can write one of your own that will depend on the toolchain.
WDYT?

ittaiz:
@borkaehw can we merge this? 
The toolchain, fail build and other tasks (if exist) can come in a follow-up PR.  
Main reason is since I want to break the API and add `external_providers` as one of the inputs to phases and I'd rather not give you a merge conflict.
Let's just make sure naming here is up-to-date with the new format?

borkaehw:
It's ready to be merged. There is no corresponding naming issue in this PR.

Thank you as always.

borkaehw:
Ah, sorry for the lint failure.

borkaehw:
Got a `GET returned 403 Forbidden`.

ittaiz:
restarted the job. will take a look in the morning. GN.

ittaiz:
@borkaehw can you PTAL at https://travis-ci.community/t/continuous-maven-repo-403/3908? Maybe we can add URL as well (import_external accepts many URLs) and that will help us be more resilient

borkaehw:
I added `https://maven-central.storage-download.googleapis.com/maven2`. Let me know if I misunderstand your suggestion.

borkaehw:
There seems to be a lot more places that have `scala_maven_import_external`, not sure if you would like to change them all in this PR.

ittaiz:
Thanks!
If you’re willing to send a follow up PR just for the URLs I’d be happy to review and merge. If not LMK and I’ll try to find the time to do it soon.

borkaehw:
I am happy to do so. I am thinking about putting the default urls in a shared file. Do you have suggesting where to put it?

ittaiz:
Shared file sounds great. No real preference.

PR #914, opened 2020-01-13
Remove providers

ittaiz:
### Description
Remove remaining attributes.

This is part of a larger effort of removing all attributes we're returning and return only providers. 
Attributes were either:
1. Moved to DefaultInfo provider
2. Removed because they were redundant (pre phases they were needed in internal communication)
3. Coverage which moved to its own provider

Followup on #913 

### Motivation
Be Bazel 3.0.0 compliant per this bazel [issue](https://github.com/bazelbuild/bazel/issues/7347) and unifying on JavaInfo.

Note that this means the minimum version is 1.1.0 because of a JavaInfo [issue](https://github.com/bazelbuild/bazel/issues/8916) that was fixed in 1.1.0 and breaks IJ integration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,19 +49,19 @@ jobs:
       env: TEST_SCRIPT=test_lint
 # Test
     - <<: *linux
-      env: TEST_SCRIPT=test_rules_scala BAZEL_VERSION=0.28.1
+      env: TEST_SCRIPT=test_rules_scala BAZEL_VERSION=1.1.0
     - <<: *linux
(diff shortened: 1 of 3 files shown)


ittaiz:
This doesn't complete the task since we need to actually return an array of providers from [run_phases](https://github.com/bazelbuild/rules_scala/blob/master/scala/private/phases/api.bzl#L73) but that will also involve changing the phases and their return value and such.
Will happen in followup PR.

BTW, this is based on #913 for ease of review

ittaiz:
fails on linting. Will fix tomorrow

PR #916, opened 2020-01-15
Return providers array instead of legacy struct

ittaiz:
### Description
Remove remaining attributes.

This is part of a larger effort of removing all attributes we're returning and return only providers. 
Attributes were either:
1. Moved to DefaultInfo provider
2. Removed because they were redundant (pre phases they were needed in internal communication)
3. Coverage which moved to its own provider

Followup on #913 
Supersedes #914 (more accurate name and better git history)

### Motivation
Be Bazel 3.0.0 compliant per this bazel [issue](https://github.com/bazelbuild/bazel/issues/7347) and unifying on JavaInfo.
The above issue is also where I got the recommendation of how to move coverage to a provider from

### DIFF
--- a/scala/private/phases/phase_final.bzl
+++ b/scala/private/phases/phase_final.bzl
@@ -7,8 +7,8 @@ def phase_binary_final(ctx, p):
     defaultInfo = DefaultInfo(
         executable = p.declare_executable,
         files = depset([p.declare_executable, ctx.outputs.jar]),
-        runfiles= p.runfiles.runfiles
-        )
+        runfiles = p.runfiles.runfiles,
+    )
     return struct(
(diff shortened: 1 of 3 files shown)


ittaiz:
Apparently I was confused and thanks to all rules calling `.final` on the output of `run_phases` then if we just change the return value to an array we're good.

ittaiz:
It seems the only other rule that was missing was scala_import so I just changed that one as well.
We should be free of legacy providers and struct but it's hard to check because Bazel itself isn't ready and so one can't run with the flag to check 😠 
More details in the linked issue.

ittaiz:
fixes #450

PR #919, opened 2020-01-15
Move declare_executable to phase file

borkaehw:
### Description

Move `declare_executable` to the phase file. There is no need for importing from other file.

### Motivation

Clean-up work of phase PR.

### DIFF
--- a/scala/private/phases/phase_declare_executable.bzl
+++ b/scala/private/phases/phase_declare_executable.bzl
@@ -5,8 +5,11 @@
 #
 load(
     "@io_bazel_rules_scala//scala/private:rule_impls.bzl",
-    "declare_executable",
+    "is_windows",
 )
 
 def phase_declare_executable(ctx, p):
-    return declare_executable(ctx)
+    if (is_windows(ctx)):
+        return ctx.actions.declare_file("%s.exe" % ctx.label.name)
+    else:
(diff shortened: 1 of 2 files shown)


borkaehw:
`is_windows` is also used by `write_executable` in `rule_impls.bzl`, but I will move that soon.

ittaiz:
The force push is because you rebased with the https?
Also, why not move the declare executable now? Because you’re trying to keep it small?

borkaehw:
> The force push is because you rebased with the https?

Yes, otherwise CI failed.

> why not move the declare executable now?

Not sure what do you mean here. It's a PR to move `declare_executable`. Do I miss something? Or do you mean `write_executable`?

ittaiz:
Yes, that’s what I meant. Thanks.
I want to suggest I’ll take the move pieces from rule_impls/common to phases.
Will make it easier for me to know where everything is.
Also this way I’ll send one PR to keep our commit history small and since you have an idea on what the changes look like it should be easier for you to review.
Most importantly is that I won’t be doing any semantic changes just moving code around.
I hope to send the PR today.
Ok?

borkaehw:
Sounds good to me. I will focus on `phase_scalafmt` and other stuff. Is there anything on the list that I can help?

Issue #918, opened 2020-01-15
Downloads from Maven broken


radix-aw:
As of [today](https://support.sonatype.com/hc/en-us/articles/360041287334), all requests to Maven are required to use https.

PR #920, opened 2020-01-15
Use https to access central.maven.org

aherrmann:
### Description
Use https to access central.maven.org.

Closes https://github.com/bazelbuild/rules_scala/issues/918

Maven now requires to use https instead of http. See https://support.sonatype.com/hc/en-us/articles/360041287334

This replaces all instances of http://central.maven.org by https://central.maven.org.

### Motivation
Code using rules_scala fails to build due to 501 errors.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -157,7 +157,7 @@ java_import_external(
     name = "org_apache_commons_commons_lang_3_5_without_file",
     generated_linkable_rule_name = "linkable_org_apache_commons_commons_lang_3_5_without_file",
     jar_sha256 = "8ac96fc686512d777fca85e144f196cd7cfe0c0aec23127229497d1a38ff651c",
-    jar_urls = ["http://central.maven.org/maven2/org/apache/commons/commons-lang3/3.5/commons-lang3-3.5.jar"],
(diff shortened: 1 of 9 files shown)


ittaiz:
Thanks! Did you see the build pass locally?
Build kite says:
scala-library-2.11.12.jar: No subject alternative DNS name matching central.maven.org found.

SrodriguezO:
We updated the urls in higherkindness from `http://central.maven.org/maven2` to `https://repo.maven.apache.org/maven2` since we couldn't find references to `https://central.maven.org/maven2`

CI is running right now, but the build phase succeeded already : )

SrodriguezO:
@aherrmann I think if you replace `https://central.maven.org/maven2` with `https://repo.maven.apache.org/maven2` it'll work

aherrmann:
@ittaiz Yes, it passed on a Linux build, but I then saw the same certificate issue on a macOS machine.
@SrodriguezO Thanks, yes, using https://repo.maven.apache.org/maven2 instead fixed the issue for us as well.
I've updated the PR

ittaiz:
only reproducibility tests are running and this didn't hurt reproducibility so merging

PR #923, opened 2020-01-16
scalatest final-remove redundant coverage handling

ittaiz:
### Description
Remove redundant coverage handling in scalatest final

### Motivation
Clean-up
Was looking over the phase architecture follow up tasks and saw the coverage smell issue.
Wanted to understand complexity and noticed that the call chain ends up leading to a constant `[]` so I just dropped it.
https://github.com/bazelbuild/rules_scala/blob/master/scala/private/phases/phase_write_executable.bzl#L28
https://github.com/bazelbuild/rules_scala/blob/master/scala/private/phases/phase_write_executable.bzl#L48
https://github.com/bazelbuild/rules_scala/blob/master/scala/private/phases/phase_write_executable.bzl#L67
https://github.com/bazelbuild/rules_scala/blob/6c16cff213b76a4126bdc850956046da5db1daaa/scala/private/rule_impls.bzl#L662

### DIFF
--- a/scala/private/phases/phase_final.bzl
+++ b/scala/private/phases/phase_final.bzl
@@ -19,11 +19,9 @@ def phase_library_final(ctx, p):
     return [defaultInfo, p.compile.merged_provider, p.collect_jars.jars2labels] + p.compile.coverage.providers
 
 def phase_scalatest_final(ctx, p):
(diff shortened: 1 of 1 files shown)


ittaiz:
@borkaehw wdyt?

borkaehw:
Sounds good, thanks.

PR #924, opened 2020-01-16
[Do not merge] checking if coverage feature is covered with tests

ittaiz:
### Description
Checking if coverage feature is covered with tests

### Motivation
Confidence


ittaiz:
yay! it's covered!

PR #926, opened 2020-01-16
chore(docs): update WORKSPACE setup for skylib

chenrui333:
### Description

Somehow we run into the Bazel fetch dependencies issue as below:

```
ERROR: Failed to load Starlark extension '@bazel_skylib//lib:dicts.bzl'.
 - @bazel_skylib
This could either mean you have to add the '@bazel_skylib' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
ERROR: cycles detected during target parsing
```

We fixed the issue by changing the order in our `WORKSAPCE` file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,19 @@ This project defines core build rules for [Scala](https://www.scala-lang.org/) t
 2. Add the following to your `WORKSPACE` file and update the `githash` if needed:
 
 ```python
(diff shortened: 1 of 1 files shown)


chenrui333:
BTW, I have still did not spend time working on the Bazel 2.0.0 yet (this happens on our Bazel 1.2.1 build)

PR #930, opened 2020-01-16
Move code from rule impls/common to phases

ittaiz:
### Description
Move all possible code (that is used by a single place) from rule_impls.bzl/common.bzl to the relevant phase.

### Motivation
Clean up post phases architecture and encapsulate logic in phases.
This isn't the whole shebang since some rules are still not in phases.
Additionally we have common code that should be extracted out or replaced with code from skylib and/or java/jvm common patterns repo.

Note that in some places I used the load(foo.bzl, _foo = "foo") pattern which I now think was redundant. I didn't want to add more noise in this PR with more commits so I'll send a follow-up PR just to handle redundant privacy.

Finally note that each commit is pretty much encapsulated and small


ittaiz:
cc @borkaehw

scala/private/phases/phase_compile.bzl
```diff
@@ -438,7 +442,7 @@ def _try_to_compile_java_jar(
         implicit_junit_deps_needed_for_java_compilation,
     )
     providers_of_dependencies += _collect_java_providers_of(
-        _get_scalac_provider(ctx).default_classpath,
```
ittaiz: note this is possibly the only interesting change in this PR since it moves something from a method call to taking it from a previous phase and still I'm confident

scala/private/phases/phase_collect_jars.bzl
```diff
@@ -65,10 +70,63 @@ def _phase_collect_jars(
         extra_deps,
         extra_runtime_deps,
         unused_dependency_checker_mode):
-    return collect_jars_from_common_ctx(
+    return _collect_jars_from_common_ctx(
```
borkaehw: Rather than calling `_collect_jars_from_common_ctx`, can we just move the logic in `_collect_jars_from_common_ctx` to `_phase_collect_jars`? `_phase_collect_jars` seems to do nothing

scala/private/phases/phase_collect_srcjars.bzl
```diff
@@ -3,12 +3,15 @@
...
-load(
-    "@io_bazel_rules_scala//scala/private:common.bzl",
-    "collect_srcjars",
-)
 
 def phase_collect_srcjars(ctx, p):
     # This will be used to pick up srcjars from non-scala library
     # targets (like thrift code generation)
-    return collect_srcjars(ctx.attr.deps)
+    return _collect_srcjars(ctx.attr.deps)
```
borkaehw: Same here, why not
```
def phase_collect_srcjars(ctx, p):
    # This will be used to pick up srcjars from non-scala library
    # targets (like thrift code generation)
    srcjars = []
    for target in ctx.attr.deps:
        if hasattr(target, "srcjars"):
            srcjars.append(target.srcjars.srcjar)
    return depset(srcjars)
```

scala/private/phases/phase_java_wrapper.bzl
```diff
@@ -47,3 +47,29 @@ def _phase_java_wrapper(
         args,
         wrapper_preamble,
     )
+
+def write_java_wrapper(ctx, args = "", wrapper_preamble = ""):
```
borkaehw: Can we combine `_phase_java_wrapper` and `write_java_wrapper`? Or at least make `write_java_wrapper` private?

scala/private/phases/phase_merge_jars.bzl
```diff
@@ -3,17 +3,42 @@
 #
 # DOCUMENT THIS
 #
-load(
-    "@io_bazel_rules_scala//scala/private:rule_impls.bzl",
-    "merge_jars",
-)
 
 def phase_merge_jars(ctx, p):
```
borkaehw: Same here. The reason I named this phase `phase_merge_jars` is because it's calling a function `merge_jars`, but they can be combined.

scala/private/phases/phase_write_executable.bzl
```diff
@@ -73,3 +79,107 @@ def _phase_write_executable(
         wrapper,
         use_jacoco,
     )
+
+def _write_executable(ctx, executable, rjars, main_class, jvm_flags, wrapper, use_jacoco):
```
borkaehw: Combine `_phase_write_executable` and `_write_executable`?

ittaiz: fixed

scala/private/phases/phase_merge_jars.bzl
ittaiz: fixed

scala/private/phases/phase_java_wrapper.bzl
ittaiz: fixed

scala/private/phases/phase_collect_srcjars.bzl
ittaiz: fixed

scala/private/phases/phase_collect_jars.bzl
ittaiz: fixed

ittaiz:
@borkaehw thanks for catching this! I was knee deep in bookkeeping of what can move and what can't and also what can we move to other repos (trying to dramatically cut down in our common code which isn't scala specific).
I have to say that I think that in at least one or two places the delegation had value in the sense that the phase function is responsible for extracting the needed variables from the ctx a& p and the other method does the actual work. Having said that because we don't have any real pattern yet I'm not sure this small value is worth the divergence in style for now

ittaiz:
Also when I moved the code I tried to be minimalistic and not change the code as much as I can (to ease the code review). There's a good chance that we should go over each phase and find places where it reads things directly from the `ctx` while it should actually read them from `p`.
If you want to make such a list we can discuss them and once we agree you'll be able to send a PR. WDYT? This will help in cementing the phases more.

ittaiz:
@borkaehw LMK if you have more inputs or I misunderstood. Thanks!

borkaehw:
Thanks for addressing my comments. I have no other concerns for this PR.

> find places where it reads things directly from the ctx while it should actually read them from p

Happy to help, but can you find an example of it?

ittaiz:
I think I was referring to [this](https://github.com/bazelbuild/rules_scala/blob/move_from_rule_impls_common_to_phases/scala/private/phases/phase_merge_jars.bzl#L14) but now I'm not sure there's a problem with it.
Maybe I was mistaken.

PR #942, opened 2020-01-19
deploy_jar creation uses args file to support long classpaths

ittaiz:
### Description
deploy_jar creation uses args file to support long classpaths

Using the `ctx.actions.args` (which I learned of from @borkaehw's PR so thanks!) to offload to params file when needed. 

### Motivation
We (at Wix) use `_deploy.jar` of `scala_library` to create ad-hoc main apps inside intellij to recreate the maven/sbt experience (and not require defining `scala_binary` targets for each fiddle). We found that on certain targets rules_scala just crashed.


scala/private/phases/phase_merge_jars.bzl
```diff
@@ -12,19 +12,27 @@ def phase_merge_jars(ctx, p):
...
+    args.add_all(runtime_jars, map_each = _fileToPath)
+
     if main_class:
-        args.extend(["--main_class", main_class])
-    args.extend(["--output", deploy_jar.path])
+        args.add_all(["--main_class", main_class])
+    args.add_all(["--output", deploy_jar.path])
+
+    args.set_param_file_format("multiline")
+    args.use_param_file("@%s")
```
or-shachar: what is that trick?

or-shachar: https://docs.bazel.build/versions/master/skylark/lib/Args.html#use_param_file
Cool!

ittaiz: Indeed :) I think it's not really clear why we need this param though so I opened this bazel issue https://github.com/bazelbuild/bazel/issues/10616

ittaiz:
Thanks! I've thought about extracting it but such a method would depend on `ctx.actions`, `main_class`, `runtime_jars` and  `deploy_jar` which feels a bit too much for the value.
WDYT?

or-shachar:
Indeed... Not worth the extraction... :/ 
Maybe we can fold in the `main_class` given the `ctx` but not sure it helps us much

PR #943, opened 2020-01-19
exposing java jar to BEP from all binary rules

ittaiz:
added a passing e2e for java jar of scala library for symmetry

### Description
exposing java jar to BEP from all binary rules

added a passing e2e for java jar of scala library for symmetry

### Motivation
When working on my bazelcon talk I realized we only expose the scala jar from binary targets. Anyone using java and scala and leveraging the BEP file will feel this. We (at Wix) mainly separate compilation from test targets so we don't feel it a lot but I expect that as people become more liberal with using scala_binary it might pop up more.


ittaiz:
I'll merge this tomorrow given no comments.

PR #944, opened 2020-01-20
Fix resource_strip_prefix of scala_libary for external repositories

simuons:
### Description
Fix `resource_strip_prefix` of `scala_libary` for external repositories.

I think this is a similar PR for `java_library` https://github.com/bazelbuild/bazel/pull/1085

### Motivation
If repository `A` has a `scala_library` with `resources` and `resource_strip_prefix` on which repository `B` depends then repository `B` fails to build complaining that resource doesn't start with correct prefix.


scala/private/phases/phase_compile.bzl
```diff
@@ -480,17 +480,20 @@ def _try_to_compile_java_jar(
         java_compilation_provider = provider,
     )
 
-def _adjust_resources_path(path, resource_strip_prefix):
+def _adjust_resources_path(resource, resource_strip_prefix):
+    path = resource.path
     if resource_strip_prefix:
-        return _adjust_resources_path_by_strip_prefix(path, resource_strip_prefix)
+        root = resource.owner.workspace_root + "/" if (resource.owner.workspace_root) else ""
```
ittaiz: maybe this logic should be in `_adjust_resources_path_by_strip_prefix`? Make it take a `resource`
I'm also bummed about having this logic here since it's very common starlark need and I have a hunch it exists already in bazel--skylib but I won't ask you now to go down that route

test/src/main/scala/scalarules/test/resources/strip/BUILD
```diff
@@ -14,3 +15,12 @@ scala_specs2_junit_test(
...
     deps = [":noSrcsWithResources"],
 )
+
+scala_specs2_junit_test(
+    name = "resouceStripPrefixFromExternalRepoTest",
+    size = "small",
+    srcs = ["ResourceStripPrefixTest.scala"],
+    suffixes = ["Test"],
+    unused_dependency_checker_mode = "off",
+    deps = ["@io_bazel_rules_scala//test/src/main/scala/scalarules/test/resources/strip:noSrcsWithResources"],
```
ittaiz: This feels a bit brittle to me. In Bazel `io_bazel_rules_scala` should be equivalent to this repo (see `experimental_remap_main_repo` flag in Bazel) and I think it's just a matter of time until it's on by default. By then this test won't mean anything.
Maybe add another `local_repository` in our WORKSPACE just for testing and add there one target with the strip resource prefix? Duplication but worth it for confidence.

scala/private/phases/phase_compile.bzl
```diff
@@ -504,12 +505,14 @@ def _add_resources_cmd(ctx):
         res_cmd.extend([line])
     return "".join(res_cmd)
 
-def _adjust_resources_path_by_strip_prefix(path, resource_strip_prefix):
-    if not path.startswith(resource_strip_prefix):
-        fail("Resource file %s is not under the specified prefix to strip" % path)
+def _adjust_resources_path_by_strip_prefix(resource, resource_strip_prefix):
+    path = resource.path
+    prefix = _paths.join(resource.owner.workspace_root, resource_strip_prefix)
```
ittaiz: Nice, thanks!

test/src/main/scala/scalarules/test/resources/strip/BUILD
simuons: @ittaiz you mean something like this https://github.com/bazelbuild/rules_scala/pull/944/commits/edafc20c260c43cebfdd8194ce4bb47117b80de0
I'm really not sure I did good job integrating test into the build

ittaiz:
I'll merge when it's green

simuons:
Thank you @ittaiz

Issue #917, opened 2020-01-15
Phase final necessity and ability for phases to expose providers for downstream targets


ittaiz:
@borkaehw @andyscott while working on #916 I started wondering about the motivation behind `run_phases` returning the global provider and then having the rules call `.final`?
The cons I see are that every rule needs to "remember" to call `.final` and additionally custom external phases cannot expose providers for downstream targets.
Is the latter by design?
   
I have a hunch that if we change the phase return value contract to be 2 dicts where one is of "internal providers" (similar to today but more typed and explicit) and a dict of "external providers" and then `run_phases` just returns the values of this dict then rules can just return the output from `run_phases` (one less thing to remember) and additionally we'll be allowing phases to expose providers (currently this isn't an option). WDYT?
 I'm saying dict to allow phases to override providers of previous phases.

andyscott:
This seems reasonable to me. I'm almost always in favor of phases being able to cleanly customize what's being returned by the rules.

PR #946, opened 2020-01-21
Custom phases expose custom providers

ittaiz:
### Description
Custom phases expose custom providers  

While doing it I found and fixed a bug in phases adjustments of first/last (they were duplicated).
I tried to send a fix for this separately but since actions are de-duped and the info passed between phases via structs is also de-duped I couldn't get a failing test. I could do it with `print` statements but I think that's too hacky.
I stumbled on to the issue in this change because Bazel forbids multiple providers of the same type. 
Btw, main implication (apart from providers) of this specific bug is a perf hit for analysis case (when adjusting phases).
Additionally this mandates bazel 2.0.0 as they have some kind of bug with providers before that (don't have an issue to ref but couldn't get the build to pass with lower bazel version)

### Motivation
Allow custom phases to expose custom providers. At the end of the day providers are a basic and powerful tool for sharing info between different rules (for example if someone writes a publishing rule internally and need metadata from our rules which only they care about).

Note:
1. This lays the ground to the removal of `phase_final`. I didn't do it here because I wanted to keep it small and will follow-up in a separate PR.
2. The mutable array of providers is so-so and what I really want is to allow for phases to also override providers of previous phases. I'll do it in a follow-up PR (again to keep it small) and then I'll change the array to be a dict.

cc @borkaehw @andyscott

### DIFF
--- a/scala/private/phases/api.bzl
+++ b/scala/private/phases/api.bzl
@@ -31,10 +31,13 @@ def _adjust_phases(phases, adjustments):
                 if needle == peer_name:
                     if relation in ["-", "before"]:
                         phases.insert(idx, (phase_name, phase_function))
+                        break
                     elif relation in ["+", "after"]:
                         phases.insert(idx + 1, (phase_name, phase_function))
+                        break
                     elif relation in ["=", "replace"]:
                         phases[idx] = (phase_name, phase_function)
+                        break
     return phases
 
 # Execute phases
--- a/scala/private/phases/api.bzl
+++ b/scala/private/phases/api.bzl
@@ -63,7 +63,7 @@ def run_phases(ctx, builtin_customizable_phases, fixed_phase):
     # A placeholder for data shared with later phases
     global_provider = {}
     current_provider = struct(**global_provider)
-    rule_providers = []
+    acculmulated_external_providers = []
     for (name, function) in adjusted_phases + [fixed_phase]:
         # Run a phase
         new_provider = function(ctx, current_provider)
@@ -71,13 +71,13 @@ def run_phases(ctx, builtin_customizable_phases, fixed_phase):
         # If a phase returns data, append it to global_provider
         # for later phases to access
         if new_provider != None:
-            if (hasattr(new_provider, "rule_providers")):
-                rule_providers.extend(new_provider.rule_providers)
+            if (hasattr(new_provider, "external_providers")):
+                acculmulated_external_providers.extend(new_provider.external_providers)
             global_provider[name] = new_provider
             current_provider = struct(**global_provider)
 
     # The final return of rules implementation
-    return rule_providers + current_provider.final
+    return acculmulated_external_providers + current_provider.final
 
 # A method to pass in phase provider
 def extras_phases(extras):
--- a/scala/private/phases/api.bzl
+++ b/scala/private/phases/api.bzl
@@ -22,18 +22,19 @@ def _adjust_phases(phases, adjustments):
     # phase_name: the name of the new phase, also used to access phase information
     # phase_function: the function of the new phase
     for (relation, peer_name, phase_name, phase_function) in adjustments:
-        for idx, (needle, _) in enumerate(phases):
-            if relation in ["^", "first"]:
-                phases.insert(0, (phase_name, phase_function))
-            elif relation in ["$", "last"]:
-                phases.append((phase_name, phase_function))
-            elif needle == peer_name:
-                if relation in ["-", "before"]:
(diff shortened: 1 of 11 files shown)


scala/private/phases/api.bzl
```diff
@@ -22,18 +22,19 @@ def _adjust_phases(phases, adjustments):
...
-            elif relation in ["$", "last"]:
-                phases.append((phase_name, phase_function))
-            elif needle == peer_name:
-                if relation in ["-", "before"]:
-                    phases.insert(idx, (phase_name, phase_function))
-                elif relation in ["+", "after"]:
-                    phases.insert(idx + 1, (phase_name, phase_function))
-                elif relation in ["=", "replace"]:
-                    phases[idx] = (phase_name, phase_function)
+        if relation in ["^", "first"]:
```
borkaehw: Even though I still can't think of a failing case of original logic, the fix looks more reasonable. Thanks for catching it!

scala/private/phases/api.bzl
```diff
@@ -22,18 +22,19 @@ def _adjust_phases(phases, adjustments):
...
+            phases.append((phase_name, phase_function))
+        else:
+            for idx, (needle, _) in enumerate(phases):
+                if needle == peer_name:
+                    if relation in ["-", "before"]:
+                        phases.insert(idx, (phase_name, phase_function))
+                    elif relation in ["+", "after"]:
+                        phases.insert(idx + 1, (phase_name, phase_function))
+                    elif relation in ["=", "replace"]:
+                        phases[idx] = (phase_name, phase_function)
```
borkaehw: How about adding a `break` in `if` statement? No need to iterate further, right?

scala/private/phases/api.bzl
```diff
@@ -59,18 +60,21 @@ def run_phases(ctx, builtin_customizable_phases, fixed_phase):
...
+    rule_providers = []
     for (name, function) in adjusted_phases + [fixed_phase]:
         # Run a phase
         new_provider = function(ctx, current_provider)
 
         # If a phase returns data, append it to global_provider
         # for later phases to access
         if new_provider != None:
+            if (hasattr(new_provider, "rule_providers")):
+                rule_providers.extend(new_provider.rule_providers)
```
borkaehw: `rule_providers = []` and `new_provider.rule_providers` have the same name. Could we have at least a little bit of difference? Like maybe `new_provider.rule_provider`.

test/phase/providers/BUILD.bazel
```diff
@@ -0,0 +1,15 @@
+load(":phase_providers_expose.bzl", "phase_expose_provider_singleton", "rule_that_needs_custom_provider", "scala_library_that_exposes_custom_provider")
+
+scala_library_that_exposes_custom_provider(
+    name = "scala_library_that_exposes_custom_provider",
+)
+
+rule_that_needs_custom_provider(
```
borkaehw: Are we just checking we can build this target?

ittaiz: Yes. The build failed before because the rule requires the custom provider to be passed and so bazel failed the build. You can take the first commit and see the failure.

scala/private/phases/api.bzl
ittaiz: I actually thought about changing the strict field name to be `inter_rule_providers` or `external_providers` and then the next phase will be to have phases also export `internal_providers` between phases (I think phases_providers is overloaded). I’ll also change the temp variable to something like what you suggested

scala/private/phases/api.bzl
ittaiz: 👍🏽 I’ll also add (in a separate PR?) handling of what happens when someone sends a non existent phase (for example if they have a typo). Right now we will silently ignore. I think we need to explicitly fail. I’ll add that as well but I think separately to keep it clean. Wdyt?

ittaiz:
It’s the first part.
Second part is inlining phase final and then allowing phases to override providers from previous phases. They might come in the same part if they can’t be separate. Need to see

scala/private/phases/api.bzl
borkaehw: Sounds like a good idea.

scala/private/phases/api.bzl
borkaehw: Thanks, it looks good.

borkaehw:
LGTM. Do we want to close #917 by merging this?

ittaiz:
I’ll close it after I send the second and third part which will remove phase final and allow overrides. Those PR should come out tonight or tomorrow

PR #947, opened 2020-01-22
final phase is inlined

ittaiz:
### Description
final phase is inlined

all phases are equal
DefaultInfo is returned by a default_info phase
Other external providers are passed by their respective phases

advances #917 

### Motivation
Symmetry, less edge cases in design


ittaiz:
cc @borkaehw 
also note that after I will finish with the third PR that allows phases to override providers of previous phases then I'll update the docs even a bit more.

ittaiz:
I also the duplication and thought to keep it until I move the internal dependencies between phases to be provider based as well (I'm planning that next and as the last big thing left for phases API).
My general idea is that instead of passing `p` to a phase I want to pass `internal_providers` dict and `external_providers` dict.
This way phases can depend on either of them.
WDYT?
Another thought is to just change the phases signature to accept `external_providers` in addition and then to solve the problem you're surfacing here and separate between the two changes.
WDYT about that? If the latter is good I'd prefer doing it after we merge this and #948 (but I'm not hellbent about it)

scala/private/phases/phase_default_info.bzl
```diff
@@ -0,0 +1,36 @@
+#
+# PHASE: default_info
+#
+# DOCUMENT THIS
+#
+def phase_binary_default_info(ctx, p):
```
andyscott: Same comment as in https://github.com/bazelbuild/rules_scala/pull/948:

> Could we name these phase_default_info_binary, phase_default_info_library, and phase_default_info_scalatest instead? I think it really improves code navigability when the file name is the prefix for all exported functions.

ittaiz:
@borkaehw waiting for your input https://github.com/bazelbuild/rules_scala/pull/947#issuecomment-577371039

borkaehw:
Sorry, I didn't see that yesterday. Yeah, I think either way looks good to me. Let's merge this PR.

PR #949, opened 2020-01-22
Bundle input protos together with generated/compiled classes

simuons:
### Description
Bundle input protos together with generated/compiled classes

### Motivation
We have lots of unresolved proto imports in intellij and this happens because intellij's protobuf plugin cannot find imported proto files on the classpath.

Bundling proto files with compiled code doesn't look like unusual practice. I mean jars produced by google carries both class files and proto files for example [protobuf-java](https://repo1.maven.org/maven2/com/google/protobuf/protobuf-java/3.11.1/)

Implementation might look a bit aggressive on resource prefix handling but I had two options ie touch only `scala/private/rule_impls.bzl` or align it with `scala/private/phases/phase_compile.bzl` and I chose later. I could do it in separate PR but I'm afraid it would look like a random refactoring by a random guy without clear purpose.

I'm not sure there is a real need for so many distinct proto_library in tests but I wasn't sure how each combination of `strip_import_prefix` and `import_prefix` work.

`PackProtosTest` could be more strict in the sense that it could check if particular proto entry exist in particular jar. On the other hand we care only that those files are on the classpath.

### DIFF
--- a/scala/private/phases/phase_compile.bzl
+++ b/scala/private/phases/phase_compile.bzl
@@ -3,7 +3,6 @@
 #
 # DOCUMENT THIS
 #
-load("@bazel_skylib//lib:paths.bzl", _paths = "paths")
 load("@bazel_skylib//lib:dicts.bzl", "dicts")
 load("@bazel_tools//tools/jdk:toolchain_utils.bzl", "find_java_runtime_toolchain", "find_java_toolchain")
 load(
--- a/scala/private/phases/phase_compile.bzl
+++ b/scala/private/phases/phase_compile.bzl
@@ -11,10 +11,10 @@ load(
 )
 load(
     "@io_bazel_rules_scala//scala/private:rule_impls.bzl",
-    _adjust_resources_path = "adjust_resources_path",
     _compile_scala = "compile_scala",
     _expand_location = "expand_location",
 )
+load(":resources.bzl", _resource_paths = "paths")
 
 _java_extension = ".java"
 
@@ -491,18 +491,9 @@ def _try_to_compile_java_jar(
     )
 
 def _add_resources_cmd(ctx):
-    res_cmd = []
-    for f in ctx.files.resources:
-        target_path = _adjust_resources_path(
-            f,
-            ctx.attr.resource_strip_prefix,
-        )
-        line = "{target_path}={res_path}\n".format(
-            target_path = target_path,
-            res_path = f.path,
-        )
-        res_cmd.extend([line])
-    return "".join(res_cmd)
+    paths = _resource_paths(ctx.files.resources, ctx.attr.resource_strip_prefix)
+    lines = ["{target}={source}\n".format(target = p[0], source = p[1]) for p in paths]
+    return "".join(lines)
 
 def _collect_java_providers_of(deps):
     providers = []
--- a/scala/private/phases/phase_compile.bzl
+++ b/scala/private/phases/phase_compile.bzl
@@ -498,8 +498,6 @@ def _add_resources_cmd(ctx):
             f,
             ctx.attr.resource_strip_prefix,
         )
-        if target_path[0] == "/":
-            target_path = target_path[1:]
         line = "{target_path}={res_path}\n".format(
             target_path = target_path,
             res_path = f.path,
--- a/scala/private/phases/phase_compile.bzl
+++ b/scala/private/phases/phase_compile.bzl
@@ -12,7 +12,7 @@ load(
 )
 load(
     "@io_bazel_rules_scala//scala/private:rule_impls.bzl",
-    _adjust_resources_path_by_default_prefixes = "adjust_resources_path_by_default_prefixes",
+    _adjust_resources_path = "adjust_resources_path",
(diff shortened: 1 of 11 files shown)


ittaiz:
1. I've re-run the job that failed due to download issues.
2. Have you verified this solves the IJ proto red code? To be sure we're talking about issues editing proto files using other proto files, right?

ittaiz:
I skimmed it and I don't understand something- I see a lot of code deleted from the scalacprocessor, is it because you encode the prefix in the input and so the processor doesn't need to do the work?
Re rule_impls.bzl- my aim is actually to delete this file eventually so adding more code into it is less than optimal. 
3 options:
1. This is specific to compilation- keep it in phase_compile
2. This is generic resource handling- should it sit in skylib or rules_java
3. This is generic in the sense that it is not only about compilation but about paths in general but closely coupled to decisions we've made in rules_scala- first we should have a good reason for this, second maybe we should have a adjust_resources.bzl file or something? Maybe it should be a phase in addition to a public function? Not sure

simuons:
> To be sure we're talking about issues editing proto files using other proto files, right?

Yes. When editing proto file which imports another proto then that import is red/unresolved.

> Have you verified this solves the IJ proto red code?

Yes. I did check it locally and imports did resolve.

> I see a lot of code deleted from the scalacprocessor, is it because you encode the prefix in the input and so the processor doesn't need to do the work?

Yes. Prefixes are now fully handled in `rule_impls.bzl` while `ScalacProcessor` just receives a mapping where to copy resources.
I've noticed that prefixes were handled in two code paths:
1. `phase_compile.bzl` - fully handled prefixes and passed `target=path` to other routines
2. `rules_impl.bzl` - handled standard prefixes and `ScalacProcessor` handled user supplied prefixes

Since `ScalacProcessor` doesn't have all the details to correctly handle prefixes and it's not used in all cases I've decided to handle prefixes in bzl file so the code was removed.

> rule_impls.bzl- my aim is actually to delete this file eventually so adding more code into it is less than optimal

Since `phase_compile.bzl` was importing functions from `rules_impl.bzl` my natural choice was to add that code to `rules_impl.bzl`. I'm not aware that `rules_impl.bzl` is going away. But sure I can move the code.

> 3 options

1. I would probably go with this option
2. Not sure I understand what you mean a) such functionality already exists in skylyb/rules_java b) I should propose these changes there
3. I'm not familiar with phases but I think we could keep these changes in `phase_compile.bzl` and later decide where to move it

ittaiz:
let's keep it as is since the other usage of `adjust_resources_path_by_default_prefixes` is inside `rule_impls` and that is the notorious `compile_scala` which can't be removed yet because not everything has moved to phases (I wonder how scalapb as an aspect will work with phases).

Super super excited about this change!!!!!!!

I'll review it now

ittaiz:
Any chance you can say a few more words about what were the inputs to the scalac action before and what are they now?

ittaiz:
Also,
Am I understanding correctly that the move is somewhat moving from an imperative "strip this prefix" to a more declarative "target=destination"? I might be wrong in my understanding

ittaiz:
Reviewed the prod code. Looks ok. WDYT about moving it to `resources.bzl` or `resources_adjustment.bzl`? It will help in the effort of killing `rule_impls.bzl` god file

simuons:
> about what were the inputs to the scalac action before and what are they now?

I assume you are asking about `ScalacProcessor` from `rule_impls.bzl`
Before:
* ResourceDests
* ResourceSrcs
* ResourceShortPaths
* ResourceStripPrefix

After:
* ResourceDests
* ResourceSrcs
* ResourceShortPaths

Actually `ResourceShortPaths` is not used and I could remove that as well

`phase_compile.bzl` same as before list of `target=source` pairs

> Am I understanding correctly that the move is somewhat moving from an imperative "strip this prefix" to a more declarative "target=destination"? I might be wrong in my understanding

Correct

> WDYT about moving it to resources.bzl or resources_adjustment.bzl? It will help in the effort of killing rule_impls.bzl god file

Sure. I can move to `resources.bzl` and build a function which produces list of `target=destination` (that would be a bit bigger refactoring I think, need to try) or just move code as is to `resources.bzl`

ittaiz:
Re ResourceShortPaths- maybe we should remove it then?
Re resources.bzl- sounds good to me

simuons:
Extracted `resources.bzl` https://github.com/bazelbuild/rules_scala/pull/949/commits/820f5dfefc2c1b893e83157b2f34ffeb0b2a50f1 now pr looks more about resource handling than what's written in title but maybe it's ok

ittaiz:
Thanks! I’ll review it later.
Did you remove passing ResourceShortPaths and from CompileOptions?

simuons:
Yes I did https://github.com/bazelbuild/rules_scala/pull/949/files#diff-d68dd15f790bd2a815ab04d3d0a320d8L166-L167 and https://github.com/bazelbuild/rules_scala/pull/949/files#diff-e70e84eddad2c764ddb01bf651a0b895L73

simuons:
Thank you!

PR #955, opened 2020-01-26
Add defs.bzl with core rules and notice about evolving APIs

andyscott:
### Description

Adds a `defs.bzl` which exports a copy of the core rules. Additionally, this adds minimal tests for the copied rules.

I propose that for the time being (perhaps the next year?) we allow breaking changes to rules in `defs.bzl` while keeping `scala.bzl` stable and documenting it as the supported API. Once we are happy with the API of `deps.bzl`, we can freeze it and promote it as the main API of rules_scala.

### Motivation

I'd like to explore changes to the core rules. Most notably, I'd like a new `scala_test` rule to enable support for any testing framework in the same manner as SBT. Having a parallel set of rules using the same phase backend allows for exploration without breaking any users of the existing rules.


scala/defs.bzl
```diff
@@ -0,0 +1,40 @@
...
+load(
+    "@io_bazel_rules_scala//scala/private:rules/scala_library.bzl",
+    _make_scala_library = "make_scala_library",
+)
+load(
+    "@io_bazel_rules_scala//scala/private:rules/scala_test.bzl",
+    _make_scala_test = "make_scala_test",
+)
+
+make_scala_library = _make_scala_library
```
andyscott: I believe these 6 APIs should be exported.

Right now it's the same exact thing exported by `scala.bzl` + the advance usage helpers, but they will likely deviate over the next year.

ittaiz:
Why is this better than the separation we already have with the advanced usage and scala.bzl?
The con I see here is that this now adds a decision point for the regular user which needs to read the docs of defs.bzl and know it’s not for them.

andyscott:
I'll make an argument for having an unstable set of rules alongside the stable rules:

- Having a parallel set of unstable rules gives us a place consolidate new development. It'll be easier to compare the unstable rules to the stable rules, and this will allow us to have meaningful conversations about how when and why to make breaking changes to the stable API. I think of it as a staging area.

- Additionally, while it's easy to add new phases I suspect it will be clunky to constantly leverage the advanced usage helpers to customize the existing rules for new (and sometimes API breaking) functionality. A set of unstable rules may help us avoid lots of ad hoc rule customizations in our tests, which may alleviate some maintenance burden.


I'm not really concerned too much with where the stable and unstable rules are located. I chose `defs.bzl` because it seems to be idiomatic spot to expose rules. Perhaps our stable rules should be exposted via `defs.bzl` and the unstable rules should be exposed somewhere else.

ittaiz:
I'm ok with `@rules_scala//scala/unstable:foo.bzl`.
My main concern is that the unstable rules will become the hot thing and the stable ones will somewhat rot until we merge it back in. Last time we did something like that it took us 2 (3?) years to merge back (who said higherkindness).
I'm not strictly opposed to this but I'd like us to be very vigilant about these experimentations being short and being merged in the mainline (unstable can be thought of like a git branch) so that we spread the value to everyone and minimize maintenance overhead.

Re `scala.bzl` vs `defs.bzl`- who is using this pattern? for us to change the public API of rules_scala there needs to be a very good reason. The API of the core rules hasn't been changed in a few years and I think this has a lot of value for users which can upgrade and for the most part things just work and APIs haven't changed.

andyscott:
https://docs.bazel.build/versions/master/skylark/deploying.html outlines that `defs.bzl` should be used as an entry point for rules. You can see a lot of the Google maintained rules have a file with this name already.

ittaiz:
Thanks and thanks for the ref.
What about the other (more important) concern? We need to be very clear about small experiments which get merged into mainline or discarded.
Wdyt?

andyscott:
I am in favor of having experiments and ultimately merging them in or getting rid of them.

I'd proposed an experiment should be merged in ASAP after it has proven value. Some experiments will bring breaking changes to the rules, so we will need to be cautious about how and when we make those merges.

ittaiz:
Great.
There is the question of what does "proven value" mean but I think that is indeed something best left for concrete discussions.
I'll merge this when green.

andyscott:
Yeah, it's hard to say. I think "proven" currently means that I can convince you that it's useful :)

Issue #769, opened 2019-06-14
Consider splitting off rules core to building and and testing scala from those that arent


softprops:
We've been bit a few times on upgrade paths due to transient dependencies of rules within this repo I feel are not core to scala support.

Specifically protobuf dependencies are always hard for us to manage. 

This could be remedied by keeping rules like scala library in a repo separate from aux rules like thrift, scalapb etc.

I understand this introduces a burden on maintainers but the reward for users feels worth it.

johnynek:
You can feel free to make a port of this code and start separate rules, and you may be right that they can be independently run well now. Since stripe has done almost all the work on thrift and did the rewrite of the proto rules to fix the issues (recompilation, duplicate classes), I don’t think we would be that motivated to take on more work. Our scala team has 3 people and they are very taxed as is.

We need more help, and our current situation isn’t maintainable. Any contributions you can make would be welcome. In the mean time, anyone is free to use the code but I don’t see much, of any, bandwidth for external requests.

This is currently best understood as a collaboration between Wix and Stripe to help our builds, and if we can help others enough to motivate them to contribute, that would be great too.

ittaiz:
+1 to Oscar's general statement.
@softprops Is it accurate to say that concretely your main pains are dependencies?
If so I think the main contribution that I'd appreciate is splitting off our "deps" pattern to separate toolchains. This will advance the rules and decouple you from the parts you don't care about.
How does that sound?

ittaiz:
Btw,
I'm working (very lazily) on moving the junit/specs2 test discovery out of this repo. Both to make it more focused and to allow java users to use it.

softprops:
Protobuf in particular yes. Thanks for the direction on the separate toolchains vs separate repo approach. I had not yet thought of that

ittaiz:
Sure.
In general Oscar and I agreed in the past about wanting to split everything
to smaller toolchain.
Didn’t get around to it :(

beala-stripe:
Last I checked, the protobufs dep is used by more than just the proto rules: https://github.com/bazelbuild/rules_scala/issues/726#issuecomment-480041245

Maybe it could be factored out, but by the looks of that bazel query, it would be non-trivial.

ittaiz:
Ok, not sure why this happened but I’m really in favor of splitting this.
@softprops any chance you want to work on identifying why we have these
deps?

johnynek:
We have the deps to proto because the bazel worker protocol uses protobuf.

[...]

PR #956, opened 2020-01-26
use worker proto file instead of precompiled java

andyscott:
### Description

Use the worker proto file instead of the pregenerated java.

### Motivation

It seems a bit cleaner to me to depend on the proto instead of the prebuilt Java. The proto file is also easier to update. The only drawback is there's an overhead to bootstrapping protobuf, but pretty much every Bazel project does that anyway, so the cost is shared.


ittaiz:
Interesting. Definitely more manageable but TBH I'm somewhat concerned about this change due to the perf implications.
I've heard many related complaints about the protoc dependency and it's overhead internally and externally (https://github.com/bazelbuild/rules_scala/issues/769, https://bazelbuild.slack.com/archives/CDCKJ2KFZ/p1579853509003000).
@simuons @sha1n wdyt?

simuons:
@ittaiz well if you ask me this looks like a reasonable thing to do in current state of rules_scala. The protoc compilation "issue" exists even without this change. Once thats issue will be resolved code in this PR would change as well.

ittaiz:
My main concern is that it will make it even harder to change our direction. If we want to change it then maybe we shouldn't continue to advance in the other direction.
Another smaller concern is the concrete increase in build times since we see internally and externally many re-compilations of proto outputs.

ittaiz:
I’m merging though there is a chance we’ll need to revert it if we get too many backlashes.

PR #958, opened 2020-01-27
unify all of the default info phases implementations

andyscott:
### Description

Unifies all of the default_info phases into a single implementation.

### Motivation

This makes it relatively straightforward for any phase to augment the files and runfiles DefaultInfo output. This is extremely useful when adding new phases for special functionality. In my case, I need functionality to support a new test discovery phase that writes a list of tests to a text file output.


scala/private/phases/phase_runfiles.bzl
```diff
@@ -38,7 +38,7 @@ def _phase_runfiles_default(ctx, p, _args = struct()):
     return _phase_runfiles(
         ctx,
         _args.transitive_files if hasattr(_args, "transitive_files") else depset(
-            [p.declare_executable, p.java_wrapper] + ctx.files._java_runtime,
```
andyscott: Last I check, the DefaultInfo's executable field is assumed part of the overall runfiles output. So this wasn't needed here.

scala/private/phases/phase_runfiles.bzl
```diff
@@ -24,7 +24,7 @@ def phase_runfiles_scalatest(ctx, p):
 
     args = struct(
         transitive_files = depset(
-            [p.declare_executable, p.java_wrapper] + ctx.files._java_runtime + runfiles_ext,
```
andyscott: Same as the below comment.

ittaiz:
Thanks. I didn't go over the few important details this PR has but in general it looks like this ended up being about two things: Unification and Dependency inversion (instead of default_info taking fields from phases you're ignoring the phase and using the field name).
The unification looks good in general (will take a look at details later) and it was an annoyance when I looked at it so nice one :)
Re dependency inversion- My plan (which I shared in a few PRs/Issues) is actually to kill `p` in favor of a new field `internal_providers` which will be a `dict` with Bazel providers which phases expose.
The fact that phases expose providers will allow clearer understanding of their semi-public API (right now you need to go and rummage through the phases to see what exactly goes on and also when the API changes it's more implicitly). In addition providers with dict can allow overrides and customizations of following phases like the `external_providers` is doing.
One thing I didn't get to implement with `external_providers` is to actually pass `external_providers` as an input so phases can copy and not only create completely new ones.

To summarize I'd love it if we could take this PR and change it so that actually we have 3 new internal providers (`FilesToExpose`,`RunFilesToExpose`, `ExecutableToExpose`) or maybe 1 new internal provider (`DefaultInfoToExpose`) that the phases take as input and can return.
Then the default_info phase can just take that from the dict.

WDYT? If it's not 100% clear I can send this PR (with the changes here) to clarify how I see this pattern.

scala/private/phases/phase_compile.bzl
```diff
@@ -165,6 +165,7 @@ def _phase_compile(
         class_jar = out.class_jar,
         coverage = out.coverage.external,
         full_jars = out.full_jars,
+        files = depset(out.full_jars),
```
ittaiz: having a depset here is probably not the best idea since depsets have their cost and especially since you want to customize this in other phases then you might have depsets which are nested for no reason.
Maybe pass array?

scala/private/phases/phase_runfiles.bzl
ittaiz: Do you think we can validate it somehow? Do you think tests should have been broken? Maybe it's good enough, not sure

andyscott: IMHO it's good enough because otherwise most of the tests covering binary and test rules would have broken! Additionally, I wouldn't have been able to iterate on #959 which incorporates these changes.

If you want, I can test this against our codebase at Stripe to be extra safe.

scala/private/phases/phase_compile.bzl
andyscott: I'm not sure. In some case we may wind up passing along existing depsets, but in many others we have just a few files that need to be added in. It's not clear to me which is more common and the impact to analysis time.

andyscott:
I'm off to bed. I'll respond to the overall comment tomorrow.

scala/private/phases/phase_runfiles.bzl
ittaiz: I think that’s a good idea

scala/private/phases/phase_compile.bzl
ittaiz: Are you replying in general or about files and runfiles? What we have now in these cases are arrays. If you’ll want to override these then you’ll need to have a nested depset.
In general I think it will depend on the case.

andyscott:
I haven't been able to check this against Stripe's codebase due to the pending fix in #963.

I'd love to see this merged to get the consolidation of the default info methods. We can then have a separate discussion about cleaning up how we pass information between phases.

But since we're here, I'm not convinced that providers are the way to go for passing phase information. Providers are typically used for passing information across rule boundaries: `rule implementation a -- target graph --> rule implementation b`. Here we are passing information between functions within the same rule implementation, i.e. `phase 1 --> phase 2, all within rule implementation c`. Perhaps we should just use well structured structs or dictionaries?

ittaiz:
This is adding tech debt (passing info via `p`) which I've been very vocal against for a few months now. We've been postponing this and keep saying that we don't want to block efforts but this can't hold a lot further. I voiced my concern about the contract between phases being too vague in a few places, namely in the big phases PR.
This PR actually hits this issue and then works around it via iteration.

I can merge this (even though I really don't like it) but then we need to stop and agree on the pattern (iteration [I don't like it but am open to hear arguments in favor of it], providers or structs/dictionaries).
WDYT?
(also see https://github.com/bazelbuild/rules_scala/issues/965 for the continuation of the discussion)

andyscott:
The tech debt is already there, this just changes it a bit 😄.

I agree with you on narrowing down the contract between phases. Part of
that process is establishing what we need phases to do, and this PR
demonstrates a need for a generic way to accumulate values for returned
providers. Basically we need some monoial computation to thread through the
phases.

I'll follow up on the issue, and perhaps do a few proof of concepts. I'd
prefer we merge this though as ties into some other work, and I like the
cleanup of the default info methods.

ittaiz:
Ok, let’s merge and then see how we agree on something on the issue before we do any more changes (I was about to send a series of PRs to move us to providers so I’m blocked until we decide).
Since you need to rebase and resolve conflicts, Wdyt about what I said about the depset? In these concrete instances it sounds like an array is the better solution

andyscott:
I tried converting the depsets to an array but we already have a lot of existing code managing the `transitive_files` in the runfiles phase as a depset. We'll have to call an extra `to_list` or do more extensive rewriting to use arrays in this PR.

smparkes:
>
> I'm not convinced that providers are the way to go for passing phase
> information. Providers are typically used for passing information across
> rule boundaries: rule implementation a -- target graph --> rule
> implementation b. Here we are passing information between functions
> within the same rule implementation, i.e. phase 1 --> phase 2, all within
> rule implementation c. Perhaps we should just use well structured structs
> or dictionaries?
>
I'm consciously a bystander on the phase stuff. I'm glad you all got this
(however it falls out).

FWIW, I did have similar feelings to the comment above.

This is what I understand (you all may already know this):

I got the impression providers were intended for "inter-target"
communication: rule calls return provider instances and those instances get
attached to targets created by the rule call which (1) can be type-enforced
by rules depending on the target (I think) parameters to label attrs and
(2) can be read by things that depend on those targets in a way that the
provider type becomes the attribute/key.

This doesn't seem to relate to phases?

Putting that special behavior aside, I did explore what behavior providers
have beyond what you get with structs. Once they're created, it doesn't
look like much. They `type` to "struct" and have `to_json` like structs do.

They do have the `provider` declaration for documentation. And they do seem
to enforce a little bit at construction: you can't add a field that isn't
declared. You can leave out fields: it doesn't do anything that a struct
constructor without that field wouldn't.

I think I'll avoid providers when I'm not using them in the rule impl/dep
role. I find the value-add low compared to the confusion it causes me
("what target/rule is this related to?")

I will say that I find the current factoring of the scalac provider stuff a
bit confusing.

`_declare_scalac_provider` runs a rule to create a target which does happen
to have the provider, but what is `scalac_default` itself? At this point,
I'm pretty comfortable saying this should be a toolchain, though I think it
needs to be factored differently. Handling multiple version toolchains will
drive this because the current global naming can't support that.

ittaiz:
@smparkes can you move this comment to #965 ? I have a reply but I’d like to have a clear discussion page about it

smparkes:
@ittaiz done

PR #953, opened 2020-01-23
Rename phases logically

borkaehw:
### Description

Refactor the phase naming convention to `phase_<FUNCTION>_<RULE>`.

### Motivation

It improves code navigability.

PR #960, opened 2020-01-27
Fix doc for new naming convention

borkaehw:
### Description

Fix doc on naming convention.

### Motivation

It should have been done with #953


borkaehw:
It happens when we have rule specific phases, I don't think we have it in the tests (since rule specific phases are not intended).

ittaiz:
ok, if you looked I trust you

PR #963, opened 2020-01-27
fix proto regression

andyscott:
### Description

Cross repository protobuf compilation seems to fail due to a regression in https://github.com/bazelbuild/rules_scala/pull/949. This PR attempts to fix it.

Note that the commit history shows:
- ca9e852 adds a test before https://github.com/bazelbuild/rules_scala/pull/949. The test passes. It's also exactly the same thing we're doing at Stripe to pull in some of the Bazel proto files.
- f816f84 merges https://github.com/bazelbuild/rules_scala/pull/949, causing the test to fail
- f3a1005 hopefully fixes the regression


andyscott:
Not sure, but the CI failure seems spurious.

andyscott:
`java.io.IOException: Error downloading [https://repo.maven.apache.org/maven2/com/github/jnr/jffi/1.2.17/jffi-1.2.17-native.jar] to /home/travis/.cache/bazel/_bazel_travis/7f7e07d167ded9af0372bc305ceff99e/external/com_github_jnr_jffi_native/jffi-1.2.17-native.jar: GET returned 403 Forbidden
`

@ittaiz any ideas about the above error? It seems unrelated to these changes.

ittaiz:
The failure might be related to https://travis-ci.community/t/continuous-maven-repo-403/3908
Maybe we should just merge and fix it in a separate PR.

andyscott:
Honestly, I didn't have a ton of time to get this done, so I just coped whatever we had at Stripe!

I've gone back and minimized it using a local repository. The commits don't show it, but if you undo the "fix" commit against the most recent simplifications then the original breaking test starts to break again.

Issue #962, opened 2020-01-27
Multiple scala version builds within the same bazel run


smparkes:
# PRs

There are a number of PRs in flight. Because they interact, they're sometimes based on one another so a diff against master will show all changes up to that point. That means it's probably easiest to review from in order.

#### Core PRs: these are cleanup/refactoring PRs for core that are blockers for multiscala

* [x] #980: use bind to remove loader-specific labels in dependencies ([smparkes:bind-dependencies](https://github.com/smparkes/rules_scala/tree/bind-dependencies))
* [ ] #978: Factor toolchains ([smparkes:factor-toolchains](https://github.com/smparkes/rules_scala/tree/factor-toolchains))
* [ ] #981: Factor scalac provider ([smparkes:factor-scalac-provider](https://github.com/smparkes/rules_scala/tree/factor-scalac-provider))
* [ ] #984: Factor scala test toolchain ([smparkes:factor-test-toolchain](https://github.com/smparkes/rules_scala/tree/factor-test-toolchain))

#### Multiscala PRs: these actually implement multiscala support. They generally require the PRs above.

* [ ] #964: multiscala configuration ([smparkes:configuration](https://github.com/smparkes/rules_scala/tree/configuration))
* [ ] #983: Skeleton for multitarget ([smparkes:multitarget](https://github.com/smparkes/rules_scala/tree/multitarget))

# tl;dr

I'm creating a new issue for this.

Many exist and I'm happy to close this in favor of another but there are (see below) so many with some amount of overlap and various comment history, personally I prefer to start with a clean slate.

I'm actively working on this.

I'd like to get buy-in early that we want to do this and on my current approach. If there is buy-in, I'd like to get feedback as I go. There are lots of choices and I'd love to get feedback/reviews early so
there's not a lot of rework.

If there's not sufficient consensus, I'm happy to close this (we don't need yet another issue) and I'll do the work independently.

# Requirements:

* Build/test targets against multiple versions of scala in the same bazel invocation
* Minimal user overhead/incompatibility if they are using a single scala version
* Maintain current API compatibility
* Opt-in until stable

# Stretch goals:

* Don't require users to enumerate, manually or with starlark, targets, e.g., one `scala_library` call will, by default, build (suffixed) targets (optional alias for default scala version)
* Some support for 2.13, e.g., support version-specific source code variants

# Key implementation choices

* Use configured toolchains rather than label dependencies
* Shared configuration available at loading phase via a repository constructed via a repository rule

[...]

PR #964, opened 2020-01-27
multiscala configuration

smparkes:
### Description

This is the first PR for #962 . It creates/uses the configuration structure necessary to configure multiple scala toolchains but at this point only uses the default toolchain.

### Changes to `multiscala`
* create `unstable/multiscala`
* combine user configuration and default configuration to create `@io_bazel_rules_scala_configuration` repo
* `maven_install` `rules_scala` dependencies based on `@io_bazel_rules_scala_configuration` 
* create `bind` aliases for `rules_scala` dependencies when there is a default scala version
* create `scala_toolchain` for configured versions
* register `scala_toolchain` if there is a configured default version
* create `unstable/multiscala/private/example` which can be run with `bazel build :all && bazel run :all`

### Changes to stable code
* add TemplateVariableInfo provider to `scala_toolchain`. Required when used as the value for the standard `toolchains` attr
* create and use a few `bind`ings so that they can also be configured in `multiscala`. Although I think long term we may want to remove all bindings, for now they're key to making `multiscala` opt-in.

### Motivation
First step in being able to configure multiple toolchains based on a configuration with multiple scala versions.

This creates the toolchains but doesn't, at this point, use them. That will come in a subsequent PR.


andyscott:
I am generally very much in favor of this functionality.

That being said, I believe providers [0]-- not toolchains-- are the right long term approach for supporting this and my fear is that merging this complicates some of the existing cleanup and feature work that's already underway.

I haven't had a chance to lay out my arguments for ditching toolchains in favor of providers, so perhaps we need to finally kick off that discussion. WDYT @ittaiz?

---

[0] I've explored the provider approach in https://github.com/higherkindness/rules_scala. I believe this approach can be improved and simplified.

smparkes:
A few thoughts:

You can think of this whole effort as a couple of big pieces. One is the tool/provider construction, execution and dependency management.

The other is the configuration and use issue, independent of how the tools are structured and managed. Things like do I have to explicitly ask for every target in every version I care about? I started with a POC where I just copied and hard-coded the versions I wanted. I moved on to explicit starlark loops for every target. That's feasible but that's when I determined that IMNSHO that would be too big a barrier: it's just too much to ask every build file to become a shim to a starlark file with a loop. I would hate that. It'd be one thing if that was the only choice but we can automate that away, so most people don't have to learn much starlark and also removes a lot of boilerplate. I'm working on a simple example of that to show soon.

The configuration and use stuff is, to a great extent, independent of how the tools are structured and managed. It has to be implemented against that stuff but that's moderately easily changeable and the API to the user doesn't need to change.

I had seen the `ScalaConfiguration` stuff though I hadn't really delved too deeply.

Looking at it now, it's not clear to me how different it is from a standard toolchain. If `declare_scala_configuration` returned a `ToolchainInfo`, I think it would, strictly speaking, be a toolchain(?) or at least a toolchain impl. Agree that conflicts with other work is always a concern. In my first stab, there really wasn't that much change on a lot of the core stuff, e.g., the phase stuff. Not sure what I might be overlooking. And there was still work underway (scalapb, scrooge, etc.) when I decided it would be better to approach things incrementally.

I think pretty much everything that can be done without toolchains can be done with toolchains in the areas we're talking about. IIUC, the primary difference is the extra behavior for toolchains built in to bazel, in particular toolchain resolution, e.g., `ctx.toolchains`. Although we can't exactly use that with CLI flags and transitions for multiple scala versions at this point (e.g., naming issues, transitions only on rules), there's support (for some definition of _support_) for an integrated `toolchains` attr in the standard API that looks useful. I'm still following up on the status of this and its behavior: looks to me like no one is using it. That said, it wouldn't be a blocker if it was dropped.

So it really comes down to factoring of dependencies and configurations, and I'm looking at that. For example, it took a bit of work to factor things so that the various tools in `src/java/io/bazel/rulesscala/` could be built against multiple versions of scala. I think there's monolithicness here that we could reduce: e.g., use more than one toolchain/provider so that test flags (dependencies?) are only pulled in if you do a test, not for libraries and binaries. But all this work could be done with toolchains or ad hoc rules/providers. Just seems like toolchains is the "right" approach since that's what they are and toolchain resolution is potentially useful.

One thing I'm not particularly comfortable/familiar with is the host/execution/target stuff. AFAICT, we're somewhat isolated from that because everything is on the JVM.

Looking forward to more/others thoughts.

smparkes:
https://github.com/smparkes/rules_scala/pull/2 is a pull request against this PR that shows early work on the `scala_*` macros.

andyscott:
I've thought about this a bit more and you've convinced me that toolchains are worth exploring for supporting multiple versions of Scala within the same build.

I only have one main note/though, mostly because I'm not fully spun up on the toolchain code in rules_scala:


> So it really comes down to factoring of dependencies and configurations, and I'm looking at that. For example, it took a bit of work to factor things so that the various tools in src/java/io/bazel/rulesscala/ could be built against multiple versions of scala.

We might want to just pin some of these things to a version of Scala used internally by the rule set. Obviously some of our code interacts directly with the Scala compiler, and in some cases this needs to be compiled against the specific Scala binary version. We could consider using an approach like SBT-- having the core code compiled against a fixed version of Scala and then using Java interfaces to bridge to code that's dynamically compiled against the specific version of Scala configured in the user's workspace.

smparkes:
> We might want to just pin some of these things to a version of Scala used internally by the rule set.

That sounds a bit like the bazel concepts of host/execution/target. I dealt with that marginally in the past with different versions of the jre.

I'm not quite sure yet how those impact but am keeping me eyes out ...

smparkes:

PR #967, opened 2020-01-28
Add AST walking mode for dependency analyzer

Jamie5:
### Description
We introduce a new dependency tracking mode for the dependency analyzer, which walks the AST.

This currently does not change any user-facing things but rather introduces this new mode only in the plugin. Later diffs will actually make use of this new mode.

This currently has known issues which will be addressed in later diffs as the fixes may be somewhat complex and/or need further exploration. This is currently not user-facing and even when it is experimental it should still be acceptable.

### Motivation
This is part of https://github.com/bazelbuild/rules_scala/issues/867 . We introduce the new mode which plus one deps will eventually be able to use.


ittaiz:
Thanks!
I took a quick look but am really short on time right now. I'll finalize review by end of weekend (🤞 for tomorrow)

ittaiz:
No word from Shachar so I’ll merge

PR #971, opened 2020-01-29
Add a per-scala-version test for dependency analyzer

Jamie5:
### Description
Add a per-scala-version test for dependency analyzer which can be run as needed.

### Motivation
As part of https://github.com/bazelbuild/rules_scala/issues/867 the dependency analyzer will become a bit more complex so we want an easy way to run some tests against all supported versions of scala. This is a bit hackish and can hopefully be replaced once https://github.com/bazelbuild/rules_scala/issues/962 progresses sufficiently.

### DIFF
--- a/dangerous_test_thirdparty_version.sh
+++ b/dangerous_test_thirdparty_version.sh
@@ -1,0 +1,201 @@
+#!/usr/bin/env bash
+
+# This test is dangerous in that it modifies the root rules_scala
+# WORKSPACE file. It attempts to restore the existing WORKSPACE file
+# but there are risks that it may not be successful.
+
+# Hence when running this test one should be sure that they do not
+# have changes in the WORKSPACE file which they cannot recover
+# from if the file gets lost.
+
+# Note that due to performance constraints this is purposely not
+# part of CI but when modifying the dependency_analyzer plugin,
+# this should be run to ensure no regressions.
+
+set -e
+
+replace_workspace() {
+  sed -i '' \
+      -e "s|scala_repositories(.*)|$1|" \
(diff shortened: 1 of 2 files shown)


dangerous_test_thirdparty_version.sh
```diff
@@ -0,0 +1,201 @@
...
+    "f2ba1550a39304e5d06caaddfa226cdf0a4cbccee189828fa8c1ddf1110c4872" \
+    "73aef1a6ccabd3a3c15cc153ec846e12d0f045587a2a1d88cc1b49293f47cb20"
+
+#######################
+# These versions 2.11.0-2.11.8 do not work as
+# it results in an error that argument -Ypartial-unification is invalid
+# which is unrelated to the plugin.
+#######################
+#$runner test_scala_version "2.11.0" \
+#    "d50dbbcc5fd79179fbe43c77560495c03c3193c38fc3ec9298802de67338d424" \
```
ittaiz: Why do we need the commented out parts?

dangerous_test_thirdparty_version.sh
```diff
@@ -0,0 +1,201 @@
...
+
+# This test is dangerous in that it modifies the root rules_scala
+# WORKSPACE file. It attempts to restore the existing WORKSPACE file
+# but there are risks that it may not be successful.
+
+# Hence when running this test one should be sure that they do not
+# have changes in the WORKSPACE file which they cannot recover
+# from if the file gets lost.
+
+# Note that due to performance constraints this is purposely not
```
ittaiz: Maybe we should add this paragraph to the contributors md or something?

PR #972, opened 2020-01-29
Separate coverage

borkaehw:
### Description

Separate coverage code from `phase_compile`.

### Motivation

I am investigating on how to integrate Zinc compiler. One possible way is to replace `phase_compile` with `phase_zinc` (some phase like this). I figure the coverage code may be unrelated to compilation and we want to run code coverage with Zinc as well.

Let me know if I am wrong.


borkaehw:
Some thoughts:
Do all 7 rules need this phase? If yes, let's keep it as is. If no, which rule can we save some computation?

borkaehw:
Suffering from Maven again. PR opened #973

borkaehw:
Force-pushed to fix CI.

andyscott:
I'm in favor of separating the coverage phase (FWIW that's exactly how I originally developed it in higherkindness/rules_scala-- as a separate phase-- but I guess you knew that 😄).

borkaehw:
Force-pushed to solve conflict.

scala/private/phases/phase_coverage.bzl
```diff
@@ -0,0 +1,78 @@
...
+    return _phase_coverage(
+        ctx,
+        p,
+        _args.srcjars if hasattr(_args, "srcjars") else depset(),
+    )
+
+def _phase_coverage(ctx, p, srcjars):
+    if len(ctx.files.srcs) + len(srcjars.to_list()) == 0 or \
+       not ctx.configuration.coverage_enabled or \
+       not hasattr(ctx.attr, "_code_coverage_instrumentation_worker"):
```
ittaiz: can/should we drop this condition about the attribute? I think it made sense when it was coupled to the mandatory compile phase.
If a rule doesn't have this attribute it shouldn't use this phase. 
Moreover we want this rule to fail and make a conscious decision, right?

scala/private/phases/phase_coverage.bzl
```diff
@@ -0,0 +1,78 @@
...
+    return _phase_coverage_default(ctx, p, args)
+
+def phase_coverage_common(ctx, p):
+    return _phase_coverage_default(ctx, p)
+
+def _phase_coverage_default(ctx, p, _args = struct()):
+    return _phase_coverage(
+        ctx,
+        p,
+        _args.srcjars if hasattr(_args, "srcjars") else depset(),
```
ittaiz: this sounds like a bug in rules_scala that we don't support srcjars for binary rules.
In any case this is in no way a call for action on this PR. Just an observation.

scala/private/phases/phase_coverage.bzl
borkaehw: This condition seems redundant since it's one of `common_attrs` so every rules except `scala_library_for_plugin_bootstrapping` should have it.

How about removing this condition and remove this phase from `scala_library_for_plugin_bootstrapping`?

ittaiz: Sounds like a great solution to me

borkaehw: Fixed.

Issue #965, opened 2020-01-28
Information passing between phases


ittaiz:
We need to define how information should be passed between phases.
Current strategy is for each phase to return a struct and then following phases get a struct with all the structs nested (keyed by phase name).
This has the following drawbacks:

1. The contract of what a phase exposes is unclear- especially if we want to allow people to depend on a semi-public API it needs to be more clear what a phase returns without rummaging through it's return statement.
2. The current strategy doesn't allow for following phases to override or augment the data returned by previous phases. One need is to add files to the exposed runfiles as shown in https://github.com/bazelbuild/rules_scala/pull/958. Note: Current mechanism does allow for phases which replace phases to override their return value by registering using the same name but that is not enough.
This means that to achieve this phases need to traverse all previous structs for named attributes.

We'd like to fix both of the above. It's likely that overrides and augmentation should be achieved by using the dependency inversion principle and having downstream phases define their needs and then upstream phases can expose this in whatever strategy we choose.

One option we can choose is to have each phase expose a dict of internal_providers (keyed by the name of the provider). This dict will also be an input to each phase and so phases will be able to both override and augment internal providers.  
Another advantage of providers is that they are a recognized in the eco system and tools will support them with respect to docs and such.
To the best of my understanding providers are not a lot more than structs.  

@andyscott has suggested we use well defined dictionaries or structs.

We should discuss these options and what value does each bring as well as other options that people might surface.

ittaiz:
cc @borkaehw

andyscott:
A few notes:

1 - We should audit the interactions between phases. I'm interested in:
- what types are commonly being returned (depset of files, array of files, flags, etc)
- reoccurring field names
- one off dependencies between phases, i.e. "the foo phase is the only phase using the bar phase's flibbit field"

At the very least we can use this to establish and document some conventions for fields.

2 - Do we have any cases where a downstream phase needs to override something from an upstream phase? AFAIK we only have two cases close to flavor:
- We only want _one_ of a particular thing defined by any phase
- Multiple phases need to add to an aggregate over the course of aggregation. This seems like a monoid to me.

We can support both of these cases without having overrides.

ittaiz:
> We should audit the interactions between the current phases

Sounds good. Can you do this audit?

> We can support both of these cases without having overrides.

I guess we can, what's the con of overrides? It sounds to me like a simple and clear pattern.

[...]

Issue #938, opened 2020-01-17
Tracking issue- tech debt


ittaiz:
This is a tracking issue for big tech debt we still have, this will contain links to tracking issues or concrete issues where applicable.
Full list can be tracked via this [search](https://github.com/bazelbuild/rules_scala/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+label%3Atech-debt+)

- [ ]  [Extract out common code](https://github.com/bazelbuild/rules_scala/issues/936)
- [ ]  Move from Label collection to manifest stamping
- [ ]  Move from "deps" patterns to defined toolchains
- [ ]  Move remaining rules to phases architecture
- [ ]  Align documentation with existing feature-set

PR #975, opened 2020-01-30
Centralize dependency logic

Jamie5:
### Description
We centralize most of the dependency-related logic into one file.

This dependency logic includes
- Dependency mode (transitive/plus one/direct)
- unused_deps_mode and strict_deps_mode
- dependency_tracking_method (currently only high-level, soon ast option as well)
- what indirect/direct jars/targets we need to compute

In theory, this should not cause any user-facing changes.

We do the following
- Change collect_jars in scala/private/common.bzl to take information about what is desired (transitivity, direct jars, etc) instead of inferring it from the various settings. Also unify some logic which hopefully makes it clearer what the difference between the various modes are.
- Add scala/private/dependency.bzl which computes the dependency settings from the various attributes that are provided. Ideally, only this logic (and not the things which call it) will need to change as we move to better +1 deps analysis.
- Replace the unused deps checker phase with a dependency phase, which includes all dependency information not just unused deps checker info.

### Motivation
Part of https://github.com/bazelbuild/rules_scala/issues/867

By centralizing the dependency logic in one place, we will be able to change things around more easily, for when we do start changing the logic to have a better +1 deps analysis.


Jamie5:
@ittaiz this is not necessarily completely polished, would like to see if this kind of approach is the way we should go.

If it is but the diff is too large and confusing I can try to extract out pieces for isolated review, though the isolated pieces may not make too much sense without having the whole in there.

ittaiz:
I’ll take a look though this is definitely a PR that I’d like to be cautious with because all of the label collection and which jars are passed to the compiler have huge perf potential impact and I want to make sure we’re not missing anything (especially since we don’t have tests about performance).

> have a better +1 deps analysis.

What does that mean?

Jamie5:
> What does that mean?

Basically the goal of https://github.com/bazelbuild/rules_scala/issues/867 , so that unused deps checker for +1 deps is accurate instead of the existing unused deps checker. (And to also have a strict deps checker for +1 deps)

ittaiz:
I think I'm missing something so I'd like to verify.
Once the AST mode is merged on the plugin side then your plan is to allow people to opt-in to it and activate it for direct, +1 and indirect? Only for some of them? Are you planning on changing the +1 collection?
Maybe we should do this discussion on the tracking issue?

scala/private/phases/phase_collect_jars.bzl
```diff
@@ -46,41 +41,32 @@ def phase_collect_jars_junit_test(ctx, p):
...
-        ctx,
-        _args.base_classpath if hasattr(_args, "base_classpath") else p.scalac_provider.default_classpath,
-        _args.extra_deps if hasattr(_args, "extra_deps") else [],
-        _args.extra_runtime_deps if hasattr(_args, "extra_runtime_deps") else [],
-        _args.unused_dependency_checker_mode if hasattr(_args, "unused_dependency_checker_mode") else p.unused_deps_checker,
+        ctx = ctx,
+        p = p,
+        base_classpath = _args.base_classpath if hasattr(_args, "base_classpath") else p.scalac_provider.default_classpath,
+        extra_deps = _args.extra_deps if hasattr(_args, "extra_deps") else [],
+        extra_runtime_deps = _args.extra_runtime_deps if hasattr(_args, "extra_runtime_deps") else [],
```
borkaehw: nit: I can see most of the cases in this repo, we are not calling functions by keywords. It would be good to make it consistent (I won't point to all the places to bombard the PR). WDYT?

scala/private/phases/phase_dependency.bzl
```diff
@@ -0,0 +1,49 @@
...
+
+def _phase_dependency_default(ctx, p, args = struct()):
+    return _phase_dependency(
+        ctx = ctx,
+        p = p,
+        unused_deps_always_off = args.unused_deps_always_off if hasattr(args, "unused_deps_always_off") else False,
+        strict_deps_always_off = args.strict_deps_always_off if hasattr(args, "strict_deps_always_off") else False,
+    )
+
+def _phase_dependency(
```
borkaehw: If we have a dedicated phase for dependency, is it possible to move all the logic in `//scala/private:dependency.bzl` to this phase? This way we can really have a centralized dependency logic.

I can see a reason to have a shared file since some of the functions are reused in other places. My vision (as the author of phase architecture) is to put code in shared files as less as possible. Hopefully we will get there when we clean up `rule_impls.bzl` and `common.bzl`.

WDYT?

scala/private/phases/phase_collect_jars.bzl
Jamie5: I think there were already inconsistencies in the past, and the benefit of keywords is that when you add and remove arguments you don't accidentally end up mixing up arguments between each other. I strongly would prefer keeping them to avoid the future confusion but if the collective will of the maintainers is to drop them then will drop them.

scala/private/phases/phase_dependency.bzl
Jamie5: Okay there are 3 places not the phase where we import dependency.bzl

- scalapb_aspect.bzl - does this actually participate in phases? If not then don't think we would be able to do this. If it does then I probably need to reread but it should be possible.
- twitter_scrooge.bzl - same thing as ^
- phase_compile.bzl - This looks probably doable - since we are just looking at the files in the rule and seeing if any end in `.java`. This does mean duplicating the logic of seeing which files are java files, or else having some place where we put that logic - where would be a good place?

scala/private/phases/phase_collect_jars.bzl
borkaehw: The reason is valid. I am also in favor of it though I wan't aware of it before. However, all the current phases are not using keywords ATM, I prefer to have a separate PR to make things consistent. Let's wait for suggestion from @ittaiz

scala/private/phases/phase_dependency.bzl
borkaehw: There is indeed a follow-up task for moving remaining rules to phase architecture #938, `scalapb_aspect.bzl` and `twitter_scrooge.bzl` may be included. But at a brief glance, I don't think they fit into existing phases.

Probably a bad idea: generate `phase_dependency_scalapb_aspect` and `phase_dependency_twitter_scrooge` and use these functions as normal. This way we are making more tech-debt but we can avoid having `//scala/private:dependency.bzl`.

We should wait for @ittaiz on this before we decide but I would like to know WDYT?

scala/private/phases/phase_collect_jars.bzl
Jamie5: I think this generally sticks to changing things where we are already changing things; if we do want consistency then I feel like doing the rest in a separate diff (but leaving such changes already in this one here) makes sense. But will defer to @ittaiz .

scala/private/phases/phase_dependency.bzl
Jamie5: If we just say right now `dependency.bzl` is the source of truth but later `phase_dependency.bzl` becomes the source of truth when some other stuff is sorted out that seems the straightforward way to me.

(Then we just need to sort out `phase_compile.bzl` - do you have suggestions on how to break out the java file separation logic?)

But if people are in favor of making `phase_dependency_scalapb_aspect` etc and the somewhat strange imports which ensure, I am fine with that.

borkaehw: > do you have suggestions on how to break out the java file separation logic?

Yes, I do think we should avoid duplication and have a function that `phase_dependency` can use to set `unused_deps_mode = "off"`. Then `phase_compile` doesn't need `new_dependency_info`. WDYT?

I can work on this.

Jamie5: Ok - so `phase_dependency` and `phase_compile` will call the function (which I guess lives in `rule_impls.bzl`) something like `get_java_sources()`, and `phase_dependency` will turn unused deps off whenever `get_java_sources()` is nonempty. Is that correct?

And in the meantime are we keeping around `dependency.bzl` or doing the thing of having `phase_dependency_scalapb_aspect` in `dependency_phase.bzl`?

borkaehw: > Ok - so phase_dependency and phase_compile will call the function (which I guess lives in rule_impls.bzl) something like get_java_sources(), and phase_dependency will turn unused deps off whenever get_java_sources() is nonempty. Is that correct?

That is correct. I believe it will be better this way.

> And in the meantime are we keeping around dependency.bzl or doing the thing of having phase_dependency_scalapb_aspect in dependency_phase.bzl?

We have to wait for maintainers. Sorry, I am not one, but I would like to help.

borkaehw: How does this PR look? Do you think it will get us what we want? #979

Jamie5: With that PR should we also have a constant for java extension? But otherwise seems like it does what we want.

I will make updates after maintainers look over the discussions.

borkaehw: > With that PR should we also have a constant for java extension?

Good idea. Updated.

borkaehw: FYI, the PR has been merged. The java extension is in `scala/private/paths.bzl`.

scala/private/phases/phase_collect_jars.bzl
ittaiz: I think consistency has value but simplicity also has its value.
If the number of arguments is small and the types differ enough then I'd probably not use keywords. 
Having said that I definitely agree that there are cases where keywords bring a lot of value.
@Jamie5 can you give a few examples (from the phases files) that were using named arguments?
I'm not against this pattern but I'm a bit concerned it's going to add a lot of verbosity

Jamie5: https://github.com/bazelbuild/rules_scala/blob/master/scala/private/phases/phase_compile.bzl#L232 and https://github.com/bazelbuild/rules_scala/blob/master/scala/private/phases/phase_compile.bzl#L413 for example

Yeah for just a few arguments it doesn't seem worthwhile but as the arg count increases it becomes riskier and also more annoying when reading the code and needing to match up arguments.

scala/private/phases/phase_dependency.bzl
Jamie5: Fixed the part with calling from phase_compile.scala, will await @ittaiz judgement for whether to do `phase_dependency_scalapb_aspect` now or later

scala/private/phases/phase_collect_jars.bzl
ittaiz: There is a trade-off here between simplicity, safety and consistency.
I'm not sure I'd like all method calls to have named arguments.
If the number of arguments is small and the types differ enough then I'd probably not use keywords. 
Having said that I definitely agree that there are cases where keywords bring a lot of value.
@Jamie5 can you stick to the current convention (of phases at least) and maybe send a separate PR only about adding named arguments? My reasoning is that I'm somewhat partial on this change and I'd rather not block this PR on this decision but at the same time I'm not saying this is completely wrong.

ittaiz: Mainly I'm concerned about verbosity and how easy it will be to get inconsistent.
Tried searching for advice in PEP8 but couldn't find any.
If you can find references that talk about one way or another as the default I'd love to see

scala/private/phases/phase_dependency.bzl
ittaiz: I'd also like for us to remove shared files but like @borkaehw says we need to finish move remaining rules. 
I'm against the synthetic phase for the aspects and I think that the right trade off for now is:
1. Have phases only use data from other phases as much as possible.
2. Where phases must use functions because non phase rules/aspects need them those functions should be in distinct bzl files with clear roles. This means eventually killing rule_impls.bzl and common.bzl in favor of smaller files.

ittaiz:
reviewed (the easy) half.
Hope to finish by tomorrow evening 🤞

scala/private/phases/phase_collect_jars.bzl
Jamie5: @ittaiz do you mean everywhere this adds/introduces keyword arguments, in `*/phases/*`, to remove them? I am fine with doing so and would probably let it lie as opposed to pushing for it in the future. For me the links due point to named args.

scala/private/phases/phase_dependency.bzl
Jamie5: I believe 1. is the case, as the only place which uses `dependency.yml` is `phase_dependency.yml` and actually all other phases are able to get info from `phase_dependency`. As for 2. I think `dependency.yml` also serves that purpose. So given the fact we don't want synthetic phases I don't think we need any changes here.

Jamie5:
@ittaiz this diff's commit message is also named badly and after merge the first line that people would see would just be `calling`. I feel like maybe force pushing would cause strange problems so maybe after all the revisions and this is ready to be merged I can do one final force push and get a reasonable commit message? Or do you have an easy way to edit the commit message during squash?

ittaiz:
Like I mentioned somewhere else- since this has multiple commits the PR title will be the first line of the commit so no worries

scala/private/common.bzl
```diff
@@ -59,60 +30,62 @@ def _collect_jars_when_dependency_analyzer_is_off(
...
+        dep_target,
+        dependency_mode):
+    if dependency_mode == "transitive":
+        return java_provider.transitive_compile_time_jars
+    elif dependency_mode == "plus-one":
+        if PlusOneDeps in dep_target:
+            plus_one_jars = [dep[JavaInfo].compile_jars for dep in dep_target[PlusOneDeps].direct_deps if JavaInfo in dep]
+            return depset(transitive = plus_one_jars + [java_provider.compile_jars])
+        else:
+            return java_provider.compile_jars
```
ittaiz: I know this is my logic but I don't fully remember the cause. Probably some dependencies that the aspect didn't reach or something but I can't remember which or why.
Any chance you can try to remove it, see which tests fail and add a comment to the code? It's a complicated piece of code (and not due to your fault)

scala/private/dependency.bzl
```diff
@@ -0,0 +1,78 @@
...
+        # all transitive dependencies are included
+        return "transitive"
+    elif not _is_plus_one_deps_off(ctx):
+        # dependencies and dependencies of dependencies are included
+        return "plus-one"
+    else:
+        # only explicitly-specified dependencies are included
+        return "direct"
+
+def _dependency_mode_for_addons(ctx):
```
ittaiz: why don't they have +1? Just because that was the existing support? Something else?

scala/private/phases/phase_dependency.bzl
```diff
@@ -0,0 +1,60 @@
...
+        unused_deps_mode = "off"
+
+    return new_dependency_info(
+        dependency_mode = get_dependency_mode(ctx),
+        unused_deps_mode = unused_deps_mode,
+        strict_deps_mode = strict_deps_mode,
+        dependency_tracking_method = "high-level",
+    )
+
+def phase_dependency_common(ctx, p):
```
ittaiz: maybe move common and bootstrapping up to the top of the file? To match formatting of other phases

scala/private/phases/phase_collect_jars.bzl
```diff
@@ -46,41 +41,32 @@ def phase_collect_jars_junit_test(ctx, p):
...
     )
 
 # Extract very common code out from dependency analysis into single place
 # automatically adds dependency on scala-library and scala-reflect
 # collects jars from deps, runtime jars from runtime_deps, and
 def _phase_collect_jars(
         ctx,
+        p,
```
ittaiz: wdyt about passing in `p.dependency`? `p` is a very big struct to pass and widens the method (as a reader I'm now more alert)

scala/private/phases/phase_compile.bzl
```diff
@@ -52,16 +52,6 @@ def phase_compile_library(ctx, p):
     )
     return _phase_compile_default(ctx, p, args)
 
-def phase_compile_library_for_plugin_bootstrapping(ctx, p):
-    args = struct(
-        unused_dependency_checker_ignored_targets = [
```
ittaiz: how come we no longer need to ignore these targets?

scala/private/rule_impls.bzl
```diff
@@ -61,73 +61,55 @@ def compile_scala(
...
+        if dependency_info.need_direct_jars:
+            direct_jars = _join_path(cjars_list)
+            optional_scalac_args_map["DirectJars"] = direct_jars
+        if dependency_info.need_direct_targets:
+            direct_targets = ",".join([str(labels[j.path]) for j in cjars_list])
+            optional_scalac_args_map["DirectTargets"] = direct_targets
+
+    optional_scalac_args = "\n".join([
+        "{k}: {v}".format(k = k, v = v)
+        for (k, v) in sorted(optional_scalac_args_map.items())
```
ittaiz: clarify in comment that `sorted` is for input stability and reproducibility?

Jamie5:
@ittaiz sorry about that, I just looked at the split view and yeah it is not very helpful. I think I'm used to this working in split view fine, but will try to avoid similar changes in the future.

scala/private/dependency.bzl
```diff
@@ -0,0 +1,78 @@
...
+        need_direct_jars = need_direct_jars,
+        need_direct_targets = need_direct_targets,
+        need_direct_info = need_direct_jars or need_direct_targets,
+        dependency_tracking_method = "high-level",
+        unused_deps_mode = unused_deps_mode,
+        strict_deps_mode = strict_deps_mode,
+        use_analyzer = is_strict_deps_on or is_unused_deps_on,
+    )
+
+def dependency_info_for_addons(ctx):
```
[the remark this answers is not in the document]
Jamie5: This refers to protobuf and scrooge as things not part of "core" scala. Is there a term for that, if not I'll maybe think of some alternative terms.

scala/private/dependency.bzl
```diff
@@ -0,0 +1,78 @@
...
+    if not hasattr(ctx.attr, "_dependency_analyzer_plugin"):
+        return "off"
+
+    # when the strict deps FT is removed the "default" check
+    # will be removed since "default" will mean it's turned on
+    if ctx.fragments.java.strict_java_deps == "default":
+        return "off"
+    return ctx.fragments.java.strict_java_deps
+
+def get_is_strict_deps_on(ctx):
```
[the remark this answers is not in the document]
Jamie5: Sounds good, _is_strict_deps_on as it's not used elsewhere

scala/private/dependency.bzl
```diff
@@ -0,0 +1,78 @@
...
+        unused_deps_mode = "off",
+        strict_deps_mode = get_strict_deps_mode(ctx),
+        dependency_tracking_method = "high-level",
+    )
+
+def get_dependency_mode(ctx):
+    if get_is_strict_deps_on(ctx):
+        # all transitive dependencies are included
+        return "transitive"
+    elif not _is_plus_one_deps_off(ctx):
```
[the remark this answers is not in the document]
Jamie5: Sounds good

scala/private/phases/phase_dependency.bzl
Jamie5: You mean that if a function is currently in `dependency.bzl` but doesn't need to be, then move it to `phase_dependency.bzl`? It seems like it splits logic between two places, but given that it is in short term plans to move everything to `phase_dependency.bzl` anyways am fine with such, so will do so.

scala/private/dependency.bzl
```diff
@@ -0,0 +1,78 @@
...
+
+def new_dependency_info(
+        dependency_mode,
+        unused_deps_mode,
+        strict_deps_mode,
+        dependency_tracking_method):
+    is_strict_deps_on = strict_deps_mode != "off"
+
+    # Currently we do not support running both strict and unused deps at the same time
+    if is_strict_deps_on:
```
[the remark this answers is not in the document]
Jamie5: readme says `Currently unused dependency checking and strict-deps can't be used simultaneously, if both are set only strict-deps will run.`, I also just tested and it seems like indeed currently we just silently ignore unused deps when strict_deps is passed on command line.

scala/private/common.bzl
Jamie5: Ok will play around and see

scala/private/dependency.bzl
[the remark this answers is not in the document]
ittaiz: ok. Since we'll allow both options soon we can keep it as is. Thanks for checking!

scala/private/dependency.bzl
Jamie5: Yeah mainly that the existing code for scrooge/protobuf only had the two options of direct and strict. I'm actually not familiar with those rules at all so opted to preserve existing behavior.

scala/private/phases/phase_dependency.bzl
ittaiz: Yes. The idea is for code in the shared files to stick out like a sore thumb.

scala/private/phases/phase_dependency.bzl
Jamie5: Sounds good and did other two phase methods too since others seem to do that

scala/private/dependency.bzl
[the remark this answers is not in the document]
ittaiz: I think there's value in understanding why they have variance (or more importantly do we think they should have variance).
Maybe they shouldn't have and then we don't need a better name we just need to align it with the other. Maybe it's because they are generated code rules and then that's the name.
Maybe it's something else, not sure

scala/private/phases/phase_collect_jars.bzl
Jamie5: Sounds good

scala/private/dependency.bzl
ittaiz: so maybe we should open an issue and add a comment about it here?
instead of trying to sugar coat it with `addons` let's say `_legacy_unclear_dependency_mode_for_protobuf_scrooge`. I know it might sound funny but it will make it clear for readers this isn't a well groomed piece but something we need to handle

scala/private/phases/phase_compile.bzl
Jamie5: Since we turn off unused deps checker anyways (which now happens in phase_dependency_library_for_plugin_bootstrapping) there's no need to collect ignore targets at all from what I can see.

As a side note unused_deps_ignored_targets was originally part of the dependency_info struct, I moved it out in part because in phase_compile we made a decision to turn off unuse deps. But now that that logic is in phase_dependency again maybe it makes sense to have phase_dependency collect the ignored targets. (But there may have been other disadvantages of doing that, I can't recall)

scala/private/rule_impls.bzl
Jamie5: Sounds good

scala/private/dependency.bzl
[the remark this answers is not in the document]
Jamie5: As per your other comment will file a ticket and will also rename to legacy_unclear_something or other

scala/private/dependency.bzl
Jamie5: Added a comment and filed https://github.com/bazelbuild/rules_scala/issues/987

Jamie5:
@ittaiz note the pushed changes do not contain the requested comments in _additional_transitive_compile_jars yet, I am just seeing how CI runs with all the other changes

scala/private/phases/phase_collect_jars.bzl
ittaiz: I missed this thread. If it's not too much trouble I'd appreciate that you indeed remove those

scala/private/common.bzl
Jamie5: Ok so the thing to determine is on the line `return depset(transitive = plus_one_jars + [java_provider.compile_jars])`, why do we need `[java_provider.compile_jars]`?

ittaiz: Why is that the thing to determine? I don't remember the specific motivation but I can definitely tell you that a clear test will fail if you will remove.
I meant the `else` for ` if PlusOneDeps in dep_target:`

Jamie5: Hmm okay have found some stuff and will add comments wrt that, hopefully it will make sense

scala/private/phases/phase_collect_jars.bzl
Jamie5: Ok will do

Jamie5:
Ok all comments should now be addressed unless I missed some

scala/private/phases/phase_collect_jars.bzl
```diff
@@ -46,41 +41,32 @@ def phase_collect_jars_junit_test(ctx, p):
...
-    )
-    return _phase_collect_jars_default(ctx, p, args)
-
 def phase_collect_jars_common(ctx, p):
     return _phase_collect_jars_default(ctx, p)
 
 def _phase_collect_jars_default(ctx, p, _args = struct()):
     return _phase_collect_jars(
         ctx,
+        p.dependency,
```
borkaehw: Would you mind to change it to just `p`? I know using `p.dependency` will be less verbose inside the function but just for the purpose of consistency.

Jamie5: This change was in response to a comment by @ittaiz . I am fine either way.

ittaiz: @borkaehw why do you think this is a better pattern in general? 
Let’s ignore consistency for a sec since maybe we’ll decide to change the pattern

borkaehw: I think we should keep the function signature as simple as possible. By the fact that `p` should contain all the information we need, we may expand the data structure within the function instead.

In my opinion,
```
phase_a(ctx, p, a)
```
```
phase_b(ctx, p, a, b)
```
```
phase_c(ctx, p, a, b, c, d, e)
```
make the phase architecture messy. When we finalize #965, I imagine the functions should all be consistent like `phase_a(ctx, internal_proviers)`.

Also, functions with a super long list of arguments like `compile_scala` makes it hard to read. I, in general, prefer to keep arguments as less as possible and no keyword is needed.

scala/private/common.bzl
ittaiz: ❤️ thanks!

scala/private/phases/phase_collect_jars.bzl
ittaiz: @borkaehw I understand the value in what you say but to be honest I actually quite disagree with it. I think that bottom line having similar APIs is less beneficial than having phases explicitly declare which internal providers they need. Something like `providers = [JavaInfo]` which we have in startlark.
Having said that I'd rather not start this debate here over this change and so for consistency I apologize @Jamie5 and I'll ask you to revert to your original change. Sorry again.

borkaehw: Sounds good. Thanks for sharing your idea.

PR #979, opened 2020-01-30
Get files with extension

borkaehw:
### Description

Move "get files with extension" logic to a helper function.

### Motivation

We have a conversation https://github.com/bazelbuild/rules_scala/pull/975#discussion_r373098148 of possible ways to reduce the use of `dependency.bzl`. We can have a helper function, which can be used in `dependency.bzl` in the future, lives in `rule_impls.bzl` to reuse duplicate code.

### DIFF
--- a/scala/private/paths.bzl
+++ b/scala/private/paths.bzl
@@ -1,0 +1,12 @@
+java_extension = ".java"
+
+scala_extension = ".scala"
+
+srcjar_extension = ".srcjar"
+
+def get_files_with_extension(ctx, extension):
+    return [
+        f
+        for f in ctx.files.srcs
+        if f.basename.endswith(extension)
+    ]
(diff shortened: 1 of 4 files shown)


borkaehw:
Sounds like a good idea. Do you mind adding it to tech-debt issue?

@Jamie5 suggests to put extension constants in a shared file. Do you want me to make a quick change with this PR?

Also, when we want to put shared code, how do we decide between `common.bzl` and `rule_impls.bzl`?

borkaehw:
Moved the extension to `rule_impls.bzl`. Let me know WDYT.

ittaiz:
Tech debt issue sounds good, maybe you’ll open it after you open the skylib issue? I’ll add the label.
Re location- I’m actually slowly trying to kill `common.bzl` and `rule_impls.bzl` as I feel they’re just trash cans of undefined behavior.
Wdyt about moving it to a distinct file? Something like `extensions.bzl`.

borkaehw:
I can create a file with path `scala/private/extensions.bzl` and move those three extension constants. How about where to put `get_files_with_extension` function?

ittaiz:
I’d put that there as well. It’s a method about extensions. Alternatively you can create a paths.bzl and put everything under that

borkaehw:
I like `paths.bzl` more. Updated.

PR #469, opened 2018-04-01
Cherry pick and merge #465 into scala 2.12 branch

johnynek:
This is a replacement of #463 by cherry-picking and merging #465 into the scala 2.12 branch ultimately closing #295 actually.

@dkomanov @ittaiz please take a look.

PR #980, opened 2020-01-30
use bind to remove loader-specific labels in dependencies

smparkes:
### Description
Create labels in `external` for loader-specific labels and use those labels as dependencies.

### Motivation
Currently some dependencies on external repos use loader-specific labels. The work in #962 is using `rules_jvm_external`, not `jvm.bzl`, so is blocked by this hard-coding.

### Comments
Extracted out of #964 (see https://github.com/bazelbuild/rules_scala/issues/962#issuecomment-580363653).

Does not necessarily handle all cases so there may be follow-ons. This covers the cases needed by #469 

A longer term approach to labels is under consideration as part of #962. See [ExternalRepositories.md](https://github.com/smparkes/rules_scala/blob/configuration/unstable/multiscala/ExternalRepositories.md). That will take some work so this unblocks #962 in the interim.


scala/private/macros/scala_repositories.bzl
```diff
@@ -199,3 +199,18 @@ def scala_repositories(
         name = "io_bazel_rules_scala/dependency/scala/guava",
         actual = "@io_bazel_rules_scala_guava",
     )
+
+    native.bind(
+        name = "io_bazel_rules_scala/dependency/scalac_rules_commons_io",
```
liucijus: What is the motivation behind having a second bind for `@scalac_rules_commons_io`?

Nit: name does not follow the same pattern as others binds.

smparkes: > second bind

This isn't a second `bind`, right? `@scalac_rules_commons_io` is a repo. `Bind`s end up in `external`? Is this what you're commenting on?

My thinking is in [ExternalRepositories](https://github.com/smparkes/rules_scala/blob/configuration/unstable/multiscala/ExternalRepositories.md). The tl;dr is that using repo names makes things loader(e.g., `jvm_import_external`, `maven_install`)-specific: different loaders create different repo names and the loaders don't make this configurable. There are a few ways to address this, including using `bind`s to map from loader-specific repo names to common bind targets.

If this approach is acceptable, it seems prudent to come up with a canonical mapping from maven coordinate to bind target. At this point, I didn't want to presume that conclusion so I was sticking with minimal changes necessary to unblock other work. I took the approach of using the existing repo name as the bind target.

liucijus: I am referring to this one
```
native.bind(
        name = "io_bazel_rules_scala/dependency/commons_io/commons_io",
        actual = "@scalac_rules_commons_io//jar",
)
```
on this line: https://github.com/bazelbuild/rules_scala/blob/0ab261f60ac82ee407a8457f84b94132ca808173/scala/private/macros/scala_repositories.bzl#L165

smparkes: Ah. Thanks. I was pushing too fast and didn't look for the bind: it wasn't being used anywhere within `rules_scala` itself.
Removed the redundant bind and updated the scalac build to use it.

ittaiz:
And @andyscott and of course @smparkes :)

PR #981, opened 2020-01-30
Factor scalac provider

smparkes:
### Description

This moves the scalac attr to the toolchain so it can be overriden by
different toolchain implementations.

It also adds a `ScalaInfo` provider level for scala_toolchain to follow the
pattern in the scala docs (rather than flattening into the ToolchainInfo).

### Motivation

Continuation of toolchain factoring in #978. Motivation is there.


smparkes:

PR #983, opened 2020-01-30
Skeleton for multitarget

smparkes:
### Description
This is a follow-on to the configuration PR #964 . It shows the basic idea of using a macro shim to create multiple targets based on the configuration.

It works for the current example builds. It should be creating and running parallel subgraphs for each version. It doesn't use toolchains correctly. That is dependent on the toolchain track in #962.

It does include documentation on the approach to give folks a chance to comment on.

I'll work on flushing out the implementation but would welcome thoughts on the approach.


smparkes:

PR #985, opened 2020-01-31
 Handle inlined literals in AST walker

Jamie5:
### Description
We update the AST-based dependency analyzer mode to properly handle literals which were inlined.

This fixes the issue only for Scala 2.12.4+. The functionality we are using to fix this only exists in 2.12.4+

We also introduce `ScalaVersion` to make handling versions easier as well as allowing a very primitive conditional compilation via macro. As we now have a macro, we make minor adjustments to make macros work in `scala_library_for_plugin_bootstrapping`.

### Motivation
This removes a false positive in the AST checker.


third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/ScalaVersionTest.scala
```diff
@@ -0,0 +1,136 @@
...
+      var hit = false
+      ScalaVersion.conditional(
+        "",
+        "500.0.0",
+        "hit = true"
+      )
+      assert(hit)
+    }
+
+    // Min bounds not hit
```
ittaiz: nit: Max bounds not hit

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/ScalaVersionTest.scala
```diff
@@ -0,0 +1,136 @@
...
+    {
+      var hit = false
+      ScalaVersion.conditional(
+        "",
+        "0.0.0",
```
ittaiz: Why do you use here `0.0.0` and not `1.0.0`?

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/ScalaVersionTest.scala
```diff
@@ -0,0 +1,136 @@
...
+        "hit = true"
+      )
+      assert(!hit)
+    }
+
+    // Min-max bound hit
+    {
+      var hit = false
+      ScalaVersion.conditional(
+        "0.0.0",
```
ittaiz: same about `0.0.0`

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/ScalaVersionTest.scala
```diff
@@ -0,0 +1,136 @@
+package third_party.dependency_analyzer.src.test.io.bazel.rulesscala.dependencyanalyzer
+
+import org.scalatest._
+import third_party.dependency_analyzer.src.main.io.bazel.rulesscala.dependencyanalyzer.ScalaVersion
+
+class ScalaVersionTest extends FunSuite {
```
ittaiz: Big big like on these tests! Well done :)

Can you add a test for the inclusive min max case?
The con for it is that it will duplicate a bit the prod code of getting the current version but I think the confidence we will get from someone accidentally changing the `>=` to `>` or `<=` to `<` is big

scala/private/rules/scala_library.bzl
```diff
@@ -161,6 +161,8 @@ _scala_library_for_plugin_bootstrapping_attrs.update(
     common_attrs_for_plugin_bootstrapping,
 )
 
+_scala_library_for_plugin_bootstrapping_attrs["build_ijar"] = attr.bool(default = True)
```
ittaiz: WDYT about changing this to a dict? To follow the existing pattern https://github.com/bazelbuild/rules_scala/blob/6ba125f449633332766b80d62fd47393589dd73f/scala/private/rules/scala_library.bzl#L42

It took me a good few seconds to understand what you're doing here

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinderTest.scala
```diff
@@ -302,6 +320,43 @@ class AstUsedJarFinderTest extends FunSuite {
     )
   }
 
+  test("inlined literal is direct") {
+    // Note: For a constant to be inlined
+    // - it must not have a type declaration such as `: Int`.
```
ittaiz: The reason you're not adding a test for the case where there is a type declaration or a lower case capital letter because you're trying to take the strictest case, right?
If other cases are inlined then this checks inlining of literals already and if they won't be inlined in the future then it's a different feature?

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinderTest.scala
```diff
@@ -318,4 +373,60 @@ class AstUsedJarFinderTest extends FunSuite {
...
+  test("java interface field and method is direct") {
+    withSandbox { sandbox =>
+      sandbox.compileJava(
+        className = "A",
+        code = "public interface A { int a = 42; }"
+      )
+      val bCode =
+        """
+          |class B {
+          |  def foo(x: A): Unit = {}
```
ittaiz: why do we need this? I think I'm missing something
Also this test seems very similar to the one below apart from this and the test name.

Jamie5:
@ittaiz so far have not addressed your comments, the only thing was force pushing the original diff and then adding a merge on top of it. The reason for the force was because I realized that the commit messages in the commits show up in the final squashed commit (instead of sticking the PR text or something like that in it), and hence at least the first message should be something better than `plugin3`.

Now (or at some point) I will actually look at your comments here and all the other places.

third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/ScalaVersion.scala
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * Runs [code] only if minVersion and maxVersion constraints are met.
+   *
+   * NOTE: This method should be used only rarely. Most of the time
+   * just comparing versions in code should be enough. This is needed
+   * only when the code we want to run can't compile under certain
+   * versions. The reason to use this rarely is the API's inflexibility
+   * and the difficulty in debugging this code.
+   *
+   * Each of minVersion and maxVersion can either be the empty string ("")
```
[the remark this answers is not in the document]
Jamie5: Shouldn't be an effect on compile time it was just hard to figure out. But figured out so it is now Option.

third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/ScalaVersion.scala
```diff
@@ -0,0 +1,112 @@
...
+   * to signify that there is no restriction on this bound.
+   *
+   * Or it can be a string of a full version number such as "2.12.10".
+   *
+   * Note only literal strings are accepted, no variables etc. i.e.
```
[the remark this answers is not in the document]
Jamie5: Added comments about otherwise not compiling. Looks like inlineable constants indeed are allowed (I guess because they get inlined before macro expansion) so added comment to that effect. If one does variables and stuff then basically that needs to be resolved at compile time which is often not possible, and even when it is (I guess a non-inlined variable which we can still analyze as having a definite value) it would be quite difficult to do so by examining the AST, AFAIK.

third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/ScalaVersion.scala
```diff
@@ -0,0 +1,112 @@
...
+    versionString.split("\\.") match {
+      case Array(superMajor, major, minor) =>
+        new ScalaVersion(superMajor.toInt, major.toInt, minor.toInt)
+      case _ =>
+        throw new Exception(s"Failed to parse version $versionString")
+    }
+  }
+
+  /**
+   * Runs [code] only if minVersion and maxVersion constraints are met.
```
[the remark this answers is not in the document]
Jamie5: Done

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/ScalaVersionTest.scala
```diff
@@ -0,0 +1,136 @@
...
+      val vb = ScalaVersion(b)
+
+      assert(!(va == vb))
+      assert(va != vb)
+      assert(!(va < vb))
+      assert(!(va <= vb))
+      assert(va > vb)
+      assert(va >= vb)
+
+      // Lesser versions
```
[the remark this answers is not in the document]
Jamie5: Went with `testOrder` just because it seemed easier to catch all the cases together

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/ScalaVersionTest.scala
Jamie5: Foxed

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/ScalaVersionTest.scala
Jamie5: No particular reason so changed

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/ScalaVersionTest.scala
Jamie5: No particular reason so changed

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/ScalaVersionTest.scala
Jamie5: Due to the way macros are implemented I don't think it's possible, since we need to use inlineable strings (which current version is not, it seems). Unless we wanted to do templating or something but that shouldn't be the case.

FWIW, hopefully anything that uses macros will be tested, and the test version checks are straightforward, so if there is a bug in the prod code it should likely be caught by unit tests when run against all versions.

scala/private/rules/scala_library.bzl
Jamie5: Done

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinderTest.scala
Jamie5: Yes (except not sure what you mean by "then it's a different feature")

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinderTest.scala
Jamie5: It is very similar, but this one does both a inlineable constant lookup and a method call. So in theory it should always work (as you can see by test `java interface method argument is direct` which has no version checks. However in practice it doesn't - in this test we need a version check of `2.12.0`. Hence there is some sort of issue which got resolved in 2.12.0 which made the checker now be able to detect this dependency.

The test below meanwhile only has an inlined constant lookup and hence needs to be 2.12.4+ for the fix.

Jamie5:
@ittiaz now updated addressing your comments. Now that I know commit messages matter I will make sure they are reasonable enough (I'm used to different workflows) so that force pushes won't be necessary in the future to fix the messages. Sorry about that.

scala/private/rules/scala_library.bzl
```diff
@@ -152,7 +152,9 @@ def _scala_library_for_plugin_bootstrapping_impl(ctx):
 # the scala compiler plugin used for dependency analysis is compiled using `scala_library`.
 # in order to avoid cyclic dependencies `scala_library_for_plugin_bootstrapping` was created for this purpose,
 # which does not contain plugin related attributes, and thus avoids the cyclic dependency issue
-_scala_library_for_plugin_bootstrapping_attrs = {}
+_scala_library_for_plugin_bootstrapping_attrs = {
+    "build_ijar": attr.bool(default = True),
```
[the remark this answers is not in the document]
Jamie5: Don't quite follow. The uses of this variable `_scala_library_for_plugin_bootstrapping_attrs` seem to be (besides adding more fields to it) seem to be only in `make_scala_library_for_plugin_bootstrapping` which seems to only be for making the rule for bootstrapping libs, and not all other rules like scala_library and so on.

third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/ScalaVersion.scala
```diff
@@ -0,0 +1,149 @@
...
+   * When set to a version number, the bounds are inclusive.
+   * For example, a maxVersion of "2.12.10" will accept version "2.12.10".
+   *
+   * Note only literal strings are accepted and inlined variables are accepted.
+   * If any non-inlined variables are passed the code will fail to compile.
+   * Inlined variables are generally those declared final on an object which
+   * do not have a type attached.
+   *
+   * valid:
+   *  conditional(minVersion = "2.12.4", maxVersion = "", code = "foo()")
```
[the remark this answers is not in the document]
Jamie5: Macros also seem not to allow named params so removed those as well while fixing the example as you point out.

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/ScalaVersionTest.scala
Jamie5: Done

Jamie5:
@ittaiz also addressed the other inline comments though I don't seem to be able to reply to them for unclear reasons

scala/private/rules/scala_library.bzl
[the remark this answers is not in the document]
ittaiz: you're right. I was wrong :)

PR #986, opened 2020-01-31
Handle default args in AST walker

Jamie5:
### Description
Remove a false positive in AST walker where a never-provided default argument is flagged.

For a default argument that a caller doesn't supply, we shouldn't record the caller as needing that argument's type. Hence we detect such cases and skip them.

### Motivation
We wish to reduce false positives in the AST checker.


Jamie5:
@ittaiz One such test is `simple composition is indirect` (though we don't explicitly mention it on the test). It looks like even in direct mode this should be marked as an unused dep so this isn't just a case of plus-one vs direct.

PR #988, opened 2020-02-04
Include location in error report

Jamie5:
### Description
For the dependency analyzer, we try to include the location for strict deps errors.

We do this by returning a map of jar -> position instead of a set of jars for the strict deps errors. Most changes to DependencyAnalyzer are just plumbing the new positions through.

We pick an arbitrary position to report, the only sorting being that we prefer when the position is defined as opposed to not defined. In practice, this seems to work well enough.

We also add tests that all strict deps errors include positions. This revealed issues where annotations didn't have positions. This was due to us relying on the synthetic tree, which did not have positions attached. Hence we now rely on the individual components of the tree.

### Motivation
It is helpful for strict deps errors to point out where in the source file the dependency is coming from. This helps people either trust the dependency analyzer more by seeing it makes sense, or else helps them file a bug report to fix some issues.


third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinderTest.scala
```diff
@@ -452,4 +462,26 @@ class AstUsedJarFinderTest extends FunSuite {
       }
     }
   }
+
+  test("classOf in class Java annotation is direct") {
```
ittaiz: What is this test?

Jamie5:
The tests now all just assert that there is a position, but not the specific position (i.e. we don't check that the line and column match up with what we expect).

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinderTest.scala
Jamie5: It makes sure that if you use a `classOf[X]` in a Java annotation, that `X` is considered a direct dependency. Scalac handles java and scala annotations differently hence the need for an explicit java test.

Should we change this to just use A/B instead of Category/UnitTests?

ittaiz: It's ok I just didn't understand why the test was added and how it relates to locations in error report.
Also,
It's for coverage, right? You didn't also add a fix here AFAICT

ittaiz:
Can you add a test that checks a specific location? Otherwise it can just regress easily

PR #989, opened 2020-02-05
Make rules_scala compatible with --incompatible_load_proto_rules_from_bzl

Yannic:
Bazel will rewrite Protobuf rules in Starlark and remove their native equivalents.
As a first step, everyone will be required to add load statements for `proto_library`.
This change adds them for rules_scala.

https://github.com/bazelbuild/bazel/issues/8922
https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/402#9082639a-6464-4821-bf47-3b93e00f37e0

### DIFF
--- a/manual_test/scalac_jvm_opts/BUILD
+++ b/manual_test/scalac_jvm_opts/BUILD
@@ -1,9 +1,7 @@
+load("@rules_proto//proto:defs.bzl", "proto_library")
 load("//scala:scala_toolchain.bzl", "scala_toolchain")
 load("//scala:scala.bzl", "scala_library")
-load(
-    "//scala_proto:scala_proto.bzl",
-    "scala_proto_library",
-)
+load("//scala_proto:scala_proto.bzl", "scala_proto_library")
 
 scala_toolchain(
     name = "failing_toolchain_impl",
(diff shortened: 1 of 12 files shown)


Yannic:
Thanks! `rules_{cc,java,python}` are used by Protobuf: https://github.com/protocolbuffers/protobuf/blob/master/BUILD#L3

ittaiz:
I see. Did you actually need them or did you add them as a precaution?
I think (but not sure) that if we don’t use python_proto_library then we won’t need python.
Can you see which of these we must have?

Yannic:
All of them are in `@com_google_protobuf//:BUILD`, so they are required as soon as you use anything from that package (e.g. protoc or Well-Known-Protos). However, at least cc, java, and proto are part of the global preamble and implicitly loaded if not defined. But I'd advocate against relying on the definitions in the global preamble because they probably won't be in there forever.

ittaiz:
I agree relying on implicit definitions is not the way.
We’ll continue with the explicit approach.
Why did you pick these versions? Based on what they had? At least for java I’d like us to use https://github.com/bazelbuild/rules_java/tree/0.1.1

@aiuto @dslomov it sounds unwise that anyone who depends on protobuf will be required to depend on java and python. Maybe it can be changed?

Yannic:
`rules_{cc,java,python}` match what Protobuf uses, `rules_proto` is at HEAD (we don't do releases yet). Using `rules_java@0.1.1` SGTM.

I'll see what I can do to remove the hard dependencies to these rules in the Protobuf repo. However, I think this should be merged for now - we can remove these rules later.

ittaiz:
@Yannic the linting failed. Can you please run `./lint.sh fix` and push?

Yannic:
@ittaiz done

ittaiz:
Thanks!

PR #990, opened 2020-02-06
Ignore scalatest in unused dependency checker

Jamie5:
### Description
For the scalatest rule, we add the scalatest library to the unused dependency checker ignore list, as that library is always imported by the rule itself.

### Motivation
This removes a false positive in unused dependency checker.

### DIFF
--- a/scala/private/phases/phase_compile.bzl
+++ b/scala/private/phases/phase_compile.bzl
@@ -99,7 +99,7 @@ def phase_compile_scalatest(ctx, p):
         unused_dependency_checker_ignored_targets = [
             target.label
             for target in p.scalac_provider.default_classpath +
-	                      [ctx.attr._scalatest] +
+                          [ctx.attr._scalatest] +
                           ctx.attr.unused_dependency_checker_ignored_targets
         ],
     )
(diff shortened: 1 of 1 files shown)


ittaiz:
Where does this false positive manifest? Maybe we can have a test that shows it? I'm just not 100% sure I understand this or how people got a long so far (since such an issue should be hard to workaround given they can't remove something they don't add) other than manually adding this ignore

Jamie5:
Ok so the very basic stupid repro is

C.scala
```
class C
```

BUILD.bazel
```
scala_test(
    name = "c",
    srcs = ["C.scala"],
    unused_dependency_checker_mode = "error",
)
```

This repros in master.

Now why would someone do something like that, probably they wouldn't. But a more complex repro comes up in this situation (which requires using AST mode, as its unused deps checker correctly points out this problem) (and also +1 mode or else it won't compile).

A.scala
```
import org.scalatest._

abstract class A extends FunSuite {
	val Number: Int

	test("number is positive") {
		assert(Number > 0)
	}
}
```

B.scala
```
class B extends A {
	override val Number: Int = 12
}
```

BUILD.bazel
```
scala_test(
    name = "a",
    srcs = ["A.scala"],
)

scala_test(
    name = "b",
    srcs = ["B.scala"],
    deps = [":a"],
)
```

In this case though B is a test it makes no reference to anything in scalatest directly.

So which repro would we prefer to add a test for? For the first one it is easy to add, for the second adding it is blocked on making it possible to turn on ast mode.

ittaiz:
I see. In this case please make sure that you add the second test when we allow turning on AST mode, ok?

Jamie5:
Ok will do

PR #991, opened 2020-02-07
Ignore base class annotation

Jamie5:
### Description
We change the dependency analyzer AST mode to ignore annotations on base classes.

### Motivation

This removes a false positive for the dependency analyzer.

### DIFF
--- a/third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinderTest.scala
+++ b/third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinderTest.scala
@@ -293,6 +293,14 @@ class AstUsedJarFinderTest extends FunSuite {
     )
   }
 
+  test("static annotation of inherited class is indirect") {
+    checkIndirectDependencyDetected(
+      aCode = "class A extends scala.annotation.StaticAnnotation",
+      bCode = "@A class B",
+      cCode = "class C extends B"
+    )
+  }
+
   test("class type parameter bound is direct") {
     checkDirectDependencyRecognized(
       aCode =
--- a/third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinder.scala
+++ b/third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinder.scala
@@ -103,7 +103,20 @@ class AstUsedJarFinder(
 
       if (shouldExamine) {
         if (tree.hasSymbolField) {
-          tree.symbol.annotations.foreach(exploreAnnotationInfo)
+          tree.symbol.annotations
+            // We skip annotations without positions. The reason for
+            // this is the case of
+            //   @SomeAnnotation class A
+            //   class B extends A
+            // Now assuming A and B are in separate packages, while
+            // examining B we will examine A as well, and hence
+            // examine A's annotations. However we don't wish to examine
+            // A's annotations as we don't care about those details of A.
+            // Hence we only examine annotations with positions (hence,
+            // they were defined in the same compilation unit and thus
+            // matter).
+            .filter(_.pos.isDefined)
+            .foreach(exploreAnnotationInfo)
         }
         if (tree.tpe != null) {
           exploreType(tree.tpe, tree.pos)


ittaiz:
Big like on where this is going since such cases are exactly what bothered us.
I want to ask- maybe this can be an issue where unused deps logic actually should differ from strict deps? If I add one more layer of transitivity then I think that scalac will demand that extra dep on the classpath but strict deps will think it’s unused, am I correct or mistaken?

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinderTest.scala
```diff
@@ -293,6 +293,14 @@ class AstUsedJarFinderTest extends FunSuite {
     )
   }
 
+  test("static annotation of inherited class is indirect") {
+    checkIndirectDependencyDetected(
+      aCode = "class A extends scala.annotation.StaticAnnotation",
```
ittaiz: Also,
This test failed originally? If so which dep did it complain about? Doesn’t StaticAnnotation come from std lib? Feel free to correct me if I’m completely off

Jamie5: The test failed originally, as we expect it to complain that A is an unused dependency, but it did not. And A should be an unused dependency because if you look at C's code it never touches A.

Jamie5:
Hmm not sure exactly what you have in mind but in cases where we are only using +1 deps and not transitive, then I can see there potentially being cases where we actually need +2 and in such a case, then yes as-is we might report a dep as unused even though it is needed to pin down a dep of it. And to handle it unused would probably need to behave somewhat differently than strict. But not sure what such a situation would be.

ittaiz:
We just described the situation :)
Take your test case, add another layer of indirection (C extends B, B extends A, A is regular class which has an annotation Foo on it).
In this case unused deps should not say that Foo should be removed but strict deps should also not require it (for example if one uses transitive).
Wdyt?

Only way out I see here is if we don’t solve it but rather say that to use this mechanism people must employ `exports` more rigorously (for example on all things mentioned in the API) and so effectively trim many edges in the graph and potentially significantly limiting parallelism

Jamie5:
So it turns out the case of static annotation might not matter - it seems like even without plus one the compiler doesn't seem to complain about the test case, and +1 handles the case you describe. Though +1 may have issues with A->B->C->D inheritance. If someone runs into many cases like that where the +1 heuristic isn't good enough, should they just be using transitive mode?

Jamie5:
Hmm there are also potential alternative solutions - if the plugin understands the graph a bit more (which presumably we could feed it) for +1 mode it could ignore unused deps which pin down some unused dep which wouldn't otherwise be included. We do still want to prune down unused deps which are unused and also not needed to pin something for compiling though hence the need for a basic understanding of the graph. So it would be somewhat complicated, though doable.

Jamie5:
Question on https://github.com/bazelbuild/rules_scala/pull/698#issuecomment-466121481 is that how the java rules work, that when B is being compiled bazel learns that its ijar needs only C, so that when A is being compiled it will only be compiled with B and whatever the ijar of B needs which is C? And this all works happily without too much cache invalidation and rebuilds and all the other good stuff?

If so I didn't know bazel was able to do such things and I wonder if we could do similar things. Or maybe I'm just confused.

ittaiz:
> If someone runs into many cases like that where the +1 heuristic isn't good enough, should they just be using transitive mode?

We have ~10M LOC of scala. I'm sure you understand we have everything you can imagine.
We still want to be able to run with +1 and unused deps. I think this change will not allow us to do this.

WDYT about having a split here in the logic? The way I see it now this is a dependency that indeed you don't want strict-deps to fail on but at the same time you don't want unused deps to make you remove it (if you're using +1).

> Hmm there are also potential alternative solutions

I'm afraid I didn't follow 100%.


Re java rules- In general that is how it works. They also try to compile like that (I think it's called limited) and if it doesn't work they run the compiler again with the entire transitive classpath (ijars of course). IIRC this is (or used to be) one action that handles this fallback in itself. 
With respect to how well this works with cache invalidation and such- I think java ijars are much much better than scala ijars and compilation times are better so they're taking the hit (though conceptually it is a hit). 
I also know that they were evaluating other strategies (similar to +1) but I don't know where this stands.
Maybe @cushon can share some info.

Jamie5:
So for +1 it sounds like we do want a category of things which are reported neither as unused deps nor strict deps. However I think generally this should be smaller - because most of the time, when something isn't a strict dep, we should indeed remove it.

Hence I think we should still have every non-strict dep reported, except in +1 mode where we need the dep to pin down some other dep. This is why it would be useful to pass a portion of the graph to the plugin.

For example if we have a subclassing relationship A -> B -> C -> D and are compiling A deps are [B, C] then the plugin can discover that the jars needed to compile are [B, C, D] and that the strict deps are [B]. Now the plugin would report [C] as an unused dep, but then it notices that [D] is currently not being brought in by any of the strict deps [B], but is being bought in by [C] so hence we should not report [C] as an unused dep.

In the example above if the deps were [B, D] we would also not report [D] as an unused dep. So the algorithm would be (for plus one mode)
1. Compute the set of required targets for compilation, R (will need to figure out how to compute this, but can just improve it slowly)
2. Compute the strict deps, S
3. Compute the targets that S reaches (i.e. itself and its plus one deps) as P
4. Let the set of unprovided targets be `UP = R - P`
5. Let targets we would report as unused be U
6. For every target in U, report it as unused unless it either is contained in UP, or one its plus one deps is contained in UP.

In the case that UP is nonempty this will potentially allow plenty of extra stuff to end up in the deps (technically any rule that points to anything in UP will be allowed), but I think in almost all cases UP will be empty so we will very rarely have false negatives.

I am still curious as to how common this actually is though and if it is rare enough just using `unused_dependency_checker_ignored_targets` would end up with needing only a tiny handful of uses?

Re java rules: They will always take the ijar hit, even when there is no need to compile with the full transitive classpath for some specific package, right? If so then that is unfortunate that it would not work well for us.

Side note - this diff doesn't actually cause the transitive issues being pointed out as far as I can see - it is the subclass of subclass of subclass of subclass which does.

Jamie5:
Re the above, I know too much configuration is bad but it would be nice at least for us to turn off the behavior selectively (or rather, given that we haven't hit this case we would probably want to just use `unused_dependency_checker_ignored_targets` and not have the above behavior run at all).

Reason is external dependencies. For which +1 deps doesn't work, and we would rather manually export the ones which a dependency needs instead of having the above algorithm run, which may lead to having lots of pruneable deps not reported as unused.

This is similar to the failure case of https://github.com/anchlovi/unused-deps-specs2 where the method described above should address the problem.

ittaiz:
Few questions:

1. If we merge this PR won't we be in a problem in the case I originally described?
2. Re the algorithm- sounds really interesting. Can you elaborate a bit more about `Compute the strict deps, S`? How do you compute that? This will be many rules which try to follow the pattern of "only things that the source files demand", right? 
3. Why doesn't +1 work for external dependencies?

Jamie5:
1. As far as I can tell, no, and not sure why. It is possibly I mistested but it did not seem to be a problem. It was only the subclassing (without annotations, and with IIRC 4 subclasses of each other) that I got to be a problem
2. The computation of S is the same strict-deps we are currently doing, so only things directly referenced in the source files. If you mean R, then probably as we discover things we will add it (for the purposes of the algorithm, if R is empty it is acceptable). For example we now would want to, for each class, put all of its superclasses (and probably supertraits, but would need to test that) into R.
3. To be specific, it does not work for external dependencies for which we can't see their deps and for which we basically just download the jar. I don't remember the exact rule that does that. For external dependencies which have regular scala_library and so on I expect it would work with plus one.

ittaiz:
Re 1- maybe we should  add a test that shows it works? If 2.12.13 (fictive version I know) will break this I’d like to know.
Re 2- I did mean S. Wanted to verify. You clarified it, thanks.
Re 3- If you can share a repro I can see if it can be supported or suggest an alternative. In any case I’m happy to add a config toggle for that.

Jamie5:
1. So we want to add a test about plus one deps that this specific scenario works? But I thought +1 was just a heuristic anyways and it may not work well from one to another version, and has no guarantees. And it just happens to work well right now.

3. Unless I'm mistaken https://github.com/anchlovi/unused-deps-specs2 is exactly such a situation (discussed https://github.com/bazelbuild/rules_scala/issues/867#issuecomment-569374730 )

Jamie5:
Just tested again and yes, the annotation of superclass of superclass is still not a problem.

ittaiz:
Thanks.
Why not add a test that passes here? I understand we can give many examples that work but I feel this one can easily happen (mainly because we don’t understand why it works)

Jamie5:
Created https://github.com/bazelbuild/rules_scala/pull/998 for checking the exact dependencies that scalc needs.

ittaiz:
Thanks!

Issue #839, opened 2019-09-02
Move from label collection to label stamping


ittaiz:
Currently we collect labels in various places (scala_library/scala_binary/scala_import/etc) and propagate them onwards.
This is valuable since when we want to report to the user on a problem with a transitive dependency (unused deps/strict deps) we have a meaningful label and not just a file path.
Java rules also used a similar (but different) approach but have moved a year or more ago to a different approach. 
Their current approach "stamps" the jar upon creation* with its label (k/v in the manifest.mf) and so if they ever need to report (which is the rare case) then and only then they pay the (higher) price of opening the jar and reading the manifest to improve the user reporting.
This doesn't actually happen on jar creation but implicitly on ijar creation.
This means that we need to explicitly stamp our outputs for scala_import and scala_macro_library since they don't go through ijar. [java_common](https://docs.bazel.build/versions/master/skylark/lib/java_common.html#stamp_jar) has a special API for this.

Note that some tests will fail with this change since our existing mechanism allows us to remap labels (if `bar` exports `foo` then when someone uses `bar` the outputs of `foo` will be listed as `bar`). This is valuable in having facade targets for logical libraries.
After lengthy discussions it was agreed that the solution for this pattern should be in a smarter tool that will use bazel query to find the fact that `bar` exports `foo` and will add `bar` to the caller.
Also the changes in #716 are needed.

ittaiz:
@long-stripe wdyt about taking this one? Shouldn't be very complicated and will be a nice win (remove a lot of noise from our code)

long-stripe:
@ittaiz Yeah if you think it's a good thing to start on! I can't promise anything timeline wise, but I will try and carve out some personal time in the coming week or so to look at this.

ittaiz:
I do :)
Re timeline it's ok- we've been wanting to do this for about a year so
another month or two won't probably change anything...

long-stripe:
FYI small heartbeat - I am still interested in tackling this, but haven't had time lately.

ittaiz:
Looking forward to it. You're not waiting for something from me, right?

long-stripe:
@ittaiz I was just gonna meander about and deep dive in when I have the time, but if you have links to relevant parts of the codebase that I should be focusing on, that would be super helpful!

ittaiz:
Unfortunately it’s spread throughout the codebase. I’d start in the lead (compiler plugin) and trace back.

PR #995, opened 2020-02-07
Update dependency configuration options

Jamie5:
### Description
We expose several dependency configuration options in the scala toolchain.

These include options for dependency mode, strict deps mode, and dependency tracking method. In particular, this exposes an option for the new strict deps mode being developed in https://github.com/bazelbuild/rules_scala/issues/867 so that people can now start trying it out.

We deprecate a few methods as described in the readme, for future removal.

### Motivation
Part of improving the strict deps/unused deps checking for plus-one dependency mode, as detailed in https://github.com/bazelbuild/rules_scala/issues/867 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,13 +158,62 @@ replacements:
         "@io_bazel_rules_scala_scala_xml//:io_bazel_rules_scala_scala_xml"
 ```
 
-## [Experimental] Using strict-deps
-Bazel pushes towards explicit and minimal dependencies to keep BUILD file hygiene and allow for targets to refactor their dependencies without fear of downstream breaking.
-Currently rules_scala does this at the cost of having cryptic `scalac` errors when one mistakenly depends on a transitive dependency or, as more often the case for some, a transitive dependency is needed to [please scalac](https://github.com/scalacenter/advisoryboard/blob/master/proposals/009-improve-direct-dependency-experience.md) itself.
-To learn more about the motivation of strict-deps itself you can visit this Bazel blog [post](https://blog.bazel.build/2017/06/28/sjd-unused_deps.html) on the subject.
+## [Experimental] Dependency options
 
-To use it just add `--strict_java_deps=WARN|ERROR` to your `bazel` invocation.
-In both cases of `WARN` or `ERROR` you will get the following text in the event of a violation:
+There are a number of dependency options which can be set in the scala toolchain. These include `dependency_mode`, `strict_deps_mode`, `unused_dependency_checker_mode`, and `dependency_tracking_method`.
+
+### [Experimental] Recommended options
+
+We recommend one of the following sets of options
+
+**Option A**
+Accept the defaults, which might work well enough for you. The defaults are
+```
+  dependency_mode = "direct",
+  strict_deps_mode = "off",
+  unused_dependency_checker_mode = "off",
+  dependency_tracking_method = "high-level",
+```
+but you do not need to include this in the toolchain as they are the defaults.
+
+**Option B**
+```
+  dependency_mode = "plus-one",
+  strict_deps_mode = "error",
+  unused_dependency_checker_mode = "error",
+  dependency_tracking_method = "ast",
+```
+
+Should the first option result in too much effort in handling build files and the like due to confusing dependencies and you becoming confused as to why some specific dependency is needed when the code being compiled never references it, consider this set of options. It will include both dependencies and dependencies of dependencies, which in practice is enough stop almost all strange missing dependency errors at the cost of somewhat more incremental compile cost in certain cases.
+
+With these settings, we also will error on dependencies which are unneeded, and dependencies which should be included in `deps` due to be directly referenced in the code, but are not.
+
+The dependency tracking method `ast` is experimental but so far proves to be better than the default for computing the direct dependencies for `plus-one` mode code. In the future we hope to make this the default for `plus-one` mode and remove the option altogether.
+
+### [Experimental] Dependency mode
+
+There are three dependency modes. The reason for the multiple modes is that often `scalac` depends on jars which seem unnecessary at first glance. Hence, in order to reduce the need to please `scalac`, we provide the following options.
+- `dependency_mode = "direct"` - only include direct dependencies during compiliation; that is, those in the `deps` attribute
+- `dependency_mode = "plus-one"` - only include `deps` and `deps` of `deps` during compiliation.
+- `dependency_mode = "transitive"` - all transitive dependencies are included during compiliation. That is, `deps`, `deps` of `deps`, `deps` of `deps` of `deps`, and so on.
+
+When using `direct` mode, there can be cryptic `scalac` errors when one mistakenly depends on a transitive dependency or, as more often the case for some, a transitive dependency is needed to [please scalac](https://github.com/scalacenter/advisoryboard/blob/master/proposals/009-improve-direct-dependency-experience.md) itself.
+
+As one goes down the list, more dependencies are included which helps reduce confusing requirements to add `deps`, at the cost of increased incremental builds due to a greater number of dependencies. In practice, using `plus-one` deps results in almost no confusing `deps` entries required while still being relatively small in terms of the number of total dependencies included.
+
+**Caveats for `plus_one` and `transitive`:**
+<ul>
+    <li>Extra builds- Extra dependencies are inputs to the compilation action which means you can potentially have more build triggers for changes the cross the ijar boundary </li>
+    <li>Label propagation- since label of targets are needed for the clear message and since it's not currently supported by JavaInfo from bazel we manually propagate it. This means that the error messages have a significantly lower grade if you don't use one of the scala rules or scala_import (since they don't propagate these labels)</li>
+    <li>javac outputs incorrect targets due to a problem we're tracing down. Practically we've noticed it's pretty trivial to understand the correct target (i.e. it's almost a formatting problem) </li>
+  </ul>
+
+### [Experimental] Strict deps mode
+We have a strict dependency checker which requires that any type referenced in a scala rule should be included in that rule's deps. To learn about the motivation for this you can visit this Bazel blog [post](https://blog.bazel.build/2017/06/28/sjd-unused_deps.html) on the subject.
+
+The option `strict_deps_mode` can be set to `off`, `warn`, or `error`. We highly recommend setting it to `error`.
+
+In both cases of `warn` or `error` you will get the following text in the event of a violation:
 ```
 ...
 Target '//some_package:transitive_dependency' is used but isn't explicitly declared, please add it to the deps.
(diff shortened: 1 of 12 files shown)


Jamie5:
This adds a bunch of logic in the toolchain - not sure if that is the best place to put it, but it does simplify code in the plus one aspect. So not sure if the goal of moving all logic to the phase_dependency will be where we end or if there should be something else happening here.

Jamie5:
A retry succeeded so suspect the issue in CI was just flaky (and the original error did look flaky)

scala/scala_toolchain.bzl
```diff
@@ -3,12 +3,74 @@ load(
...
 def _scala_toolchain_impl(ctx):
+    if ctx.attr.plus_one_deps_mode != "":
+        print(
+            "Setting plus_one_deps_mode on toolchain is deprecated." +
+            "Use 'dependency_mode = \"plus-one\"' instead",
+        )
+    if ctx.attr.dependency_mode != "" and ctx.attr.plus_one_deps_mode != "":
+        fail("Cannot set both dependency_mode and plus_one_deps_mode on toolchain")
+
+    if ctx.fragments.java.strict_java_deps != "default" and ctx.fragments.java.strict_java_deps != "off":
```
ittaiz: this is for backward compatibility?

ittaiz: also, does this mean that I'll be able to run with +1 only if I set strict_java_deps to warn/error? Sounds like a bug

README.md
```diff
@@ -158,31 +158,71 @@ replacements:
...
+```
+  dependency_mode = "plus-one",
+  strict_deps_mode = "error",
+  unused_dependency_checker_mode = "error",
+  dependency_tracking_method = "ast",
 ```
-Note that if you have `buildozer` installed you can just run the last line and have it automatically apply the fix for you.
 
-**Caveats:**
+Should the first option result in too much effort in handling build files and the like due to confusing dependencies and you becoming confused as to why some specific dependency is needed when the code being compiled never references it, consider this set of options. It will include both dependencies and dependencies of dependencies, which in practice is enough stop almost all strange missing dependency errors at the cost of somewhat more incremental compile cost in certain cases.
```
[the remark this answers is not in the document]
Jamie5: Will fix

README.md
```diff
@@ -158,31 +158,71 @@ replacements:
...
+
+With these settings, we also will error on dependencies which are unneeded, and dependencies which should be included in `deps` due to be directly referenced in the code, but are not.
+
+The dependency tracking method `ast` is experimental but so far proves to be better than the default for computing the direct dependencies for `plus-one` mode code. In the future we hope to make this the default for `plus-one` mode and remove the option altogether.
+
+### [Experimental] Dependency mode
```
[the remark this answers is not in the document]
Jamie5: You mean that for external dependencies they ought to export the dependencies their ijar needs, or something else?

README.md
```diff
@@ -158,31 +158,71 @@ replacements:
...
+- `dependency_mode = "direct"` - only include direct dependencies during compiliation; that is, those in the `deps` attribute
+- `dependency_mode = "plus-one"` - only include `deps` and `deps` of `deps` during compiliation.
+- `dependency_mode = "transitive"` - all transitive dependencies are included during compiliation. That is, `deps`, `deps` of `deps`, `deps` of `deps` of `deps`, and so on.
+
+When using `direct` mode, there can be cryptic `scalac` errors when one mistakenly depends on a transitive dependency or, as more often the case for some, a transitive dependency is needed to [please scalac](https://github.com/scalacenter/advisoryboard/blob/master/proposals/009-improve-direct-dependency-experience.md) itself.
+
+As one goes down the list, more dependencies are included which helps reduce confusing requirements to add `deps`, at the cost of increased incremental builds due to a greater number of dependencies. In practice, using `plus-one` deps results in almost no confusing `deps` entries required while still being relatively small in terms of the number of total dependencies included.
+
+**Caveats for `plus_one` and `transitive`:**
```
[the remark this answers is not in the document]
Jamie5: Ok, will add something

README.md
```diff
@@ -158,31 +158,71 @@ replacements:
...
     <li>Label propagation- since label of targets are needed for the clear message and since it's not currently supported by JavaInfo from bazel we manually propagate it. This means that the error messages have a significantly lower grade if you don't use one of the scala rules or scala_import (since they don't propagate these labels)</li>
     <li>javac outputs incorrect targets due to a problem we're tracing down. Practically we've noticed it's pretty trivial to understand the correct target (i.e. it's almost a formatting problem) </li>
   </ul>
 
-Note: Currently strict-deps is protected by a feature toggle but we're strongly considering making it the default behavior as `java_*` rules do.
+### [Experimental] Strict deps mode
+We have a strict dependency checker which requires that any type referenced in a scala rule should be included in that rule's deps. To learn about the motivation for this you can visit this Bazel blog [post](https://blog.bazel.build/2017/06/28/sjd-unused_deps.html) on the subject.
```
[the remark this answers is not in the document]
Jamie5: Will fix

scala/scala_toolchain.bzl
Jamie5: Yes for backwards compatibility.

Don't think it would mean you need to set strict_java_deps to run with +1. As this is saying, if strict_java_deps is set to warn/error then do this set of specific things which match previous behavior. And if not, then do the whole thing about reading from dependency_mode/plus_one_deps_mode to compute the dependency mode, etc, etc.

README.md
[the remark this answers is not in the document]
ittaiz: No, I mean that exports also get added on the classpath. It’s not 100% related here but your description is so detailed and awesome so it can fill in the last part (or close to last)

scala/scala_toolchain.bzl
ittaiz: You’re right. I misread.

README.md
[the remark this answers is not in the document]
Jamie5: Ok will try to add something

test/toolchains/BUILD.bazel
```diff
@@ -0,0 +1,77 @@
...
+scala_toolchain(
+    name = "ast_plus_one_deps_unused_deps_warn_impl",
+    dependency_mode = "plus-one",
+    dependency_tracking_method = "ast",
+    unused_dependency_checker_mode = "warn",
+    visibility = ["//visibility:public"],
+)
+
+toolchain(
+    name = "ast_plus_one_deps_unused_deps_warn",
```
ittaiz: is this used? Can't see it and I think maybe unused deps warn isn't tested
I think this [line](https://github.com/bazelbuild/rules_scala/pull/995/files#diff-23f6e936cb498d27c3ef44340314fd82R53) was supposed to be a different test that used it but was forgotten

test/toolchains/BUILD.bazel
```diff
@@ -0,0 +1,77 @@
...
+scala_toolchain(
+    name = "ast_plus_one_deps_strict_deps_warn_impl",
+    dependency_mode = "plus-one",
+    dependency_tracking_method = "ast",
+    strict_deps_mode = "warn",
+    visibility = ["//visibility:public"],
+)
+
+toolchain(
+    name = "ast_plus_one_deps_strict_deps_warn",
```
ittaiz: Also this isn't used

test/shell/test_unused_dependency.sh
```diff
@@ -47,8 +47,41 @@ test_unused_dependency_fails_even_if_also_exists_in_plus_one_deps() {
...
+    exit 1
+  fi
+
+  local expected="warning: Target '//test_expect_failure/plus_one_deps/with_unused_deps:c' is specified as a dependency to //test_expect_failure/plus_one_deps/with_unused_deps:a but isn't used, please remove it from the deps."
+
+  echo "$output" | grep "$expected"
+  if [ $? -ne 0 ]; then
+    echo "Expected output:[$output] to contain [$expected]"
+    exit 1
+  fi
```
ittaiz: this is needed because `action_should_fail` doesn't allow to assert on the output? Maybe it should? If not maybe it should be extracted to a private method? It was a bit hard to understand what's going on (and maybe indeed I misunderstood)

README.md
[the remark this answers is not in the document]
ittaiz: You didn't forget about this one, right?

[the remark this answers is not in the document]
Jamie5: This is the `Note: the last two issues are bugs which will be addressed by [https://github.com/bazelbuild/rules_scala/issues/839].` below the list itself

README.md
```diff
@@ -197,6 +197,8 @@ There are three dependency modes. The reason for the multiple modes is that ofte
 - `dependency_mode = "plus-one"` - only include `deps` and `deps` of `deps` during compiliation.
 - `dependency_mode = "transitive"` - all transitive dependencies are included during compiliation. That is, `deps`, `deps` of `deps`, `deps` of `deps` of `deps`, and so on.
 
+Note when a dependency is included, that means its jar is included on the classpath, along with the jars of any rules that it exports.
```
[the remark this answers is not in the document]
Jamie5: will do

test/toolchains/BUILD.bazel
Jamie5: Yes it was supposed to be used in test_plus_one_ast_analyzer_unused_deps

test/toolchains/BUILD.bazel
Jamie5: I think this is used in test_deps.sh on the last line of test_plus_one_ast_analyzer_strict_deps

test/shell/test_unused_dependency.sh
Jamie5: Ok so this was supposed to be for warn. So it's puzzling that the test passed and I need to look into that.

The logic was copied from test_unused_dependency_checker_mode_warn and therefore it is 100%  appropriate that we don't need to do anything like pull it into a function or something. (Okay the real reason is these days I do all my utility scripts in python and am terrified about bash. But will try to see if we can indeed extract a function)

Jamie5: Ok so action_should_fail does `exit 0` on success so only the first test was run -_-. Not sure why we want that but will separate these tests out.

Luckily test_expect_failure_or_warning_on_missing_direct_deps_with_expected_message does not `exit 0` on success so we don't have to copy the messages around into separate functions. Though admittedly if the policy is to do an `exit 0` then its kind of risky.

ittaiz: I missed the exit 0 part. If you can reiterate that would be great.
Also two more small points:
1. Do you also need to split the tests in test_deps.sh?
2. I’m glad you’re terrified from bash, it’s awful :) In general please make sure to triple check every test you’ve added/changed fails and if you change it refactor the smallest thing please make sure it still fails by changing/commenting out the prod code. We’ve had multiple issues in the past with bash tests silently failing...

Jamie5: What do you mean by reiterate?

1. No because the command we run there doesn't call exit 0
2. Ok will do so in the future. If the situation is so bad wouldn't even just converting all existing tests to something where at least you won't be constantly surprised be a good idea? e.g. python or something.

ittaiz: We don’t change them that often so the price is usually contained.
Also what we’d really like to do is move these tests inside to Bazel (using https://github.com/bazelbuild/bazel-integration-testing) but that’s a complicated task

PR #998, opened 2020-02-13
Add test for scalac dependencies

Jamie5:
### Description
We add some tests that scalac needs the dependencies that we think they need.

### Motivation
As part of https://github.com/bazelbuild/rules_scala/issues/867 it will be useful to know whenever our assumptions is violated by e.g. a new scala version.

### DIFF
--- a/third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/ScalacDependencyTest.scala
+++ b/third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/ScalacDependencyTest.scala
@@ -13,6 +13,10 @@ import third_party.utils.src.test.io.bazel.rulesscala.utils.TestUtil
  * dependencies. That is, for given scenarios, we want to make sure
  * that scalac requires the given set of dependencies; no more and
  * no less.
+ *
+ * To clarify: these tests do not reflect the end result of strict/unused
+ * deps as we are considering alternatives of how to mitigate scalac's
+ * limitations.
  */
 class ScalacDependencyTest extends FunSuite {
   private def withSandbox(action: Sandbox => Unit): Unit = {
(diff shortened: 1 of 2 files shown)


third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/ScalacDependencyTest.scala
```diff
@@ -0,0 +1,105 @@
...
+      sandbox.compile("class A extends scala.annotation.StaticAnnotation")
+      sandbox.compile("@A class B")
+      sandbox.checkExactDepsNeeded(
+        code = "class C extends B",
+        deps = List("B")
+      )
+    }
+  }
+
+  test("superclass of superclass needed") {
```
ittaiz: I’m confused; I thought this is the poster child of us not wanting scalac’s limitations for spill over, no? The fact that scalac needs A doesn’t mean C actually uses A

Jamie5: Yes, this is testing scalac's behavior. In this case that it needs A anyways. These dependencies affect set `R` in the previous discussion. And should this test break it means that the computation of `R` would need to change. (not that we have a computation of R in the code yet, I guess we should figure if that approach is indeed what we want)

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/ScalacDependencyTest.scala
```diff
@@ -0,0 +1,105 @@
...
+import org.apache.commons.io.FileUtils
+import org.scalatest._
+import third_party.utils.src.test.io.bazel.rulesscala.utils.JavaCompileUtil
+import third_party.utils.src.test.io.bazel.rulesscala.utils.TestUtil
+
+/**
+ * Test that the scalac compiler behaves how we expect it to around
+ * dependencies. That is, for given scenarios, we want to make sure
+ * that scalac requires the given set of dependencies; no more and
+ * no less.
```
ittaiz: WDYT about adding a sentence at the end saying something like:
```
To clarify- these tests do not reflect the end result of strict-deps/unused-deps as we're considering alternatives of how to mitigate scalac's limitations.
```
Suggesting this mainly since the purpose of this test wasn't 100% clear to me

ittaiz:
Thanks!

PR #1001, opened 2020-02-16
remove spotify from adopters

ittaiz:
### Description
remove spotify from adopters

They looked at Bazel but didn't adopt

### Motivation
Up-to-date adopters list

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,7 +317,6 @@ Here's a (non-exhaustive) list of companies that use `rules_scala` in production
 * [Grand Rounds](http://grandrounds.com/)
 * [Kitty Hawk](https://kittyhawk.aero/)
 * [Meetup](https://meetup.com/)
-* [Spotify](https://www.spotify.com/)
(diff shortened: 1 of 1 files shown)


ittaiz:
@nevillelyh I'm not mistaken, right?

mandrean:
@ittaiz There are 2-3 teams using it and a few more are considering it. But most of the company will remain on whatever is "default" for a particular language, like e.g. Maven, for the foreseeable future.

mandrean:
And the only reason why I found this PR is because I'm working on an internal project and needed to pin `rules_scala` to the latest commit hash 😂

ittaiz:
Thanks! I met with @nevillelyh in NYC and understood no one uses it.
Happy to hear some teams are getting value from it.
Re-adding you.

nevillelyh:
@ittaiz thanks. You’re correct that we (the scala infra team) don’t use it but there are a few downstream (feature teams) that do.

PR #1002, opened 2020-02-17
 specify deps for specs2 and specs2-junit deps 

ittaiz:
### Description
 specify deps for specs2 and specs2-junit deps 

### Motivation
Strict-deps and unused-deps https://github.com/bazelbuild/rules_scala/issues/867#issuecomment-586413125

### DIFF
--- a/test/aspect/aspect.bzl
+++ b/test/aspect/aspect.bzl
@@ -57,8 +57,8 @@ def _rule_impl(ctx):
     }
     content = ""
     for target in ctx.attr.targets:
-        visited = sorted(target.visited)
-        expected = sorted(expected_deps[target.label.name])
+        visited = depset(sorted(target.visited)).to_list()
+        expected = depset(sorted(expected_deps[target.label.name])).to_list()
         if visited != expected:
             content += """
             echo Expected these deps from {name}: 1>&2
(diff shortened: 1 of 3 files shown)


ittaiz:
@anchlovi would love your review.
@Jamie5 FYI, I think this is what was missing

test/aspect/aspect.bzl
```diff
@@ -53,8 +57,8 @@ def _rule_impl(ctx):
     }
     content = ""
     for target in ctx.attr.targets:
-        visited = sorted(target.visited)
-        expected = sorted(expected_deps[target.label.name])
+        visited = depset(sorted(target.visited)).to_list()
+        expected = depset(sorted(expected_deps[target.label.name])).to_list()
```
Jamie5: Stupid question, is the point of doing this to dedup these lists by using the depset which will not return duplicates, and I guess the order of iteration will be identical in this case due to the sorting? Is the identical order guaranteed by depset, based on the docs wasn't fully sure.

Jamie5:
Oh yes makes sense, did not know you could use deps here but actually it's right there in the docs and for exactly this reason.

test/aspect/aspect.bzl
ittaiz: The default order is undefined but deterministic which IIUC gives us the equality we’re seeking for.
I think this test is obsolete but I’m not sure and would rather not delete it just this instance.

ittaiz: Waiting for your 👍🏽👎🏽

Jamie5: Oh I thought deterministic just meant that for a given specific input the iteration order would be the same across different calls. But I guess if it works it works.

PR #1006, opened 2020-02-23
Fix source pathing for code coverage

d-haxton:
### Description

Fixes code coverage output to more aptly point to the source file. This allows tools like intellij's code coverage to work without any other issues.

Should also allow java coverage that's pulled into a scala library.

I'm open to making changes here, but the main idea is that bazel / jacoco allows for a file in the jar that ends with `-paths-for-coverage.txt` and rather than stitching together a file name from its package it will leverage these source 'hints' we provide it. 

### Motivation

I wanted code coverage in intellij.

### DIFF
--- a/scala/private/phases/phase_coverage.bzl
+++ b/scala/private/phases/phase_coverage.bzl
@@ -36,9 +36,11 @@ def _phase_coverage(ctx, p, srcjars):
         output_jar = ctx.actions.declare_file(
             "{}-offline.jar".format(input_jar.basename.split(".")[0]),
         )
+        
         srcs_paths = []
-        for src in ctx.files.srcs: 
+        for src in ctx.files.srcs:
             srcs_paths = srcs_paths + [src.path]
+
         in_out_pairs = [
             (input_jar, output_jar, srcs_paths),
         ]
--- a/scala/private/phases/phase_coverage.bzl
+++ b/scala/private/phases/phase_coverage.bzl
@@ -38,26 +38,26 @@ def _phase_coverage(ctx, p, srcjars):
         )
         srcs_paths = []
         for src in ctx.files.srcs:
-            srcs_paths = srcs_paths + [src.path]
-        in_out_pairs = [
+            srcs_paths.append(src.path)
+        records = [
             (input_jar, output_jar, srcs_paths),
         ]
 
         args = ctx.actions.args()
-        args.add_all(in_out_pairs, map_each = _jacoco_offline_instrument_format_each)
+        args.add_all(records, map_each = _jacoco_offline_instrument_format_each)
         args.set_param_file_format("multiline")
         args.use_param_file("@%s", use_always = True)
 
         ctx.actions.run(
             mnemonic = "JacocoInstrumenter",
-            inputs = [in_out_pair[0] for in_out_pair in in_out_pairs],
-            outputs = [in_out_pair[1] for in_out_pair in in_out_pairs],
+            inputs = [record[0] for record in records],
+            outputs = [record[1] for record in records],
             executable = ctx.attr._code_coverage_instrumentation_worker.files_to_run,
             execution_requirements = {"supports-workers": "1"},
             arguments = [args],
         )
 
-        replacements = {i: o for (i, o, _) in in_out_pairs}
+        replacements = {i: o for (i, o, _) in records}
         provider = _coverage_replacements_provider.create(
             replacements = replacements,
         )
@@ -75,5 +75,5 @@ def _phase_coverage(ctx, p, srcjars):
             },
         )
 
-def _jacoco_offline_instrument_format_each(in_out_pair):
-    return (["%s=%s=%s" % (in_out_pair[0].path, in_out_pair[1].path, ",".join(in_out_pair[2]))])
+def _jacoco_offline_instrument_format_each(records):
+    return (["%s=%s=%s" % (records[0].path, records[1].path, ",".join(records[2]))])
--- a/scala/private/phases/phase_coverage.bzl
+++ b/scala/private/phases/phase_coverage.bzl
@@ -36,7 +36,7 @@ def _phase_coverage(ctx, p, srcjars):
         output_jar = ctx.actions.declare_file(
             "{}-offline.jar".format(input_jar.basename.split(".")[0]),
         )
-        src_paths = [src.path for src in ctx.files.srcs]
+        srcs_paths = [src.path for src in ctx.files.srcs]
         records = [
             (input_jar, output_jar, srcs_paths),
         ]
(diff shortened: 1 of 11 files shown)


test/coverage/BUILD
```diff
@@ -62,20 +62,10 @@ scala_library(
         "A2.scala",
     ],
     deps = [
-        # TODO :: Understand why referencing a local java library breaks coverage
```
johnynek: Nice to fix this!

src/java/io/bazel/rulesscala/coverage/instrumenter/JacocoInstrumenter.java
```diff
@@ -98,6 +99,13 @@ private FileVisitResult actuallyVisitFile(Path inPath, BasicFileAttributes attrs
                 } else {
                     Files.copy(inPath, outPath);
                 }
+
+                if(outPath.toString().endsWith(".class")) {
```
ittaiz: isn't this condition always true when `inPath.toString().endsWith(".class")`? If so why add another condition and not add it to the above? Probably best to extract the current code there to a private method and also the body of the `then` here to a private method for clarity.

src/java/io/bazel/rulesscala/coverage/instrumenter/JacocoInstrumenter.java
```diff
@@ -98,6 +99,13 @@ private FileVisitResult actuallyVisitFile(Path inPath, BasicFileAttributes attrs
                 } else {
                     Files.copy(inPath, outPath);
                 }
+
+                if(outPath.toString().endsWith(".class")) { 
+                    Files.write(
+                        outFS.getPath((outPath.toString() + "-paths-for-coverage.txt")),
```
ittaiz: There's something I'm not sure I understand.
IIUC `srcs` contains the list of all sources for the jar we're processing, right?
At the same time `outPath` represents a single `.class` file (in the output jar but I'm not sure it's important), right?
If so why do we add another txt file per each .class file and in the .txt write all of the sources?

d-haxton: Ah! You are correct. I was incorrectly making an assumption off of a comment in bazel which read:

```
If a jar contains uninstrumented classes it will also contain a txt file with the paths of each of these classes, one on each line.
```

To which I interpreted as must have a corresponding `-paths-for-coverage.txt`, but that is incorrect, it just needs a single file named as such.

src/java/io/bazel/rulesscala/coverage/instrumenter/JacocoInstrumenter.java
d-haxton: This should be moot now.

src/java/io/bazel/rulesscala/coverage/instrumenter/JacocoInstrumenter.java
```diff
@@ -69,6 +70,11 @@ private void processArg(Instrumenter jacoco, String arg) throws Exception {
                     throw new RuntimeException(e);
                 }
             });
+
+            Files.write(
+                outFS.getPath("-paths-for-coverage.txt"),
```
ittaiz: maybe add a link to the bazel doc or jacoco doc (or both) that explain about this feature?

src/java/io/bazel/rulesscala/coverage/instrumenter/JacocoInstrumenter.java
ittaiz: I see.
My main concern then is that no test failed when the implementation actually didn't work, right?
Or do both versions work?
And finally- if I change the filename to `-paths-blah-for-coverage.txt` will any test break? Can we add such a test? I think it can be super valuable.

d-haxton: Both implementations did actually work. 

What happens in bazel/jacoco is quite crude. It essentially scans all JARs for `-paths-for-coverage.txt` and adds them all to a single set. Regardless of what jar it comes from. 

Yes, if you do change to that file name it will fail (see the two commits I just pushed).

src/java/io/bazel/rulesscala/coverage/instrumenter/JacocoInstrumenter.java
d-haxton: Ha! Unfortunately, this is fairly undocumented in bazel / bazel's implementation of jacoco. I updated the comments to reflect this as best I could. Let me know if I should add some more color.

ittaiz: This is awesome, thanks!
I think one more thing we should do (I can do it if you prefer) is to open a bazel issue for documentation m. The main reason I want this is to have them acknowledge this is a public API and not a hack they will break.

ittaiz:
I want to see the failure and see that it makes sense but after that I’m good.
@johnynek, on honorable visitor 😉, any more issues on your side?

d-haxton:
As mentioned in the other link: https://travis-ci.org/bazelbuild/rules_scala/jobs/654605424?utm_medium=notification&utm_source=github_status

That is the failing coverage test. Does that make sense to you? Otherwise curious what else you're looking for.

src/java/io/bazel/rulesscala/coverage/instrumenter/JacocoInstrumenter.java
d-haxton: Filed in: https://github.com/bazelbuild/bazel/issues/10857

test/coverage/BUILD
```diff
@@ -62,20 +62,10 @@ scala_library(
...
-        # ":b2",
+        ":b2",
     ],
 )
 
-#
```
andyscott: Very happy to see this comment go away

scala/private/phases/phase_coverage.bzl
```diff
@@ -36,25 +36,28 @@ def _phase_coverage(ctx, p, srcjars):
         output_jar = ctx.actions.declare_file(
             "{}-offline.jar".format(input_jar.basename.split(".")[0]),
         )
-        in_out_pairs = [
-            (input_jar, output_jar),
+        srcs_paths = []
+        for src in ctx.files.srcs:
```
andyscott: Can we do `src_paths = [src.path for src in ctx.files.srcs]`? IMHO it's a bit cleaner and more readable to do a list comprehension than a loop with mutation.

d-haxton:
Also note that I changed the package in the test files to workaround a bazel bug. I've filed an issue here: https://github.com/bazelbuild/bazel/issues/10863 and opened a PR that should address that.

PR #1007, opened 2020-02-25
update to the latest worker protobuf file

andyscott:
### Description

Updates our copy of the worker protobuf file from ~https://github.com/bazelbuild/rules_scala/blob/master/third_party/bazel/src/main/protobuf/worker_protocol.proto~ https://github.com/bazelbuild/bazel/blob/master/src/main/protobuf/worker_protocol.proto.

### Motivation

The worker protobuf protocol has been updated to support multiplexing workers which can handle multiple concurrent requests in a single process. We can eventually leverage this to support parallel Scalac unit compilation without running extra JVM processes.

### DIFF
--- a/third_party/bazel/src/main/protobuf/worker_protocol.proto
+++ b/third_party/bazel/src/main/protobuf/worker_protocol.proto
@@ -38,14 +38,25 @@ message WorkRequest {
   // The inputs that the worker is allowed to read during execution of this
   // request.
   repeated Input inputs = 2;
+
+  // To support multiplex worker, each WorkRequest must have an unique ID. This
+  // ID should be attached unchanged to the WorkResponse.
+  int32 request_id = 3;
 }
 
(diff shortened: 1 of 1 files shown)


ittaiz:
Thanks! Multiplexing can be really valuable.. Devs at Wix indeed complain about the persistent workers overhead. Where is this proto copied from?

andyscott:
> Where is this proto copied from?

🙄 

> Updates our copy of the worker protobuf file from https://github.com/bazelbuild/rules_scala/blob/master/third_party/bazel/src/main/protobuf/worker_protocol.proto.

ittaiz:
I think I’m missing something. The path you gave is the path you’re updating. What’s the source?

andyscott:
Whoops

andyscott:
https://github.com/bazelbuild/bazel/blob/master/src/main/protobuf/worker_protocol.proto

PR #1009, opened 2020-02-26
Don't unnecessarily use fullyExploreTree

Jamie5:
### Description
We remove unnecessary usage of foreach with fullyExploreTree in the AST used jar finder and simplify the interface.

### Motivation
In two places in the code we were calling `Tree.foreach(fullyExploreTree)`. This is redundant because `fullyExploreTree` calls `foreach` on the tree already. So we change those usages to just `fullyExploreTree(Tree)`.

We also change `fullyExploreTree` to not call `exploreTree` on its parameter, because this is redundant with `tree.foreach`.

Nobody should be calling `exploreTree` as it only partially explores the tree and everyone should just be fully exploring the tree, so we move `exploreTree` into `fullyExploreTree` and rename it `visitNode` (which is still not the best of names).


Jamie5:
This was tested against a relatively large internal codebase (in additional to normal tests) and that did not turn up any issues.

ittaiz:
Can you rebase post #1008?

PR #1010, opened 2020-02-27
Return resource's short_path when fail to find relative

d-haxton:
### Description

There was a breaking change in: https://github.com/bazelbuild/rules_scala/pull/949/files which when failing to find the relative path for resources it would return the entire path, which typically included `/bazel-out/[...]/<short_path>`

I'm assuming there is also going to be some breakage if someone does similar to this and has the literal folder name `resources` in their path. But maybe that's for a different PR / issue.

I don't really know the best approach to this, so I'm just checking to see if it starts with `bazel-out`. Either way, I've added a test in which I'd expect to be the correct functionality.
Alternatively, we could always do the short-path and use the old way for `external`:

```
        // paths to external repositories, for some reason, start with a leading ../	
        // we don't want to copy the resource out of our temporary directory, so	
        // instead we replace ../ with external/	
        // since "external" is a bit of reserved directory in bazel for these kinds	
        // of purposes, we don't expect a collision in the paths.
```

### Motivation
Unable to upgrade to newest version of rules_scala due to this bug.


d-haxton:
Right, so actually the test case that I have is "wrong" and actually showing that breakage with `resources`! 

I would actually expect the assert to look like:
`getClass.getResource("/test/src/main/scala/scalarules/test/resources/py_resource_binary")`, but because the BUILD path is `test/src/main/scala/scalarules/test/resources/`, it just strips everything out before resources and calls it a day. 

This will only break you when you're trying to reference a resource that is the output bazel compilation. If you do not attempt to do that you should be fine. If that doesn't make sense, I'll create an issue later where I can give it some more thought on explaining. 

I'll update the class name and test case again in a little bit (this change unblocks us at Asana), so I think it should be fine to fix this immediate issue and follow-up with the resources in path.

ittaiz:
I’d love to understand what functionality broke and what functionality you think your fix will break.
Is there urgency about you upgrading? Not that it should take weeks or even days but to understand if you need something really fast

d-haxton:
I don't think there's an immediate need to upgrade. I have some bandwidth at the moment and figured I'd try to get it done. 

The functionality that broke was we have a typescript (empty / dummy) library that generates several json files based off a single yaml file. From there we load that target as a resource into a scala_library. Then to access those json files we were calling `getClass.getResource("<short_path>/<target>.json")`. The problem we're having is that with the changes to how resource pathing works, what we end up getting is `bazel-out/darwin-fastbuild/.../<short_path>/<target>.json` which is unusable for us as that breaks between mac and linux builds anyways. 

With regards to the PR I have up: 
- The fix is valid and I don't think it should break anything new
- The test case I have is invalid because it's being affected by the `resources` in path bug.

ittaiz:
I see.
Sounds good and thank you for the explanation.
If you can fix the test case to show this problem (and see that it fails informatively like you expect before the fix) that would be great and I’ll merge it.

To verify- is there a resources issue or is it non existent because we have specific handling for “resources”?

d-haxton:
I've updated the test to get out of the `resources` folder so that we're not complicating things right now. 

The expected test cases (https://github.com/bazelbuild/rules_scala/pull/1010/files#diff-f439e50ed247a4809838a206d3761ec2R5) will fail against master: https://buildkite.com/bazel/rules-scala-scala/builds/1837

They should now be passing with latest commit. 

There is a resources issue, and I will follow up and open another issue with a better repro so we're not conflating these two issues.

test/ScalaTestResourcesFromLocalTargetTest.scala
```diff
@@ -0,0 +1,8 @@
+package scalarules.test
+
+class ScalaTestResourcesFromLocalTargetTest extends org.scalatest.FlatSpec {
+  "scala_test's resources when referencing local target" should "assert that local target is not placed in bazel-out, but rather next to the generated code" in {
+    assert(getClass.getResource("/bazel-out/darwin-fastbuild/bin/test/py_resource_binary") == null)
```
ittaiz: If I'm correctly reading the test failure you linked to previously this assertion passed and only the next one failed.
What is its purpose?
Not saying you should remove it but I'm not 100% sure about it

d-haxton: Huh. That _should_ never pass. I pushed up a branch with just that assert and it seems to fail. Not sure what you were looking at, but I'd be very curious to know how that happened.

ittaiz: 👍🏽 I was looking at the test before the line break so line 6 was the second assertion.

ittaiz:
Thanks! Looking forward to the other issue

arjantop:
@d-haxton Was just debugging this issue (before your fix) and I can confirm that external resources were located at bazel-out/... in the `_deploy.jar`. But with this fix they are not even present in the `_deploy.jar`. Is this related to the next issue that you mentioned? Have you had a time to work on it yet?

PR #1013, opened 2020-02-28
Fixed protobuf URL in README

pawelaw:
### Description

Fixed protobuf URL in README

### Motivation

With the last protobuf version update in the README, the URL needs to include the 'v' prefix before the version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ protobuf_version_sha256="cf754718b0aa945b00550ed7962ddc167167bd922b842199eeb6505
 
 http_archive(
     name = "com_google_protobuf",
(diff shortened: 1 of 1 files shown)


ittaiz:
What about the strip prefix part? Also CLA is needed

pawelaw:
@googlebot I signed it!

pawelaw:
> What about the strip prefix part? Also CLA is needed

The strip_prefix value is relevant only to the content of the archive - inside the archive there is a top level directory protobuf-3.11.3 (without the 'v'). Seems like there's some inconsistency. Nevertheless, the download URL requires 'v' prefix.

gergelyfabian:
Thanks @pawelaw for fixing the bug I introduced... :)

PR #1016, opened 2020-02-28
Add tests for imports

Jamie5:
### Description
Add tests around the behavior of imports in the ast dependency checker

### Motivation
This documents existing behavior of the ast dependency checker.

### DIFF
--- a/third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinderTest.scala
+++ b/third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinderTest.scala
@@ -50,6 +50,13 @@ class AstUsedJarFinderTest extends FunSuite {
       )
     }
 
+    private def verifyAndConvertDepToClass(dep: String): String = {
+      val classPath = tmpDir.resolve(s"$dep.class")
+      // Make sure the dep refers to a real file
+      assert(classPath.toFile.isFile)
+      classPath.toString
+    }
+
     def checkStrictDepsErrorsReported(
       code: String,
       expectedStrictDeps: List[String]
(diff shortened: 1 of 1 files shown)


third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinderTest.scala
```diff
@@ -421,6 +421,76 @@ class AstUsedJarFinderTest extends FunSuite {
...
+        s"""
+           |package foo.bar
+           |
+           |object A
+           |""".stripMargin
+      )
+      sandbox.checkUnusedDepsErrorReported(
+        code =
+          s"""
+             |import foo.bar
```
ittaiz: if we had here `foo.bar._` you would have received the type information we need, right?

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinderTest.scala
```diff
@@ -421,6 +421,76 @@ class AstUsedJarFinderTest extends FunSuite {
...
+        s"""
+           |package foo
+           |
+           |object A
+           |""".stripMargin
+      )
+      sandbox.checkUnusedDepsErrorReported(
+        code =
+          s"""
+             |import foo.A
```
ittaiz: wdyt about having the same object in this test and the previous one and so to emphasize that the difference is the way people import it? Maybe in all three tests the A can be identical

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinderTest.scala
Jamie5: No because `foo.bar` does not have a type attached. Added a case

third_party/dependency_analyzer/src/test/io/bazel/rulesscala/dependencyanalyzer/AstUsedJarFinderTest.scala
Jamie5: Sounds good

PR #1018, opened 2020-02-29
Enable coverage testing for java_test

d-haxton:
### Description
Fixes coverage `java_test` and adds a test to make sure it works / continues to work.

### Motivation

Since we can now support `java_test`, might as well test it and make sure it stays supported.


ittaiz:
The build here fails. 
Also I think that maybe it's better to split to two tests? And have the target be a bit more descriptive?

ittaiz:
@d-haxton are you picking this up?

liucijus:

PR #1026, opened 2020-03-31
upgrading scalapb

tian000:
### Description

Upgrading scalapb to the latest version supported for Scala 2.11

Since this is a minor version upgrade in scalapb, there may be binary, and perhaps API incompatibilities. 

### Motivation

https://github.com/bazelbuild/rules_scala/issues/667
So that I can use sparksql-scala and lenses without binary incompatibilities.

There are some tests that are failing can someone  give me some pointers on why they are failing.

In particular, it looks like this target `//test/proto:test_service-fast` is failing to build but running `bazel build //test/proto:test_service-fast` says that the target does not exist.

Here's the full error message:
```
(base) nudomsak-pro:rules_scala napasudomask$ bazel build //test/...
DEBUG: /private/var/tmp/_bazel_napasudomask/27b4d30503e20c142fbe6514959c12c2/external/bazel_skylib/lib.bzl:30:1: WARNING: lib.bzl is deprecated and will go away in the future, please directly load the bzl file(s) of the module(s) needed as it is more efficient.
INFO: Analyzed 1303 targets (12 packages loaded, 1727 targets configured).
INFO: Found 1303 targets...
INFO: From scala //src/scala/scripts:scalapb_generator_lib [for host]:
warning: there were three deprecation warnings; re-run with -deprecation for details
one warning found
ERROR: /Users/napasudomask/gcode/rules_scala/test/proto/BUILD:102:1: scala //test/proto:test_service-fast failed (Exit 1)
bazel-out/darwin-fastbuild/bin/test/proto/tmp2917373157101204084/test/proto/test_service/TestServiceGrpc.scala:4: error: type MethodDescriptor is not a member of package io.grpc
  val METHOD_TEST_METHOD1: _root_.io.grpc.MethodDescriptor[_root_.test.proto.test.TestRequest, _root_.test.proto.test2.TestResponse1] =
                                          ^
bazel-out/darwin-fastbuild/bin/test/proto/tmp2917373157101204084/test/proto/test_service/TestServiceGrpc.scala:14: error: type MethodDescriptor is not a member of package io.grpc
  val METHOD_TEST_METHOD2: _root_.io.grpc.MethodDescriptor[_root_.test.proto.test.TestRequest, _root_.test.proto.test3.TestResponse2] =
                                          ^
bazel-out/darwin-fastbuild/bin/test/proto/tmp2917373157101204084/test/proto/test_service/TestServiceGrpc.scala:24: error: type ServiceDescriptor is not a member of package io.grpc
[...]

PR #1029, opened 2020-04-08
fixes after scalapb upgrade

darl:
### Description
Build fixes for scala 2.12 and strict_deps enabled

### Motivation
Build failures in my current project after last PR (#1026)

### DIFF
--- a/scala_proto/private/scala_proto_default_repositories.bzl
+++ b/scala_proto/private/scala_proto_default_repositories.bzl
@@ -133,7 +133,7 @@ def scala_proto_default_repositories(
     _scala_maven_import_external(
         name = "scala_proto_rules_scalapb_fastparse",
         artifact = _scala_mvn_artifact(
-            "com.lihaoyi:fastparse:" + "2.1.2" if int(scala_version.split(".")[1]) < 12 else "2.1.3",
+            "com.lihaoyi:fastparse:" + ("2.1.2" if int(scala_version.split(".")[1]) < 12 else "2.1.3"),
             major_version,
         ),
         artifact_sha256 = scala_version_jar_shas["scalapb_fastparse"],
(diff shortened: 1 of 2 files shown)


darl:
I think it's a problem:

CI fails with `dependency_tracking_method = 'default'` and without explicit protobuf dependency;

my local build fails with `dependency_tracking_method = 'high-level'` and with explicit dependency from master.

ittaiz:
1. Sorry for the delay. Corona times are wrecking a havoc in my availability.
2. I think that the best thing to do is to revert your change and then push a new commit that only adds the failure that you're seeing locally to CI. Then we can try and devise a solution.
I'll try to do it now.

cc @Jamie5 I have a hunch that somewhere in the flag changes we lost some of the e2e tests we used to have

ittaiz:
btw, worst case I think we'll be able to add the problematic dep to the ignore targets of `scalapb_generator_lib`.
Let's hope we wont' reach this.

ittaiz:
Also right [now](https://github.com/bazelbuild/rules_scala/blob/9c84950f5d6e8f3cb9fdac06d9a9fdaff37ae4ad/scala/scala_toolchain.bzl#L26) `dependency_tracking_method='default'` means `dependency_tracking_method='high-level'`.
It would really help to get a repro or even just your full setup.
What toolchain are you using (one of the default ones or a custom ones), what scala_repositories and which target fails (I'm assuming it's the same one that's failing right now but would love to make sure).

ittaiz:
btw, it would also help to see the exact failure

ittaiz:
Closing in favor of #1030 which has a workaround for both issues.
Feel free to reopen if you have strong objections.

ittaiz:
btw, this isn't time critical or anything, right? 
Asking since I'd like to give Jamie time to review #1030 and we don't have anything newer than the scalapb upgrade so I'm guessing you can wait but would still like to verify

darl:
I can wait for #1030 to be merged

PR #1033, opened 2020-04-15
Remove strict_java_deps backwards compatibility

Jamie5:
### Description
We remove the backwards compatibility feature that setting
strict_java_deps means that scala code is compiled transitively
and with the special override settings. Instead strict_java_deps
affects only java code.

### Motivation
As previously planned, we are removing this backwards compatibility
piece.


test_rules_scala.sh
```diff
@@ -16,13 +16,12 @@ $runner bazel build test/...
 #$runner bazel build "test/... --all_incompatible_changes"
 $runner bazel test test/...
 $runner bazel test third_party/...
-# UnusedDependencyChecker doesn't work with strict_java_deps
```
Jamie5, line 19: @ittaiz not sure why we needed `-test:UnusedDependencyChecker` but as the comment that unused/strict doesn't go together isn't relevant anymore, so removed them in this block of code. Though it seems even in the past it would still compile and just not run unused deps checker (and there weren't unused deps in that rule as far as I could tell).

test_rules_scala.sh
```diff
@@ -16,13 +16,12 @@ $runner bazel build test/...
...
-#$runner bazel build "--strict_java_deps=ERROR --all_incompatible_changes -- test/... -test:UnusedDependencyChecker"
-$runner bazel test "--strict_java_deps=ERROR -- test/... -test:UnusedDependencyChecker"
-$runner bazel test "--extra_toolchains=//scala:minimal_direct_source_deps -- test/... -test:UnusedDependencyChecker"
-$runner bazel build "test_expect_failure/missing_direct_deps/internal_deps/... --strict_java_deps=warn"
+$runner bazel build "--extra_toolchains=//test/toolchains:high_level_transitive_deps_strict_deps_error -- test/..."
+$runner bazel build "--extra_toolchains=//scala:minimal_direct_source_deps -- test/..."
+#$runner bazel build "--extra_toolchains=//test/toolchains:high_level_transitive_deps_strict_deps_error --all_incompatible_changes -- test/..."
+$runner bazel test "--extra_toolchains=//test/toolchains:high_level_transitive_deps_strict_deps_warn -- test/..."
+$runner bazel test "--extra_toolchains=//scala:minimal_direct_source_deps -- test/..."
+$runner bazel build "test_expect_failure/missing_direct_deps/internal_deps/... --strict_java_deps=warn --extra_toolchains=//test/toolchains:high_level_transitive_deps_strict_deps_warn"
```
Jamie5, line 24: @ittiaz We do need both `strict_java_deps` and the toolchain. Because the toolchain controls strictness/deps of scala code under compile, while strict_java_deps controls the strictness/deps of java code under compile (even if the java code is in scala_library). Due to `scala/private/phases/phase_compile.bzl` still using strict_java_deps for java.

At the time I did do this intentionally but now wonder if it was a good idea given that it sometimes leads to things like this (there's a few other places where we don't change strict_java_deps to toolchain because they were testing strict deps on java source files). Anyways this is the existing behavior at least.

README.md
```diff
@@ -302,12 +302,6 @@ It can be daunting to turn on strict deps checking or unused dependency mode che
 
 We recommend turning on strict_deps_mode first, as rule `A` might have an entry `B` in its `deps`, and `B` in turn depends on `C`. Meanwhile, the code of `A` only uses `C` but not `B`. Hence, the unused dependency checker, if on, will request that `B` be removed from `A`'s deps. But this will lead to a compile error as `A` can no longer depend on `C`. However, if strict dependency checking was on, then `A`'s deps is guaranteed to have `C` in it.
 
-### [Experimental] Migrating from deprecated configurations
-
-There are a few deprecated configuration methods which we will be removing in the near future.
```
ittaiz, line 307: maybe we should leave a note somewhere to say that the java code can be controlled via strict_java_deps like java_library?

test_rules_scala.sh
```diff
@@ -16,13 +16,12 @@ $runner bazel build test/...
...
-$runner bazel build "--strict_java_deps=ERROR -- test/... -test:UnusedDependencyChecker"
-$runner bazel build "--extra_toolchains=//scala:minimal_direct_source_deps -- test/... -test:UnusedDependencyChecker"
-#$runner bazel build "--strict_java_deps=ERROR --all_incompatible_changes -- test/... -test:UnusedDependencyChecker"
-$runner bazel test "--strict_java_deps=ERROR -- test/... -test:UnusedDependencyChecker"
-$runner bazel test "--extra_toolchains=//scala:minimal_direct_source_deps -- test/... -test:UnusedDependencyChecker"
-$runner bazel build "test_expect_failure/missing_direct_deps/internal_deps/... --strict_java_deps=warn"
+$runner bazel build "--extra_toolchains=//test/toolchains:high_level_transitive_deps_strict_deps_error -- test/..."
+$runner bazel build "--extra_toolchains=//scala:minimal_direct_source_deps -- test/..."
+#$runner bazel build "--extra_toolchains=//test/toolchains:high_level_transitive_deps_strict_deps_error --all_incompatible_changes -- test/..."
+$runner bazel test "--extra_toolchains=//test/toolchains:high_level_transitive_deps_strict_deps_warn -- test/..."
```
ittaiz, line 22: Pretty sure `high_level_transitive_deps_strict_deps_warn` should have been `high_level_transitive_deps_strict_deps_error` since the [original line](https://github.com/bazelbuild/rules_scala/pull/1033/files#diff-9e850e1e0893c9ea027f87c0b45dffabL23) had `ERROR`

test_rules_scala.sh
ittaiz: I think this is actually important because this means that people who have java code in both scala_library and java_library control them together

Jamie5: Ok, fine with me

test_rules_scala.sh
Jamie5: Oops will fix

README.md
Jamie5: Ok will add something

PR #1034, opened 2020-04-15
[WIP] Handle a unused_dependency_checker_ignored_targets

Jamie5:
### Description
WIP, just here for the CI

### Motivation
WIP

### DIFF
--- a/src/java/io/bazel/rulesscala/specs2/BUILD
+++ b/src/java/io/bazel/rulesscala/specs2/BUILD
@@ -7,13 +7,9 @@ scala_library(
         "Specs2RunnerBuilder.scala",
         "package.scala",
     ],
-    unused_dependency_checker_ignored_targets = [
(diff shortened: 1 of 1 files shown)


Jamie5:
@ittaiz I suspect this is an issue with the high-level mode, because looking the source I don't see references to xml. And it seems this does work with ast mode. So once we switch transitive to use ast instead of high-level by default (and change the tests to all use ast instead of high-level), this should be solved.

ittaiz:
SGTM,
Do we want to have an issue open for removing this instance of `unused_dependency_checker_ignored_targets` and link it to the discussion of making `high-level` the default?

Jamie5:
I added a mention in the first post of https://github.com/bazelbuild/rules_scala/issues/867 (which contains the other TODOs as well), is that sufficient?

ittaiz:
Indeed. Can we close this PR for now?

PR #1036, opened 2020-04-16
Fix deduplication error on FocusedZipImporter

blorente:
### Description
Add a `toString` method to `FocusedZipImporter` that creates a unique representation of the instance, in order to allow `MultiImporter` to handle deduplication correctly.

When compiling thrift files with dependencies, `FocusedZipImporter`s get created and merged into a `MultiImporter` ([coderef](https://github.com/bazelbuild/rules_scala/blob/44438c6205f5c4a9bf5b9b762715b88530356ea9/src/scala/io/bazel/rules_scala/scrooge_support/Compiler.scala#L125)). However, `MultiImporter` uses `toString` to decide whether it should deduplicate importers when merging them ([coderef](https://github.com/twitter/scrooge/blob/2c8db36fa38bc4450d7ce0264407c40b37a9aff6/scrooge-generator/src/main/scala/com/twitter/scrooge/frontend/Importer.scala#L213)). Therefore, the merging of importers mentioned above had no effect.

### Motivation
I have added a folder with a reproduction of the case that doesn't work without the changes, but works with them.

Note that the test case is not using `absolute_prefixes`. It's possible that those would solve the issue, but I think this change is relevant regardless, as currently [this `+:` operation](https://github.com/bazelbuild/rules_scala/blob/44438c6205f5c4a9bf5b9b762715b88530356ea9/src/scala/io/bazel/rules_scala/scrooge_support/Compiler.scala#L125) has no effect.

### DIFF
--- a/src/scala/io/bazel/rules_scala/scrooge_support/FocusedZipImporter.scala
+++ b/src/scala/io/bazel/rules_scala/scrooge_support/FocusedZipImporter.scala
@@ -60,6 +60,12 @@ case class FocusedZipImporter(focus: Option[File], zips: List[File], zipFiles: L
       FileContents(importer, Source.fromInputStream(zipFile.getInputStream(entry), "UTF-8").mkString, Some(entry.getName))
     }
 
+  // Instances of this class are supposed to be mergeable into MultiImporters
+  // However, MultiImporters use `toString` to deduplicate importers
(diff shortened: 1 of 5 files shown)


ittaiz:
Hi,
Thanks for this!
I'll take a look once the build will pass and the CLA will be signed, ok?

Make sure to ping me here so I'll know we're good to go

blorente:
@googlebot I signed it!

blorente:
Closing this PR, given that the problem is definitely in an unreleased version of `scrooge`, and therefore I'll fix it there before it's even released, thus never affecting this repo.

ittaiz:
👍 thanks

PR #1037, opened 2020-04-16
[WIP] Jar stamping

Jamie5:

### DIFF
--- a/scala/scala_import.bzl
+++ b/scala/scala_import.bzl
@@ -1,5 +1,25 @@
 load("@io_bazel_rules_scala//scala:jars_to_labels.bzl", "JarsToLabelsInfo")
 
+def _handle_jar_stamping(ctx, jar):
+    source_file = jar
+    if ctx.label.package != jar.dirname:
+        symlink = ctx.actions.declare_file(jar.basename)
+        source_file = symlink
+        ctx.actions.run(
+            inputs = [jar],
+            outputs = [symlink],
+            # XXX need to make this relative to deal with remote caching etc
+            executable = "ln",
+            arguments = [jar.path, symlink.path],
+        )
+
+    return java_common.stamp_jar(
+        actions = ctx.actions,
+        jar = source_file,
+        target_label = ctx.label,
+        java_toolchain = ctx.attr._java_toolchain[java_common.JavaToolchainInfo],
+    )
+
 #intellij part is tested manually, tread lightly when changing there
 #if you change make sure to manually re-import an intellij project and see imports
 #are resolved (not red) and clickable
@@ -12,11 +32,14 @@ def _scala_import_impl(ctx):
         current_target_compile_jars,
         intellij_metadata,
     ) = (target_data.code_jars, target_data.intellij_metadata)
-    current_jars = depset(current_target_compile_jars)
+    current_target_stamped_compile_jars = [
+        _handle_jar_stamping(ctx, jar)
+        for jar in current_target_compile_jars
+    ]
+    current_jars = depset(current_target_stamped_compile_jars)
     exports = java_common.merge([export[JavaInfo] for export in ctx.attr.exports])
     transitive_runtime_jars = \
-        java_common.merge([dep[JavaInfo] for dep in ctx.attr.runtime_deps]) \
-            .transitive_runtime_jars
+        java_common.merge([dep[JavaInfo] for dep in ctx.attr.runtime_deps]).transitive_runtime_jars
     jars2labels = {}
     _collect_labels(ctx.attr.deps, jars2labels)
     _collect_labels(ctx.attr.exports, jars2labels)  #untested
@@ -26,8 +49,8 @@ def _scala_import_impl(ctx):
         jars2labels,
     )  #last to override the label of the export compile jars to the current target
 
-    if current_target_compile_jars:
-        current_target_providers = [_new_java_info(ctx, jar) for jar in current_target_compile_jars]
+    if current_target_stamped_compile_jars:
+        current_target_providers = [_new_java_info(ctx, jar) for jar in current_target_stamped_compile_jars]
     else:
         # TODO(#8867): Migrate away from the placeholder jar hack when #8867 is fixed.
         current_target_providers = [_new_java_info(ctx, ctx.file._placeholder_jar)]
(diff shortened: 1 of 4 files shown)


Jamie5:
@ittaiz so far I am lost in a muddle and probably won't have enough time to investigate so am putting this up for you

Jamie5:
@ittaiz wondering if you have any easy ideas for this issue, if not I will keep looking.

`bazel build --extra_toolchains=//test/toolchains:high_level_transitive_deps_strict_deps_error -- //src/java/io/bazel/rulesscala/specs2:specs2_test_discovery` fails with the following

```ERROR: /home/jamie/shared/src/rules_scala/src/java/io/bazel/rulesscala/specs2/BUILD:4:1: scala //src/java/io/bazel/rulesscala/specs2:specs2_test_discovery failed (Exit 1)
error: Target '@io_bazel_rules_scala_scala_reflect//:io_bazel_rules_scala_scala_reflect' is used but isn't explicitly declared, please add it to the deps.
You can use the following buildozer command:
buildozer 'add deps @io_bazel_rules_scala_scala_reflect//:io_bazel_rules_scala_scala_reflect' //src/java/io/bazel/rulesscala/specs2:specs2_test_discovery
error: Target '@io_bazel_rules_scala_scala_library//:io_bazel_rules_scala_scala_library' is used but isn't explicitly declared, please add it to the deps.
You can use the following buildozer command:
buildozer 'add deps @io_bazel_rules_scala_scala_library//:io_bazel_rules_scala_scala_library' //src/java/io/bazel/rulesscala/specs2:specs2_test_discovery
```

Looking at `specs2_test_discovery_scalac_worker_input` and focusing on scala-reflect we see that under Classpath and IndirectJars entries, we have the following
- bazel-out/k8-fastbuild/bin/external/io_bazel_rules_scala_scala_reflect/scala-reflect-2.11.12-stamped.jar
- bazel-out/host/bin/external/io_bazel_rules_scala_scala_reflect/scala-reflect-2.11.12-stamped.jar

while under DirectJars we have only `bazel-out/host/bin/external/io_bazel_rules_scala_scala_reflect/scala-reflect-2.11.12-stamped.jar`

Whereas without this change, everyone has just `external/io_bazel_rules_scala_scala_reflect/scala-reflect-2.11.12.jar`

So the stamping moves the jar into both `k8-fastbuild` and `host` 
"configurations"? (don't understand this part very well). Which seems undesirable as jar stamping ought to be independent of the config, but maybe that's not how bazel likes it.

I guess probably the direct rule is depending on the `host` configuration, while it is depending on `//external:io_bazel_rules_scala/dependency/specs2/specs2` which is on the `k8-fastbuild` configuration.

But I guess it's already a bad thing that the classpath contains both jars, right? Because while it might work we'd probably not want to have that in the first place.

(changing all of `cfg = "host"` to `cfg = "target"` does not help; it was just something I tried for funsies)

ittaiz:
Not sure and indeed a problem.
I have an idea- Wdyt about hacking bazel itself just to see if it will happen also via java_common.stamp_jar? Hack bazel and then build rules_scala with the hacked version.
Also two more thoughts- 
1. Is this the only instance? Maybe we should  print in the plugin (just for now) all classpath entries which have more than one instance
2. Does this happen only with bind? Maybe we can move this to toolchain. To check maybe just change rules_scala to explicit dependency and not via bind

Jamie5:
@ittaiz

This happened with at least scala-library and scala-reflect that I observed.

Removing native.bind did help, though just doing scala-library/reflect/compiler was not enough - also had to do scala-xml, junit, specs2, specs2_junit (not sure which of these last 4 were necessary and which were optional).

In https://github.com/bazelbuild/bazel/issues/1952 it sounds like it is reasonable to try to get rid of native.bind anyways, even if it wasn't causing an issue right now. I am not 100% sure that getting rid of bind will solve all remaining problems or that it won't cause strange issues.

But assuming you are fine with it and there aren't backwards compatibility issues (I couldn't find any docs on native.bind - the linked issue did suggest that people might have wanted to remove it from docs, which may have happened), I'd like to just get rid of rules_scala usage of native.bind just to not have to think about it and getting confused by it when trying to solve this issue.

In my experiment I just replaced all of `//external:io_bazel_rules_scala/dependency/scala/scala_compiler` with `@io_bazel_rules_scala_scala_compiler` and similar. Based on your comment it sounds like you want something else, though I don't 100% understand.

Do you mean that we want to add to `scala_toolchain` e.g.

```
        "scala_compiler": attr.label(
            default = "@io_bazel_rules_scala_scala_compiler",
        ),```
(or maybe as a provider so we can group all the different libraries together into one attr) (and would we just stick all the libraries on here, including specs2 and jmh and guava and everything else which uses native.bind)

And then access this toolchain everywhere (though I don't think we always have ctx so that's a problem e.g.

```scala_binary(
    name = "scaladoc_generator",
    main_class = "scala.tools.nsc.ScalaDoc",
    runtime_deps = [
        "//external:io_bazel_rules_scala/dependency/scala/parser_combinators",
        "//external:io_bazel_rules_scala/dependency/scala/scala_library",
        "@io_bazel_rules_scala_scala_compiler",
        "@io_bazel_rules_scala_scala_xml",
    ],
)```)

Jamie5:
@ittaiz ^ in case the notification was missed

ittaiz:
Thanks for the ping.
I didn't miss it per-se but was chewing on the problem and got sidetracked. 
The reason we use bind is to offer a layer of indirection where users can decide to use their own deps (different jars, different loaders) without rules_scala needing to change.
The alternative (very roughly speaking) is toolchains in the sense that this is a different mechanism of indirection which allows user configuration.
@liucijus started to dive into binds & toolchains for the cross compilation saga.
Vaidas,
do you have any thoughts on the matter?

To be honest I'm very uncertain on how to play this (since I remember someone poking holes into  the applicability of toolchains, maybe it was because they arrive too late in the lifecycle).

cc @johnynek in case he has some words of wisdom

liucijus:
I don't know the problem with binds deeply, but I think one of the biggest problems is as @Jamie5 described with depending directly on labels that should come via toolchain. I think it prevents rules_scala from getting rid of binds. Maybe there is something more. Grouping Scala libraries with provider is a good abstraction as long as user can override it easily.

liucijus:
Closing this as stale, we already have jar stamping in rules scala

PR #1041, opened 2020-04-30
Move transitive dependency mode tests to use ast dependency method

Jamie5:
### Description
We change the transitive dependency mode tests to use the ast dependency method.

This allows us to use an instance of unused_dependency_checker_ignored_targets which was needed to a deficiency in the high-level analyzer.

### Motivation
As we are moving towards using the ast dependency method for transitive dependency mode, we wish to make the tests align with that. We also wish to not use `unused_dependency_checker_ignored_targets` when not necessary.

### DIFF
--- a/src/java/io/bazel/rulesscala/specs2/BUILD
+++ b/src/java/io/bazel/rulesscala/specs2/BUILD
@@ -7,13 +7,9 @@ scala_library(
         "Specs2RunnerBuilder.scala",
         "package.scala",
     ],
-    unused_dependency_checker_ignored_targets = [
-        "//external:io_bazel_rules_scala/dependency/scala/scala_xml",
-    ],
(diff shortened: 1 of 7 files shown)


Jamie5:
@ittaiz this happens to address https://github.com/bazelbuild/rules_scala/pull/1034 .

Is it too early to stop running high-level transitive? If so I guess we'd have to leave this for a later point.

ittaiz:
I feel it's too soon yeah.
Maybe we should have these tests run both modes for now? Won't help with the unused_dependency_checker_ignored_targets but will give more inputs

Jamie5:
Hmm seems that would add more time to test running (not sure how much). So far not seeing any issues which would need to push removing the dependency_tracking_method past July 30 so seems probably fine to wait until nearer to then.

PR #1047, opened 2020-05-20
Document IntelliJ plugin incompatibility with custom rule names

gergelyfabian:
### Description

Adding documentation about IntelliJ plugin incompatibility as requested in https://github.com/bazelbuild/intellij/issues/1824.

### Motivation
We need to document that custom rule names won't work with IntelliJ which is not obvious.

### DIFF
--- a/docs/customizable_phase.md
+++ b/docs/customizable_phase.md
@@ -6,6 +6,7 @@
 *  [As a consumer](#as-a-consumer)
 *  [As a contributor](#as-a-contributor)
    *  [Phase naming convention](#phase-naming-convention)
+*  [Cooperation with IntelliJ plugin](#cooperation-with-intellij-plugin)
 
 ## Overview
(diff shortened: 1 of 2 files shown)


docs/customizable_phase.md
```diff
@@ -150,3 +151,30 @@ Function names in `phase_<PHASE_NAME>.bzl`
  - `_phase_<PHASE_NAME>`: private function with the actual logic
 
 See `phase_compile.bzl` for example.
+
+## Cooperation with IntelliJ plugin
+
+Bazel IntelliJ plugin has hard-coded the names of rules_scala targets that it detects as Scala targets:
+
+https://github.com/bazelbuild/intellij/blob/master/scala/src/com/google/idea/blaze/scala/ScalaBlazeRules.java#L32-L37
```
ittaiz, line 159: Maybe use a permalink in case it moves?

gergelyfabian: Fixed.

PR #1051, opened 2020-05-27
Fixes to make test_version.sh work

blorente:
### Description
Update the `test_version.sh` machinery to work with recent versions of the repo.

Changes:
- There were some missing dependencies and setup steps in `WORKSPACE.template`.
- The path to `test/shell/test_runner.sh` from `test_version.sh` was wrong.
- One target used the invalid `jvm_flags` attribute. Changed it to `scalac_jvm_flags`. **Question:** Should these flags be added as `javac_jvm_flags` as well?

Result:
`./test_version.sh` now passes.

Questions:
- Would it be possible to add `./test_version.sh` to travis somewhere, so that it doesn't regress again?

### Motivation
It would be useful to extend this machinery to test In #1050.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,10 +52,14 @@ jobs:
       env: TEST_SCRIPT=test_rules_scala BAZEL_VERSION=2.0.0
     - <<: *linux
       env: TEST_SCRIPT=test_reproducibility BAZEL_VERSION=2.0.0
+    - <<: *linux
+      env: TEST_SCRIPT=test_version BAZEL_VERSION=2.0.0
     - <<: *osx
       env: TEST_SCRIPT=test_rules_scala BAZEL_VERSION=2.0.0
     - <<: *osx
       env: TEST_SCRIPT=test_reproducibility BAZEL_VERSION=2.0.0
+    - <<: *osx
+      env: TEST_SCRIPT=test_version BAZEL_VERSION=2.0.0
 
 before_install:
   - |
--- a/test_version.sh
+++ b/test_version.sh
@@ -37,7 +37,7 @@ test_scala_version() {
 
 dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 # shellcheck source=./test_runner.sh
-. "${dir}"/test_runner.sh
+. "${dir}"/test/shell/test_runner.sh
 runner=$(get_test_runner "${1:-local}")
 
 $runner test_scala_version "2.11.12" \
--- a/test_version.sh
+++ b/test_version.sh
@@ -35,6 +35,11 @@ test_scala_version() {
 
 }
 
+if ! bazel_loc="$(type -p 'bazel')" || [[ -z "$bazel_loc" ]]; then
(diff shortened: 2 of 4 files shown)


blorente:
Thanks for the review!

Answering your questions:

1. No, I copied everything from the top level README
2. The top level README worked correctly :) The only thing that is not on the README is initialising the protobuf toolchains [ref](https://github.com/bazelbuild/rules_scala/pull/1051/files#diff-2ff29d9c3dd1bbe21bf545ad417dc1c3R44-R46), which is not the base config I think.

Will make a change with the other two comments.

ittaiz:
Awesome that the readme was your go to!
Two other comments is enabling CI and updating the protobuf version in the readme?

blorente:
Almost: Enabling CI and adding the `jvm_flags` to `javac_jvm_flags`.

ittaiz:
Oh. I meant that what you did about jvm flags sound right (and indeed not what you’re suggesting).
Can you do the protobuf readme update? Do you prefer a separate Pr?

blorente:
I'm sorry, I'm doing too many things at once and didn't respond as well as I should have. The only lines in `WORKSPACE.template` that are not straight from the `README` are:
```
load("@io_bazel_rules_scala//scala_proto:toolchains.bzl", "scala_proto_register_toolchains")

scala_proto_register_toolchains()
```

We only need these because the tests happen to compile some protobuf, similar to how it needs the `unused_deps` for its own purposes:
```
load("@io_bazel_rules_scala//scala:toolchains.bzl", "scala_register_unused_deps_toolchains")

scala_register_unused_deps_toolchains()
```

This is not something that should be in "the minimum set of things a user needs in their WORKSPACE to use rules_scala", so I don't think it should go in the example in the README.

However, I agree that those toolchains are hard to find, so I can open a follow-up PR with documentation updates to document what toolchains are available (although not in the immediate future).

ittaiz:
Sounds good.
Can you explain the change to test_version? why is it needed? Can we solve it differently?
I don’t like these conditions

blorente:
If you mean [these changes](https://github.com/bazelbuild/rules_scala/pull/1051/files#diff-af9f5a9ff6dcf95828c50792eb95701fR38-R42):
```
if ! bazel_loc="$(type -p 'bazel')" || [[ -z "$bazel_loc" ]]; then
  export PATH="$(cd "$(dirname "$0")"; pwd)"/tools:$PATH
  echo 'Using ./tools/bazel directly for bazel calls'
fi
```
This is used to find the `bazel` executable in CI. It's copy-pasted from [test_rules_scala.sh](https://github.com/bazelbuild/rules_scala/blob/64c810541911707afa17728161ab0e3e13c97a65/test_rules_scala.sh#L5) and [test_reproducibility.sh](https://github.com/bazelbuild/rules_scala/blob/3dd2e9f7ba1e94acbfbb40bc0e0d48c5efe1c6cb/test_reproducibility.sh#L5), the other two scripts that are run in `.travis.yml`.

Issue #1049, opened 2020-05-26
scalacopts parsing issue with comma


wisechengyi:
Per scalac doc https://docs.scala-lang.org/overviews/compiler-options/index.html: there could be options like `-Ywarn-unused:WARNING1,WARNING2`

However, scala options are joined by comma then split by comma:
https://github.com/bazelbuild/rules_scala/blob/master/scala/private/rule_impls.bzl#L142
https://github.com/bazelbuild/rules_scala/blob/master/src/java/io/bazel/rulesscala/scalac/CompileOptions.java#L39

so if I use `-Ywarn-unused:-implicits,-locals,-explicits,-patvars,-privates` as scalaopts, the build will fail with
```
scalac error: bad option: '-locals'
  scalac -help  gives more information
one error found
java.lang.RuntimeException: Build failed
	at io.bazel.rulesscala.scalac.ScalacProcessor.compileScalaSources(ScalacProcessor.java:256)
	at io.bazel.rulesscala.scalac.ScalacProcessor.processRequest(ScalacProcessor.java:76)
	at io.bazel.rulesscala.worker.GenericWorker.runPersistentWorker(GenericWorker.java:45)
	at io.bazel.rulesscala.worker.GenericWorker.run(GenericWorker.java:111)
	at io.bazel.rulesscala.scalac.ScalaCInvoker.main(ScalaCInvoker.java:41)
```

wisechengyi:
A couple of options:
1. Write the scalac options into another file to be read by `ScalaCInvoker` which separates the options by newline:
```
-Yabc:x,y,z
-Yefg:alpha,beta
```
2. Use another delimiter such `:::`

@ittaiz @oscar-stripe preferences here?

ittaiz:
Let’s go with a different delimiter.
Chances of `:::` colliding are really low, if that will happen we’ll add the complexity of the file.
Any chance you want to send a PR (with a test of course)?

wisechengyi:
Thanks! Ofc. Just wanted to make sure the approach first 😁

PR #1053, opened 2020-06-02
Use ::: as scalacopts delimiter

wisechengyi:
### Description

As suggested in #1049, using `:::` as the delimiter instead of `,` to separate scalac options, in order to avoid collision with `,`. 

### Motivation

Per scalac doc https://docs.scala-lang.org/overviews/compiler-options/index.html: there could be options like `-Ywarn-unused:WARNING1,WARNING2`

For example before this change, if
```
scalaopts = ['-Ywarn-unused:-implicits,-locals,-explicits,-patvars,-privates']
```
the build would fail with
```
scalac error: bad option: '-locals'
  scalac -help  gives more information
one error found
java.lang.RuntimeException: Build failed
	at io.bazel.rulesscala.scalac.ScalacProcessor.compileScalaSources(ScalacProcessor.java:256)
	at io.bazel.rulesscala.scalac.ScalacProcessor.processRequest(ScalacProcessor.java:76)
```


wisechengyi:
@ittaiz sorry just saw the message in the slack.

Just added
1. `test/scalacopts`, so will be covered by `bazel build test/...`
2. unit test on option parsing: `src/java/io/bazel/rulesscala/scalac/CompileOptionsTest.java`

Let me know if further adjustment is needed. Thanks!

ittaiz:
ok, I think the current test indeed guards us but IMHO in a mechanical white box way.
I think a more semantic test could both help regression and demonstrate the feature set more.
For example we could add a target that has `scalacopts = [ "-language:implicitConversions, postfixOps",  -language:higherKinds, -Xfatal-warnings]" and I'd have the scala file of that target use these three features. I've added the fatal-warnings since I think (not sure) that you can use them right now without the flags but they will generate a warning so the fatal makes sure that they're actually used.

I'd be happy to push a commit to this PR with the test if the above isn't clear.

Btw,
We can also do an integration test by having a scala file with warnings, use your examples to turn on these warnings and use again fatal-warnings. then the integration test asserts the build fails.
I think the first option is simpler and faster (unit tests are cached) but just wanted to give it for completeness. Both check actual feature.

wisechengyi:
Thanks! Makes sense. My initial plan was to introduce an unused import in the code then use `scalacopts` to fail on unused import, but ran into backwards incompatibility issue with 2.11. 

Let me see if I can somehow only test this with 2.12 via `test_versions.sh`

ittaiz:
If you fail on an unused import then this needs to be an external shell integration test (otherwise the build won’t pass).

If you hit a wall with my offer LMK and I’ll help. You can even push a half baked commit.

Re integration tests- There is work in the bazel space to allow such integration tests to be part of the bazel build.
I have an OSS library (bazel integration testing) which does this but it’s not trivial to use and has its costs. There is a chance that whatever will come from bazel will be easier.

test/scalacopts/A.scala
```diff
@@ -0,0 +1,15 @@
...
+import scala.language.experimental.macros
+
+object Macros {
+  def hello: String = macro macroSettings
+
+  def macroSettings(c: blackbox.Context): c.Expr[String] = {
+    import c.universe._
+    // c.settings are the values from scalac's -Xmacro-settings
+    val s = c.settings.mkString(",")
+    c.Expr(q"""${s}""")
```
wisechengyi, line 13: here is the relevant part using scalac options

test/scalacopts/BUILD
```diff
@@ -0,0 +1,21 @@
...
+    srcs = ["A.scala"],
+)
+
+scala_test(
+    name = "EchoMacroSettings",
+    srcs = ["EchoMacroSettings.scala"],
+    scalacopts = [
+        # The macro settings are used in the test itself.
+        "-Xmacro-settings:name=a,b,c,d",
+        # These two settings are not important
```
ittaiz, line 14: Can we remove them?

test/scalacopts/BUILD
```diff
@@ -0,0 +1,21 @@
...
+    name = "A",
+    srcs = ["A.scala"],
+)
+
+scala_test(
+    name = "EchoMacroSettings",
+    srcs = ["EchoMacroSettings.scala"],
+    scalacopts = [
+        # The macro settings are used in the test itself.
+        "-Xmacro-settings:name=a,b,c,d",
```
ittaiz, line 13: Can we move “d” to  another option?
Are two different “Xmacro-settings” options accumulative?
If not maybe add another one with “name2=A”?
My idea is to have the test show the combination of two top level options where one of them has commas

src/java/io/bazel/rulesscala/scalac/BUILD
```diff
@@ -37,3 +37,14 @@ filegroup(
     ],
     visibility = ["//visibility:public"],
 )
+
+java_test(
+    name = "compile_options_test",
+    srcs = [
+        "CompileOptionsTest.java",
```
ittaiz, line 44: Maybe we should drop this test? I feel it’s too white box and the other test you added would give the same confidence with more value

test/scalacopts/BUILD
wisechengyi: done

test/scalacopts/BUILD
wisechengyi: converted to 
```
    scalacopts = [
        # The macro settings are used in the test itself.
        "-Xmacro-settings:name=Mike,location=US",
        "-Xmacro-settings:hobby=basketball,soccer",
    ],
```

src/java/io/bazel/rulesscala/scalac/BUILD
wisechengyi: done.

PR #1055, opened 2020-06-05
Implement `scrooge_java_library` to compile thrift into java

blorente:
### Description
Implement a `language` attribute in `scrooge_scala_library`, to allow compiling thrift to Java.

A more detailed description to come when I finish it.

For now, it generates Java code, but it doesn't even try to compile it (`twitter_scrooge.bzl:_compile_jvm_code` doesn't even try to compile the Java).


ittaiz:
Question- why use an additional attribute as opposed to a different rule? We can of course share a lot of the implementation internally?
I'm not saying we must and I'm not 100% we should but have you thought about it?

My main reason for asking is because this feels like proto being compiled to different languages and there the bazel idiom is different rules (`java_proto_library`, `scala_proto_library`, etc).

And to clarify- yes I fully accept the need.
Question is about design (and of course later on, functionality, readability, coverage, etc).

ittaiz:
Next time btw, I think it might be easier if you open an issue detailing your problem/need so we can discuss design and thoughts there instead of the PR.
Can also reduce waste if we decide to go down a very different path :)

No worries for this time though...

blorente:
Thank you for the amazing feedback!
>  why use an additional attribute as opposed to a different rule? We can of course share a lot of the implementation internally?
I... hadn't thought of that! But it could make sense.
I like the idea of a separate rule. However, the only place where these two would be separate (either calling the scala or java compiler), the `scala_library` already calls both, that's why I implemented it like this in the first place.
I'll experiment with the idea of separate rules and report back, to see what shape they take.

> Next time btw, I think it might be easier if you open an issue detailing your problem/need so we can discuss design and thoughts there instead of the PR.

Definitely, I should have done this a long time ago, that's my bad. We have a couple more changes we would like to implement, I'll open issues soon for both of them.

blorente:
Oh, good catch! I think this line (https://github.com/bazelbuild/rules_scala/pull/1055/files#diff-4fe70c7ad76a7189b88d4792f15a77ffR305) needs to be `ctx.rule.attr.language` instead, so that it will get the language attribute from the rule.

Another point in favour of making separate rules though. I haven't had time to explore that yet, but will soon.

blorente:
@ittaiz I'm investigating ways to do this with multiple rules, and I think I have it boiled down to two options:
- We modify the separation between aspect and rule implementation: Right now, the aspect both compiles the thrift into scala and the scala into bytecode. We can make the aspect take the language flag and only take care of compiling from thrift to <jvm langauge>, and then each rule would compile its own code from their respective languages to bytecode.

I like this one because the "common" code is still inside the aspect, which is "call scrooge with this thrift", and the specifics of each rule are inside each rule. The downside is that I'm not sure it's possible, and it doesn't follow the convention set by `scala_proto_library`.

If we end up going this way, I'd like to do 3 PRs:
- one (probably small one) to factor out the aspect
- one (probably small one) to add the language flag.
- one to add the `java_scrooge_library` rule. If this is not something that should go in `rules_scala`, the interface left would be flexible enough for us to add this rule on our end.

- Alternatively, we could keep compiling from jvm source to bytecode in the aspect, introduce a conditional on "language" in the aspect and compile either scala of java there. We could then have the `java_scrooge_library` rule just be a facade that sets the correct value for `language`.

blorente:
On a second thought, I don't think option 1 is possible, since we need to be able to access the value of `aspect(r)` (the generated scala/java code) on rule `r`, which I don't think is possible.

ittaiz:
I agree option 1 doesn’t sound feasible.
I’m not sure I understand option 2.
Does it mean we’ll run 2 aspects?
The way I see it is either we run 1 aspect which generates scala/java/both for all Scrooge targets in the workspace (might be ok) or we have 2 aspects (sitting on top of shared code) and each have a configuration of language.
Wdyt?

blorente:
My option 2 was the same as your option 1, just very poorly explained on my part. I prefer your option 2 if the aspects can be factored out enough that they end up roughly looking like:
```
def _scrooge_java_aspect_impl(...):
  (data1, data2, data3...) = _do_common_prep()
  java_info = _compile_the_java_code(...)
  return _form_common_output(java_info, ...)
```
As in, the only thing differring is the compilation function.

I'll timebox trying to do that, and implement option 1 if I'm not satisfied.

Thanks for your help!

ittaiz:
Sounds good.
Good luck!

blorente:
I've implemented it as two aspects, this is now ready for review. The CI failures are mac/linux compatibility issues (I think)

blorente:
It turns out the incompatibility is with the jdks we use to compile the generated java. Working on it :)

blorente:
I think the problem is gone now :)

blorente:
The travis error seems to be a timeout, for some reason the OSX shards take twice as much as the linux ones.

blorente:
I can't restart it, @ittaiz  could you do it for me?

ittaiz:
I'm sorry, I completely missed this. Will review it tomorrow morning, it's a bit beefy so I hope I'll be able to do it in one pass.

scala/private/rule_impls.bzl
```diff
@@ -209,6 +210,29 @@ StatsfileOutput: {statsfile_output}
         ] + ["@" + argfile.path],
     )
 
+def compile_java(ctx, source_jars, source_files, output, extra_javac_opts, providers_of_dependencies):
```
ittaiz, line 213: in general this change is in the opposite direction of the repo.
We've moved most of our rules to phases and trying to kill rule_impls and such.
To be honest scrooge is one of the big rules which haven't moved.
Partly because the person initiating the move didn't know scrooge or had a need for it and partly because it's aspect based and I think we need to crack how to do phases with aspect.

I think that the right thing to do is to accept this to not block this PR but I'd really appreciate it if you could take a deep look at phases and try to come up with a suggestion on how scrooge moves to be phase based.

wdyt?

src/scala/io/bazel/rules_scala/scrooge_support/BUILD
```diff
@@ -8,7 +8,9 @@ scala_library(
     visibility = ["//visibility:public"],
     deps = [
         ":focused_zip_importer",
+        "//external:io_bazel_rules_scala/dependency/scala/guava",
```
ittaiz, line 11: is this addition and the line below needed? I don't see any source changes correlating to it

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -121,6 +124,20 @@ def twitter_scrooge(
         actual = util_logging,
     )
 
+    mustache_name = "io_bazel_rules_scala_mustache"
```
ittaiz, line 127: I'm also not seeing where this is used (since I didn't understand the reference in `src/scala/io/bazel/rules_scala/scrooge_support/BUILD`

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -247,26 +276,36 @@ def _compile_scala(
...
+    return _create_java_info_provider(scrooge_jar, all_deps, output)
+
 def _concat_lists(list1, list2):
     all_providers = []
     all_providers.extend(list1)
     all_providers.extend(list2)
     return all_providers
 
-####
-# This is applied to the DAG of thrift_librarys reachable from a deps
```
ittaiz, line 266: why did you remove this comment?

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -281,99 +320,158 @@ def _scrooge_aspect_impl(target, ctx):
...
+
+    scrooge_file = ctx.actions.declare_file(
+        target.label.name + "_scrooge_{}.srcjar".format(language),
+    )
 
     # we sort so the inputs are always the same for caching
     compile_thrifts = sorted(target_ti.srcs.to_list())
-    imps = [j[JavaInfo] for j in ctx.attr._implicit_compile_deps]
-    if compile_thrifts:
-        # we sort so the inputs are always the same for caching
```
ittaiz, line 289: I think this comment was lost

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -164,11 +181,12 @@ def _compile_to_scala(ctx, label, compile_thrifts, include_thrifts, jar_output):
             # always add finagle option which is a no-op if there are no services
             # we could put "include_services" on thrift_info, if needed
             "--with-finagle",
+            "--language={}".format(language),
         ]),
     )
 
     argfile = ctx.actions.declare_file(
-        "%s_worker_input" % label.name,
+        "{}_{}_worker_input".format(label.name, language),
```
ittaiz, line 189: this is needed if we have two targets with the same label but for java/scala scrooge? I think this isn't legal bazel wise to have targets with the same name for different rules.
If this serves a different purpose I'd love to know.
I got to it btw from trying to find a test that needs this and couldn't find it so if it does exist I'd love to know

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -281,99 +320,158 @@ def _scrooge_aspect_impl(target, ctx):
...
+    )
+    return scrooge_file
+
+def _common_aspect_implementation(target, ctx, language, compiler_function):
+    """Aspect implementation to generate code from thrift files in a language of choice, and then compile it.
+    Takes in a `language` (either "java" or "scala") and a function to compile the generated sources.
+    """
+    allowed_languages = ["java", "scala"]
+    if language not in allowed_languages:
+        fail("Trying to compile thrift to language {}, when only {} are allowed".format(language, allowed_languages))
```
ittaiz, line 368: We don't have any test for this and honestly we can't because it's a private function and on one can call it.
Why do we need this protection btw?

ittaiz:
And forgot to mention I finished reviewing all files

src/scala/io/bazel/rules_scala/scrooge_support/BUILD
blorente: Scrooge needs Guava and mustache to generate java code from thrift.

twitter_scrooge/twitter_scrooge.bzl
blorente: We don't actually. I'll remove it.

twitter_scrooge/twitter_scrooge.bzl
blorente: I meant to write a better one, but I didn't. My bad, will unremove.

scala/private/rule_impls.bzl
blorente: I haven't had time to really look into this, but we've already found a usecase where phases would be really good for us internally, so definitely something we'll have to crack sooner rather than later.

I'm happy to block this PR until we come up with at least a general approach to phases in scrooge, I definitely want us to move in the direction of the repo.

src/scala/scripts/ScroogeWorker.scala
```diff
@@ -58,8 +58,13 @@ object ScroogeWorker extends Worker.Interface {
 
     val scrooge = new Compiler
 
-    if (additionalFlags.contains("--with-finagle")) {
-      scrooge.flags += WithFinagle
+    additionalFlags.foreach {
+      case "--with-finagle" => {
+        scrooge.flags += WithFinagle
+      }
+      case f if f.startsWith("--language=") => {
```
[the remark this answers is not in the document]
blorente: I'm personally not a fan of silently ignoring flags, but can definitely ignore them if we prefer to keep semantics. Or add a better error message.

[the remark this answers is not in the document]
ittaiz: Better error message sounds ok.
MatchError sounds like a bad compromise for me since our users will need to dig into our code to understand what happened.
AFK so don’t remember- can the flags be passed by our users or does the rule fully control them? If they can be passed then we need a test for this error case

scala/private/rule_impls.bzl
ittaiz: I’d rather not block it because I’d like to separate concerns (java language and Scrooge aspect to phases).
If you’re on it that’s great. Once you have thoughts we can discuss them over an issue or POC PR.

src/scala/io/bazel/rules_scala/scrooge_support/BUILD
ittaiz: But where is it used exactly?

twitter_scrooge/twitter_scrooge.bzl
blorente: This is used by a [ApacheJavaGenerator.scala](https://github.com/twitter/scrooge/blob/develop/scrooge-generator/src/main/scala/com/twitter/scrooge/java_generator/ApacheJavaGenerator.scala) in Scrooge, to generate java from mustache templates. Since our scrooge wrapper inserts itself in scrooge in a funny way, it needs these dependencies declared for it.

blorente: Also, mustache hasn't changed in a very long time, so I didn't think it needed to be customizable like the other targets here, but happy to change that.

src/scala/io/bazel/rules_scala/scrooge_support/BUILD
blorente: I honestly don't see any explicit guava imports, but the `ApacheJavaGenerator` class mentioned in the comment about mustache fails without this import.

src/scala/scripts/ScroogeWorker.scala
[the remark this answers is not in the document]
blorente: The rule fully controls them, there's no sort of "extra scrooge flags" argument.

src/scala/io/bazel/rules_scala/scrooge_support/BUILD
blorente: Oh, ApacheJavaGenerator apparently relies on `com/google/common/base/Function` for some reason. This line fails: https://github.com/twitter/scrooge/blob/15d061372d11cef341bc00eabd9efdfa79920cf0/scrooge-generator/src/main/scala/com/twitter/scrooge/java_generator/ApacheJavaGenerator.scala#L297

With the following error:
```
Exception parsing: test/src/main/scala/scalarules/test/twitter_scrooge/thrift/thrift2/thrift3/Thrift3.thrift
com.twitter.scrooge.frontend.FileParseException: Exception parsing: test/src/main/scala/scalarules/test/twitter_scrooge/thrift/thrift2/thrift3/Thrift3.thrift
Caused by: java.lang.NoClassDefFoundError: com/google/common/base/Function
        at com.twitter.scrooge.java_generator.ApacheJavaGenerator.$anonfun$apply$2(ApacheJavaGenerator.scala:297)
        at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
        at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
        at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
        at com.twitter.scrooge.java_generator.ApacheJavaGenerator.apply(ApacheJavaGenerator.scala:294)
        at io.bazel.rules_scala.scrooge_support.Compiler.$anonfun$run$6(Compiler.scala:149)
        at io.bazel.rules_scala.scrooge_support.Compiler.$anonfun$run$6$adapted(Compiler.scala:120)
        at scala.collection.Iterator.foreach(Iterator.scala:943)
        at scala.collection.Iterator.foreach$(Iterator.scala:943)
        at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
        at io.bazel.rules_scala.scrooge_support.Compiler.run(Compiler.scala:120)
        at scripts.ScroogeWorker$.work(ScroogeWorker.scala:93)
        at io.bazel.rulesscala.worker.Worker.persistentWorkerMain(Worker.java:92)
        at io.bazel.rulesscala.worker.Worker.workerMain(Worker.java:46)
        at scripts.ScroogeWorker$.main(ScroogeWorker.scala:15)
        at scripts.ScroogeWorker.main(ScroogeWorker.scala)
Caused by: java.lang.ClassNotFoundException: com.google.common.base.Function
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:583)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
```

twitter_scrooge/twitter_scrooge.bzl
blorente: I don't think so, it was duplicated to begin with :) it's here https://github.com/bazelbuild/rules_scala/pull/1055/files/c346443ab84f4f60b6178bdd2899102951d3519b#diff-4fe70c7ad76a7189b88d4792f15a77ffR340

twitter_scrooge/twitter_scrooge.bzl
blorente: In this code, `target` will be a thrift target, and we may need to generate both java and scala out of it. The same reason to parametrize by language here: https://github.com/bazelbuild/rules_scala/pull/1055/files/c346443ab84f4f60b6178bdd2899102951d3519b#diff-4fe70c7ad76a7189b88d4792f15a77ffR336

src/scala/scripts/ScroogeWorker.scala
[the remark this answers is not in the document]
blorente: Added error message on match failed.

twitter_scrooge/twitter_scrooge.bzl
blorente: I merged that comment into a broader explanation of the aspect: https://github.com/bazelbuild/rules_scala/pull/1055/commits/ad320eb37c7ecf1e5245af070df601403bf96276#diff-4fe70c7ad76a7189b88d4792f15a77ffR363-R368

scala/private/rule_impls.bzl
blorente: Thanks! Opened https://github.com/bazelbuild/rules_scala/issues/1064 to track. It may not be me personally working on this, but it's something we will probably want soon :)

blorente:
Responded/addressed comments :)

twitter_scrooge/twitter_scrooge.bzl
ittaiz: do we have an example of trying to generate both scala and java from a single thrift target?
I think we do but I also think that I changed this line back and nothing broke. 
If you can validate it I'd be grateful.

twitter_scrooge/twitter_scrooge.bzl
ittaiz: ok, I now think I know what's wrong with the current setup.
I think that the generator deps should be specified to it.
This [call](https://github.com/bazelbuild/rules_scala/blob/c346443ab84f4f60b6178bdd2899102951d3519b/twitter_scrooge/twitter_scrooge.bzl#L107) should pass in the runtime_deps (`scala_maven_import_external` takes in `deps` and `runtime_deps` attributes). This will clarify to the reader that our code doesn't need it but it's a requirement of the generator (and I'd maybe say it's a requirement to generate java code)

blorente: Wow, that makes a lot of sense, and actually helps me understand another problem!
Made the change, thanks a lot for the help.

ittaiz:
Thanks!

ittaiz:
btw, why did you restore external targets? because it didn't work you when you tried to consume it?
I don't like this situation where we don't have coverage of how it is to consume us.
Probably something we should tackle.
We (Wix) got burnt on this a few times in the past

blorente:
>btw, why did you restore external targets?
It did work fine, but linking to internal targets means that we don't consume whatever targets people bind, like we do in twitter-scrooge.
If we consumed `@io_bazel_rules_scala_guava` and someone for some reason (I know the repo doesn't work like that now, but still) wanted to use their own in-source implementation of guava (like we do with `scrooge-generator` and the other things in #1050 ), we wouldn't be able to see it from here.

It's a minor thing, and I doubt it will ever be relevant, but it was good practice to work through the tradeoffs in my head. Sorry for the commit history confusion.

PR #1066, opened 2020-07-06
Fix JOpt Simple version

remyhaemmerle-da:
### Description

The PR switches the version of `JOpt Simple` from `5.0.3` to `4.6`.

### Motivation

The `JMH` support introduces a dependency to the version `5.0.3` of `JOpt Simple`. However the `JMH` library is built against the version `4.6` of `JOpt Simple`. In current state the command line parsing functionality of JMH is partially broken (see test `//test/jmh:jmh_command_line_parsing_test`).

### DIFF
--- a/jmh/jmh.bzl
+++ b/jmh/jmh.bzl
@@ -56,8 +56,8 @@ def jmh_repositories(maven_servers = _default_maven_server_urls()):
     )
     _scala_maven_import_external(
         name = "io_bazel_rules_scala_net_sf_jopt_simple_jopt_simple",
-        artifact = "net.sf.jopt-simple:jopt-simple:5.0.3",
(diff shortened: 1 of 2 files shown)


ittaiz:
thanks!
does jmh have a newer version? In general I prefer rolling forward.
Having said that the right thing to do is probably merge this PR and then try to upgrade JMH itself (if relevant) in a separate PR.
Also can you please take care of the CLA?

remyhaemmerle-da:
> does jmh have a newer version? In general I prefer rolling forward.

Unfortunately, it seems the last version of JMH (v1.23 from last January) is still depending on `JOpt Simple` 4.6.
https://github.com/openjdk/jmh/blob/1.23/pom.xml#L269

> Also can you please take care of the CLA?

I will take care of that.

remyhaemmerle-da:
@googlebot I signed it!

Issue #1014, opened 2020-02-28
Slow specs2 file filtering when many tests are available


laurynasl-wix:
It seems that there is an issue with test filtering performance when a package contains many tests. If you have around > 100 tests, the overhead from filtering itself usually defeats any gain received by not running other tests. I'll attach a pull request with a repro case.

PR #1070, opened 2020-07-20
Fix slow specs2 filtering

simuons:
### Description
Fixes https://github.com/bazelbuild/rules_scala/issues/1014


simuons:
I'll open a new PR with only own commit so I could skip google cla

PR #1071, opened 2020-07-20
Fix slow specs2 filtering

simuons:
### Description
Filtering of specs2 test classes with many examples is slow because of inefficient construction of list of examples that needs to be run. For each included example all examples are scanned in test class.

### Motivation
Fixes https://github.com/bazelbuild/rules_scala/issues/1014

### DIFF
--- a/src/java/io/bazel/rulesscala/specs2/Specs2RunnerBuilder.scala
+++ b/src/java/io/bazel/rulesscala/specs2/Specs2RunnerBuilder.scala
@@ -94,24 +94,14 @@ class FilteredSpecs2ClassRunner(testClass: Class[_], testFilter: Pattern)
     createDescriptionTree(ee).toTree.flattenLeft.toMap
 
   /**
-    * Retrieves an original (un-sanitized) text of an example fragment,
-    * used later as a regex string for specs2 matching.
-    *
-    * This is done by matching the actual (sanitized) string with the sanitized version
-    * of the original example text.
-    */
-  private def specs2Description(desc: String)(implicit ee: ExecutionEnv): String = {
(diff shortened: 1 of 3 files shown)


ittaiz:
Thanks.
Can you take a look at the build? It's broken.
See contributors md for instructions on running the e2e tests (it's a bash in the root of the repo)

or-shachar:
@ianoc  - Any idea why good old `test_disappearing_class` can fail on this PR... ?

or-shachar:
Can we / should we connect this with https://github.com/bazelbuild/rules_scala/pull/1015 ?

or-shachar:
Another evidence: https://github.com/bazelbuild/rules_scala/pull/310#issuecomment-337661962

ianoc:
We don't use specs2, so I don't know much about it/why it would cause it to fail  I'm afraid, in that other case of #310 the test caught a possible insidious issue (scalac hiding information away so build outputs would not be reflective of the source tree invalid), so its a kinda great test from that pov. But maybe its something in scala versions being used now its possible or some interaction with specs.

simuons:
@ittaiz infamous test_disappearing_class passed with this [commit](https://github.com/bazelbuild/rules_scala/pull/1071/commits/5bdc749c2dede566961c52392f9a9e3e655ee1af). wdyt? I can move this commit to separate PR.
cc @or-shachar

ittaiz:
I think moving it is a good idea. Can you explain what was the problem and how you addressed it? Can be in the separate PR

ittaiz:
@simuons you'll rebase and force push?

simuons:
@ittaiz cleaned up commits, rebased, force pushed. Now it's green.

PR #1072, opened 2020-07-21
Toolchain deps infra

liucijus:
This is the first actual PR from Deps Toolchains Infrastructure (#1067)

- Adds reusable infra to develop toolchains for deps
- Adds docs
- Minimizes deps required for minimal Scala setup

Opinionated decisions (welcome to discuss!):
- common deps are added on existing scala toolchain (proposal had a separate deps toolchain) - this is done for simplicity
- guava (moved to scrooge setup) and commons io were removed. The reusability gain was minimal, but has added more complexity for users. cc @blorente 

Question: maybe we also want to migrate ScalacProvider to DepsInfo map?

**N.B. This is a breaking change for users who do not use repositories macros**


ittaiz:
Thanks for this! I think there's a lot of sense in having a unified mechanism.
Not sure if there are cons from moving ScalacProvider itself.
WDYT about sending a commit that moves it and we can understand from it if there are cons?

scala/toolchains/README.md
```diff
@@ -0,0 +1,153 @@
+# Developing Toolchains for Deps
```
ittaiz, line 1: @katre is there a chance you can take a look at this toolchains pattern?
I think this might be valuable for rulesets in general and not only rules_scala.
If you think we're abusing toolchains or just misunderstanding it we'd also love to know that.
One concrete question is whether using a map [here](https://github.com/bazelbuild/rules_scala/pull/1072/files#diff-4315eb5f77e2ca3a6bb2a3af7f783366R19) is the best solution?

katre: I looked over the PR and the docs, and I'm afraid I don't fully understand what's being implemented. You are adding dependencies to the toolchain that will then also be provided to other, non-toolchain-aware rules, via the normal transitive dependencies? Can you give a concrete example of where this would be useful?

As for the map, I am also confused about why it is a map from a label (which seems to always be a declare_deps_provider) to a string (which is a unique provider id?). Why this direction and not a map of provider id to label?

Lastly, if you are updating your toolchains, have you seen the design for [Toolchain Transitions](https://github.com/bazelbuild/proposals/blob/master/designs/2019-02-12-toolchain-transitions.md), including the [migration steps](https://github.com/bazelbuild/proposals/blob/master/designs/2020-02-07-toolchain-transition-migration.md)? This will allow you to declare some toolchain dependencies as being in the target config (for the target being built with the toolchain), and others in the exec config, which may be useful with these deps.

liucijus: > I looked over the PR and the docs, and I'm afraid I don't fully understand what's being implemented. You are adding dependencies to the toolchain that will then also be provided to other, non-toolchain-aware rules, via the normal transitive dependencies? Can you give a concrete example of where this would be useful?

Example: [scalac](https://github.com/bazelbuild/rules_scala/blob/master/src/java/io/bazel/rulesscala/scalac/BUILD#L23) target is `java_binary` which isn't aware of `scala_toolchain`.

> 
> As for the map, I am also confused about why it is a map from a label (which seems to always be a declare_deps_provider) to a string (which is a unique provider id?). Why this direction and not a map of provider id to label?

I haven't found a type for string to label mapping (https://docs.bazel.build/versions/master/skylark/lib/attr.html). I don't like having inverted dict as a compromise, but I failed to find a better alternative.

> Lastly, if you are updating your toolchains, have you seen the design for [Toolchain Transitions](https://github.com/bazelbuild/proposals/blob/master/designs/2019-02-12-toolchain-transitions.md), including the [migration steps](https://github.com/bazelbuild/proposals/blob/master/designs/2020-02-07-toolchain-transition-migration.md)? This will allow you to declare some toolchain dependencies as being in the target config (for the target being built with the toolchain), and others in the exec config, which may be useful with these deps.

I've looked at toolchain transitions, but I do not see how it fits into our case. Are there good examples to learn from?

liucijus:
I have migrated `scalac_provider_attr` to `dep_providers`, but come to conclusion that leaving concept of ScalacProvider is quite useful abstraction, so I left it inside phases.

scala/toolchains/README.md
katre: If scalac is a java_binary, why doesn't it have a complete set of deps? Can it only be built as a dependency of a scala target? Or have I misunderstood the build graph?

Instead of a string_to_label dict, the typical approach is to use a label_list, and when you get the DepInfo providers (is that the right provider from this attribute?) from the attribute, each DepsInfo has the provider id. Are you ever actually using the same target as different provider ids in different toolchains?

Toolchain transitions are new, and we're still in the migration period. The basic idea is that it allows your toolchain to separate out the tool dependencies (like scalac), which need to run on the execution platform, from your library dependencies (which are actually linked into your final target), and which need to be in the target configuration.

ittaiz:
>  have migrated scalac_provider_attr to dep_providers, but come to conclusion that leaving concept of ScalacProvider is quite useful abstraction, so I left it inside phases.

Can you elaborate on why?

liucijus:
> > have migrated scalac_provider_attr to dep_providers, but come to conclusion that leaving concept of ScalacProvider is quite useful abstraction, so I left it inside phases.
> 
> Can you elaborate on why?

Mostly to have less changes in current phases. Also scala deps have "special" meaning in scala rules which are design around them. Maybe this needs to be redesigned but I would prefer to do it outside toolchains scope, which is already quite big.

ittaiz:
To make sure I understand- the problem is only priorities and breaking down tasks? I'm treating scalac provider as another validation of your design and want to understand if we got negative signals

scala/toolchains/README.md
liucijus: I'll move provider id to DepsInfo. I don't like this approach, but it is semantically more correct. Mapping ids to labels feels more declarative and would be good for mapping the same targets more than once, but inverted map does not allow the same label twice anyway.

liucijus:
I have refactored dep_providers to label list

liucijus:
> To make sure I understand- the problem is only priorities and breaking down tasks? I'm treating scalac provider as another validation of your design and want to understand if we got negative signals

The question is if we want to change current phases. Now there's single phase which brings three depsets with ScalacProvider. Maybe it makes sense to split it into three new phases per depset and get rid of ScalacProvider. I didn't do it as I have no strong opinion about it and it was very easy to hook into existing phase. In general it's no a complicated change, but will touch quite a few places in the current code.

scala/providers.bzl
```diff
@@ -7,20 +7,26 @@ ScalacProvider = provider(
...
-declare_scalac_provider = rule(
-    implementation = _declare_scalac_provider,
+declare_deps_provider = rule(
+    implementation = _declare_deps_provider,
     attrs = {
-        "default_classpath": attr.label_list(allow_files = True),
-        "default_repl_classpath": attr.label_list(allow_files = True),
-        "default_macro_classpath": attr.label_list(allow_files = True),
+        "deps": attr.label_list(allow_files = True),
+        "depset_id": attr.string(mandatory = True),
```
katre, line 30: Unless this is actually somehow a [depset](https://docs.bazel.build/versions/master/skylark/lib/depset.html), I would suggest renaming.

scala/toolchains/README.md
katre: Thanks, I agree that the previous labe->id map was odd. Feel free to reach out to the Starlark API team on bazel-users and ask about this, but I suspect the answer will be that the label_list system you have is more "bazel-y".

In general, I really like this approach for solving the problem of adding transitive deps to things that consume scala targets. I think linking future toolchain authors to your docs will definitely be useful. Would it be possible for the author to also write a blog post or some other article with a high-level discussion to point other rule authors to? I realize this is even more work, but it could really help the community.

scala/toolchains/README.md
```diff
@@ -0,0 +1,163 @@
...
+    deps = ["@dep1", "@dep2"],
+    visibility = ["//visibility:public"],
+)
+```
+- rule `declare_deps_toolchain` - used to declare toolchains for deps providers. Eg.:
+```starlark
+declare_deps_toolchain(
+    name = "my_toolchain_impl",
+    dep_providers = {
+        ":my_compile_deps_provider": "compile_deps",
```
simuons, line 135: I think this is outdated

liucijus: Thanks for catching, I will update docs

scala/providers.bzl
liucijus: Thanks, I will rename it

ittaiz:
seems as lint fails

liucijus:
@ittaiz, I have fixed lint error and issues mentioned in the comments

PR #1073, opened 2020-07-21
[dummy] checking test_disappearing_class

or-shachar:
https://github.com/bazelbuild/rules_scala/pull/1071 is failing on `test_disappearing_class` and we have a hunch that it's not related to the change but to travis cache or something.

I ran the same test on my fork and [it didn't fail](https://travis-ci.org/github/or-shachar/rules_scala/jobs/710459641).

FYI @simuons @ittaiz


or-shachar:
:-( passed... but it also couldn't download cache so maybe it's related...

simuons:
@or-shachar since https://github.com/bazelbuild/rules_scala/pull/1081 is merged and now https://github.com/bazelbuild/rules_scala/pull/1071 passed I think this can be closed.

or-shachar:
Thanks!

PR #1080, opened 2020-07-23
[travisci] dry check latest bazel version

or-shachar:
### Description
Adding another automatic test (that doesn't fail the build) that will always check the repo against latest release of bazel.
This can help us identify breaking changes as early as possible and prepare mitigations and adjust the code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,6 +61,12 @@ jobs:
     - name: "[osx] Test scala version"
       <<: *osx
       env: TEST_SCRIPT=test_version XDG_CACHE_HOME=~/xdg_cache
+    - name: "[linux] Dry test rules_scala + latest bazel version"
+      <<: *linux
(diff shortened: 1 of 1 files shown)


ittaiz:
Build fails

or-shachar:
`test_disappearing_class`... :) I hope #1081 would solve it

ittaiz:
Can you rebase and force push?

or-shachar:
done... Once this is merged - I'll open some PRs to try to adjust the code to latest release. 

What is our policy ? If the code builds with bazel 3.4.1 - do we update the repo to be built with it or do we stay with older version to keep backward compatibility?

ittaiz:
We had a few different iterations of it.
I think our last decision (which we didn't keep up to) was to test using two versions- oldest possible and newest possible.
At some point we needed to break compatibility and move everything to latest and keep one version.
To be honest since Travis build times are so horrible I think we were so happy with one version we just stuck with it. We should probably change back to the 2 versions policy (running everything on both versions)

PR #1082, opened 2020-07-27
Add executable as a direct dependency

cocreature:
This PR fixes a bug where `rootpath` on a `scala_binary` resolved to the JAR instead of the wrapper binary.

This fixes a bug introduced by
https://github.com/bazelbuild/rules_scala/pull/958. Before #958,
`rootpath` on a Scala binary resolved to the wrapper binary so other
rules could treat it as an opaque executable. After that PR `rootpath`
now resolves to the JAR instead of the wrapper binary.

Adding the executable back as a direct dependency fixes that.

Happy to try and add a test for this but I need some guidance how and where to test this.

### DIFF
--- a/scala/private/phases/phase_default_info.bzl
+++ b/scala/private/phases/phase_default_info.bzl
@@ -7,6 +7,7 @@
 def phase_default_info(ctx, p):
     executable = None
     files = []
+    direct = None
     runfiles = []
 
     phase_names = dir(p)
@@ -18,6 +19,7 @@ def phase_default_info(ctx, p):
         if hasattr(phase, "executable"):
             if executable == None:
                 executable = phase.executable
+                direct = [executable]
             else:
                 fail("only one executable may be provided")
 
@@ -31,11 +33,11 @@ def phase_default_info(ctx, p):
         external_providers = {
             "DefaultInfo": DefaultInfo(
                 executable = executable,
-                files = depset(transitive = files),
+                files = depset(direct = direct, transitive = files),
                 # TODO:
                 # Per Bazel documentation, we should avoid using collect_data. The core phases need to be updated
                 # before we can make the adjustment.
-                runfiles = ctx.runfiles(transitive_files = depset(transitive = runfiles), collect_data = True),
+                runfiles = ctx.runfiles(transitive_files = depset(direct = direct, transitive = runfiles), collect_data = True),
             ),
         },
     )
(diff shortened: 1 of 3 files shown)


cocreature:
@googlebot I signed it!

scala/private/phases/phase_default_info.bzl
```diff
@@ -31,11 +33,11 @@ def phase_default_info(ctx, p):
         external_providers = {
             "DefaultInfo": DefaultInfo(
                 executable = executable,
-                files = depset(transitive = files),
+                files = depset(direct = direct, transitive = files),
                 # TODO:
                 # Per Bazel documentation, we should avoid using collect_data. The core phases need to be updated
                 # before we can make the adjustment.
-                runfiles = ctx.runfiles(transitive_files = depset(transitive = runfiles), collect_data = True),
+                runfiles = ctx.runfiles(transitive_files = depset(direct = direct, transitive = runfiles), collect_data = True),
```
cocreature, line 40: Following https://groups.google.com/g/bazel-discuss/c/oEyEIr7vaF4/m/0E5QaK46AQAJ it also needs to be in runfiles.

ittaiz:
Thanks!
What were you trying to do and failed? That's usually the test we want :)
If the bug means you can't run scala binaries then it sounds we roughly have two options here:
1. We add an sh_test that depends on a scala_binary and tries to run that. I think that will work but not 100% sure and might require some tweaking. The plus side of this is that this test is part of bazel.
2. We add an e2e bash test (see our bash tests in the root of the repo) which can call bazel run on a binary and see that it printed something you expect.

 If it's something else (because you mention something about other rules) then this again relates to my first sentence.

I want us to merge this with a test.

cocreature:
`sh_test` should be good enough. That roughly mirrors what we were trying to do. I’ll see if I can get something working.

cocreature:
I added a testcase. It is slightly convoluted since `rootpath` in a `sh_test` has some special treatment  in Bazel so we need to call it in a `genrule` to observe the difference. Without this patch you will only get the JAR in the rootpaths which was also the behavior before the PR I linked to.

simuons:
Hi @cocreature, could you run lint (ci is failing on that)?

cocreature:
Done, sorry I missed the failure.

simuons:
@ittaiz I think this is ready to be merged.

PR #1084, opened 2020-07-28
[bazel-3.4.1] adjustments

or-shachar:
Not a lot of changes towards complying with bazel 3.4.1


ittaiz:
Actually, should the readme have changed as well?

PR #1087, opened 2020-08-05
Add protobuf deps to "Getting Started"

### LINKED ISSUE #1083: default instructions are incomplete
If you follow:

https://github.com/bazelbuild/rules_scala#getting-started

and you use `com_google_protobuf` you will get an error like:
```
ERROR: /home/oscar/oss/bosatsu/cli/src/main/protobuf/bosatsu/BUILD:18:1: every rule of type proto_library implicitly depends upon the target '@com_google_protobuf//:protoc', but this target could not be found because of: error loading package '@com_google_protobuf//': Unable to find package for @rules_python//python:defs.bzl: The
repository '@rules_python' could not be resolved.
ERROR: Analysis of target '//cli/src/main/protobuf/bosatsu:bosatsu_proto' failed; build aborted: error loading package '@com_google_protobuf//': Unable to find package for @rules_python//python:defs.bzl: The repository '@rules_python' could not be resolved.
```
So, it seems the user needs to also set up a dependency on a (compatible?) version of rules_python.

I fixed this by adding:
```
load("@com_google_protobuf//:protobuf_deps.bzl", "protobuf_deps")
protobuf_deps()
```

blorente:
### Description
As described in #1083, there's no official guideline yet on how to set up transitive repository deps.
In the interim, this PR modifies the "Getting Started" section of the `Readme.md` (and the example `WORKSPACE.template`) to add the necessary dependencies.

### Motivation
We would like the "Getting Started" workflow to be as close to "copy, paste and go" as possible. This means that we need to include those dependencies in the Readme, even if some users end up having to modify it afterwards to fit their project needs.

Fixes #1083

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,11 @@ http_archive(
     strip_prefix = "protobuf-%s" % protobuf_version,
     sha256 = protobuf_version_sha256,
 )
+
+# Dependencies needed for google_protobuf.
+# You may need to modify this if your project uses google_protobuf for other purposes.
+load("@com_google_protobuf//:protobuf_deps.bzl", "protobuf_deps")
+protobuf_deps()
 ```
 
 This will load the `rules_scala` repository at the commit sha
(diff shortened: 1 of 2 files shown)


ittaiz:
waiting until green

PR #1090, opened 2020-08-07
Add javax annotations-api so it works with jdk 11

beala-stripe:
### Description

This adds support for compiling twitter scrooge files with jdk11.

scrooge makes use of the `@javax.annotation.Generated` annotation when code gen'ing `union`s. The location of this annotation moved between jdk8 and jdk11. The `javax.annotation-api` dep I added shims this definition in.

You can see the scrooge project making the same change here: https://github.com/twitter/scrooge/commit/78b06e92b9a5268f1c684eb186f7bff1dd429886
Related issue: https://github.com/twitter/scrooge/issues/299

### DIFF
--- a/test/shell/test_twitter_scrooge.sh
+++ b/test/shell/test_twitter_scrooge.sh
@@ -1,0 +1,11 @@
+# shellcheck source=./test_runner.sh
+dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+. "${dir}"/test_runner.sh
+. "${dir}"/test_helper.sh
+runner=$(get_test_runner "${1:-local}")
+
+scrooge_compile_with_jdk_11() {
+    bazel build --javacopt='--release 11' test/src/main/scala/scalarules/test/twitter_scrooge/...
+}
+
+$runner scrooge_compile_with_jdk_11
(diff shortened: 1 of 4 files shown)


beala-stripe:
Yup that's right. You can see the move here:

Java 8: https://docs.oracle.com/javase/8/docs/api/javax/annotation/Generated.html
Java 11: https://docs.oracle.com/en/java/javase/11/docs/api/java.compiler/javax/annotation/processing/Generated.html

I'll take a look at the test failure and ping you when it's fixed. It wasn't happening locally, so I suspect it has something to do with the JDK installed on my laptop vs travis.

ittaiz:
Why do you need a newer bazel version? I'd really rather we keep the minimal supported bazel version to as low as we can (assuming no huge value is needed)

beala-stripe:
@ittaiz Right now I'm throwing spaghetti against the wall and seeing what sticks. I can't repro locally, so I'm having to push up my experiments. I was wondering if the default jdk in 2.0.0 was jdk 8 only. I'll remove that change and clean up my commits once it's working. I'll set this to a draft PR in the meantime.

beala-stripe:
@ittaiz It looks like the build will pass at this point, assuming I didn't break something else. Mind taking another look? I've reverted the change to the bazel version.

ittaiz:
Thanks!

PR #1098, opened 2020-08-27
Revert "Disable non-functional version check in rule_impl.bsl"

comius:
This reverts commit 9a50e1b22d986d4c4fa292659c923960f48e1f6f.

### Description
This change was not needed. Even with previous commit rules_scala builds on downstream tests (with latest unreleased bazel version): https://buildkite.com/bazel/rules-scala-scala/builds/2265


ittaiz (approved):
Thanks!

Issue #1091, opened 2020-08-14
Upgrade to bazel 2.2.0


tian000:
I would like to use Go and Scala in the same monorepo, but the bazel minimum version for rules_go is 2.2.0, but the max version for rules_scala (according to README) is 2.0.0.

ittaiz:
I think it's not the max but rather the minimal we know works.
Would you be interested in sending a Travis PR to add the latest bazel version to all builds? This will allow us to know the minimal and the max

simuons:
@tian000 as @ittaiz said `2.0.0` is min version build is run. But there is a build for `latest` bazel version (which currently is `3.4.1`) please see here https://travis-ci.org/github/bazelbuild/rules_scala/jobs/720924785#L217

PR #1099, opened 2020-08-27
Clarify readme on compatible bazel versions

simuons:
### Description
Make a note in readme that `2.0.0` is a minimal supported bazel version and that there is a PR checks against latest bazel version

### Motivation
Resolves https://github.com/bazelbuild/rules_scala/issues/1091

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ for an example workspace using another scala version.
 | 0.14.x | 3b9ab9be31ac217d3337c709cb6bfeb89c8dcbb1 |
 | 0.13.x | 3c987b6ae8a453886759b132f1572c0efca2eca2 |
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -169,6 +169,8 @@ for an example workspace using another scala version.
 | 0.14.x | 3b9ab9be31ac217d3337c709cb6bfeb89c8dcbb1 |
 | 0.13.x | 3c987b6ae8a453886759b132f1572c0efca2eca2 |
 
+PRs are also built with latest released bazel version (see [travis config](https://github.com/bazelbuild/rules_scala/blob/master/.travis.yml#L66))
```
ittaiz, line 172: Wdyt about changing this to say with highest supported bazel version (right now it's latest but there were times where this wasn't true).
We can say (see Travis config for the exact highest version)

ittaiz (approved):
Thanks!

Issue #940, opened 2020-01-18
Tracking issue- move from deps to toolchains


ittaiz:
Description:
We'd like to move from using the deps/repositories pattern and bind to toolchains.
A great deal of the work has been done in #399 #530 and finally in #544 but we only handled scala itself and left the other parts of the repo untouched. This means that using a different proto version or specs2 is still difficult. 
Motivation:
For one bind is deprecated, it causes issues with labels (strict-deps/unused-deps) and also the configurability roadmap (which toolchains belongs to) specifies that one should be able to run bazel with more than one toolchain configured (so we will be able to have a single bazel build cross building scala versions for example).

- [ ] [proto](https://github.com/bazelbuild/rules_scala/issues/403)
- [ ] [scrooge](https://github.com/bazelbuild/rules_scala/issues/402)
- [ ] [specs2](https://github.com/bazelbuild/rules_scala/issues/401)
- [ ] [scala_test](https://github.com/bazelbuild/rules_scala/issues/400)
- [ ] tut
- [ ] jmh

smparkes:
FYI, more than anything:

We're looking at what it will take to be able to compile with multiple versions of scala in the same build. We only need to support different major versions, we don't see the need to support different minor versions.

We need to be able to build the same (for some definition of "same") library/binary against different versions of scala and also different versions of dependencies. Doing something like external templating of `WORKSPACE` and/or `BUILD` files is a very last resort for us.

We end up defining `BUILD/.bzl` pairs where targets are defined in macros that take an environment parameter and create targets for that environment. We then call that macro once for each environment. We implement this a bit like scala does with version suffixes. This lets us build the "same" target e.g., `//package/:some_library_{env}` for different runtime environment (i.e., classpaths) where we don't have control over all the dependences.

The macros handle all the muck around scala version suffixes and also clean up some of the junk that we, along with @johnynek, found annoying with `rules_jvm_external`. We write deps that look like
```
        "@{environment}//com.fasterxml.jackson.module:jackson_module-scala_{scala}",
```
and  use macro code to do the conversion to labels. We do find the pinning/version resolution in `rules_jvm_external`/`coursier ` really important.

[...]

PR #1104, opened 2020-09-11
ScalaPB deps toolchain

liucijus:
Refactors scalapb toolchain to use DepsInfo providers to pass required dependencies. Part of #940

### DIFF
--- a/docs/scala_proto_library.md
+++ b/docs/scala_proto_library.md
@@ -34,6 +34,7 @@ To configure ScalaPB options, configure a different `scala_proto_toolchain` and 
 
 ```python
 load("@io_bazel_rules_scala//scala_proto:scala_proto_toolchain.bzl", "scala_proto_toolchain")
+load("@io_bazel_rules_scala//scala:providers.bzl", "declare_deps_provider")
 
 scala_proto_toolchain(
     name = "scala_proto_toolchain_configuration",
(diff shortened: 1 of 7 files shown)


ittaiz:
is this a breaking change?

liucijus:
> is this a breaking change?

If I'm not missing anything, it should not be breaking (it refactors private attributes)

ittaiz (approved):

ittaiz:
Build is broken

liucijus:
@ianoc-stripe do you remember why [this](https://github.com/bazelbuild/rules_scala/blob/master/test/shell/test_deps.sh#L19) test is needed? Toolchains automatically include host paltform as an [execution platform](https://docs.bazel.build/versions/master/toolchains.html#toolchain-resolution). Is it a problem?

liucijus:
Moving this PR to draft, due to:
Generators like [`@io_bazel_rules_scala//src/scala/scripts:scalapb_worker`](https://github.com/bazelbuild/rules_scala/blob/e612d3e670ef1a8363488ab043139512c517750e/src/scala/scripts/BUILD#L53) depend on the same libs that come via proto toolchain, which ends up in a cyclic resolution [here](https://github.com/bazelbuild/rules_scala/blob/master/scala_proto/scala_proto_toolchain.bzl#L32). I assume we  need to introduce another toolchain for proto deps, so generators can get deps before their are resolved from the proto toolchain.

liucijus:
We may need to upgrade to at least Bazel 3.5.0 address toolchain limitations (#801 ). @ittaiz WDYT?

ittaiz:
I can live with that if I understand the exact value gain and reason.
Value gain is #751 ?
If so why do we need 3.5.0 for exactly? What does toolchain limitations mean? Best if we can have a text and a link to bazel issue + PR

liucijus:
Limitation is that with a current state of toolchains, all deps are added as host: https://github.com/bazelbuild/rules_scala/issues/797. It's address with an ongoing work on execution transitions (https://groups.google.com/g/bazel-discuss/c/vXrG6uFL5V8/m/RjK8_chaEgAJ). 
Even with 3.5.0 it is problematic as it requires flag `--incompatible_override_toolchain_transition` to be used. It would be reasonable to wait until the full support in later Bazel versions, but then we need an alternative way to address #751

ianoc-stripe:
@liucijus I might be missing something but how does the deps coming via the bind or the toolchain effect #751 ?

liucijus:
> @liucijus I might be missing something but how does the deps coming via the bind or the toolchain effect #751 ?

Bind should be ok, but toolchain deps will be added as host deps, no?

ianoc-stripe:
@liucijus having them in the bind keeps it off the host deps. But the issues around having deps in the toolchain or not is unconnected with #751 as far as i know. You can add/change the deps via the bind (indeed we bind these against local java_libraries iirc to setup the deps we use locally).  It would be nice to move the deps on the toolchain since it seems like thats how it should be done, but afaik it won't solve any of our other issues today

liucijus:
this PR is ready to be merged

PR #1108, opened 2020-09-17
Reduce duplication in directory delete utils

liucijus:
There are multiple implementation of recursive directory deletion utils. Let's have single implementation.


src/java/io/bazel/rulesscala/io_utils/DeleteRecursively.java
```diff
@@ -1,26 +0,0 @@
-package io.bazel.rulesscala.io_utils;
-
-import java.io.IOException;
-import java.nio.file.*;
-import java.nio.file.attribute.BasicFileAttributes;
-
-public class DeleteRecursively {
```
ittaiz, line 7: why not keep this implementation? Sounds better without the sort

test/coverage/DirectoryUtilsTest.scala
```diff
@@ -1,25 +0,0 @@
-package coverage;
-import org.scalatest._
-import java.nio.file.Files
-import java.io.File
-import io.bazel.rulesscala.coverage.instrumenter.DirectoryUtils
-
-class TestAll extends FlatSpec with Matchers {
```
ittaiz, line 7: should we keep a unit test for this?

ittaiz (commented):
Very much in favor of the PR in general. See two questions inside

src/java/io/bazel/rulesscala/io_utils/DeleteRecursively.java
liucijus: > why not keep this implementation? Sounds better without the sort

I left the stream based version, because I felt it has less boilerplate code

test/coverage/DirectoryUtilsTest.scala
liucijus: yes, I'll restore it

simuons (approved):

src/java/io/bazel/rulesscala/io_utils/DeleteRecursively.java
ittaiz: On the other hand it sorts and this one doesn't so it sounds more performant and we're already trying to optimize our perf

ittaiz:
@liucijus if you can minimize force pushing it would really help me review the code (github has better abilities to see new changes)

liucijus:
Unfortunately I had to force push to not to mess on top of orginal authorship of the code.

ittaiz (approved):

ittaiz:
failure is due to github

PR #1109, opened 2020-09-18
Reintroduce implicit_compile_deps

purkhusid:
## Why
Reverts #801

## Why
ScalaPB features like extra imports and type mappers don't work with the current implementation of scala_proto_library.
This adds back `implicit_compile_deps` tha allows the user to add extra implicit dependencies added to all scala_proto_library targets.

## How
Reverted some of the changes in #801. This should work with the latest versions of Bazel since toolchain transitions have been introduced.

This is not the ideal solution since changes to theses dependencies will trigger a rebuild of all the proto targets.

But it can be an escape hatch for people already using these features and are looking to migrate to Bazel.


liucijus:
There's a similar work related to proto toolchains and ability to specify custom deps via toolchain here #1104 . But it's blocked by lack of execution transitions support in older bazel versions.

purkhusid:
Closing in favour of #1104

PR #1110, opened 2020-09-21
ScalaTest deps to toolchain

liucijus:
ScalaTest deps on toolchain, part of #940

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -1,6 +1,20 @@
 ## Testing toolchain configuration
 
-Toolchain type `testing_toolchain_type` is used to set up test dependencies. 
+Toolchain type `testing_toolchain_type` is used to set up test dependencies. You can customize
+test dependencies by defining a custom testing toolchain.
+
+In your `WORKSPACE` default repositories and toolchains can be loaded via:
+```starlark
+# JUnit 4
+load("@io_bazel_rules_scala//testing:junit.bzl", "junit_repositories", "junit_toolchain")
+junit_repositories()
+junit_toolchain()
+
+# ScalaTest
+load(""@io_bazel_rules_scala//testing:scalatest.bzl", "scalatest_repositories", "scalatest_toolchain")
+scalatest_repositories()
+scalatest_toolchain()
+```
 
 ### Example to set up JUnit dependencies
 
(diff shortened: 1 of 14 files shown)


docs/testing.md
```diff
@@ -38,8 +38,51 @@ declare_deps_provider(
 `junit_classpath_provider` (deps_id `junit_classpath`) is where classpath required for junit tests
 is defined.
 
-Toolchain must be registerd in your `WORKSPACE` file: 
+ScalaTest support can be enabled by configuring a provider with an id `scalatest_classpath`:
```
ittaiz, line 41: unrelated to this change but this doc seems to suggest that in order to use scalatest I need to configure it. Why doesn't it show the easy defaults way?

ittaiz (approved):

docs/testing.md
liucijus: Not sure I understood your comment, there are no defaults when setting up custom toolchain

ittaiz: Is `docs/testing.md` only about custom testing toolchains?

liucijus: Probably not, but I think it requires a lot more effort to rethink/rewrite whole setting up rules scala docs. I think it should be done outside this PR.

ittaiz: My point- configuring custom deps became more complicated (but more powerful and we're ok with it).
If the readme before caused people to think they need to do X to use it now it seems like one needs to do 5X.
Can you please add a note here clarifying that this is needed only for the custom case?

liucijus: I have reorganized doc and added a note about custom deps

ittaiz: Thanks. I appreciate it

ittaiz (approved):

testing/scalatest.bzl
```diff
@@ -0,0 +1,6 @@
+def scalatest_repositories():
+    # currently ScalaTest dependencies are already loaded via //scala:scala.bzl#scala_repositories()
+    pass
+
+def scalatest_toolchain():
+    native.register_toolchain("//testing:scalatest_toolchain")
```
TheKaur, line 6: @liucijus Is this supposed to be `register_toolchains`? 
Added the following in my WORKSPACE file:
```
load("@io_bazel_rules_scala//testing:scalatest.bzl", "scalatest_toolchain")

scalatest_toolchain()
```
and it errored out with:
```
Error: no native function or rule 'register_toolchain'
```

liucijus: Thanks for reporting, here's the fix: https://github.com/bazelbuild/rules_scala/pull/1130

PR #1113, opened 2020-09-24
Keep all maven deps in the central place for easier version management

liucijus:
I've been thinking how to make external maven deps management simpler and decided to move all maven deps into single location per Scala version. I hope this will bring better visibility on which deps we are using and help update them more routinely. 

My hope is that this refactoring helps towards:
- an easier manual or automated update of versions
- adding support of new Scala versions, eg #809
- a unified user experience when consuming rules with deps from repository macros
- less code duplication

**This is a breaking change for some repository users (eg. scrooge).** It's also not fully tested yet, marking PR as a draft and calling for your opinions.


blorente (commented):
In general, this looks really good and a change in a direction I really love to see, thank you so much!

I've left a comment regarding some implementation details on the scrooge side. Don't really mind breaking changes, as long as we can keep the functionality described in some form.

Let me know if I can help any more!

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -29,148 +23,72 @@ load(
...
         scala_version = _default_scala_version(),
         maven_servers = _default_maven_server_urls(),
-        # These target labels need maven_servers to compute sensible defaults.
-        # Therefore we leave them None here.
-        libthrift = None,
-        scrooge_core = None,
-        scrooge_generator = None,
-        util_core = None,
-        util_logging = None):
+        overriden_artifacts = {}):
```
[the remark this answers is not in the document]
liucijus: @blorente Do you want to keep using override approach, or would you consider moving dependencies to a toolchain? I have a branch for scrooge already, but I will make a PR after other toolchains PRs will progress: https://github.com/liucijus/rules_scala/tree/scrooge-toolchain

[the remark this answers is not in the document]
blorente: Moving to a toolchain would be great, as long as we keep avoiding the resolution for targets that we override.

We'd be happy to migrate to toolchains, looking forward to that PR, we can discuss the specifics there :)

I guess, if this PR lands before the scrooge toolchains PR, we would be blocked on upgrading `rules_scala` until the scrooge toolchains PR lands, but I think that is not a big problem :)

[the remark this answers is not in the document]
liucijus: I hope to get with the toolchain first, otherwise I will leave previous interface

ittaiz:
build still fails

junit/junit.bzl
```diff
@@ -2,31 +2,23 @@ load(
...
 
-def junit_repositories(maven_servers = _default_maven_server_urls()):
-    _scala_maven_import_external(
-        name = "io_bazel_rules_scala_junit_junit",
-        artifact = "junit:junit:4.12",
-        artifact_sha256 = "59721f0805e223d84b90677887d9ff567dc534d7c502ca903c0c2b17f05c116a",
-        licenses = ["notice"],
-        server_urls = maven_servers,
+def junit_repositories(
+        maven_servers = _default_maven_server_urls(),
```
simuons, line 8: `maven_servers` is not passed to `repositories`

scala/scalafmt/scalafmt_repositories.bzl
```diff
@@ -19,484 +15,39 @@ def scalafmt_default_config(path = ".scalafmt.conf"):
...
-    )
-
-    _scala_maven_import_external(
-        name = "com_geirsson_metaconfig_typesafe_config",
-        artifact = "com.geirsson:metaconfig-typesafe-config_{major_version}:{extra_jar_version}".format(
-            major_version = major_version,
-            extra_jar_version = scala_version_extra_jars["com_geirsson_metaconfig_typesafe_config"]["version"],
-        ),
-        artifact_sha256 = scala_version_extra_jars["com_geirsson_metaconfig_typesafe_config"]["sha256"],
+        maven_servers = _default_maven_server_urls(),
```
simuons, line 48: I think `maven_servers` should be taken from prams of `scalafmt_repositories `

junit/junit.bzl
liucijus: Thanks for catching, @simuons!

scala_proto/private/scala_proto_default_repositories.bzl
```diff
@@ -5,439 +5,197 @@ load(
...
-    _scala_maven_import_external = "scala_maven_import_external",
-)
+load("//third_party/repositories:repositories.bzl", "repositories")
 
 def scala_proto_default_repositories(
         scala_version = _default_scala_version(),
-        maven_servers = _default_maven_server_urls()):
+        maven_servers = _default_maven_server_urls(),
+        overriden_artifacts = {}):
     major_version = _extract_major_version(scala_version)
```
simuons, line 14: I think `major_version` is not passed to `repositories`

specs2/specs2.bzl
```diff
@@ -5,84 +5,28 @@ load(
...
 
 def specs2_version():
     return "4.4.1"
 
 def specs2_repositories(
         scala_version = _default_scala_version(),
-        maven_servers = _default_maven_server_urls()):
+        maven_servers = _default_maven_server_urls(),
+        overriden_artifacts = {}):
     major_version = _extract_major_version(scala_version)
```
simuons, line 17: Maybe this `major_version = _extract_major_version(scala_version)` could be in `repositories` function? It would be possible to remove this duplicated line from other `*_repositories`

simuons (approved):
👍

specs2/specs2.bzl
```diff
@@ -2,87 +2,28 @@ load(
...
-        artifact = _scala_mvn_artifact(
-            "org.specs2:specs2-matcher:" + specs2_version(),
-            major_version,
-        ),
-        artifact_sha256 = scala_version_jar_shas["specs2_matcher"],
-        deps = ["@io_bazel_rules_scala_org_specs2_specs2_common"],
-        licenses = ["notice"],
-        server_urls = maven_servers,
+        scala_version = scala_version,
+        maven_servers = maven_servers,
```
or-shachar, line 24: If we're using the default servers all the time - why can't it be just the default value and skip the explicit parameter?

third_party/repositories/repositories.bzl
```diff
@@ -0,0 +1,38 @@
...
+
+artifacts_by_major_scala_version = {
+    "2.11": _artifacts_2_11,
+    "2.12": _artifacts_2_12,
+}
+
+def repositories(
+        scala_version = default_scala_version(),
+        for_artifact_ids = [],
+        maven_servers = default_maven_server_urls(),
```
or-shachar, line 22: Oh I see you already call here for default value - so can we remove the explicit parameter from caller site?

third_party/repositories/repositories.bzl
```diff
@@ -0,0 +1,38 @@
...
+    "@io_bazel_rules_scala//scala:scala_maven_import_external.bzl",
+    _scala_maven_import_external = "scala_maven_import_external",
+)
+
+artifacts_by_major_scala_version = {
+    "2.11": _artifacts_2_11,
+    "2.12": _artifacts_2_12,
+}
+
+def repositories(
```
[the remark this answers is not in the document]
or-shachar: I understand why you chose repositories... To keep current naming convention - but maybe it's an opportunity to name it better? `define_maven_external_dependencies` or something similar?

third_party/repositories/scala_2_11.bzl
```diff
@@ -0,0 +1,416 @@
+artifacts = {
```
or-shachar, line 1: Awesome!! How did you create this list? Did you use some kind of generator?
What happens when the next dev wants to update something?

or-shachar (commented):
Looks very good Vaidas!!  I think once this is merged we should prioritize creating a base list and have some sort of tool that will generate those files automatically for all scala flavors.

third_party/repositories/repositories.bzl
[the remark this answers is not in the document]
liucijus: I prefer shorter names, I think finding specific names with underscores is harder for users. I think name that includes `repositories` is right as it aligns with its only usecase repository bzl pattern. I don't expect any other usecases.

specs2/specs2.bzl
liucijus: Users may provide other server urls

third_party/repositories/repositories.bzl
liucijus: Users will need to be able to provide their own urls

third_party/repositories/scala_2_11.bzl
liucijus: Currently it's a manual procedure to update deps

liucijus:
@blorente, as while we have discussion on #1116, I have restored existing Twitter Scrooge repositories interface. I would appreciate if you could take a look.
Otherwise this PR is in ready to be merged state.

blorente:
@liucijus Looks good to me, thanks for putting so much thought into scrooge!

ittaiz (approved):

PR #1114, opened 2020-09-24
Extended proto rule

purkhusid:
## What?
Adds a new rule for compiling protobuf to scala. This rule is not aspect based and that allows the user to add extra dependencies to the compilation.

## Why?
The current proto rule does not support adding extra dependencies becaue it's aspect based.

## How?
Extracted some of the functionality of the aspect based rule and reused it in the new rule.

### DIFF
--- a/scala_proto/private/compile_proto.bzl
+++ b/scala_proto/private/compile_proto.bzl
@@ -1,0 +1,82 @@
+load("//scala/private:rule_impls.bzl", "compile_scala")
+load("//scala/private:common.bzl", "write_manifest_file")
+load("//scala/private:dependency.bzl", "legacy_unclear_dependency_info_for_protobuf_scrooge")
+
+def _concat_lists(list1, list2):
+    all_providers = []
+    all_providers.extend(list1)
+    all_providers.extend(list2)
+    return all_providers
+
+def compiled_jar_file(actions, scalapb_jar):
+    scalapb_jar_name = scalapb_jar.basename
+
+    # ends with .srcjar, so remove last 6 characters
+    without_suffix = scalapb_jar_name[0:len(scalapb_jar_name) - 6]
+
+    # this already ends with _scalapb because that is how scalapb_jar is named
+    compiled_jar = without_suffix + "jar"
+    return actions.declare_file(compiled_jar, sibling = scalapb_jar)
+
+def compile_proto(
+        ctx,
+        scalac,
+        label,
+        output,
+        scalapb_jar,
+        deps_java_info,
+        implicit_deps,
+        resources,
+        resource_strip_prefix):
+    manifest = ctx.actions.declare_file(
+        label.name + "_MANIFEST.MF",
+        sibling = scalapb_jar,
+    )
+    write_manifest_file(ctx.actions, manifest, None)
+    statsfile = ctx.actions.declare_file(
+        label.name + "_scalac.statsfile",
+        sibling = scalapb_jar,
+    )
+    diagnosticsfile = ctx.actions.declare_file(
+        label.name + "_scalac.diagnosticsproto",
+        sibling = scalapb_jar,
+    )
+    merged_deps = java_common.merge(_concat_lists(deps_java_info, implicit_deps))
+
+    # this only compiles scala, not the ijar, but we don't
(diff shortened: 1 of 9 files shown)


purkhusid:
This is my first draft of this rule. I just wanted to see if there was interest in getting it into `rules_scala`.

scala_proto/private/compile_proto.bzl
```diff
@@ -0,0 +1,82 @@
...
+    scalapb_jar_name = scalapb_jar.basename
+
+    # ends with .srcjar, so remove last 6 characters
+    without_suffix = scalapb_jar_name[0:len(scalapb_jar_name) - 6]
+
+    # this already ends with _scalapb because that is how scalapb_jar is named
+    compiled_jar = without_suffix + "jar"
+    return actions.declare_file(compiled_jar, sibling = scalapb_jar)
+
+def compile_proto(
```
ianoc, line 21: this isn't really compiling proto, this is compiling scala code. The scala has been generated at this point.

scala_proto/scala_proto_extended.bzl
```diff
@@ -0,0 +1,98 @@
...
+        imps,
+        compile_protos,
+        "" if ctx.attr.proto[ProtoInfo].proto_source_root == "." else ctx.attr.proto[ProtoInfo].proto_source_root,
+    )
+    return [
+        java_info,
+        DefaultInfo(files = outs),
+    ]
+
+scala_proto_library_extended = rule(
```
simuons, line 75: I wouldn't call it "extended" it doesn't extend original rule but rather changes behaviour significantly. Sorry I know naming is hard and I don't have anything to offer

purkhusid: Haha, yeah. This was mainly a placeholder while I tune it.

simuons:
@purkhusid have you tried this rule on your code base? I'm wondering how many other requirements would pop up.
I think with this rule you would need to replicate whole `proto_library` hierarchy with `scala_proto_library_extended` as in your example https://github.com/bazelbuild/rules_scala/issues/751#issuecomment-698903655 
I would suggest to have that rule locally and try it on your codebase. This would reveal all your requirements.
In my opinion this might be a good addition but also a heavy one. In a sense that there would be two distinct ways how to generate scala from proto.

purkhusid:
> @purkhusid have you tried this rule on your code base? I'm wondering how many other requirements would pop up.
> I think with this rule you would need to replicate whole `proto_library` hierarchy with `scala_proto_library_extended` as in your example [#751 (comment)](https://github.com/bazelbuild/rules_scala/issues/751#issuecomment-698903655)
> I would suggest to have that rule locally and try it on your codebase. This would reveal all your requirements.
> In my opinion this might be a good addition but also a heavy one. In a sense that there would be two distinct ways how to generate scala from proto.

Yeah, I'll try it out in my codebase first and see what pops up. There is one thing that I would like to fix but I'm unable to figure it out, how do I only pass top level dependencies to the compilation? E.g. in the test I created in this PR the `custom_import_lib`target is pulled transitively to the `proto2_test` compilation. Any pointers on how to to limit the transitive dependencies in the compilation?

simuons:
> Any pointers on how to to limit the transitive dependencies in the compilation?

I think this is the place where all transitive deps gets merged https://github.com/bazelbuild/rules_scala/pull/1114/files#diff-65fac07922564e63374c6db4871ef5dcR44

ittaiz:
I agree with @simuons in general that this is a heavy addition that should be done only if it's applicable to enough use cases (and definitely only after you see it fully helps your use case).
Additionally the direction of having different strategies for messages and services (like java) makes a lot of sense.

ianoc:
I do really like the separation of the services out from the message generation (we have extra/special services generation stuff similar to wix, finagle based services services, and then also some other ones for doing http 1 requests where needed). 

I think those transitive deps in the compilation bit fwiw (though I don't know if its needed/should go away in general since the aspect should be aware of the direct deps I believe), but those aren't the issue of the old rules I'm pretty sure -- those are jar dependencies rather than the old problem is multiple sets of sources. (difference here is that jar dependencies won't get merged into the output jar, where as a source dependency would be in the output jar)

liucijus:

PR #1117, opened 2020-09-28
Fix scalac label for scala_proto_library

simuons:
### Description
When building `scala_proto_library` in users workspace then local labels are incorrectly prefixed with `@io_build_rules_scala//`.

### Motivation
After bazel upgrade we noticed that some of our action labels have `@io_build_rules_scala` prefix. Probably something changed in [label construction](https://github.com/bazelbuild/rules_scala/compare/master...simuons:fix-scalac-label-for-scala_proto_library?expand=1#diff-9fefd61d81bab6c5a07945a029e0737aR77). External proto targets seems to be fine.


simuons:
@ianoc why this label conversion is needed https://github.com/bazelbuild/rules_scala/blob/master/scala_proto/private/scalapb_aspect.bzl#L77?

ianoc:
@simuons I think the label would otherwise be the same as the proto itself -- this is the same thing used in the java proto rules, could see if the same issue regarding location applies there?

ianoc:
(tbc, I've no idea if anything bad would happen if we didn't update the label.. I was attempting to following the the java one)

ittaiz:
@simuons the build is failing because your test is downloading bazel. Might be related to bazelisk and aquery 🤷 
can you take a look?

ittaiz:
@simuons did you try to fix the label construction instead of dropping the conversion? using the various flags of `Label` and such.

simuons:
@ittaiz 
> did you try to fix the label construction instead of dropping the conversion? using the various flags of Label and such.

Looking at [Label constructor](https://docs.bazel.build/versions/master/skylark/lib/Label.html#Label) there is a single property `relative_to_caller_repository` I could play with, but it's marked as `Deprecated. Do not use.`

Maybe I can pass it as a string and keep that suffix. I'll try.

> the build is failing because your test is downloading bazel. Might be related to bazelisk and aquery

Yeah test looks too complicated for the actual change but I'm out of ideas how to make it simpler.

ittaiz:
@or-shachar can you lend a hand?

simuons:
Sorry for lengthy comment.

So I've added `java_proto_library(name = "java", deps = [":proto"])` in https://github.com/bazelbuild/rules_scala/pull/1117/files#diff-c4c272791f5491a9f362a30f3ce26d8f and ran `bazel aquery --include_aspects --include_commandline=false --include_artifacts=false '//...'`  to compare `java_proto_library` and `scala_proto_library` actions.

This is the output I got
```
action 'Generating Descriptor Set proto_library //:proto'
  Mnemonic: GenProtoDescriptorSet
  Target: //:proto
  Configuration: darwin-fastbuild
  ActionKey: 252ea033fa61d96f766e88d8738bb319447d857406f320b00ed6fe04d8af1e9b

action 'Generating Java (Immutable) proto_library //:proto'
  Mnemonic: GenProto
  Target: //:proto
  Configuration: darwin-fastbuild
  AspectDescriptors: [BazelJavaProtoAspect()]
  ActionKey: 98eb68106d14fa6dd788b7cae34cbc600d4ac46e98831a766c93e4faf4f542f2

action 'Building libproto-speed.jar (1 source jar)'
  Mnemonic: Javac
  Target: //:proto
  Configuration: darwin-fastbuild
  AspectDescriptors: [BazelJavaProtoAspect()]
  ActionKey: e2c493aae9f08aa36247061dfcfa64e6740c2951c16e4e1343ae3a4b3fc00f92
  Environment: [LC_CTYPE=en_US.UTF-8]
  ExecutionInfo: {supports-workers: 1}

action 'Compiling Java headers libproto-speed-hjar.jar (1 source jar)'
  Mnemonic: Turbine
  Target: //:proto
  Configuration: darwin-fastbuild
  AspectDescriptors: [BazelJavaProtoAspect()]
  ActionKey: 6638a2e21f2e8e9b718aa55e0f56d55a38ac64a2f59139bee7b26a647ef4d7d1
  Environment: [LC_CTYPE=en_US.UTF-8]

action 'Writing file proto_worker_input'
  Mnemonic: FileWrite
  Target: //:proto
  Configuration: darwin-fastbuild
  AspectDescriptors: [@io_bazel_rules_scala//scala_proto/private:scalapb_aspect.bzl%scalapb_aspect()]
  ActionKey: bd6419786df86e406dc1aa65f8d5fff7bc05b57e0e2d9b6fe00c0ae9605a7be4

action 'creating scalapb files //:proto'
  Mnemonic: ProtoScalaPBRule
  Target: //:proto
  Configuration: darwin-fastbuild
  AspectDescriptors: [@io_bazel_rules_scala//scala_proto/private:scalapb_aspect.bzl%scalapb_aspect()]
  ActionKey: cd2d1ae74cd16c6adee9a2ea812ffc1630dcd76ef331985476a13255be4254f7
  ExecutionInfo: {supports-workers: 1}

action 'Writing file proto_MANIFEST.MF'
  Mnemonic: FileWrite
  Target: //:proto
  Configuration: darwin-fastbuild
  AspectDescriptors: [@io_bazel_rules_scala//scala_proto/private:scalapb_aspect.bzl%scalapb_aspect()]
  ActionKey: 0815f59e853458a9262cf4b4cb89646cc04807ecb1d508bc2b13c9ebb8ba8822

action 'Writing file proto_scalac_worker_input'
  Mnemonic: FileWrite
  Target: //:proto
  Configuration: darwin-fastbuild
  AspectDescriptors: [@io_bazel_rules_scala//scala_proto/private:scalapb_aspect.bzl%scalapb_aspect()]
  ActionKey: 2d90f877b57ab43197f7e23659a7adbcc3f94618690d838cc546856a7d6c80d9

action 'scala //:proto'
  Mnemonic: Scalac
  Target: //:proto
  Configuration: darwin-fastbuild
  AspectDescriptors: [@io_bazel_rules_scala//scala_proto/private:scalapb_aspect.bzl%scalapb_aspect()]
  ActionKey: 40052f449ca31dfcf3efa5cb7aa263b543de750665ed53b15f1f96a728f5a7ec
  ExecutionInfo: {supports-workers: 1}
```

So as you see all actions refer to the same target `//:proto`, all of them have unique mnemonics.
`action` is displayed to user and can have anything. As you see java compilation targets have output jars in there instead of targets they refer to.

I think we can use original label (without conversion), but we need to make sure outputs are unique (since generation/compilation happens in `proto_library` context). And I think they are because most of them have `scalac` or `scalapb` in their name. Or we can use output jar base name in action string.

simuons:
btw build is failing because `test_scala_version` uses latest bazel and is not related to this pr

ittaiz (approved):

ittaiz:
Lint fails

third_party/test/proto/.bazelversion
```diff
@@ -0,0 +1 @@
+2.0.0
```
or-shachar, line 1: Better do [like I did in test_scala_version](https://github.com/bazelbuild/rules_scala/commit/ecae01705dc188724950a4b6a7d2904a38a5c326#diff-af9f5a9ff6dcf95828c50792eb95701fR105) and propagate either the USER_BAZEL_VERSION in or the content of root folder `.bazelversion` and the evaluated .bazelrc flags.

third_party/test/proto/WORKSPACE
```diff
@@ -0,0 +1,43 @@
+workspace(name = "proto")
```
or-shachar, line 1: Is it the first place we hardcode a WORKSACE in one of the tests? Maybe we better use one generator to ease on us? Not saying that it's mandatory for this PR but can be nice to open an issue and care for it later.

I know that using[ `bazel-integration-testing` ](https://github.com/bazelbuild/bazel-integration-testing)can be challenging for something larger then a rule that is contained in single bzl file but I wish we had a better solution to do proper e2e tests on this rather than this.

or-shachar (changes requested):
Thanks Simonas for caring for this! 🙏 

One mandatory change. few general comments.

ittaiz:
lint still fails

third_party/test/proto/.bazelversion
simuons: Done https://github.com/bazelbuild/rules_scala/pull/1117/commits/279a41fa8cf0cb7a74b04f09df00a1a644e7d157

PR #1121, opened 2020-10-14
Pass scalac as executable

simuons:
### Description
Mark `_scalac` implicit attribute as `executable = True` and pass it to `ctx.actions.run(executable)`

It used to be like so in the beginning but I think it changed in https://github.com/bazelbuild/rules_scala/pull/544 and I couldn't find reason why. Maybe someone has an explanation and we can just close this PR.

### Motivation
I stumbled upon this when implementing my own rule which required `scalac`. I looked at the `scalac` target and saw it is `java_binary` so I declared attribute like so 
```"_scalac": attr.label(default = @io_bazel_rules_scala//src/java/io/bazel/rulesscala/scalac", executable = True, cfg = "host")```

but it failed when passed to `compile_scala` with

> Error in run: Found tool(s) 'bazel-out/host/bin/external/io_bazel_rules_scala/src/java/io/bazel/rulesscala/scalac/scalac' in inputs. A tool is an input with executable=True set. All tools should be passed using the 'tools' argument instead of 'inputs' in order to make their runfiles available to the action. This safety check will not be performed once the action is modified to take a 'tools' argument. To temporarily disable this check, set --incompatible_no_support_tools_in_action_inputs=false.


simuons:
Answer why it was done so can be found here:

- https://github.com/bazelbuild/rules_scala/pull/530
- https://github.com/bazelbuild/bazel/issues/5463

liucijus (approved):

ittaiz:
Is it still draft?

simuons:
Note: this might be a breaking change for user who uses `compile_scala` directly. The way how to get `scalac` changes ie `ctx.attr._scala` to `ctx.executable._scalac`. But `compile_scala` is in `private` directory so maybe this is not an issue, in other words it is not a public api. Otherwise it is not a breaking change for regular rules users.

ittaiz (approved):

ittaiz:
@simuons please resolve conflicts

simuons:
@ittaiz resolved

PR #1127, opened 2020-10-23
Make coverage testing version independent

simuons:
### Description
Assert on which files were covered instead of comparing whole report.

### Motivation
Coverage report changed in bazel 3.7.0 see failing test https://travis-ci.org/github/bazelbuild/rules_scala/jobs/737941438


ittaiz:
But I think the point was actually to see that coverage works and produces results and not just sees which files were covered.
This change is a bit problematic

simuons:
Opened an issue https://github.com/bazelbuild/rules_scala/issues/1132 as proposed approach isn't sufficient.

PR #1130, opened 2020-10-27
Fix toolchain registration call for ScalaTest

liucijus:
Call correct function to register toolchain

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -4,6 +4,15 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 load("@bazel_tools//tools/build_defs/repo:jvm.bzl", "jvm_maven_import_external")
 
+skylib_version = "1.0.3"
+
+http_archive(
+    name = "bazel_skylib",
+    sha256 = "1c531376ac7e5a180e0237938a2536de0c54d93f5c278634818e0efc952dd56c",
+    type = "tar.gz",
+    url = "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/releases/download/{}/bazel-skylib-{}.tar.gz".format(skylib_version, skylib_version),
+)
+
 _build_tools_release = "3.3.0"
 
 http_archive(
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -128,8 +128,11 @@ format_repositories()
 
 http_archive(
     name = "io_bazel_rules_go",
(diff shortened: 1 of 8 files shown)


ittaiz:
How come it passed until now?

liucijus:
> How come it passed until now?

In `rules_scala` we use another toolchain, which has configurations for all testing frameworks: https://github.com/bazelbuild/rules_scala/blob/master/WORKSPACE#L50

ittaiz:
Do you think it's a good idea to keep it like this? Feels like a recipe for future bugs

liucijus:
@ittaiz I agree it's an opportunity for bugs. Added an example workspace which serves as an integration test for ScalaTest repository/toolchain helpers.

WORKSPACE
```diff
@@ -119,8 +128,11 @@ format_repositories()
 
 http_archive(
     name = "io_bazel_rules_go",
-    sha256 = "45409e6c4f748baa9e05f8f6ab6efaa05739aa064e3ab94e5a1a09849c51806a",
-    url = "https://github.com/bazelbuild/rules_go/releases/download/0.18.7/rules_go-0.18.7.tar.gz",
```
ittaiz, line 123: Why did you upgrade rules_go? Because of skylib?

testing/scalatest.bzl
```diff
@@ -3,4 +3,4 @@ def scalatest_repositories():
     pass
 
 def scalatest_toolchain():
-    native.register_toolchain("//testing:scalatest_toolchain")
+    native.register_toolchains("@io_bazel_rules_scala//testing:scalatest_toolchain")
```
ittaiz, line 6: Small question- your previous CI (IIRC) was without `@io_bazel_rules_scala` right? Why the change? What pushed it?

ittaiz (commented):
Thanks for the test workspace! Looks really good :) left few questions

WORKSPACE
liucijus: yes

testing/scalatest.bzl
liucijus: It was a mistake - all toolchains have to be registered with external name otherwise they may be treated as different

ittaiz: So it's precaution rather than something actually failed? (Ok with either, trying to understand which)

liucijus: I tried with the example repo - it doesn't even find the toolchain:
```
ERROR: Analysis of target '//example:example' failed; build aborted: invalid registered toolchain '//testing:scalatest_toolchain': no such package 'testing': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.

```

ittaiz: Amazing. Great that we have this test workspace

PR #1133, opened 2020-11-05
Introduce Scala config

liucijus:
This is a breaking change!

To support multiple versions of Scala within rules codebase we need a way to check for current Scala version in rule and macro code. For example, to support Scala 2.13 properly we need to be able to pass different scalac opts to some targets. This will allow to do conditionals like this: 
```starlark
load("@io_bazel_rules_scala_config//:config.bzl", "SCALA_MAJOR_VERSION")

scala_library(
  scalacopts = [
     # 2.13 options
  ] if SCALA_MAJOR_VERSION == "2.13" else [ 
     # options for other versions
  ],
)
``` 

In addition, I moved `bazel_version` under `scala_config` to simplify things `rules_scala` users need to configure. It will allow us to add more configuration options with default values without breaking an existing user code.

How to upgrade:
In `WORKSPACE` file replace
```
load("@io_bazel_rules_scala//:version.bzl", "bazel_version")
bazel_version(name = "bazel_version")
```
with
```
load("@io_bazel_rules_scala//:scala_config.bzl", "scala_config")
scala_config()
```
or if non-default Scala version is in use, with explicit Scala version specified:
```
load("@io_bazel_rules_scala//:scala_config.bzl", "scala_config")
scala_config(scala_version = "2.11.12")
```


scala_config.bzl
```diff
@@ -0,0 +1,36 @@
+load("@bazel_skylib//lib:versions.bzl", "versions")
+load("//scala:scala_cross_version.bzl", "extract_major_version")
+
+def _default_scala_version():
+    """return the scala version for use in maven coordinates"""
+    return "2.12.11"
+
+def _store_config(repository_ctx):
+    bazel_version = versions.get()
```
simuons, line 9: Maybe replace `versions.get()` with `native.bazel_version`? This is what `skylib` does under the hood. As it is now it requires `skylib` to be declared before `scala_config`. Maybe not an issue though.

README.md
```diff
@@ -50,8 +50,9 @@ http_archive(
     sha256 = "8c48283aeb70e7165af48191b0e39b7434b0368718709d1bced5c3781787d8e7",
 )
 
-load("@io_bazel_rules_scala//:version.bzl", "bazel_version")
-bazel_version(name = "bazel_version")
+#
```
simuons, line 53: Leftover?

scala_config.bzl
```diff
@@ -0,0 +1,36 @@
...
+_config_repository = repository_rule(
+    implementation = _store_config,
+    attrs = {
+        "scala_version": attr.string(
+            mandatory = True,
+        ),
+    },
+)
+
+def scala_config(scala_version = _default_scala_version()):
```
simuons, line 32: I'm not sure about `scala_config` naming. It contains only versions so maybe `versions` would be more appropriate. But as I understand it is a placeholder for other things as well. Maybe just `config` as it contains non scala related thing like `BAZEL_VERSION`? Either way I'm fine as it is currently. So you can just ignore this comment.

simuons (approved):

scala_config.bzl
liucijus: I have intentionally named it config to allow other configuration to be stored in the future.

README.md
liucijus: Actually a mistake - I forgot to add documentation text here

scala_config.bzl
liucijus: Thanks, good point!

WORKSPACE
```diff
@@ -177,7 +171,7 @@ jvm_maven_import_external(
     name = "org_typelevel__cats_core",
     artifact = scala_mvn_artifact(
         "org.typelevel:cats-core:0.9.0",
-        default_scala_major_version(),
+        SCALA_MAJOR_VERSION,
```
ittaiz, line 174: Why the change from method to const?

ittaiz (commented):
Maybe add a comment with a snippet on the PR with what people need to do to upgrade?

WORKSPACE
liucijus: Is there a reason to wrap the const inside a function? The default version functions are removed, ant the stored version should be a single source of truth.

liucijus:
> Maybe add a comment with a snippet on the PR with what people need to do to upgrade?

Done

ittaiz (approved):

katre:
The directions for project setup in README.md are now wrong: the suggested version a2f5852902f5b9f0302c727eead52ca2c7b6c3e2 does not contain scala_config.bzl.

This makes it pretty confusing for even a very savvy Bazel user to get started with rules_scala.

ittaiz:
Thanks! Is it only the version that's wrong? cc @liucijus

katre:
I updated to head and fixed the sha, and it seemed to work.

Have any good scala tutorials to point me at? :)

ittaiz:
It's a big question and I really don't know why you need it and how much time you want to allocate :) 
Small 
https://twitter.github.io/scala_school/
http://twitter.github.io/effectivescala/
Bigger
https://www.handsonscala.com/ (this is a book many colleagues said was helpful to dive into scala by real code and topics)
https://www.udemy.com/course/rock-the-jvm-scala-for-beginners/
https://courseware.epfl.ch/courses/course-v1:EPFL+progfun1+2018_T1/about

Some Scala exercises (like coding katas) https://www.scala-exercises.org/std_lib/asserts

Hope it helps

katre:
Thanks! I've been wanting to learn, so this year I'm planning to do https://adventofcode.com/ in Scala as a forcing function.

ittaiz:
Awesome! Welcome aboard to our community :)
Good luck and have fun...

PR #1142, opened 2020-11-18
Use released protoc binaries

simuons:
### Description
Based on change in rules_proto
https://github.com/bazelbuild/rules_proto/pull/36

`rules_proto_dependencies()` declares a `@com_google_protobuf//:protoc`
which points to released `protoc` binary.

The tricks is to load that before `scala_repositories()` which has
conditional load for `@com_google_protobuf`

### Motivation
Avoid compiling `protoc` during `bazel build`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,23 @@ http_archive(
     sha256 = "2ef429f5d7ce7111263289644d233707dba35e39696377ebab8b0bc701f7818e",
 )
 
+# rules_scala depend on rules_proto and @com_google_protobuf//:protoc
+http_archive(
+    name = "rules_proto",
+    sha256 = "8e7d59a5b12b233be5652e3d29f42fba01c7cbab09f6b3a8d0a57ed6d1e9a0da",
+    strip_prefix = "rules_proto-7e4afce6fe62dbff0a4a03450143146f9f2d7488",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/rules_proto/archive/7e4afce6fe62dbff0a4a03450143146f9f2d7488.tar.gz",
+        "https://github.com/bazelbuild/rules_proto/archive/7e4afce6fe62dbff0a4a03450143146f9f2d7488.tar.gz",
+    ],
+)
+
+load("@rules_proto//proto:repositories.bzl", "rules_proto_dependencies", "rules_proto_toolchains")
+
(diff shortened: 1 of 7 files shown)


liucijus (approved):

ittaiz (commented):
Thanks!
2 questions:
1. If existing clients don't have this will they break?
2. Update the readme?

simuons:
> If existing clients don't have this will they break?

Clients won't break. `rules_proto` and `com_google_protobuf` are defined in `rules_scala_setup` and guarded by `if`.  So they will get current defaults or whatever they configure.

> Update the readme?
Ok I'll remove current recommendation for `com_google_protobuf` from README and add `rules_proto` setup after `skylib`.

liucijus:
> Ok I'll remove current recommendation for `com_google_protobuf` from README and add `rules_proto` setup after `skylib`.

I think it's more important to explain in the REDAME what's going on. I would say that this change in the first place is intended for `rules_scala` development. I would keep this change out user attention until we properly refactor `repository.bzl`'s - meaning no update  to README or setup instructions.

ittaiz:
Why? Maybe I'm missing the point here. Is this really only meant for rules_scala development or do you just want to keep it internal until the repository.bzl refactor?

liucijus:
At the moment `rules_proto` are also loaded in `scala_repositories.bzl`. Until this duplication is refactored (which isn't a trivial change), I consider it will be confusing to users.

ittaiz:
Then why do we need to load them twice?

simuons:
It's loaded twice because I didn't want to change existing macro and break users. Well technically it isn't loaded twice because definition in macro is guarded by the if. I had to define rules_proto before scala_repositories macro because it contains definition for com_google_protobuf and I want to use definition of com_google_protobuf from rules_proto.

Guys can we agree that I'll revert readme change. And will treat this PR as an optimization for rules_scala build. And I'll look at repository.bzl refactoring to align setup of rules_scala and what user have to setup.

simuons:
Or feel free to close this PR if you think it should be done as a "proper refactoring of repositories.bzl"

ittaiz:
Let's revert the PR and merge.
Maybe just add a comment about why our WORKSPACE setup is different than the readme? If there is already a comment and I forgot about it then 👍🏽

PR #1145, opened 2020-11-21
Add Gemini as rules_scala user

jakemcc:
### Description

Updates README to add Gemini as a user of rules_scala.

### Motivation

Just trying to be transparent about our usage in hopes it can help others make the choice to also use `rules_scala`


ittaiz (approved):
🙏🏾 happy to hear!
Thanks for taking the time!

ittaiz:
Failure unrelated. @liucijus @simuons can you please take a look at why this failed?

Issue #1042, opened 2020-05-04
Exclude test targets from instrumentation when running coverage


kien-truong:
Currently, when I run coverage, `scala_test` targets are also instrumented and included in the coverage reports. 

Test targets should not be instrumented unless we set `--instrument_test_targets=true`

ittaiz:
That sounds right.
Are you interested in sending a PR?

kien-truong:
Sorry, I don't think I know enough about bazel and rules_scala to submit a fix for this yet.

liucijus:
Solved in #1147

PR #1147, opened 2020-11-23
Support --instrument_test_targets flag in coverage

liucijus:
Support `--instrument_test_targets` (default value is False) flag. Solves #1042 and partially addresses instability with rules_scala coverage testing when test libraries change (by excluding test target coverage).

### DIFF
--- a/scala/private/phases/phase_coverage_runfiles.bzl
+++ b/scala/private/phases/phase_coverage_runfiles.bzl
@@ -14,7 +14,7 @@ def phase_coverage_runfiles(ctx, p):
     if ctx.configuration.coverage_enabled and _coverage_replacements_provider.is_enabled(ctx):
         coverage_replacements = _coverage_replacements_provider.from_ctx(
             ctx,
-            base = p.coverage.replacements,
+            #            base = p.coverage.replacements,
         ).replacements
 
         rjars = depset([
--- a/scala/private/phases/phase_coverage_runfiles.bzl
+++ b/scala/private/phases/phase_coverage_runfiles.bzl
@@ -14,7 +14,7 @@ def phase_coverage_runfiles(ctx, p):
     if ctx.configuration.coverage_enabled and _coverage_replacements_provider.is_enabled(ctx):
         coverage_replacements = _coverage_replacements_provider.from_ctx(
             ctx,
-            #            base = p.coverage.replacements,
+            base = p.coverage.replacements if ctx.coverage_instrumented() else {},
         ).replacements
 
         rjars = depset([
--- a/test/shell/test_coverage.sh
+++ b/test/shell/test_coverage.sh
@@ -19,9 +19,5 @@ test_coverage_includes_test_targets() {
     grep -q "SF:test/coverage/TestAll.scala" $(bazel info bazel-testlogs)/test/coverage/test-all/coverage.dat
 }
 
-xmllint_test() {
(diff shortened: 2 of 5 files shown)


test/shell/test_coverage.sh
```diff
@@ -0,0 +1,27 @@
...
+
+test_coverage_includes_test_targets() {
+    bazel coverage \
+          --extra_toolchains="//scala:code_coverage_toolchain" \
+          --instrument_test_targets=True \
+          //test/coverage/...
+    grep -q "SF:test/coverage/TestAll.scala" $(bazel info bazel-testlogs)/test/coverage/test-all/coverage.dat
+}
+
+xmllint_test() {
```
blorente, line 22: This seems not used in this file. Did you mean to copy it over from `test/shell/test_misc.sh`?

scala/private/phases/phase_coverage_runfiles.bzl
```diff
@@ -14,7 +14,7 @@ def phase_coverage_runfiles(ctx, p):
     if ctx.configuration.coverage_enabled and _coverage_replacements_provider.is_enabled(ctx):
```
blorente, line 14: I'm very not familiar with the implementation, but given that this runs per-rule, it feels strange that we're not checking for `ctx.coverage_instrumented()` in this conditional and skipping the entire body (or providing explicitly empty results in an else branch) if it's not enabled.

Feel free to ignore this comment if there's complexity I'm missing.

blorente (commented):
I'm not familiar enough with the implementation to comment on `phase_coverage_runfiles.bzl` with any kind of confidence, but thanks for working on this!

test/shell/test_coverage.sh
liucijus: Thanks for catching!

scala/private/phases/phase_coverage_runfiles.bzl
liucijus: I may be wrong, but it still has to collect instrumented files from deps. Seems that appending `and ctx.coverage_instrumented()` produces an empty `coverage.dat`.

scala/private/phases/phase_coverage_runfiles.bzl
```diff
@@ -14,7 +14,7 @@ def phase_coverage_runfiles(ctx, p):
...
         coverage_replacements = _coverage_replacements_provider.from_ctx(
             ctx,
-            base = p.coverage.replacements,
+            base = p.coverage.replacements if ctx.coverage_instrumented() else {},
```
ittaiz, line 17: I think I'm missing something- this is the only prod code change, right?
If so can you explain how it solves the test targets?

ittaiz (commented):
Looks good but to be honest I just don't understand this change

scala/private/phases/phase_coverage_runfiles.bzl
liucijus: It means - if current rule is filtered out don't include its sources. `--instrument_test_targets` controls if by default test target sources should be filtered out.

ittaiz: Thanks! I missed the original issue.
I thought we never instrumented and that you added that under the flag.
Now I understand we always instrumented and now you use the flag to toggle it.
Thanks!

PR #1153, opened 2020-11-26
Coverage should use configuration instead of toolchain flag

liucijus:
Having a toolchain flag seems to be redundant and requires a separate toolchain for coverage. This complicates integration with tools like IDEs.
Let's use `ctx.configuration.coverage_enabled` to check if this Bazel run does coverage instrumentation.


scala/private/phases/phase_write_executable.bzl
```diff
@@ -109,7 +109,7 @@ def _write_executable_non_windows(ctx, executable, rjars, main_class, jvm_flags,
         wrapper.short_path,
     )
 
-    if use_jacoco and _coverage_replacements_provider.is_enabled(ctx):
+    if use_jacoco and ctx.configuration.coverage_enabled:
```
simuons, line 112: Probably unrelated to this PR but what `use_jacoco` stands for? Isn't it covered with `ctx.configuration.coverage_enabled`? I think I saw somewhere a code like this `use_jacoco = ctx.configuration.coverage_enabled`

simuons (approved):
👍

scala/private/phases/phase_write_executable.bzl
liucijus: Thanks, let me see if it can be removed

scala/private/phases/phase_write_executable.bzl
```diff
@@ -29,7 +29,6 @@ def phase_write_executable_scalatest(ctx, p):
             "-DRULES_SCALA_MAIN_WS_NAME=%s" % ctx.workspace_name,
             "-DRULES_SCALA_ARGS_FILE=%s" % p.runfiles.args_file.short_path,
         ] + expand_location(ctx, final_jvm_flags),
-        use_jacoco = ctx.configuration.coverage_enabled,
```
ittaiz, line 32: just so we're clear- you just broke the implicit API of this phase in case someone depends on this to csutomize this behavior of toggling coverage on/off.
I'm ok with it but I think that if we had providers you would have known that (I think you missed it but maybe I'm wrong) and more importantly extenders could easily know that because that provider would have changed and not a nitty gritty in the impl.

test/shell/test_coverage.sh
```diff
@@ -5,15 +5,12 @@ dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 runner=$(get_test_runner "${1:-local}")
 
 test_coverage_on() {
-    bazel coverage \
```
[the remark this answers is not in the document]
liucijus: Coverage is collected and controlled by Bazel, I doubt we can get much value out of such test

scala/private/phases/phase_write_executable.bzl
liucijus: Isn't this field private to this phase?

ittaiz: It seems I was mistaken and it is private

PR #1159, opened 2020-12-03
Fix failed build reporting for Scala 2.11

### LINKED ISSUE #1120: Failing compilations succeed in Bazel
I am trying to bazelify SpinalHDL/VexRiscV, a sbt project, with the following rule:
```
load("@io_bazel_rules_scala//scala:scala.bzl", "scala_binary", "scala_library")

scala_library(
        name = "VexRiscV",
        srcs = glob(
            ["src/main/scala/**/*.scala"],
            exclude = ["src/main/scala/demo/**/*"]
        ),
        plugins = ["@maven//:com_github_spinalhdl_spinalhdl_idsl_plugin_2_11"],
        deps = [
            "@maven//:com_github_spinalhdl_spinalhdl_core_2_11",
            "@maven//:com_github_spinalhdl_spinalhdl_sim_2_11",
            "@maven//:com_github_spinalhdl_spinalhdl_lib_2_11",
            "@maven//:com_github_spinalhdl_spinalhdl_idsl_plugin_2_11",
            "@maven//:org_yaml_snakeyaml",
            "@maven//:commons_io_commons_io",
            "@io_bazel_rules_scala_scala_compiler//:io_bazel_rules_scala_scala_compiler",
            "@io_bazel_rules_scala_scala_library//:io_bazel_rules_scala_scala_library"
            ],
)
```
Attempting to build this rule results in a lot of errors (unrelated to this issue), but **bazel reports the build as successful**:
```INFO: Analyzed target //:VexRiscV (65 packages loaded, 1292 targets configured).
INFO: Found 1 target...
INFO: From scala //:VexRiscV:
src/main/scala/spinal/lib/bus/wishbone/Wishbone.scala:53: error: Symbol 'type spinal.idslplugin.ValCallback' is missing from the classpath.
This symbol is required by 'trait spinal.core.ValCallbackRec'.
Make sure that type ValCallback is in your classpath and check for conflicting dependencies with `-Ylog-classpath`.
A full rebuild may help if 'ValCallbackRec.class' was compiled against an incompatible version of spinal.idslplugin.

...
many more errors here
...
src/main/scala/vexriscv/plugin/MmuPlugin.scala:254: error: type mismatch;
 found   : spinal.core.spinal.core.Area
 required: spinal.core.spinal.core.Area
    fenceStage plug new Area{
                    ^
Target //:VexRiscV up-to-date:
  bazel-bin/VexRiscV.jar
INFO: Elapsed time: 11.593s, Critical Path: 10.27s
INFO: 2 processes: 2 worker.
INFO: Build completed successfully, 7 total actions
```
The jar is created empty - contains just a MANIFEST.
From my basic bisecting it appears this was introduced in 2aaa7bcc104d40e.

liucijus:
fixes #1120

### DIFF
--- a/src/java/io/bazel/rulesscala/scalac/ReportableMainClass.java
+++ b/src/java/io/bazel/rulesscala/scalac/ReportableMainClass.java
@@ -4,26 +4,26 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import scala.tools.nsc.reporters.Reporter;
 import scala.tools.nsc.Settings;
 import scala.tools.nsc.reporters.ConsoleReporter;
 import scala.tools.nsc.MainClass;
 import scala.tools.nsc.Global;
 
-
-public class ReportableMainClass extends MainClass{
+public class ReportableMainClass extends MainClass {
     private Global compiler;
     private final CompileOptions ops;
 
-    public ReportableMainClass(CompileOptions ops){
+    public ReportableMainClass(CompileOptions ops) {
         this.ops = ops;
     }
 
     @Override
     public Global newCompiler() {
+        if (!ops.enableDiagnosticsReport)
+            return super.newCompiler();
+
         if (compiler == null) {
             Settings settings = super.settings();
-            ConsoleReporter reporter;
             Path path = Paths.get(ops.diagnosticsFile);
             try {
                 Files.deleteIfExists(path);
@@ -31,14 +31,10 @@ public class ReportableMainClass extends MainClass{
             } catch (IOException e) {
                 throw new RuntimeException("Could not delete/make diagnostics proto file", e);
             }
-            if (!ops.enableDiagnosticsReport) {
-                reporter = new ConsoleReporter(settings);
-            } else {
-                reporter = new ProtoReporter(settings);
-            }
+            ConsoleReporter reporter = new ProtoReporter(settings);
+
             compiler = new Global(settings, reporter);
         }
         return compiler;
     }
 }
-
--- a/src/java/io/bazel/rulesscala/scalac/ReportableMainClass.java
+++ b/src/java/io/bazel/rulesscala/scalac/ReportableMainClass.java
@@ -19,22 +19,29 @@ public class ReportableMainClass extends MainClass {
 
     @Override
     public Global newCompiler() {
-        if (!ops.enableDiagnosticsReport)
+        if (!ops.enableDiagnosticsReport) {
+            createDiagnosticsFile();
             return super.newCompiler();
+        }
 
         if (compiler == null) {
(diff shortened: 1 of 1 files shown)


src/java/io/bazel/rulesscala/scalac/ReportableMainClass.java
```diff
@@ -4,41 +4,44 @@
...
 
-    public ReportableMainClass(CompileOptions ops){
+    public ReportableMainClass(CompileOptions ops) {
         this.ops = ops;
     }
 
     @Override
     public Global newCompiler() {
+        if (!ops.enableDiagnosticsReport) {
+            createDiagnosticsFile();
```
simuons, line 23: This looks a bit confusing to me. Diagnostics file is created even when diagnostics is not enabled.
Diagnostics file is created in two branches but doesn't seem to be used. Maybe it's a wrong place to create this file?

liucijus: Diagnostics is still defined as rule output, so it expects this file to be produced - this is how original implementation works.

liucijus:
To clarify important bits regarding this change:
- this fixes broken 2.11 support - I think it's important to get this fixed ASAP
- to do it properly, diagnostics reporter should be fixed also when it is enabled for all supported versions of Scala. This means we need to have failing test scenarios for all major Scala versions for error reporting. Unfortunately I have no time to look into it myself. Any chances @andrefmrocha you can look into it?

andrefmrocha:
> to do it properly, diagnostics reporter should be fixed also when it is enabled for all supported versions of Scala. This means we need to have failing test scenarios for all major Scala versions for error reporting. Unfortunately I have no time to look into it myself. Any chances @andrefmrocha you can look into it?

Thanks for looking into this!

I unfortunately wasn't able to look into this earlier due to hardware issues that did not allow me to work on this. However, this hardware issues should be fixed by the start of next week, and I'll gladly look into this.

Just to make sure I understood, your changes guarantee that compilation fails whenever the diagnostics reporter is not used, meaning that it will still break when the diagnostics reporter is used. Is this correct?

liucijus:
> Just to make sure I understood, your changes guarantee that compilation fails whenever the diagnostics reporter is not used, meaning that it will still break when the diagnostics reporter is used. Is this correct?

Yes, I think so. I only short circuited to the old reporting mechanism when diagnostics reporter is off.

ittaiz:
I hate merging without a test but maybe we should make an exception given @andrefmrocha is committed to add tests.
What I'm still not sure is why your fix fixes anything. Do you know?

liucijus:
> I hate merging without a test but maybe we should make an exception given @andrefmrocha is committed to add tests.
> What I'm still not sure is why your fix fixes anything. Do you know?

ittaiz:
ok, does that answer the question?

PR #1160, opened 2020-12-04
Stamp scala_import jars

liucijus:
Probably very naive approach on stamping `scala_import` jars. @ittaiz WDYT?


liucijus:
Few problems:
- `java_common.stamp_jar` seems to be duplicating some Manifest attributes
- due to symlinking, all rules that depend on toolchain deps need to have `incompatible_use_toolchain_transition = True,` set to prevent host and target dependencies mixing. I'll do it with a separate PR.

third_party/dependency_analyzer/src/test/BUILD
```diff
@@ -4,8 +4,8 @@ load("//scala:scala.bzl", "scala_junit_test", "scala_test")
 
 common_jvm_flags = [
     "-Dplugin.jar.location=$(location //third_party/dependency_analyzer/src/main:dependency_analyzer)",
-    "-Dscala.library.location=$(location //external:io_bazel_rules_scala/dependency/scala/scala_library)",
-    "-Dscala.reflect.location=$(location //external:io_bazel_rules_scala/dependency/scala/scala_reflect)",
+    "-Dscala.library.location=$(rootpath //external:io_bazel_rules_scala/dependency/scala/scala_library)",
```
ittaiz, line 7: don't remember the difference between location and rootpath. Do you remember?

ittaiz (commented):
looks good in general. see one question inside.
FYI for other people watching- I was curious on why I didn't use symlink before and found out it was only introduced in 1.0 and even then it was experimental and required a bazel flag

third_party/dependency_analyzer/src/test/BUILD
liucijus: I don't think they are good for runtime information, but as we use them there's some info on them:

https://docs.bazel.build/versions/master/be/make-variables.html#predefined_label_variables

and 

https://github.com/bazelbuild/bazel/issues/2475#issuecomment-339318016

liucijus:
Status of various stamping techniques:
- `java_common.stamp_jar` calls `ijar` tool, which does incorrectly handles MANIFEST sections
- `singlejar` does not preserve existing attributes
- custom `ijar` works fine, but requires to bring in modified ijar sources

Currently the only way is to proceed with custom `ijar` until https://github.com/bazelbuild/bazel/issues/12730 is fixed

ittaiz:
I think we should advance with custom ijar.
How complicated will it be to only have custom ijar on opt-in?
I'm concerned that if some orgs are really strict on security then they might have a problem with the custom ijar and they might be blocked on upgrades.
If we can have this ability (and so also custom ijar requirement) be opt-in then maybe we have the best of both worlds.

liucijus:
> I think we should advance with custom ijar.
> How complicated will it be to only have custom ijar on opt-in?
> I'm concerned that if some orgs are really strict on security then they might have a problem with the custom ijar and they might be blocked on upgrades.
> If we can have this ability (and so also custom ijar requirement) be opt-in then maybe we have the best of both worlds.

Custom ijar source is under third_party, I don't think it's a security concern. I don't want to have opt-in, as it would be very confusing for users (buggy stamping, etc).

ittaiz:
ok

scala/scala_import.bzl
```diff
@@ -1,5 +1,40 @@
...
+
+    ctx.actions.run(
+        executable = ctx.executable._ijar,
+        inputs = [symlink_file],
+        outputs = [stamped_file],
+        arguments = [
+            "--nostrip_jar",
+            "--target_label",
+            ctx.label.name,
+            symlink_file.path,
```
simuons, line 30: Why symlink is needed? Can't you just pass `jar.path`?

third_party/java_tools/ijar/BUILD
```diff
@@ -0,0 +1,132 @@
...
+        "platform_utils.h",
+    ],
+    deps = [
+        "//third_party/java_tools/src/main/cpp/util:errors",
+        "//third_party/java_tools/src/main/cpp/util:filesystem",
+        "//third_party/java_tools/src/main/cpp/util:logging",
+    ],
+)
+#
+#cc_binary(
```
simuons, line 50: Maybe remove this?

third_party/java_tools/ijar/BUILD
```diff
@@ -0,0 +1,132 @@
...
+    ],
+    deps = [":zip"],
+)
+
+filegroup(
+    name = "srcs",
+    srcs = glob(["**"]),
+)
+
+#filegroup(
```
simuons, line 71: I'd say remove commented coded

simuons (approved):
Looks good to me.
I didn't look at ijar/zlib code.
I though there is a need only for ijar, why there is code for zlib?

scala/scala_import.bzl
liucijus: Because bazel does not allow modifications outside package

third_party/java_tools/ijar/BUILD
liucijus: I left it commented out to have it easier to update if needed. Otherwise it's a bit harder to pick what's needed to be changed.

liucijus:
> I though there is a need only for ijar, why there is code for zlib?

It's a source dependency used to process jar archives

simuons:
@liucijus thanks for your answers. LGMT

ittaiz (approved):
Thanks!

PR #1162, opened 2020-12-09
Use toolchain deps in specs2_test_discovery

liucijus:
Use dependencies from testing toolchain


ittaiz (commented):
Sorry for my ignorance but is this the PR that will break existing users? If not is there a reason we didn't do it before or we just missed it?

liucijus:
> Sorry for my ignorance but is this the PR that will break existing users? If not is there a reason we didn't do it before or we just missed it?

I missed it. It won't break users - at this point they should have toolchain in use.

ittaiz (approved):

Issue #877, opened 2019-11-15
Investigate -Youtline


johnynek:
See: https://github.com/scala/scala/pull/7712

We could potentially use outlines rather than ijar for better scala support.

Related to: https://github.com/twitter/rsc

It may be possibly useful as a tool for splitting targets as well.

ittaiz:
Definitely. I'll take a look at the related issue.
I have been in contact with the twitter people about rsc and waiting for
some loose ends there.
These possibilities are really exciting :)

PR #1165, opened 2020-12-09
POC: build pipelining with -Youtline

simuons:
Resolves to https://github.com/bazelbuild/rules_scala/issues/877

Step 0: Generate sig jars for scala_library

### DIFF
--- a/scala/private/experiments/pipeline/BUILD.bazel
+++ b/scala/private/experiments/pipeline/BUILD.bazel
@@ -1,8 +1,8 @@
 load("@rules_java//java:defs.bzl", "java_binary", "java_library", "java_test")
 
 java_library(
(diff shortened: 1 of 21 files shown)


ittaiz:
I'm super **excited** about this work! 😁
To make sure we're aligned-
The main advantages we're trying to achieve from using this are:
1. Pipelining/outlining (not sure which is the exact term)- Given B depends on A we can compile B before A's compilation finishes (we just need to wait for A's ijar to be ready).
2. Better ijar- we have suspected for a long time that the scala ijar support is basic and that ijars are invalidated to easily. Hopefully using this tool by the scala compiler people will show better results.

I think it can be very interesting if we can have stats on the above goals for this feature.

liucijus:

PR #1166, opened 2020-12-11
make_scala_test: attrs keyword argument to enable default arguments for custom scala_test rules

gergelyfabian:
### Description
Custom default attributes for make_scala_test to override rules_scala defaults.

extras already enables custom arguments but it's strongly coupled with phases.
Sometimes you may want to override some default arguments for your custom
scala_test rule created by make_scala_test.

Enabled keyword arguments on make_scala_test.
Use `attrs` to provide your own default attrs.

### Motivation
For certain attributes of your custom scala_test rules you may want to provide other defaults then foreseen by rules_scala. #1163 could be even more useful with this PR, as you could provide `jacocorunner` just once in your make_scala_test definition.

### DIFF
--- a/scala/private/rules/scala_test.bzl
+++ b/scala/private/rules/scala_test.bzl
@@ -102,10 +102,11 @@ _scala_test_attrs.update(common_attrs)
 
 _scala_test_attrs.update(_test_resolve_deps)
 
-def make_scala_test(*extras):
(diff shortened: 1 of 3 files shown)


gergelyfabian:
Closing this one as #1172 is a better solution.

PR #1168, opened 2020-12-14
Fix maven_servers parameter passing in scala_repositories

gergelyfabian:
### Description
scala_repositories should use the passed parameter value instead of providing _default_maven_server_urls as a value passed downstream.

### Motivation
scala_repositories does not currently work with custom maven_servers attribute as the implementation uses the default anyhow.
This blocks rules_scala if the default maven repositories are unavailable (e.g. in restricted company networks).


liucijus (approved):

gergelyfabian:
One of the tests failed, I don't really understand why (on timeout). Can the test be re-triggered?

gergelyfabian:
Thank you! Tests are passing now.

liucijus:
Thanks, @gergelyfabian!

PR #1172, opened 2020-12-17
Use toolchain for providing jacocorunner

gergelyfabian:
### Description
Use toolchains to provide jacocorunner for scala_test.
This is to enable overriding jacocorunner.

### Motivation
Following the discussion in #1166 and #1163, this is the latest proposal to solve #1056.

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -61,7 +61,7 @@ register_toolchains('//my/package:testing_toolchains_with_junit')
 `junit_classpath_provider` (deps_id `junit_classpath`) is where classpath required for junit tests
 is defined.
 
-### ScalaTest dependencies can be configured by decalring a provider with an id `scalatest_classpath`:
+### ScalaTest dependencies can be configured by declaring a provider with an id `scalatest_classpath`:
 
 ```starlark
 # my/package/BUILD
--- a/test/shell/test_scala_jacocorunner.sh
+++ b/test/shell/test_scala_jacocorunner.sh
@@ -1,0 +1,16 @@
+# shellcheck source=./test_runner.sh
+dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+. "${dir}"/test_runner.sh
+. "${dir}"/test_helper.sh
+runner=$(get_test_runner "${1:-local}")
+
+test_scala_jacocorunner_from_scala_toolchain_passes() {
+  bazel coverage --extra_toolchains="//manual_test/scala_test_jacocorunner:passing_scala_toolchain" //manual_test/scala_test_jacocorunner:empty_test
+}
+
+test_scala_jacocorunner_from_scala_toolchain_fails() {
+  action_should_fail coverage --extra_toolchains="//test_expect_failure/scala_test_jacocorunner:failing_scala_toolchain" //test_expect_failure/scala_test_jacocorunner:empty_test
+}
+
+$runner test_scala_jacocorunner_from_scala_toolchain_passes
+$runner test_scala_jacocorunner_from_scala_toolchain_fails
--- a/test_expect_failure/scala_test_jacocorunner/BUILD
+++ b/test_expect_failure/scala_test_jacocorunner/BUILD
@@ -1,0 +1,26 @@
+load("//scala:scala_toolchain.bzl", "scala_toolchain")
+load("//scala:scala.bzl", "scala_test")
+
+scala_toolchain(
+    name = "failing_toolchain_impl",
+    # This will fail because :failing_jacocorunner is not a valid JacocoRunner.
+    jacocorunner = ":failing_jacocorunner",
+    visibility = ["//visibility:public"],
+)
+
+toolchain(
+    name = "failing_scala_toolchain",
+    toolchain = "failing_toolchain_impl",
+    toolchain_type = "@io_bazel_rules_scala//scala:toolchain_type",
+    visibility = ["//visibility:public"],
+)
+
+scala_test(
+    name = "empty_test",
+    srcs = ["EmptyTest.scala"],
+)
+
+filegroup(
+    name = "failing_jacocorunner",
+    srcs = ["EmptyTest.scala"],
+)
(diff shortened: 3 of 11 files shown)


gergelyfabian:
@ittaiz @liucijus please let me know whether this is a good direction and what am I missing?

I tried using simply `testing_toolchain`, but it did not really work out, only moving jacocorunner to scala toolchain worked for me.

docs/testing.md
```diff
@@ -61,7 +61,7 @@ register_toolchains('//my/package:testing_toolchains_with_junit')
 `junit_classpath_provider` (deps_id `junit_classpath`) is where classpath required for junit tests
 is defined.
 
-### ScalaTest dependencies can be configured by decalring a provider with an id `scalatest_classpath`:
+### ScalaTest dependencies can be configured by declaring a provider with an id `scalatest_classpath`:
```
liucijus, line 64: :100:

testing/toolchain/toolchain.bzl
```diff
@@ -14,5 +15,8 @@ scala_testing_toolchain = rule(
             ],
             providers = [_DepsInfo],
         ),
+        "jacocorunner": attr.label(
```
liucijus, line 18: If we have `jacocorunner` on `scala_toolchain`, then it's not needed in `testing_toolchain`

liucijus (changes requested):
@gergelyfabian, yes, `jacocorunner` should be placed on `scala_toolchain`.

testing/toolchain/toolchain.bzl
gergelyfabian: Sure. Will fix.

gergelyfabian:
I'll shortly fix tests.

gergelyfabian:
@liucijus I made all necessary fixes, please re-review.

liucijus (approved):

test/shell/test_scala_jacocorunner.sh
```diff
@@ -0,0 +1,16 @@
...
+test_scala_jacocorunner_from_scala_toolchain_passes() {
+  bazel test --extra_toolchains="//manual_test/scala_test_jacocorunner:passing_scala_toolchain" //manual_test/scala_test_jacocorunner:empty_test
+}
+
+test_scala_jacocorunner_from_scala_toolchain_fails() {
+  action_should_fail test --extra_toolchains="//test_expect_failure/scala_test_jacocorunner:failing_scala_toolchain" //test_expect_failure/scala_test_jacocorunner:empty_test
+}
+
+$runner test_scala_jacocorunner_from_scala_toolchain_passes
+$runner test_scala_jacocorunner_from_scala_toolchain_fails
```
ittaiz, line 16: Small question- does this test fail only when you use the target in the toolchain? If I were to add jacocorunner to the toolchain and not use it in the actual code or maybe even better use it without actually passing it to the actions (still using the _jacocorunner) does it fail? If bazel is completely lazy it won't fail and then this test is good. Otherwise it fail fast and the test is problematic.

ittaiz (commented):
Looks good, thanks!
See question about the test

test/shell/test_scala_jacocorunner.sh
gergelyfabian: Good point. I have checked this test with 1. only adding jacocorunner option on scala_toolchain, 2. not using it anywhere.

Unfortunately Bazel fails fast as the target I've given it does not exist.

I'll change the failing tests configuration to use a valid target, so if jacocorunner is not used from the toolchain then it passes.

gergelyfabian: Indeed this is critical, as if jacocorunner from the toolchain is not used, then the positive test passes anyhow (does not detect the problem), but the negative test should also "pass" to detect that jacocorunner from the toolchain was not used.

gergelyfabian:
@ittaiz I've pushed a version that includes a fix for the test, but has the usage of the `jacocorunner` from the `scala_toolchain` reverted.
This should fail. Then I'll remove the revert of the functionality, and the test should pass. This should prove that the test works correctly.

test/shell/test_scala_jacocorunner.sh
gergelyfabian: Fixed.

test_expect_failure/scala_test_jacocorunner/BUILD
```diff
@@ -0,0 +1,26 @@
+load("//scala:scala_toolchain.bzl", "scala_toolchain")
+load("//scala:scala.bzl", "scala_test")
+
+scala_toolchain(
+    name = "failing_toolchain_impl",
+    # This will fail because :failing_jacocorunner is not a valid JacocoRunner.
+    jacocorunner = ":failing_jacocorunner",
```
gergelyfabian, line 7: This is an existing target.

test/shell/test_scala_jacocorunner.sh
gergelyfabian: Thank you once more... Your comment uncovered another issue. With the negative case fixed it came out that the positive case is also wrong. It's not testing coverage at all, as it runs `bazel test` instead of `bazel coverage`... My bad.

gergelyfabian: Fixed again :)

ittaiz:
Awesome, I'll try to take a look later today (AFK for most of the day)

gergelyfabian:
With the reverted functionality the build fails:

https://travis-ci.org/github/bazelbuild/rules_scala/jobs/750354582

```
"bazel coverage --extra_toolchains=//test_expect_failure/scala_test_jacocorunner:failing_scala_toolchain //test_expect_failure/scala_test_jacocorunner:empty_test" should have failed but passed. 

Test "test_scala_jacocorunner_from_scala_toolchain_fails" failed  (3 sec) 
```

gergelyfabian:
Removed the revert of the functionality, now the build should pass.

test/shell/test_scala_jacocorunner.sh
ittaiz: Happy to help. Testing is hard, bash testing is even harder.

ittaiz (approved):

Issue #1161, opened 2020-12-04
Fix Diagnostics reporter exit codes


andrefmrocha:
To clarify important bits regarding this change:
- this fixes broken 2.11 support - I think it's important to get this fixed ASAP
- to do it properly, diagnostics reporter should be fixed also when it is enabled for all supported versions of Scala. This means we need to have failing test scenarios for all major Scala versions for error reporting. Unfortunately I have no time to look into it myself. Any chances @andrefmrocha you can look into it?

_Originally posted by @liucijus in https://github.com/bazelbuild/rules_scala/issues/1159#issuecomment-738802911_

With #1159 , whenever the diagnostic reporter flag is turned off, it is assured that the exit code is correct. However, when the `ProtoReporter` is used, the exit code is not properly outputted. This should be addressed and several test cases for this should be created.

PR #1176, opened 2021-01-11
Reporter tests

andrefmrocha:
### Description
Added a series of tests for the different reporters used throughout different scala versions.

I have also taken the liberty to uncomment the tests for the `DiagnosticsReporter`, since the toolchain flag allows for the usage of this without a development version of bazel (I think leaving them uncommented beforehand was a mistake). Do let me know what you think of that :smile: 

### Motivation
Related to #1161 . This guarantees that different changes made to how the build is interpreted do not break error reporting.

I don't think this should close the issue, since it doesn't yet address problems in error reporting on scala 2.11 when using the `DiagnosticsReporter`.


andrefmrocha:
@googlebot I signed it!

andrefmrocha:
I've also managed to fix the error exit code on 2.11 . Is it fine for me to use this pull request to close everything?

liucijus:
please move **failing** tests under `test_expect_failure`, any other tests should go under `test`

liucijus:
> I've also managed to fix the error exit code on 2.11 . Is it fine for me to use this pull request to close everything?

If this fix does not overlap/depend on this PR, I prefer a smaller separate PR.

liucijus:
> please move **failing** tests under `test_expect_failure`, any other tests should go under `test`

Because these test are version based, maybe they can be moved to a package under version tests, and be excluded in version test case to avoid failure, WDYT @andrefmrocha?

andrefmrocha:
I've gone ahead and made the changes in question. I've decided to keep the package outside since the `version_specific_tests_dir` but still inside the `test_version` since, even though I do believe there is a flag so that a specific package is not built with commands such as `bazel build //...`, I think this would make the tests quite more complicated to read.

Do let me know what you think!

liucijus (approved):
LGTM!

ittaiz (approved):
Thanks!

PR #1178, opened 2021-01-13
Stamp scalac jars

liucijus:
ScalacWorker stamps jars with target label

### DIFF
--- a/src/java/io/bazel/rulesscala/jar/JarCreator.java
+++ b/src/java/io/bazel/rulesscala/jar/JarCreator.java
@@ -27,6 +27,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Map;
 import java.util.TreeMap;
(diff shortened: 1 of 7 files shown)


src/java/io/bazel/rulesscala/jar/JarCreator.java
```diff
@@ -265,8 +266,18 @@ public static void buildJar(String[] args) throws IOException {
       manifestFile = args[1];
       idx = 2;
     }
+
+    String targetLabel = null;
```
simuons, line 270: Probably usage message needs to be updated. Current impl requires flags in specific order ie `-m ... -t ...` maybe need to be more flexible here?

liucijus: I tried to do a minimal change here - it's a copy/paste from bazel: https://github.com/bazelbuild/bazel/blob/master/src/java_tools/buildjar/java/com/google/devtools/build/buildjar/jarhelper/JarCreator.java

I think better solution would be to research and find a way to depend on the original code and extend/override its functionality. But that's probably too big for this PR.

liucijus: Added info to usage string

ittaiz (approved):
one question- is the name “Target-Label” also what java uses?

liucijus:
> one question- is the name “Target-Label” also what java uses?

https://github.com/bazelbuild/bazel/blob/master/src/java_tools/buildjar/java/com/google/devtools/build/buildjar/jarhelper/JarHelper.java#L65

ittaiz:
👍 looks good. Resolve the conflict and merge

PR #1181, opened 2021-01-14
Fix typo `classapath` -> `classpath` in specs2

liucijus:


ittaiz (approved):
LGTM.
This is a breaking change right?
Should we document it somehow?

liucijus:
No, it's an internal change

PR #1182, opened 2021-01-15
Provide jni usage example

lgirault:
### Description

A working example of jni usage thanks to @liucijus with a small documentation referencing it

### Motivation

It is not obvious how to use jni, hence a documentation can be helpful


lgirault:
@googlebot I signed it!

examples/jni/WORKSPACE
```diff
@@ -0,0 +1,57 @@
+workspace(name = "jni_example")
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
```
ittaiz, line 3: Why should this be a separate workspace? If it were under the main workspace we would get build and test for free

ittaiz (commented):
Looks great in general.
I'm a bit concerned about how to keep this up to date since the CI doesn't build and test it.
I added a question inside

examples/jni/WORKSPACE
lgirault: I followed the pattern I saw in `examples/testing/scalatest_repositories` and `examples/testing/specs2_junit_repositories/`. I thougth that the idea was to make the examples self-contained but I agree that it comes with the question of keeping them up to date.
Maybe this can be mitigated by using something like 
```
local_repository(
     name = "io_bazel_rules_scala",
     path = "../..",
)
```
?

ittaiz: I think that those are tested by our e2es but I might be wrong. @liucijus are they tested?

ittaiz: Also those are different since their WORKSPACE file is different (toolchain).
Is this different in any way?

lgirault: I don't think so. I'll put it in the main workspace

ittaiz:
Two good news and one bad.
Good:
1. CI now checks examples folders (those which don't have their own WORKSPACE file)
2. JNI examples pass on Linux 
Bad:
1. JNI examples don't pass on OS X.
@lgirault any chance you can fix the example to be cross platform?

lgirault:
I'm sorry, I've never use OS X so I may not be the best fitted for this task :-/

ittaiz:
Understood.
@liucijus do you think you could modify this in a few minutes or we're talking hours?

liucijus:
Unfortunately I don't see where the problem is, and I'm also on Linux machine.

ittaiz:
for some reason mac build passes on my machine :(
not sure why and I've held up this simple PR for too long.
Removed the building of the example and I hope it won't grow too stale too fast.
Thanks for your efforts and for your patience!

ittaiz:
@googlebot I consent

ittaiz:
@googlebot I consent.

ittaiz:
@liucijus the test version is failing on BK but passing locally for me.
Sent https://github.com/bazelbuild/rules_scala/pull/1197 to check both it and the linter.
The linter fails there as well.
Can you take a look into it? I ran ./lint.sh fix already btw and it didn't fix the issue

ittaiz:
test version passes also on the other PR I opened.
It seems like a core crash which is a fluke. Wondering if we should run this test something like 100 times to see if it happens.

lgirault:
> Thanks for your efforts and for your patience!

Thank YOU for your persistence !
Sorry this doesn’t go more smoothly.

liucijus:
> @liucijus the test version is failing on BK but passing locally for me.
> Sent #1197 to check both it and the linter.
> The linter fails there as well.
> Can you take a look into it? I ran ./lint.sh fix already btw and it didn't fix the issue

#1197 seems to work correctly:
```
~/projects/rules_scala_prs$ ./lint.sh 
INFO: Analyzed target //tools:lint_fix (0 packages loaded, 1 target configured).
INFO: Found 1 target...
Target //tools:lint_fix up-to-date:
  bazel-bin/tools/lint_fix.bash
INFO: Elapsed time: 0.200s, Critical Path: 0.01s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action

~/projects/rules_scala_prs$ git status 
On branch ittaiz-patch-9
Your branch is up to date with 'origin/ittaiz-patch-9'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   WORKSPACE

```

ittaiz:
@liucijus can you try to run also this PR? lint check and fix pass and nothing to push

liucijus:
> @liucijus can you try to run also this PR? lint check and fix pass and nothing to push

This PR has only part of linting changes merged, please add this commit too:
`git cherry-pick c3750b0b3186ce7b1c52492632b4447f9a7f90f8`

ittaiz:
@liucijus can you please consent and then merge the PR?

ittaiz (approved):

liucijus (changes requested):
1. I would prefer code under examples with separate WORKSPACEs, I think this example should go as a test somewhere  under `//tests`
2. It looks this code example is not executed under BK. I think someone needs to bring properly working solution for Macs, otherwise it's not really clear what is the value of this example.
3. I have hard time following merges into this PR. @ittaiz which code have you added to this PR? Can you rebase, so it's easier to follow on Github.

ittaiz:
@liucijus re 2- do you think that if no one will take the mac support then we should not merge it?

liucijus:
> @liucijus re 2- do you think that if no one will take the mac support then we should not merge it?

This example incorrectly addresses architectural differences, I believe they need to be solved first before it is a valuable example.

ittaiz:
@lgirault it seems @liucijus and me are in a bit of a disagreement here and our default is to reject a conflicted change.
Are you interested in tackling the mac area and the other changes requested by @liucijus ?
If not I completely understand that and I apologize for the time it took us to get to this point

lgirault:
Hi, sorry I've never used a mac environment and I don't really have the time nor the will to learn it in addition to actually solve the problem. Actually I'm sorry to have caused all this :sweat_smile: 
Not sure what are the other points ? Moving the code under `//tests` ? that I can do, but I'm not sure it will be useful if it's not working on mac.
Thank you for all the time you both took on this matter.

ittaiz:
@lgirault i think that for Vaidas running on Mac is essential. I suggest we close this PR for now. Wdyt?

lgirault:
This is fine by me ^_^

ittaiz:
Thanks And sorry again

lgirault:
No problem :sweat_smile:

PR #1184, opened 2021-01-21
Remove specs2 binds

liucijus:
Removes binds from specs2 test rules. Users should use testing toolchain deps to customize specs2 dependencies.

### DIFF
--- a/specs2/BUILD
+++ b/specs2/BUILD
@@ -9,7 +9,6 @@ java_import(
         "//testing/toolchain:specs2_classpath",
     ],
     deps = [
-        "//scala/private/toolchain_deps:scala_library_classpath",
         "//scala/private/toolchain_deps:scala_xml",
     ],
 )
--- a/specs2/specs2.bzl
+++ b/specs2/specs2.bzl
@@ -22,10 +22,5 @@ def specs2_repositories(
         overriden_artifacts = overriden_artifacts,
     )
 
-    native.bind(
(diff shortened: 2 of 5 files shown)


ittaiz (approved):
LGTM.
Should we have a docs section about commits with breaking changes and maybe links to the code migration suggestion?

liucijus:
> LGTM.
> Should we have a docs section about commits with breaking changes and maybe links to the code migration suggestion?

Maybe then we should move to releases? Otherwise, it's hard to follow what happened when, and which version is following which.

ittaiz:
We can. I feel releases might be a bit too big of a hammer for this need but I'm open to it.
Also I'm not sure I'd like to do it here and now.
I think we should discuss this in slack and get some feedback. Or maybe just link to a github issue about it.
Concretely do you think we should merge as is without documenting this change?
Also cc @simuons

liucijus:
> We can. I feel releases might be a bit too big of a hammer for this need but I'm open to it.
> Also I'm not sure I'd like to do it here and now.
> I think we should discuss this in slack and get some feedback. Or maybe just link to a github issue about it.
> Concretely do you think we should merge as is without documenting this change?

I think this change is compatible with existing [docs](https://github.com/bazelbuild/rules_scala/blob/master/docs/testing.md), probably it's ok to merge. It should not be hard to migrate to toolchains.

PR #1185, opened 2021-01-23
Move version tests to buildkite

liucijus:
- moves version tests from Travis to Buildkite
- gives names to Buildkite tasks
- adds ubuntu2004
- add config validation


liucijus:
@or-shachar please review

or-shachar:
Looks amazing actually :) I'd follow up on this and see if we can move more and more tests to BK...

or-shachar (approved):

ittaiz (approved):

PR #1191, opened 2021-01-27
Linter script

liucijus:
- adds linting tool with reduced number of warning rules (I've disabled ones which are harder to fix quickly)
- lint starlark files

### DIFF
--- a/lint.sh
+++ b/lint.sh
@@ -2,6 +2,4 @@
 
 set -eou pipefail
 
-bazel run //tools:buildifier@fix
-
-bazel run //tools:buildifier@lint_fix
+bazel run //tools:lint_fix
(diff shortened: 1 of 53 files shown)


lint.sh
```diff
@@ -2,4 +2,6 @@
 
 set -eou pipefail
 
-./tools/bazel run //tools:buildifier@fix
+bazel run //tools:buildifier@fix
```
simuons, line 5: Is there really a need for two commands? Why not have one with lint and format?

liucijus: moved to a single command

simuons (approved):

ittaiz (approved):

PR #1193, opened 2021-01-27
scala_proto: configure ScalaPbCodeGenerator instead of worker

simuons:
### Description
Allows to configure main/default generator for the [ScalaPBWorker](https://github.com/bazelbuild/rules_scala/blob/master/src/scala/scripts/ScalaPBWorker.scala#L44)

Currently if you need to customise scalapb generator, you have to change whole persistent worker. Imo worker should be implementation detail and not exposed to users. 

### Motivation
Resolves https://github.com/bazelbuild/rules_scala/issues/1095


src/scala/scripts/ScalaPBWorker.scala
```diff
@@ -41,7 +46,7 @@ object ScalaPBWorker extends Worker.Interface {
 
     val code = ProtocBridge.runWithGenerators(
       protoc = exec(extractRequestResult.protoc),
-      namedGenerators = namedGeneratorsWithTypes ++ Seq("scala" -> ScalaPbCodeGenerator),
+      namedGenerators = namedGeneratorsWithTypes ++ Seq("scala" -> MainGenerator),
```
liucijus, line 49: Can single element be appended with `:+` here?

simuons: Done

liucijus (approved):

ittaiz (approved):

PR #1196, opened 2021-01-30
Move test_rules_scala and test_reproducibility to BK, remove Travis

liucijus:


liucijus:
@ittaiz do you know how to disable Travis? I deleted its config, but it's not enough.

ittaiz (approved):
Looks good. 
Awesome that bazel buildkite allows to run arbitrary bash scripts.
Can you please also edit the readme file to remove the travis badge?
Re travis- I removed it from being a required check. Not sure how to remove it completely.

ittaiz:
I also paused the webhook so I think that we should be ok.
I think we should edit the readme to remove the travis badge, merge this and then open a dummy PR to make sure

ittaiz (approved):

PR #1198, opened 2021-02-01
Fix action_should_fail_with_message sh helper

simuons:
### Description
Add missing `exit 1` to make test actually fail.

Had to remove `jvm_opts = ["-Xbootclasspath/p:$(location
@bazel_tools//third_party/java/jdk/langtools:javac_jar)"],`
because test fails for wrong reason `-Xbootclasspath/p is no longer a
supported option.`

### Motivation
Tried to write a test using `action_should_fail_with_message` and couldn't make it fail with a wrong error message.


liucijus (approved):

simuons:
Also fixes https://github.com/bazelbuild/rules_scala/issues/855 and duplicates https://github.com/bazelbuild/rules_scala/pull/860

ittaiz (approved):

PR #1204, opened 2021-02-03
bump bazel to v4.0.0

chenrui333:


liucijus:
Thanks for PR, @chenrui333! What is the motivation behind this update? I think it's reasonable to support 3.5.0 as long we don't have breaking changes in rules_scala between bazel versions.

PR #1205, opened 2021-02-03
Restored check with latest version

or-shachar:
### Description
In travis we used to have a non-failing test_rules_scala that checked if tests passs when working with latest bazel version

I wish to restore it.

Please kindly look only at the final changes and not by commits as I prefer to squash and merge it if it's accepted.

If latest version fails - it would add annotation like this:
image
(but won't fail the build)
Else - no annotation would be added.

I see that we're failing with `test_rules_scala.sh`, maybe we simply want to do the same thing with `bazel test //test/...` ?

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -57,8 +57,9 @@ tasks:
     shell_commands:
     - "./test_rules_scala.sh"
   test_rules_scala_linux_latest:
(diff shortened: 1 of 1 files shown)


liucijus (changes requested):
Great feature! Can the message be tuned a bit to not scare users who are not aware what this test is about? Something like:
`Non mandatory build with the latest Bazel version failed` (please improve the wording).

liucijus:
@or-shachar ping

liucijus (approved):

ittaiz (approved):

PR #1206, opened 2021-02-08
Fixing scalap version for Scala 2.11.12

gergelyfabian:
### Description

The hash and version for scalap for Scala 2.11.12 conflicted.
Set up both version and hash for scalap to be for Scala 2.11.12 (as other Scala components are set for Scala 2.11).

### Motivation

The sha256 is set for scalap:2.11.12, but the version set was 2.11.10 which
caused a hash conflict.

$ sha256sum scalap-2.11.12.jar
a6dd7203ce4af9d6185023d5dba9993eb8e80584ff4b1f6dec574a2aba4cd2b7  scalap-2.11.12.jar
$ sha256sum scalap-2.11.10.jar
3f6f352fce91c33055398a7081e35e5091fd5e095130905633e72f52124c1d27  scalap-2.11.10.jar

### DIFF
--- a/third_party/repositories/scala_2_11.bzl
+++ b/third_party/repositories/scala_2_11.bzl
@@ -110,7 +110,7 @@ artifacts = {
         "sha256": "b5f1d6071f1548d05be82f59f9039c7d37a1787bd8e3c677e31ee275af4a4621",
     },
     "org_scala_lang_scalap": {
-        "artifact": "org.scala-lang:scalap:2.11.10",
+        "artifact": "org.scala-lang:scalap:2.11.12",
         "sha256": "a6dd7203ce4af9d6185023d5dba9993eb8e80584ff4b1f6dec574a2aba4cd2b7",
         "deps": [
             "@io_bazel_rules_scala_scala_compiler",


gergelyfabian:
This is seems to be a regression in commit: d8f5bd96f8f1719bf24ebab765e5025d037349b1.

```
-            "org_scala_lang_scalap": {
-                "version": "2.11.12",
-                "sha256": "a6dd7203ce4af9d6185023d5dba9993eb8e80584ff4b1f6dec574a2aba4cd2b7",
-            },
```

```
+    "org_scala_lang_scalap": {
+        "artifact": "org.scala-lang:scalap:2.11.10",
+        "sha256": "a6dd7203ce4af9d6185023d5dba9993eb8e80584ff4b1f6dec574a2aba4cd2b7",
+        "deps": [
+            "@io_bazel_rules_scala_scala_compiler",
+        ],
+    },
```

liucijus (approved):

ittaiz (approved):
Thanks!

ittaiz:
@liucijus do you think we should add a CI run that is without any caches? That would have caught this issue

liucijus:
> @liucijus do you think we should add a CI run that is without any caches? That would have caught this issue

I guess we do not run full test suite for all versions (this is 2.11 ). Not sure if we should do that though - it's expensive CI run and is relatively easy to fix on the user side.

gergelyfabian:
> > @liucijus do you think we should add a CI run that is without any caches? That would have caught this issue
> 
> I guess we do not run full test suite for all versions (this is 2.11 ). Not sure if we should do that though - it's expensive CI run and is relatively easy to fix on the user side.

I do agree that this kind of test is very expensive. Getting all jars, to propagate the repository cache... That's kind of though :)

ittaiz:
Would we have caught it for 2.12?

gergelyfabian:
> Would we have caught it for 2.12?

I don't think so, because this was that kind of regression that you retrieved the proper file with the proper hash once (before the regression), then you change the file's url, leaving the same hash, so Bazel won't try to refetch it. It will only be found on a clean repository cache (exactly this happened in our project, this bug was found only when trying to build on clean environments).

ittaiz:
I think it's worth it to run such a clean test on 2.12 only but it's not that important to me

PR #1213, opened 2021-02-11
scala_proto_library code generation with FileDescriptorSets

simuons:
`proto_library` outputs a [FileDescriptorSet](https://developers.google.com/protocol-buffers/docs/techniques#self-description
 ) which contains all `.proto` files directly mentioned in srcs. 
Transitive descriptor sets are available through the `[ProtoInfo].transitive_descriptor_sets`

Please refer to https://docs.bazel.build/versions/master/be/protocol-buffer.html#proto_library

This change relies on outputs of `proto_library` instead of tracking/collecting transitive sources which imo simplifies code as it remove complexity of setting up `import_paths`.

Other changes made:
- calculate static values in toolchain instead doing so each time aspect is applied
- pass static data to worker once on startup via `jvm_flags`
- `ctx.actions.args` is reduced to contain data related to aspect application
- `srcjars` are produced by `protoc` instead of `JarCreator` (requires `protoc` > `2.12.0`)
- moved `protoc` to toolchain

Possible future changes:
- `scala_proto_library` should accept only `ProtoInfo` as srcs. `JavaInfo` made sense when `with_java` flag was supported, but it was removed by aspects
- remove named generators. Although code is generated by separate `protoc` plugins into separate `srcjars` at the end they are packed into single `JavaInfo`. There is no way to use individual generator outputs.


scala_proto/private/scalapb_aspect.bzl
```diff
@@ -2,255 +2,162 @@ load("@rules_proto//proto:defs.bzl", "ProtoInfo")
...
-        "_protoc": attr.label(
-            executable = True,
-            cfg = "exec",
-            default = "@com_google_protobuf//:protoc",
+        "_java_toolchain": attr.label(
+            default = Label("@bazel_tools//tools/jdk:current_java_toolchain"),
+        ),
+        "_host_javabase": attr.label(
+            default = Label("@bazel_tools//tools/jdk:current_java_runtime"),
+            cfg = "host",
```
liucijus, line 158: should be exec?

liucijus (approved):
LGTM, please check the comment regarding `cfg`

scala_proto/private/scalapb_aspect.bzl
simuons: I think not. It's same as https://github.com/bazelbuild/rules_scala/blob/5aecfd9b1b06d22ae549cd96101ec1242b7566bf/scala/private/common_attributes.bzl#L95-L98 
I assume `_host_javabase` implies `cfg=host`

simuons: Changed `cfg` to `exec`

ittaiz (approved):

PR #1214, opened 2021-02-11
Scala Import stamping

liucijus:
Another to address broken signed jars after stamping with Target-Label their manifests. 

- stamps only compile jars
- original jars are passed for runtime/full compile jars
- adds setting to disable stamping in case there are issues

### DIFF
--- a/docs/scala_import.md
+++ b/docs/scala_import.md
@@ -1,6 +1,6 @@
 # scala_import
 
-```python
+```starlark
 scala_import(
     name,
     jars,
@@ -8,7 +8,8 @@ scala_import(
     runtime_deps,
     exports,
     neverlink,
-    srcjar
+    srcjar,
+    stamp,
 )
 ```
 
@@ -17,6 +18,13 @@ like `scala_library`, similar to `java_import` from Java rules.
 
 This rule reimplements `java_import` without support for ijars, which break Scala macros.
 Generally, ijars don’t help much for external dependencies, which rarely change.
+
+The jar's compile MANIFEST.MD is stamped with a Target-Label attribute for dependency tracking 
+reporting. This behaviour can be changed per `scala_import` target with an attribute or globally 
+with a setting:
+```
+bazel build //my:target --//scala/settings:stamp_scala_import=False
+```
 
 ## Attributes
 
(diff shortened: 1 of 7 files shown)


docs/scala_import.md
```diff
@@ -18,6 +19,13 @@ like `scala_library`, similar to `java_import` from Java rules.
 This rule reimplements `java_import` without support for ijars, which break Scala macros.
 Generally, ijars don’t help much for external dependencies, which rarely change.
 
+The jar's compile MANIFEST.MD is stamped with a Target-Label attribute for dependency tracking
```
simuons, line 22: .MD -> .MF

ittaiz (approved):

test/src/main/scala/scalarules/test/scala_import/ScalaImportExposesJarsTest.scala
```diff
@@ -20,6 +20,11 @@ class ScalaImportExposesJarsTest extends SpecificationWithJUnit {
       success
     }
 
+    "stamp jars with a target label" in {
+      findManifest[Cache[String, String]] must haveTargetLabel
+      findManifest[ArrayUtils] must haveTargetLabel
+    }.pendingUntilFixed("runtime jars are not stamped")
```
simuons, line 26: Will this test be ever fixed? Maybe just remove it or change assertion and test name to state that runtime jars are not stamped?

liucijus: Ideally we want stamps in runtime jars too. Currently this test serves as a gatekeeper if compile deps leak by accident to runtime. If you have better ideas how to express this situation, I would be happy to change this test. Please note that test like "jars must not be stamped" loses meaning of compile deps vs runtime deps information. If `pendingUntilFixed` fails, it would indicate potential mistake in the code.

simuons: > Ideally we want stamps in runtime jars too. 

Why? There is no need for this and not possible for signed jars

> Currently this test serves as a gatekeeper if compile deps leak by accident to runtime. 

As I understand this test checks just that https://github.com/bazelbuild/rules_scala/pull/1214/files#diff-815ff37ec22a035579a0711eb9d9ff114a7cede2c7e4ffa73af10ae158994035R28-R43

> Please note that test like "jars must not be stamped" loses meaning of compile deps vs runtime deps information. 

Not sure what do you mean. How this meaning is preserved now? To me `pendingUntilFixed` looks like a fancy way to invert assertion.

> If pendingUntilFixed fails, it would indicate potential mistake in the code.

For me it reads like a TODO which will never be fixed. If mistake is made then test should fail assertion and not pass it.


In any case this is really a minor thing and should not block the merge. It's just hard for me to understand the intent of this test. Feel free to merge.

simuons (approved):

test/src/main/scala/scalarules/test/scala_import/ScalaImportExposesJarsTest.scala
liucijus: Removed tests

PR #1222, opened 2021-02-18
Dep analyzer: add pattern base filtering

liucijus:
Not sure if this is a feature which will land into rules_scala, but **I'm calling for opinions**. This is something that I'm experimenting on production codebase to find best way to gradually move to strict deps/unused deps.

### DIFF
--- a/scala/BUILD
+++ b/scala/BUILD
@@ -1,5 +1,5 @@
 load("@rules_java//java:defs.bzl", "java_import", "java_library")
-load("//scala:scala_toolchain.bzl", "scala_toolchain")
+load("//scala:scala_toolchain.bzl", "scala_toolchain", "dependency_tracking_patterns_config")
 load("//scala:providers.bzl", "declare_deps_provider")
 
 toolchain_type(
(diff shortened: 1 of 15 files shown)


simuons:
@liucijus if I understand correctly those `dependency-tracking-*-patterns` controls whether `dependency-analyser-plugin` should be applied on target or not.  Maybe this can be done in starlark and if target doesn't satisfy criteria plugin should not be passed to scalac. This would keep plugin simpler. wdyt?

liucijus:
> @liucijus if I understand correctly those `dependency-tracking-*-patterns` controls whether `dependency-analyser-plugin` should be applied on target or not. Maybe this can be done in starlark and if target doesn't satisfy criteria plugin should not be passed to scalac. This would keep plugin simpler. wdyt?

Yeah, that sounds like a good improvement! Thanks, @simuons

liucijus:
I have managed to migrate large repo to strict/unused error mode using filtering. A few takeaways:
1. You need separate settings for strict and unused filters. It's much easier to start with migration to strict error setting first and then do unused deps. Eg. something like this.:
```
strict_deps_checker_patterns = ["//"],
unused_deps_checker_patterns = ["//foo/bar", "//baz/zoo"],
```
The good thing is that changing patterns has scoped impact on other target caching. I think this can be useful for other folks who want to gradually apply dependency tracking on their repos. I will open a separate issue to start discussion.

2. I also tested `local_only` setting which only takes into account target dependencies which are local (does not start with "@"). It may be useful for some workflows (eg. migrate only your target deps on your repo), but the logic is heavier and is in the compiler plugin. Probably not worth adding such feature.

PR #1229, opened 2021-02-23
Add Domino Data Lab as a rules_scala adopter

urianchang:
### Description
Adds [Domino Data Lab](https://www.dominodatalab.com/) as an adopter of `rules_scala`.

### Motivation
We're currently using `rules_scala` to build and test code in our monorepo. Hopefully our adoption encourages others to also use `rules_scala`.


liucijus (approved):
Welcome!

ittaiz (approved):
Indeed!

PR #1230, opened 2021-02-25
Simplify scalac worker args format

simuons:
### Description
Avoid args concatenation in starlark. Dump all values line by line as is into args file.

### Motivation
There where three separators used to pass list values: `:` or `;`, `:::` and `,`.
Most of the concatenated values were split on java side.
Responsibilities were scattered through tree places.
Now starlark just simply writes all values line by line as is.
`CompileOptions` parses args file.
`ScalacWorker` does validation and formatting of args as required by `scalac` (previously some work was done in all three places).


scala/private/rule_impls.bzl
```diff
@@ -85,28 +82,28 @@ def compile_scala(
...
-    args.add_joined("--ResourceTargets", [p[0] for p in resource_paths], join_with = ",")
-    args.add_joined("--ResourceSources", [p[1] for p in resource_paths], join_with = ",")
-    args.add_joined("--ResourceJars", resource_jars, join_with = ",")
-    args.add_joined("--ScalacOpts", scalacopts, join_with = ":::")
-    args.add_joined("--SourceJars", all_srcjars, join_with = ",")
+    args.add_all("--Classpath", compiler_classpath_jars)
+    args.add_all("--ClasspathResourceSrcs", classpath_resources)
+    args.add_all("--Files", sources)
+    args.add_all("--Plugins", plugins)
+    args.add_all("--ResourceTargets", [p[0] for p in resource_paths])
```
liucijus, line 89: @simuons can you explain why do we need `--ResourceTargets`?

liucijus (approved):

scala/private/rule_impls.bzl
simuons: `ResourceSources` are files to be included and `ResourceTargets` are paths under which they need to be included in final jar (with removed prefixes and workspace relative parts). This is handled https://github.com/bazelbuild/rules_scala/blob/master/scala/private/resources.bzl#L2 
And yes there is that weird contract that positions need to match in distinct lists.

ittaiz (approved):

Issue #1233, opened 2021-02-27
Question on behavioral changes with scalac and setting of `--javabase`/`--host_javabase`


adam-singer:
### Context

I'm working off of a fork of `scala_rules` https://github.com/twitter-forks/rules_scala. Changes to that fork are focused around scrooge changes https://github.com/bazelbuild/rules_scala/compare/master...twitter-forks:master. When attempting to rebase to latest bazelbuild/rules_scala I think I've noticed some change in behavior around which JDK `JAVABIN` is used durning scalac actions. Would like to clarify what used to happen and what is expected to happen.

I'm noticing that the behavior of how `--javabase` and `--host_javabase` might have changed between the upgrade where `scalac` action is (latest master) using the `--host_javabase` when generating the `execroot/__main__/bazel-out/darwin-fastbuild/bin/external/io_bazel_rules_scala/src/java/io/bazel/rulesscala/scalac/scalac` script. In the previous behavior `scalac` action was using the `--javabase` (afaict). 

This was problematic for because of having two different JDK versions for `--javabase` and `--host_javabase` (ex `build --javabase=//tools/jdks:twitter_jdk8 --host_javabase=//tools/jdks:twitter_jdk11`) can lead to incompatiblities in generated bytecode. For example I'm seeing some byte code generated where interfaces have known to be broken for example https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-4774077 / https://github.com/AdoptOpenJDK/openjdk-jdk9/commit/d9d7e875470bf478110b849315b4fff55b4c35cf , is happening on the latest master where it didn't happen the past

```
1) testExtractByteBuffer(com.twitter.io.BufCompilationTest)
java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;
    at com.twitter.io.Buf$ByteBuffer$Shared$.copy(Buf.scala:953)
    at com.twitter.io.Buf$ByteBuffer$Shared$.apply(Buf.scala:957)
    at com.twitter.io.Bufs.sharedBuf(Bufs.java:72)
    at com.twitter.io.BufCompilationTest.testExtractByteBuffer(BufCompilationTest.java:126)
2) testByteBuffer(com.twitter.io.BufCompilationTest)
java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;
[...]

PR #1236, opened 2021-03-03
Reproduce issue #1233

simuons:
Identical java code passes with `--javabase=:jdk8 --host_javabase=:jdk11` while scala fails.

Pass: `bazel clean && bazel run --sandbox_debug --verbose_failures --toolchain_resolution_debug --javabase=:jdk8 --host_javabase=:jdk11 :MainJava`
Fail: `bazel clean && bazel run --sandbox_debug --verbose_failures --toolchain_resolution_debug --javabase=:jdk8 --host_javabase=:jdk11 :MainScala`


simuons:
Checked repro with commit https://github.com/bazelbuild/rules_scala/commit/f28107017c5f2a3857d983988b663ea37cd4d00f

Indeed it passes, but fails when adding `cfg="exec"` at  https://github.com/bazelbuild/rules_scala/blob/master/scala/private/common_attributes.bzl#L104 as it is in master.

liucijus:
Do we know why `cfg="exec"` is not correct configuration for `scalac`? Are we missing separate scala toolchain for tools?

simuons:
`cfg="exec"` chooses shell script of `scalac` binary which uses `host_javabase` in this case jdk11. jdk > 8 broke compatibility meaning if you want to run your code on jdk8 you have to compile it with jdk8. 

> Are we missing separate scala toolchain for tools?
Not sure what do you mean and how it would help.

scala/private/rule_impls.bzl
```diff
@@ -82,6 +83,8 @@ def compile_scala(
     args.add("--StatsfileOutput", statsfile)
     args.add("--EnableDiagnosticsReport", enable_diagnostics_report)
     args.add("--DiagnosticsFile", diagnosticsfile)
+    args.add("--Release", java_toolchain.target_version)
```
wisechengyi, line 86: would this be able to accommodate using jdk 11 to compile scala targeting 8 (so `-release 8`), then run on 11?

simuons: This is still a wip. I don't think `-release` flag is a solution (especially this is not available in scala 2.11). I'm trying `javabootclasspath` now.

However I think you could add `scalacopts = ["-release", "8"]` to your `scala_toolchain` to get unblocked (referring to https://github.com/bazelbuild/rules_scala/issues/1233)

wisechengyi: If 
> add scalacopts = ["-release", "8"]

and 
```
    args.add("--Release", java_toolchain.target_version)
```

Are they the same thing, or does one take precedence over the other?

simuons: `args.add("--Release", java_toolchain.target_version)` translates to `"-release", "x"` [here](https://github.com/bazelbuild/rules_scala/pull/1236/files#diff-ee9bbb81b4acc0673a726250f8fee1a76e466b308c5e29148fdac091d3f37657R246)

Of course if we implicitly add `-release` it might conflict with user supplied flag (and we need to handle that).
`-release 8` passes test case `bazel run --javabase=:jdk8 --host_javabase=:jdk11 :MainScala` but is not enough for all jdk and scala versions.

wisechengyi: I see. Thanks for clarifying! that works.

liucijus:
@simuons is this PR still relevant?

simuons:
I don't think it's relevant, since bazel have changed the way java runtime, source and target versions are selected.

Issue #1238, opened 2021-03-04
`scala_junit_test` not working with `bazel coverage`


virusdave:
Description of the problem / feature request:
Add coverage_runfiles phase and _lcov_merger attr to scala_junit_test

Feature requests: what underlying problem are you trying to solve with this feature?
bazel coverage on bazel 4.0 barfs on scala_junit_test target types, seemingly because it's missing this phase.

Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.
Create a scala_junit_test target
Run bazel coverage on that target

This can be seen in this repo via `bazel coverage //test:JunitTestWithDeps`

What operating system are you running Bazel on?
OSX Catalina

What's the output of bazel info release?
release 4.0.0

Extra info
I've manually modified the rules_scala sources to do this, and it suddenly starts to work. I've also used the custom phase mechanism to build a custom scala_junit_test rule which does the same thing, and this also works. It's what I'm using now internally until upstream has a similar change.

liucijus:
@virusdave, thanks for reporting! Will you be able to prepare a PR too?

liucijus:
> @virusdave, thanks for reporting! Will you be able to prepare a PR too?

Oh, I see there is #1239, thanks!

liucijus:
Solved by #1250

PR #1239, opened 2021-03-04
Add `coverage_runfiles` phase to `scala_junit_test`.

virusdave:
See title.

Previously:
`bazel coverage //test:JunitTestWithDeps` would fail.

Now it succeeds as expected.

Relates to: https://github.com/bazelbuild/rules_scala/issues/1238

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,7 +48,7 @@ Note that Skylint failures are ignored and that the fix
 command will modify your files in place.
 
 ### Additional Tests to Run
-Some changes reqiring running additional tests which are not currently
+Some changes require running additional tests which are not currently
 part of the CI pipeline.
 
 When editing code in `./third_party`, please run `./dangerous_test_thirdparty_version.sh`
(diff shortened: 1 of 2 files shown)


liucijus:
@virusdave, thanks for looking into this!

Does `bazel coverage //test:JunitTestWithDeps` produce coverage file with data for you? I get an empty file. I guess more information needs to be passed in some phases.

Also it would be great if you could add a test under `//test/coverage` package. Please note that `expected-coverage.dat` is generated by Bazel 3.5.0 (the lowest version rules_scala supports).  It's tested with ` test/shell/test_coverage.sh` script.

virusdave:
Ha - I didn't actually check if the result file contained useful output,
just that the bazel command no longer failed outright :)
Shame on me.

I'll investigate a bit more when I have time to devote to it.

   -dave-

liucijus:
Seems to duplicated effort by #1250

virusdave:
> Seems to be duplicated effort by #1250

Perfect, and that one looks more comprehensive.  Closing this PR.

Issue #1240, opened 2021-03-04
Add target filtering for strict/unused dep tracking


liucijus:
I did experiment with dep tracking filtering (#1222) and got great [results](https://github.com/bazelbuild/rules_scala/pull/1222#issuecomment-790428620). I think adding filtering would benefit other `rules_scala` users.

I'm thinking of adding the following attributes to the scala toolchain:
```starlark
"dependency_tracking_strict_deps_patterns": attr.string_list(
        doc = "List of target prefixes included for strict deps analysis",
        default = [""],
),
"dependency_tracking_unused_deps_patterns": attr.string_list(
        doc = "List of target prefixes included for unused deps analysis",
        default = [""],
),
```
Example configuration:
```starlark
dependency_tracking_strict_deps_patterns: [
    "@foo",      # track all targets which start with "@foo"
    "//",        # track every local target starting with "//"
    "-//:bar",   # but ignore "//:bar"
],
dependency_tracking_unused_deps_patterns: [
    "//foo/bar", # track only targets starting with "//foo/bar"
],
```

default configuration would be `[""]`, which matches every target.

Tracking decision would be made in starlark and is relatively cheap (check if target label `startswith` over the list items) and will cache well.

Calling for all types of input (naming, potential problems, design improvements, etc).

liucijus:
Addresses #1046

liucijus:
Solved with #1241

PR #1241, opened 2021-03-05
Dep tracking filtering

liucijus:
Implements #1240


scala/scala_toolchain.bzl
```diff
@@ -48,13 +48,43 @@ def _scala_toolchain_impl(ctx):
 
     enable_diagnostics_report = ctx.attr.enable_diagnostics_report
 
+    all_strict_deps_patterns = ctx.attr.dependency_tracking_strict_deps_patterns
+
+    strict_deps_include_patterns = [
```
simuons, line 53: Maybe extract a function to partition patterns into include/exclude lists

scala/private/phases/phase_dependency.bzl
```diff
@@ -64,3 +78,9 @@ def _get_unused_deps_mode(ctx):
         return ctx.attr.unused_dependency_checker_mode
     else:
         return ctx.toolchains["@io_bazel_rules_scala//scala:toolchain_type"].unused_dependency_checker_mode
+
+def _is_target_included(target, includes, excludes):
+    if len([exclude for exclude in excludes if target.startswith(exclude)]) > 0:
```
simuons, line 83: Maybe instead of building list and checking length you could use loop with an if? For example
```
for exclude in excludes:
  if target.startswith(exclude):
    return False
```

Or maybe `if any([target.startswith(exclude) for exclude in excludes])` but I guess you would building new list each time however it might not be an issue.

simuons (approved):
LGTM. Just couple minor comments/suggestions.

scala/private/phases/phase_dependency.bzl
liucijus: I'm leaning towards simple loop

Issue #1234, opened 2021-03-01
scala_proto_library gets incorrectly stamped with proto_library label


liucijus:
`scala_proto_library` gets incorrectly stamped with `proto_library` label. With strict deps it becomes an invalid `buildozer` command:

```
You can use the following buildozer command:
buildozer 'add deps @foo//proto:proto' //bar:bar
```

Given such targets:
```
# proto/BUILD
proto_library(
    name = "proto",
    srcs = ["foo.proto"],
)

scala_proto_library(
    name = "scala_proto",
    deps = [":proto"],
)
```

Buildozer command should suggest `@foo//proto:scala_proto`

liucijus:
The same issue with `java_proto_library` https://github.com/bazelbuild/bazel/issues/4990

liucijus:
The root cause of incorrect stamping is that proto compilation happens in the aspect context on proto deps, where `scala_proto_library` label is not available (which makes sense, as there may be multiple `scala_proto_library` depending on the same `proto_library` target). 
Possible ways (though none of them are without significant compromises) to mitigate stamping problem:
- move compilation from aspects to `scala_proto_library` - this means every `scala_proto_library` will be fully generating and  rebuilding all proto dependecies.
- stamp according `injecting_rule_kind` and generate labels according to [convention (described under "code organization")](https://docs.bazel.build/versions/master/be/protocol-buffer.html#proto_library). This means that aspect can rewrite their label from `foo_proto` to `foo_scala_proto` under the same package. This idea can be stretched even further by refactoring into phases and adding a phase to allow users to have their own labeling convention.

I'm leaning towards conventional labeling option. @simuons, @ittaiz WDYT?

ittaiz:
Conventional means the first option?
Because that sounds too problematic IMHO performance wise

liucijus:
> Conventional means the first option?
> Because that sounds too problematic IMHO performance wise

No, the second one - which would assume users follow recommended code organization described in the Bazel docs

ittaiz:
SGTM

simuons:
I'd vote for second option (conventional labeling option)

liucijus:
Solved with #1243

PR #1243, opened 2021-03-09
Scala proto conventional stamping

liucijus:
- refactored `scala_proto_library` rule/aspect into phases
- added label stamping phase with a toolchain attribute to control conventional stamping

implementation of the second option in https://github.com/bazelbuild/rules_scala/issues/1234#issuecomment-791628830


scala/private/phases/api.bzl
```diff
@@ -42,7 +42,7 @@ def _adjust_phases(phases, adjustments):
     return phases
 
 # Execute phases
-def run_phases(ctx, builtin_customizable_phases):
+def run_phases(ctx, builtin_customizable_phases, target = None):
```
simuons, line 45: Maybe add a comment that `target` is given only in aspects context

scala_proto/private/scala_proto_aspect.bzl
```diff
@@ -122,27 +133,61 @@ def _compile_sources(ctx, toolchain, proto, src_jars, deps):
         runtime_deps = compile_deps,
     )
 
-####
-# This is applied to the DAG of proto_librarys reachable from a deps
-# or a scalapb_scala_library. Each proto_library will be one scalapb
-# invocation assuming it has some sources.
-def _scalapb_aspect_impl(target, ctx):
+def _phase_proto_provider(ctx, p):
+    target = p.target
```
simuons, line 137: Please inline this

scala_proto/private/scala_proto_aspect.bzl
```diff
@@ -122,27 +133,61 @@ def _compile_sources(ctx, toolchain, proto, src_jars, deps):
...
+    return target[ProtoInfo]
+
+def _phase_deps(ctx, p):
+    return [d[ScalaProtoAspectInfo].java_info for d in ctx.rule.attr.deps]
+
+def _phase_generate_and_compile(ctx, p):
```
simuons, line 143: Would it be possible to split this phase into two without `and`?

scala_proto/private/scala_proto_aspect.bzl
```diff
@@ -111,6 +121,7 @@ def _compile_sources(ctx, toolchain, proto, src_jars, deps):
         scalac = toolchain.scalac,
         dependency_info = legacy_unclear_dependency_info_for_protobuf_scrooge(ctx),
         unused_dependency_checker_ignored_targets = [],
+        stamp_target_label = stamp_label,
```
simuons, line 124: Why not to pass `stamp_label` instead of `ctx.label`? I think in this context `ctx.label` is effectively unused because dependency analyser is disabled.

test_expect_failure/missing_direct_deps/scala_proto_deps/customized_scala_proto.bzl
```diff
@@ -0,0 +1,34 @@
+load("//scala_proto:scala_proto.bzl", "make_scala_proto_aspect", "make_scala_proto_library")
+load("@io_bazel_rules_scala//scala:advanced_usage/providers.bzl", "ScalaRulePhase")
+
+def _phase_custom_stamping_convention(ctx, p):
+    rule_label = str(p.target.label)
+    toolchain = ctx.toolchains["@io_bazel_rules_scala//scala_proto:toolchain_type"]
+
+    if toolchain.stamp_by_convention:
+        return rule_label + "_scala"
```
simuons, line 9: Maybe you could use suffix that stands out better. It was hard to spot difference between `some_scala_proto` and `some_proto_scala`

simuons (changes requested):
Please find my comments in PR.

I have couple other questions:
* how granular phases should be. For example is `_phase_proto_provider` really needed?
* Are stamp phase implementations required to respect `stamp_by_convention` flag on toolchain? It could lead to confusion if `stamp_by_convention = True` and phase was removed.

scala/private/phases/api.bzl
liucijus: I'll do this, I think it's more clear:
```
def run_phases(ctx, builtin_customizable_phases):
    return _run_phases(ctx, builtin_customizable_phases, target = None)

def run_aspect_phases(ctx, builtin_customizable_phases, target):
    return _run_phases(ctx, builtin_customizable_phases, target)

def _run_phases(ctx, builtin_customizable_phases, target):
    ...
```

scala_proto/private/scala_proto_aspect.bzl
liucijus: It's possible, I didn't not do it to avoid multiple checks accross phases for this conditional decision:
```starlark
if proto.direct_sources and _code_should_be_generated(ctx, toolchain):
    src_jars = _generate_sources(ctx, toolchain, proto)
    java_info = _compile_sources(ctx, toolchain, proto, src_jars, deps, stamp_label)
    return java_info
else:
    # this target is only an aggregation target
    return java_common.merge(deps)
```
Do you think it's worth splitting it?

scala_proto/private/scala_proto_aspect.bzl
liucijus: `ctx.label` is unused, but I think `stamp_label` and `ctx.label` have different semantics. I can't think of very strong case against having just one, but let's say we want to log actual target while compiling scala. Maybe parameter `target_label` should be removed if it is not used and can be taken from ctx.

liucijus:
>     * how granular phases should be. For example is `_phase_proto_provider` really needed?
I introduced this phase as I was thinking to split generation and compilation, but then left both in one phase. I think it makes sense to have it as a separate responsibility, but no code needs it at the moment.

>     * Are stamp phase implementations required to respect `stamp_by_convention` flag on toolchain? It could lead to confusion if `stamp_by_convention = True` and phase was removed.

It would be great if phases respected this attribute, but I can't think of a way to enforce it

scala_proto/private/scala_proto_aspect.bzl
simuons: It's not a blocker for sure. It just feels unbalanced. Some phases are fine grained and this one is coarse grained doing multiple things. Maybe phases are missing some "short-circuit" functionality. So let's keep it that way if it's simpler.

simuons (approved):

ittaiz (approved):

scala_proto/private/scala_proto_aspect.bzl
```diff
@@ -122,27 +133,60 @@ def _compile_sources(ctx, toolchain, proto, src_jars, deps):
...
+def _phase_stamp_label(ctx, p):
+    rule_label = str(p.target.label)
+    toolchain = ctx.toolchains["@io_bazel_rules_scala//scala_proto:toolchain_type"]
 
-scalapb_aspect = aspect(
-    implementation = _scalapb_aspect_impl,
-    attr_aspects = ["deps"],
-    incompatible_use_toolchain_transition = True,
+    if toolchain.stamp_by_convention and rule_label.endswith("_proto"):
+        return rule_label.rstrip("_proto") + "_scala_proto"
```
darl, line 168: @liucijus 
Does starlark's `rstrip` differ from python's one?
In python it strips all characters from argument in any order:
```
>>> "otorp_proto".rstrip("_proto")
''
```

liucijus: @darl thanks for pointing this out - yes this is incorrect implementation

PR #1247, opened 2021-03-15
Stop setting BAZEL_VERSION

illicitonion:
### Description
Stop setting BAZEL_VERSION

### Motivation
It appears to be no longer used, but may pollute cache keys.


liucijus (approved):
Thanks, @illicitonion!

ittaiz (approved):

PR #1248, opened 2021-03-16
Add multi framework example

liucijus:


examples/testing/multi_frameworks_toolchain/BUILD
```diff
@@ -0,0 +1,61 @@
+load("@io_bazel_rules_scala//scala:providers.bzl", "declare_deps_provider")
+load("@io_bazel_rules_scala//testing/toolchain:toolchain.bzl", "scala_testing_toolchain")
+
+scala_testing_toolchain(
+    name = "testing_toolchains_with_all_deps_impl",
+    dep_providers = [
+        ":junit_classpath_provider",
```
simuons, line 7: Maybe reuse existing deps providers from rules_scala? Example would be smaller and focused.

simuons (changes requested):
Could you add a link to this example in docs/testing.md?

examples/testing/multi_frameworks_toolchain/BUILD
liucijus: I'm actually thinking how to make providers private to rules_scala - they aren't meant for external reuse. Ideally this example should come with different loader like `rules_jvm_external` with different external names to fully demonstrate custom toolchain setup.

simuons (approved):

ittaiz (approved):

PR #1249, opened 2021-03-16
Example with multiple testing frameworks

simuons:
### Description

Add example with multiple testing frameworks in same workspace.

TODO: update testing.md

### Motivation

From current docs it's not clear how to achieve that as we are getting same question multiple times.


simuons:
Duplicates https://github.com/bazelbuild/rules_scala/pull/1248

Issue #1257, opened 2021-03-29
Bad classpath elements: scala-library.jar and scala-reflect.jar


samolisov:
When I build a Scala project on a Windows 10 machine, I get the following warnings:

```sh
INFO: From Building external/io_bazel_rules_scala/src/java/io/bazel/rulesscala/scalac/scalac.jar (5 source files):      
warning: [path] bad path element "C:\users\...\bwtymqfd\execroot\scala_example\external\io_bazel_rules_scala_scala_compiler\scala-reflect.jar": no such file or directory
warning: [path] bad path element "C:\users\...\bwtymqfd\execroot\scala_example\external\io_bazel_rules_scala_scala_compiler\scala-library.jar": no such file or directory 
```

The directory `io_bazel_rules_scala_scala_compiler` contains only the following jar: `scala-compiler-2.12.11.jar` while the files `scala-library-2.12.11.jar` and `scala-reflect-2.12.11.jar` are in the `external\io_bazel_rules_scala_scala_library` and `external\io_bazel_rules_scala_scala_reflect` directories, respectively.

liucijus:
@samolisov could you check if warnings disappear if you pass `--@io_bazel_rules_scala//scala/settings:stamp_scala_import=False` setting?

samolisov:
@liucijus I tried the following command line:

```sh
bazel build --@io_bazel_rules_scala//scala/settings:stamp_scala_import=False --jobs 4 example-target
```

But get the following error:

```
ERROR: Skipping '@io_bazel_rules_scala//scala/settings:stamp_scala_import': no such package '@io_bazel_rules_scala//scala/settings': BUILD file not found in directory 'scala/settings' of external repository @io_bazel_rules_scala. Add a BUILD file to a directory to mark it as a package
```

Maybe I'm doing something wrong?

liucijus:
@samolisov would you be able to make a small git repository for reproduction?

samolisov:
@liucijus Please have a look: https://github.com/samolisov/bazel-scala-hello

```sh
bazel run --jobs 4 hello
```

simuons:
Looks similar to https://github.com/bazelbuild/bazel/issues/12968

PR #1258, opened 2021-03-30
Remove bad path element warnings in bazel 4

simuons:
### Description

Remove `[path] bad path element` warnings during build in bazel 4.

Bazel 4.0.0 sets `-Xlint:path` for `javac` invocations, which validates `Class-Path` entries and emits warnings.

This PR takes same approach as [rules_jvm_external](https://github.com/bazelbuild/rules_jvm_external/pull/518) ie remove `Class-Path` entry from imported jar's manifest.

Reproduce with:

`USE_BAZEL_VERSION=4.0.0 bash -c 'bazel build //src/java/io/bazel/rulesscala/scalac:scalac 2>&1 | grep "bad path element"'`

Related issues:
* https://github.com/bazelbuild/bazel/issues/12968
* https://github.com/bazelbuild/rules_jvm_external/pull/518

### Motivation

Resolves https://github.com/bazelbuild/rules_scala/issues/1257

### DIFF
--- a/third_party/java_tools/ijar/BUILD
+++ b/third_party/java_tools/ijar/BUILD
@@ -146,31 +146,30 @@ EOF""",
 
 genrule(
     name = "test-input-jar",
-    outs = ["input.jar"],
-    cmd = "$(JAVABASE)/bin/jar -cfm $@ $(location :test-input-manifest)",
     srcs = [":test-input-manifest"],
+    outs = ["input.jar"],
+    cmd = "$(JAVABASE)/bin/jar -cfm $@ $<",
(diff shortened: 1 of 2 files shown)


simuons:
This ijar fork is a temporary solution which must go away in favour of library from bazel. Therefore investing more effort in this code is hardly justified. 
Future steps: discuss this issue with bazel and probably fix ijar to remove `Class-Path` from manifest in `--nostrip_jar` branch.
Alternative solution: replace ijar with own tool which manipulates manifest (this is the only feature we need from ijar as we do not strip classes for scala).
cc @liucijus

PR #1264, opened 2021-05-04
Remove tut_rule

simuons:
### Description

Remove `tut_rule` with a proposed migration https://github.com/bazelbuild/rules_scala/releases/tag/tut

### Motivation

According to https://github.com/tpolecat/tut#tut tut is not maintained and binaries are not available anymore as of May 1, 2021

Builds are failing with
```
Error downloading [https://dl.bintray.com/tpolecat/maven/org/tpolecat/tut-core_2.12/0.6.13/tut-core_2.12-0.6.13.jar]
GET returned 403 Forbidden
```


liucijus:
I would vote for complete removal of tut from Rules Scala. It does not even work Scala 2.13. WDYT, @simuons?

simuons:
@liucijus I'm definitely for removal. I just was worried that people might actually use it. How about now?

liucijus (approved):

PR #1266, opened 2021-05-10
Removes scala_version

Marcus-Rosti:
### Description
scala_proto_repositories does not accept scala_version

### Motivation

Clean up docs


liucijus (approved):
Thanks, @Marcus-Rosti!

PR #1275, opened 2021-07-16
Extra jmh options: testonly, runtime_deps

donce:
### Description

Adding 2 options to JMH rules:
- `testonly` to allow marking JMH as `testonly` and consuming `testonly` dependencies, which allows re-using various test-kits and test helpers.
- `runtime_deps`, which might (and is in our case) required for benchmark dependencies

Both are optional, so should be fully backwards-compatible.

### Motivation

We have many test-kits and test helpers marked with `testonly`, because we're using them for testing mainly and `testonly` protects us from getting them into production. They are very useful for benchmarks as well, which is why we need this flag.

### DIFF
--- a/jmh/jmh.bzl
+++ b/jmh/jmh.bzl
@@ -81,6 +81,7 @@ def scala_benchmark_jmh(**kw):
     data = kw.get("data", [])
     generator_type = kw.get("generator_type", "reflection")
     lib = "%s_generator" % name
+    testonly = kw.get("testonly", False)
     scalacopts = kw.get("scalacopts", [])
     main_class = kw.get("main_class", "org.openjdk.jmh.Main")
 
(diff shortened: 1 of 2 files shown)


jmh/jmh.bzl
```diff
@@ -77,10 +77,12 @@ scala_generate_benchmark = rule(
 def scala_benchmark_jmh(**kw):
     name = kw["name"]
     deps = kw.get("deps", [])
+    runtime_deps = kw.get("runtime_deps", [])
```
liucijus, line 80: @donce I want to understand what are usecases when runtime deps are needed here. Do you have a real world example?

liucijus (changes requested):
Thanks, @donce! Please add targets to test `test_only` under `test/jmh`. Other than that LGTM.

jmh/jmh.bzl
donce: We're benchmarking logging, which requires a custom logging config. The custom logging config is a static `xml` file with `resources` target, which is added to the benchmark.

liucijus (approved):

donce:
I don't have permissions to merge the PR - is there anything else needed from me, or should I just wait?

liucijus:
We need @simuons approval

simuons (approved):

Issue #1277, opened 2021-07-27
Custom JacocoRunner scripts fail


BillyAutrey:
If you run these scripts as is, you will get a few problems.

1. The bazel master branch is now much further ahead.  This causes the patch to be irrelevant, because the existing jacoco 0.8.6 jars in the bazel target directory are the ones used to build the JAR.
2. The `coverage_local_jacocorunner/test.sh` script assumes that tests are in the `coverage` directory.  The code that is tested has since moved.
3. The `expected-coverage.dat` file in the same folder needs to be updated for the new test directory, as well.

Fixes:

1. Change the bazel portion of the compilation script to check out the `4.1.0` tag instead of `master`.
2. Change the `test_coverage_with_local_jacocorunner()` function in `coverage_local_jacocorunner/test.sh`  to use `//test/coverage_scalatest:test-scalatest` instead.
3. Change all references to `coverage` in `expected-coverage.dat` to `coverage_scalatest`.

BillyAutrey:
I have fixes for this, PR incoming.

liucijus:
fixed in #1278

PR #1278, opened 2021-07-27
Fix build jacoco scripts

BillyAutrey:
### Description
Fixes for #1277.  Updated jacoco build script and test script to use specific tags and new test locations.  

### Motivation
I attempted to run these scripts, and ran into several issues.  After a lot of troubleshooting and debugging, I figured out the root causes and a few potential fixes.  Decided to submit a PR with these fixes, in case it helps anyone else attempting the same thing.

### DIFF
--- a/manual_test/coverage_local_jacocorunner/expected-coverage.dat
+++ b/manual_test/coverage_local_jacocorunner/expected-coverage.dat
@@ -1,12 +1,12 @@
-SF:test/coverage/A1.scala
-FN:-1,coverage/A1$::<clinit> ()V
-FN:3,coverage/A1$::<init> ()V
-FN:5,coverage/A1$::a1 (Z)Lcoverage/B1$;
-FN:-1,coverage/A1::a1 (Z)Lcoverage/B1$;
-FNDA:1,coverage/A1$::<clinit> ()V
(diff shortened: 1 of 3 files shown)


BillyAutrey:
@googlebot I signed it!

liucijus:
Thanks @BillyAutrey! 

@gergelyfabian your review would be very valuable here

gergelyfabian:
It looks all good to me (assuming it also works, I did not have a chance to
check that). This is a beautiful example of why open source is nice, you
once contribute and then someone else fixes your code (when it stops
working due to other changes).

Great thanks @BillyAutrey <https://github.com/BillyAutrey>!

Vaidas Pilkauskas ***@***.***> ezt írta (időpont: 2021. júl.
28., Sze 13:54):

BillyAutrey:
I ran all of the scripts, and they worked for me locally.  The test script works as well.  I do invite others to dry run these scripts, however.  I'd be happy to correct them if any of my assumptions were wrong.

liucijus (approved):

liucijus:
I've tried the test script - it finished successfully

simuons (approved):

PR #1281, opened 2021-08-20
Fix missing directories in JaCoCo instrumented jars

### LINKED ISSUE #1280: Coverage runner fails to get resources from ClassLoader
# Cause
JaCoCoInstrumenter does not build instrumented jars with paths in the offline jar.  For instance, if I have a `resource.jar` with files in `com/example/resourcejar/` in my `src/main/scala` and `src/main/resources` directories, I get the following discrepancies.

**resource.jar**
```
META-INF/
META-INF/MANIFEST.MF
com/
com/example/
com/example/resourcejar/
com/example/resourcejar/Util$.class
com/example/resourcejar/Util.class
com/example/resourcejar/example.sql
```

**resource-offline.jar**
```
META-INF/
META-INF/MANIFEST.MF
-paths-for-coverage.txt
com/example/resourcejar/Util$.class
com/example/resourcejar/Util$.class.uninstrumented
com/example/resourcejar/Util.class
com/example/resourcejar/Util.class.uninstrumented
com/example/resourcejar/example.sql
```

# Effect
Suppose I have a `Consumer` class that depends on `resource.jar`.  If I have code that does the following:
```scala
import java.lang.Thread
import java.net.URL
import java.util.Enumeration

class Consumer {
  val classLoader = Thread.currentThread.getContextClassLoader
  val resourceList: Enumeration[URL] = classLoader.getResources("com/example/resourcejar")
}
```

and a test that looks like this:

```scala
import org.scalatest._

import java.net.URL

class ConsumerSpec extends FlatSpec with Matchers {

    "ConsumerSpec" should "have valid classloaders" in {
      val fixture = new Consumer()
      fixture.resourceList.hasMoreElements should be (true)
      val element = fixture.resourceList.nextElement
      element shouldBe a [URL]
    }

}
```

`bazel test //...` will succeed, and the test will pass.  But `bazel coverage //...` will fail.  `fixture.resourceList` will be an empty `Enumeration`.

BillyAutrey:
### Description
Added directories to the `packagename-offline.jar` artifacts that get created by [JaCoCoInstrumenter.java](src/java/io/bazel/rulesscala/coverage/instrumenter/JacocoInstrumenter.java).

### Motivation
Fixes #1280.  This fix allows calls for `ClassLoader.getResources("path/in/jar")` to succeed.  Any code that tries to fetch resources at a given path will get no results in `bazel coverage`, even if `bazel test` allows the path to be explored properly.

### DIFF
--- a/test/coverage_scalatest_resources/consumer/BUILD.bazel
+++ b/test/coverage_scalatest_resources/consumer/BUILD.bazel
@@ -2,15 +2,15 @@ load("//scala:scala.bzl", "scala_library", "scala_test")
 
 scala_library(
     name = "consumer",
+    srcs = ["src/main/scala/com/example/consumer/Consumer.scala"],
     deps = ["//test/coverage_scalatest_resources/resource"],
-    srcs = ["src/main/scala/com/example/consumer/Consumer.scala"]
 )
 
 scala_test(
     name = "tests",
-    deps = ["//test/coverage_scalatest_resources/resource"],
     srcs = [
         "src/main/scala/com/example/consumer/Consumer.scala",
         "src/test/scala/com/example/consumer/ConsumerSpec.scala",
-        ]
+    ],
+    deps = ["//test/coverage_scalatest_resources/resource"],
 )
--- a/test/coverage_scalatest_resources/consumer/BUILD.bazel
+++ b/test/coverage_scalatest_resources/consumer/BUILD.bazel
@@ -1,0 +1,14 @@
+load("@rules_java//java:defs.bzl", "java_library")
+load("//scala:scala.bzl", "scala_library", "scala_test")
+
+scala_library(
+    name = "consumer",
+    deps = ["//test/coverage_scalatest_resources/resource"],
+    srcs = ["src/main/scala/com/example/consumer/Consumer.scala"]
+)
+
+scala_test(
+    name = "tests",
+    deps = ["//test/coverage_scalatest_resources/resource"],
+    srcs = glob(include = ["{}/**".format("src/main/scala")]) + glob(include = ["{}/**".format("src/test/scala")])
+)
--- a/test/coverage_scalatest_resources/consumer/BUILD.bazel
+++ b/test/coverage_scalatest_resources/consumer/BUILD.bazel
@@ -10,5 +10,8 @@ scala_library(
 scala_test(
     name = "tests",
     deps = ["//test/coverage_scalatest_resources/resource"],
-    srcs = glob(include = ["{}/**".format("src/main/scala")]) + glob(include = ["{}/**".format("src/test/scala")])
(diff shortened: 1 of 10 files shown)


BillyAutrey:
@gergelyfabian Since you contributed here, you might double check this as well.

liucijus:
@BillyAutrey thanks for PR, I'm currently on vacation, I will look into the changes at the end of next week if you can wait.

BillyAutrey:
> @BillyAutrey thanks for PR, I'm currently on vacation, I will look into the changes at the end of next week if you can wait.

@liucijus that's fine.  I was hoping to use this fix to resolve a few things on another project, but I can wait until you're back.  Hope you're having a good summer holiday!  Stay safe.  :)

test/coverage_scalatest_resources/consumer/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
+load("//scala:scala.bzl", "scala_library", "scala_test")
+
+scala_library(
+    name = "consumer",
+    srcs = ["src/main/scala/com/example/consumer/Consumer.scala"],
+    deps = ["//test/coverage_scalatest_resources/resource"],
+)
+
+scala_test(
```
liucijus, line 9: Can test depend on lib instead of sources? eg.:

```
scala_test(
    name = "tests",
    srcs = [
        "src/test/scala/com/example/consumer/ConsumerSpec.scala",
    ],
    deps = [":consumer"],
)
```

test/shell/test_coverage_scalatest_resources.sh
```diff
@@ -0,0 +1,20 @@
...
+runner=$(get_test_runner "${1:-local}")
+
+test_coverage_succeeds_resource_call() {
+    bazel coverage //test/coverage_scalatest_resources/consumer:tests
+    diff test/coverage_scalatest_resources/expected-coverage.dat $(bazel info bazel-testlogs)/test/coverage_scalatest_resources/consumer/tests/coverage.dat
+}
+
+test_coverage_includes_resource_test_targets() {
+    bazel coverage \
+          --instrument_test_targets=True \
```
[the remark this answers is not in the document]
liucijus: > Is `--instrument_test_targets=True` needed?

Ignore this comment

test/coverage_scalatest_resources/consumer/BUILD.bazel
BillyAutrey: Indeed, it does work!  It alters the diff a bit, but in a good way.  Pushing this change now.

simuons (approved):

liucijus (approved):

liucijus:
Thanks, @BillyAutrey!

PR #1285, opened 2021-09-06
Remove customized ijar code

liucijus:
ijar patch was upstreamed and is available since Bazel 4.1.0 (https://github.com/bazelbuild/bazel/commit/ba60c0b3f9bbd00975c984244839b155e84b4c5d)


simuons (approved):
🥇

PR #1289, opened 2021-09-17
Add directory scanning support to JUnit test discovery

liucijus:
I'm trying to make Intellij's [Fast Build](https://ij.bazel.build/docs/fast-builds.html) work with Specs2 tests, but currently JUnit test discovery only knows how to scan jars and fails id directories are passed. For fast build to work we need to prepend directories to the classpath. 
This PR adds support for listing class files from directories.

### DIFF
--- a/src/java/io/bazel/rulesscala/test_discovery/ArchiveEntries.scala
+++ b/src/java/io/bazel/rulesscala/test_discovery/ArchiveEntries.scala
@@ -24,7 +24,7 @@ object ArchiveEntries {
 
   private def jarEntries(jarInputStream: JarInputStream): Stream[String] =
     Stream.continually(getJarEntryOrCloseStream(jarInputStream))
-      .takeWhile(_.isDefined)
+      .takeWhile(_.isEmpty)
       .flatten
       .map(_.getName)
 
--- a/src/java/io/bazel/rulesscala/test_discovery/ArchiveEntries.scala
+++ b/src/java/io/bazel/rulesscala/test_discovery/ArchiveEntries.scala
@@ -6,7 +6,7 @@ import java.util.jar.{JarEntry, JarInputStream}
 object ArchiveEntries {
   def listClassFiles(file: File): Stream[String] = {
     val allEntries = if (file.isDirectory)
-      directoryEntries(file).map(_.stripPrefix(file.toString).stripPrefix("/"))
+      directoryEntries(file).map(_.stripPrefix(file.getPath).stripPrefix("/"))
     else
       jarEntries(new JarInputStream(new FileInputStream(file)))
 
--- a/src/java/io/bazel/rulesscala/test_discovery/ArchiveEntries.scala
+++ b/src/java/io/bazel/rulesscala/test_discovery/ArchiveEntries.scala
@@ -1,0 +1,36 @@
+package io.bazel.rulesscala.test_discovery
+
+import java.io.{File, FileInputStream}
+import java.util.jar.{JarEntry, JarInputStream}
+
+object ArchiveEntries {
+  def listClassFiles(file: File): Stream[String] = {
+    val allEntries = if (file.isDirectory)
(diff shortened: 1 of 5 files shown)


src/java/io/bazel/rulesscala/test_discovery/ArchiveEntries.scala
```diff
@@ -0,0 +1,36 @@
+package io.bazel.rulesscala.test_discovery
+
+import java.io.{File, FileInputStream}
+import java.util.jar.{JarEntry, JarInputStream}
+
+object ArchiveEntries {
+  def listClassFiles(file: File): Stream[String] = {
+    val allEntries = if (file.isDirectory)
+      directoryEntries(file).map(_.stripPrefix(file.toString).stripPrefix("/"))
```
simuons, line 9: Maybe instead of `file.toString` use `file.getPath`

src/java/io/bazel/rulesscala/test_discovery/ArchiveEntries.scala
```diff
@@ -0,0 +1,36 @@
...
+    val entry = Option(jarInputStream.getNextJarEntry)
+
+    if (entry.isEmpty)
+      jarInputStream.close()
+
+    entry
+  }
+
+  private def jarEntries(jarInputStream: JarInputStream): Stream[String] =
+    Stream.continually(getJarEntryOrCloseStream(jarInputStream))
```
simuons, line 26: Maybe instead of `getJarEntryOrCloseStream` wrap whole stream in `try` and close stream in `finally`?

test/src/main/scala/scalarules/test/junit/test_discovery/ArchiveEntriesTest.scala
```diff
@@ -0,0 +1,25 @@
...
+import io.bazel.rulesscala.test_discovery.ArchiveEntries.listClassFiles
+import org.specs2.mutable.SpecWithJUnit
+
+import java.io.File
+import java.nio.file.{Files, Paths}
+
+class ArchiveEntriesTest extends SpecWithJUnit {
+  "List only class files from a directory" in {
+    val dir = Files.createTempDirectory("temp")
+    Files.createFile(Paths.get(dir.toString, "SomeFile"))
```
simuons, line 12: I think instead of `Paths.get(dir.toString, "SomeFile")` you could use `dir.resolve("SomeFile")`

simuons (approved):
LGTM. Just few minor comments

src/java/io/bazel/rulesscala/test_discovery/ArchiveEntries.scala
```diff
@@ -0,0 +1,36 @@
...
+    else
+      jarEntries(new JarInputStream(new FileInputStream(file)))
+
+    allEntries.filter(_.endsWith(".class"))
+  }
+
+  private def getJarEntryOrCloseStream(jarInputStream: JarInputStream): Option[JarEntry] = {
+    val entry = Option(jarInputStream.getNextJarEntry)
+
+    if (entry.isEmpty)
```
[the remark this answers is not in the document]
liucijus: It came from the original code, I will unify them

src/java/io/bazel/rulesscala/test_discovery/ArchiveEntries.scala
liucijus: Is that possible with a lazy stream? Or am I missing the point in what you're saying?

simuons: Right `lazy`... Then please ignore this comment. However I think this is a bit dangerous. What if stream is not consumed till the end, but maybe this is not a problem in this case.

liucijus: Yes, it's a risk, but I think impact is small. Current code also won't close in case exception is thrown before the stream is closed

PR #1293, opened 2021-09-27
implement multiplexer for scalac

wix-andriusb:
### Description
Add multiplexer support for the scalac worker.

This creates a thread pool inside of the scalac worker and uses threads in order to use a singular process.

### Motivation
Potential speed improvement and memory usage improvement because we use a single JVM process which warms up faster. 

### Implementation
The current implementation passes dedicated output streams to each worker which collect the output and then returns it to bazel.

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -125,7 +125,7 @@ def compile_scala(
         executable = scalac,
         mnemonic = "Scalac",
         progress_message = "scala %s" % target_label,
-        execution_requirements = {"supports-workers": "1"},
+        execution_requirements = {"supports-workers": "1", "supports-multiplex-workers": "1"},
         #  when we run with a worker, the `@argfile.path` is removed and passed
         #  line by line as arguments in the protobuf. In that case,
         #  the rest of the arguments are passed to the process that
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -119,13 +119,18 @@ def compile_scala(
     # scalac_jvm_flags passed in on the target override scalac_jvm_flags passed in on the toolchain
     final_scalac_jvm_flags = first_non_empty(scalac_jvm_flags, toolchain.scalac_jvm_flags)
 
+    if toolchain.multiplex_worker:
+        execution_requirements = {"supports-workers": "1", "supports-multiplex-workers": "1"}
+    else:
+        execution_requirements = {"supports-workers": "1"}
+
     ctx.actions.run(
         inputs = ins,
         outputs = outs,
         executable = scalac,
         mnemonic = "Scalac",
         progress_message = "scala %s" % target_label,
-        execution_requirements = {"supports-workers": "1", "supports-multiplex-workers": "1"},
+        execution_requirements = execution_requirements,
         #  when we run with a worker, the `@argfile.path` is removed and passed
         #  line by line as arguments in the protobuf. In that case,
         #  the rest of the arguments are passed to the process that
--- a/src/java/io/bazel/rulesscala/scalac/ReportableMainClass.java
+++ b/src/java/io/bazel/rulesscala/scalac/ReportableMainClass.java
@@ -1,9 +1,7 @@
 package io.bazel.rulesscala.scalac;
 
-import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
-import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.PrintStream;
 import java.nio.file.Files;
@@ -32,8 +30,6 @@ public class ReportableMainClass extends MainClass {
     if (!ops.enableDiagnosticsReport) {
       createDiagnosticsFile();
       Settings settings = super.settings();
-      InputStream inputStream = new ByteArrayInputStream(new byte[0]);
-      BufferedReader br = new BufferedReader(new InputStreamReader(inputStream));
       reporter = new StreamReporter(settings, out, err);
       Global global = new Global(settings, reporter);
 
--- a/src/java/io/bazel/rulesscala/scalac/ReportableMainClass.java
+++ b/src/java/io/bazel/rulesscala/scalac/ReportableMainClass.java
@@ -1,6 +1,11 @@
 package io.bazel.rulesscala.scalac;
 
+import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.PrintStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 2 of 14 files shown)


scala/private/rule_impls.bzl
```diff
@@ -123,7 +123,7 @@ def compile_scala(
         executable = scalac,
         mnemonic = "Scalac",
         progress_message = "scala %s" % target_label,
-        execution_requirements = {"supports-workers": "1"},
+        execution_requirements = {"supports-multiplex-workers": "1"},
```
wix-andriusb, line 126: supports-workers is still relevant

src/java/io/bazel/rulesscala/worker/Worker.java
```diff
@@ -11,22 +11,25 @@
...
  * <p>Worker implementations should implement the `Worker.Interface` interface and provide a main
  * method that calls `Worker.workerMain`.
  */
 public final class Worker {
 
+  private static ExecutorService executorService = Executors.newCachedThreadPool();
+
   public static interface Interface {
-    public void work(String[] args) throws Exception;
+    public void work(String[] args, PrintStream out, PrintStream err) throws Exception;
```
simuons, line 32: @wix-andriusb why worker interface takes two `PrintStream`s? Implementations are not allowed to touch `stdout` so I think the choice should not be given. And you are passing `stderr` twice anyway. I think minimal worker interface should be `(args, stderr) -> exit_code`

simuons (changes requested):
@wix-andriusb according to https://docs.bazel.build/versions/main/multiplex-worker.html multiplex-worker is still an experimental feature. So maybe this should be hidden behind toggle?

Also maybe it would be a good idea to have two different interfaces for simplex and multiplex workers? Currently all implementations were changed for all workers and this could give false assumption that it is enough to add `execution_requirements = {"supports-multiplex-workers": "1"},` which is not because current implementations ignores supplied `stderr` and might lead to strange behaviour.

src/java/io/bazel/rulesscala/worker/Worker.java
```diff
@@ -78,37 +74,64 @@ public void checkPermission(Permission permission) {
...
+    try {
+      out.flush();
+      WorkerProtocol.WorkResponse response =
+          WorkerProtocol.WorkResponse.newBuilder()
+              .setExitCode(code)
+              .setOutput(outStream.toString())
+              .setRequestId(request.getRequestId())
+              .build();
+
+      synchronized (lock) {
```
simuons, line 123: Maybe it would be possible to synchronize on `System.out`? That would reveal intent better. Or maybe syncronozed method? Anyway lock object is also fine.

wix-andriusb: I read that having a private lock object is the safer way of doing synchronization, because locking on a public object like System.out might create dead locks if someone else does the same.

src/java/io/bazel/rulesscala/worker/Worker.java
wix-andriusb: It needs a dedicated PrintStream for stdout, otherwise if a worker would do multiple job requests in parallel the std output would interleave.

Are you suggesting to use one PrintStream for both stdout and stderr?

wix-andriusb:
@simuons I created a dedicated MultiplexWorker.Interface and made only ScalacWorker utilize it, this means other workers can migrate independently to this new interface.

src/java/io/bazel/rulesscala/worker/Worker.java
simuons: Yes I suggest to use single output stream since you are passing same output stream for both parameters anyway https://github.com/bazelbuild/rules_scala/pull/1293/files#diff-ffcbe22d947b0629a710679137348a0a1c7a6b32e9b8cf284803c549ae194b97R99 
Unless I'm completely missing something.

wix-andriusb: fixed this

src/java/io/bazel/rulesscala/scalac/ReportableMainClass.java
```diff
@@ -13,25 +18,33 @@
...
 
   @Override
   public Global newCompiler() {
     if (!ops.enableDiagnosticsReport) {
       createDiagnosticsFile();
-      Global global = super.newCompiler();
-      reporter = global.reporter();
+      Settings settings = super.settings();
+      InputStream inputStream = new ByteArrayInputStream(new byte[0]);
+      BufferedReader br = new BufferedReader(new InputStreamReader(inputStream));
```
johnynek, line 36: this isn't used is it? Then neither is `inputStream`?

src/java/io/bazel/rulesscala/scalac/StreamReporter.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
+import java.io.InputStreamReader;
+import java.io.PrintStream;
+import java.io.PrintWriter;
+import scala.tools.nsc.Settings;
+import scala.tools.nsc.reporters.ConsoleReporter;
+
+class StreamReporter extends ConsoleReporter {
```
johnynek, line 11: isn't this the exact same code as the `211` code? Can't we share or at least document why it is the same or how it is subtly different? (I understand they will be compiled with different versions of scala, but that's always the case when we change versions).

src/java/io/bazel/rulesscala/scalac/StreamReporter.java
```diff
@@ -0,0 +1,24 @@
...
+  public StreamReporter(Settings settings, PrintStream out, PrintStream err) {
+    super(
+        settings,
+        new BufferedReader(new InputStreamReader(new ByteArrayInputStream(new byte[0]))),
```
johnynek, line 15: why do we need to buffer an empty input stream?

wix-andriusb: [The inherited constructor from ConsoleReporter expects a BufferedReader ](https://github.com/scala/scala/blob/8a2cf63ee5bad8c8c054f76464de0e10226516a0/src/compiler/scala/tools/nsc/reporters/ConsoleReporter.scala#L21)

src/java/io/bazel/rulesscala/scalac/StreamReporter.java
wix-andriusb: The signature of the constructor is different for [2.11](https://github.com/scala/scala/blob/2e2f65a201b0b06e01a39fec4ae232c5b94efcca/src/compiler/scala/tools/nsc/reporters/ConsoleReporter.scala#L17) and for the ve[rsions after that](https://github.com/scala/scala/blob/ce02613de571004cf7d0e4cc22ae919d825c947f/src/compiler/scala/tools/nsc/reporters/ConsoleReporter.scala#L21)

JackTreble:
Thanks @wix-andriusb for working on this I was starting to investigate how to do this change, 
@johnynek @simuons what is changes are outstanding before this can be approved? :smile:

wix-andriusb:
It did not make the compilation faster.

My guess is that it is bottlenecked by re-packing the jars.

liucijus:

PR #1295, opened 2021-09-29
Add flattenLeft function in Specs2RunnerBuilder to support newer specs2 versions

gaudrius:
Add flattenLeft function in Specs2RunnerBuilder to support newer specs2 versions.

`flattenLeft` was removed from Trees trait with ([fix: #892 always use flattenLeft for trees (including when calling size)
](https://github.com/etorreborre/specs2/commit/ba47c8f37c18aa27190a7510a00984ae9696460d#diff-162da17eeac842c31a79882aa0fc7e5536bbf91f9ee0a9890498d5f0481b84dbL70) commit.


liucijus (approved):
Thanks, @gaudrius

simuons (approved):

PR #1298, opened 2021-10-05
Add flag to disable writing volatile bits of statsfile

adam-singer:
### Description
Writing the milliseconds bit of the statsfile can cause hash sum differences in the disk cache. Adding a flag to disable writing the timestamp in attempt in improving cache hit rate when distributed with remote caching. 

### Motivation
We are trying to improve the cache hit rate for the action graph. 

### Manual Test

Doing the following should yield zero differences between the different disk caches when `scala_toolchain` `enable_stats_file = False`

```
rm -rf $DC1 $DC2
DC1=/tmp/disk_cache_15
DC2=/tmp/disk_cache_16
TARGET=<sometarget>
mkdir -p $DC1 && mkdir -p $DC2 && ./bazel clean && ./bazel build --disk_cache=$DC1 $TARGET && ./bazel clean && ./bazel build --disk_cache=$DC2 $TARGET && diff -r $DC1 $DC2
```


test_statsfile/BUILD
```diff
@@ -0,0 +1,25 @@
...
+    deps = [],
+)
+
+scala_binary(
+    name = "SimpleNoStatsFile",
+    srcs = ["Simple.scala"],
+    main_class = "scalarules.test.Simple",
+    print_compile_time = True,
+    deps = [],
+)
```
wisechengyi, line 21: how are these two targets different in terms of whether statsfile is enabled? I am unable to spot the difference.

test/toolchains/BUILD.bazel
```diff
@@ -104,3 +104,16 @@ toolchain(
...
+
+scala_toolchain(
+    name = "enable_stats_file_disabled_toolchain_impl",
+    enable_stats_file = False,
+    visibility = ["//visibility:public"],
+)
+
+toolchain(
+    name = "enable_stats_file_disabled_toolchain",
+    toolchain = "enable_stats_file_disabled_toolchain_impl",
```
wisechengyi, line 116: does it need to be `:enable_stats_file_disabled_toolchain_impl`?

test_statsfile/BUILD
adam-singer: `check_statsfile`/`check_statsfile_empty` expand into genrules and if the same name is use they would collide

test/toolchains/BUILD.bazel
adam-singer: It doesn't have to be, but I think using `:` would be qualifying, I followed the style used in  https://github.com/bazelbuild/rules_scala/blob/master/test/toolchains/BUILD.bazel

wisechengyi:
> How is [76f5950](https://github.com/bazelbuild/rules_scala/commit/76f59507f0bda29091edb1fecbc95b17d6aa9bc4) related to cache hit rate?

That was a conservative approach to make sure bazel treats the files with the same content to be same, rather than possibly distinguish them due to their last modified time, causing churns on AC and CAS.

liucijus (changes requested):
I'm ok with statsfile flag, but I think https://github.com/bazelbuild/rules_scala/commit/76f59507f0bda29091edb1fecbc95b17d6aa9bc4 must be another PR with a proper backing to the problem (it must be clear if such problem exists and what needs to be done to reproduce it - essential information for maintainers and contributors).

simuons:
Hi @adam-singer,

I like the idea of disabling `stats_file` (because not everyone cares or knows about it). But maybe it would be possible to completely stop generating that file if `enable_stats_file = False` (instead of generating empty file)?  Maybe this could be done by removing `stats_file` from "predeclared outputs" (defined via `common_outputs`) and conditionally add it to "default outputs" via `DefaultInfo(files)`? That way you wouldn't need to `touch` it. WDYT?

darl:
It can be done with additional output group:
Return `OutputGroupInfo(scala_stats_files = files)` from rule 
and request them via [--output_groups](https://docs.bazel.build/versions/main/command-line-reference.html#flag--output_groups) flag: `--ouptut_groups=+scala_stats_files`

simuons:
@darl that would be ideal but I have two concerns with this approach:
- potentially that could break some users (maybe not a an issue)
- stats file comes from same `scalac` action and file would be generated regardless if output group was requested or not (not sure there is a way to check which output groups ware requested in rule implementation)

adam-singer:
> I'm ok with statsfile flag, but I think [76f5950](https://github.com/bazelbuild/rules_scala/commit/76f59507f0bda29091edb1fecbc95b17d6aa9bc4) must be another PR with a proper backing to the problem (it must be clear if such problem exists and what needs to be done to reproduce it - essential information for maintainers and contributors).

ok, let me separate this patch, we have tested these changes internally while debugging remote caching. https://github.com/bazelbuild/rules_scala/commit/76f59507f0bda29091edb1fecbc95b17d6aa9bc4 is a defensive measure, but I would need to unbundle and do some testing to understand if we can benefit from it. 



> Hi @adam-singer,
> 
> I like the idea of disabling `stats_file` (because not everyone cares or knows about it). But maybe it would be possible to completely stop generating that file if `enable_stats_file = False` (instead of generating empty file)? Maybe this could be done by removing `stats_file` from "predeclared outputs" (defined via `common_outputs`) and conditionally add it to "default outputs" via `DefaultInfo(files)`? That way you wouldn't need to `touch` it. WDYT?

@simuons @darl I wanted to originally stop generating the file altogether but couldn't fully figure out what constructs were required to support that, I would need guidance in that approach. The backwards compatible approach I took was "opt out if disabled" but leaves the file in place and nulled. I wonder if a more ideal situation is to emit (if possible) this information to the overall bazel profiler and leave it out of the build graph altogether. In terms of approach and feasibility I wouldn't have much context on who depends on the stats file, I know we don't.

adam-singer:
One of the tests fails with `xargs: xmllint: No such file or directory`, is that some missing binary on the test agents?

simuons:
@adam-singer 

> One of the tests fails with xargs: xmllint: No such file or directory, is that some missing binary on the test agents?

try to rebase on master this was fixed in https://github.com/bazelbuild/rules_scala/commit/fd997e6f3dd4f9af7033116be1383a1be1c41554

liucijus (approved):

simuons (approved):

PR #1300, opened 2021-10-08
Document that it's possible to override scala artifacts while calling scala_repositories()

gergelyfabian:
### Description

I tried specifying a later minor version for Scala 2.12, but it failed due to circular dependencies with rules_jvm_external. I found this way that I could upgrade Scala 2.12 to a later version.
This is available in the code, but I did not find documentation for it.

### Motivation

I need to upgrade Scala 2.12 to a later version, due to a bug:

https://akka.io/blog/news/2021/02/23/akka-2.6.13-released

If you use a new Akka version, then you'll get strange errors with Scala <2.12.13

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,8 @@ Please check [coverage.md](docs/coverage.md) for more details on coverage suppor
 
 ## Selecting Scala version
 
+### With toolchains
+
 Rules scala supports the last two released minor versions for each of Scala 2.11, 2.12, 2.13.
 Previous minor versions may work but are supported only on a best effort basis.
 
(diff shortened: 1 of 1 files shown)


gergelyfabian:
Here is the Scala version upgrade with a toolchain, where I'm providing scala libraries with a repository rule that downloads those jars without rules_jvm_external:

https://github.com/gergelyfabian/bazel-scala-example/blob/33a0ed4444df9b94c2d81e540d1a97ef0601e184/WORKSPACE#L43-L64
https://github.com/gergelyfabian/bazel-scala-example/blob/33a0ed4444df9b94c2d81e540d1a97ef0601e184/tools/scala.bzl
https://github.com/gergelyfabian/bazel-scala-example/blob/33a0ed4444df9b94c2d81e540d1a97ef0601e184/tools/jdk/BUILD.bazel

If this is better, then question, is it documented currently how to use `scala_maven_import_external()` to download scala jars on our own? (I don't see such documentation, is it expected to stay public?)

The example for which I had the circular dependency error was coming from: https://github.com/liucijus/rules-scala-proto-toolchain/blob/main/BUILD (there `@maven//:..` was used as a target for the scala jars).

gergelyfabian:
Ok, I have no idea what happened in my earlier example with circular deps, but now I could make it work, using your proto toolchain example, no circular deps and I'm using targets generated by rules_jvm_external:

https://github.com/gergelyfabian/bazel-scala-example/blob/5b5e36db3eb49ca823316fe11fa5dcc6c5d20e40/tools/jdk/BUILD.bazel

I guess I'll need to sort out my other repo, comparing it with bazel-scala-example.

So in this case, if you wish, maybe this PR should be closed?

liucijus:
@gergelyfabian there are a few ways with this PR:
- merge PR as is - it's a valid doc
- add a line saying that toolchains are more flexible way to configure dependencies. Maybe also a line saying that overrides will go away in the future.
- close it

BTW, one of the reasons why deps toolchains were created is for loaders like rules_jvm_external, and it should work fine.

PR #1301, opened 2021-10-08
Upgrading Scala 2.12 to 2.12.14

gergelyfabian:
Scalap version for 2.12 was desynchronized from 2.12.11, also upgraded that.

### Description
Upgrading Scala 2.12 default version from 2.12.11 to 2.12.14.

### Motivation
A bug in Akka/Scala, that makes using 2.12.11 a pain with Akka:

https://akka.io/blog/news/2021/02/23/akka-2.6.13-released

If you use a new Akka version with Scala 2.12.11, then you'll get errors like:

```
Symbol 'type scala.annotation.nowarn' is missing from the classpath
```

### DIFF
--- a/third_party/dependency_analyzer/src/main/BUILD
+++ b/third_party/dependency_analyzer/src/main/BUILD
@@ -1,5 +1,5 @@
 load("//scala:scala.bzl", "scala_library_for_plugin_bootstrapping")
-load("@io_bazel_rules_scala_config//:config.bzl", "SCALA_MAJOR_VERSION")
+load("@io_bazel_rules_scala_config//:config.bzl", "SCALA_MAJOR_VERSION", "SCALA_VERSION")
 
 licenses(["notice"])  # 3-clause BSD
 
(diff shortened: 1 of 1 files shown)


gergelyfabian:
This unit test error is really strange. It seems that for this Scala version it returns non-deterministically either `UnitTests` or `Category` as a dependency error (checked locally), it's never both of them.

gergelyfabian:
Added documentation in test case.

gergelyfabian:
Edited unit tests and documentation to say that we need special handling for the unit tests starting from 2.12.13.
Test versions for:

* 2.12.15: https://github.com/gergelyfabian/rules_scala/commit/12fd7ced992f30d490406aca639be66ae0ed4cfb
* 2.12.13: https://github.com/gergelyfabian/rules_scala/commit/d481065e9e56eb27de1be93696a55306741af44a
* 2.12.12: https://github.com/gergelyfabian/rules_scala/commit/dfb45d909d7cfb9d337f6ce618df29ab9bcfd01c

liucijus:
Is this the same reporter issue in ast tests as with [2.13](https://github.com/bazelbuild/rules_scala/blob/master/third_party/dependency_analyzer/src/main/BUILD#L30)? It's important that all dep tracking error are reported in the same build. I that's the case, you can reuse existing reporter, just change the version ckeck in BUILD file.

gergelyfabian:
I think it may be the same issue. In this case I'm removing the test changes and modifying the BUILD file as you suggested.

Issue #1322, opened 2021-12-11
Extremely long file names


caseyduquettesc:
After #1315, extremely long file names can result in exceptions.

```
File name too long
        at sun.nio.fs.UnixException.translateToIOException(UnixException.java:91)
        at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
        at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
        at sun.nio.fs.UnixFileSystemProvider.createDirectory(UnixFileSystemProvider.java:384)
        at java.nio.file.Files.createDirectory(Files.java:674)
        at java.nio.file.Files.createAndCheckIsDirectory(Files.java:781)
        at java.nio.file.Files.createDirectories(Files.java:767)
        at io.bazel.rulesscala.scalac.ScalacWorker.extractSourceJars(ScalacWorker.java:118)
        at io.bazel.rulesscala.scalac.ScalacWorker.work(ScalacWorker.java:37)
        at io.bazel.rulesscala.worker.Worker.persistentWorkerMain(Worker.java:84)
        at io.bazel.rulesscala.worker.Worker.workerMain(Worker.java:39)
        at io.bazel.rulesscala.scalac.ScalacWorker.main(ScalacWorker.java:25)
```

_Originally posted by @caseyduquettesc in https://github.com/bazelbuild/rules_scala/pull/1315#r767089451_

PR #1323, opened 2021-12-11
Fix long file names in scalac action when handling srcjars

simuons:
### Description
Instead of using full path of srcjar as a destination directory
prepend index to srcjar file name.
Index corresponds to order in which file was passed to scalac.

Also considered computing hash of full path, but full path
contains platform specific labels ie darwin, linux etc
so not sure that would be a good approach.

### Motivation
Resolves https://github.com/bazelbuild/rules_scala/issues/1322


caseyduquettesc (approved):
FWIW, LGTM!

liucijus (approved):

PR #1325, opened 2021-12-13
Update deps of my_scala_xml_provider

ahe-humio:
### Description
The example didn't work for me, but using io_bazel_rules_scala_scala_xml worked.

### Motivation
The example appears to be wrong.

### DIFF
--- a/docs/scala_toolchain.md
+++ b/docs/scala_toolchain.md
@@ -78,7 +78,7 @@ scala_register_toolchains()
     declare_deps_provider(
         name = "my_scala_xml_provider",
         deps_id = "scala_xml",
-        deps = ["@scala_xml_dep"],
+        deps = ["@io_bazel_rules_scala_scala_xml"],
     )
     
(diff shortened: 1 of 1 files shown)


ahe-humio:
Sorry, I'm not gonna sign the CLA.

liucijus:
@ahe-humio thanks for PR. I know this is confusing, but I don't recommend using rules_scala external names outside rules_scala repository (I mean all deps prefixed with `@io_bazel_rules_scala`), because they are subject to internal implementation change. Therefore most of the deps in the example should be replaced with something else. Those deps should come from user's WORKSPACE, probably using some maven deps loader, like https://github.com/bazelbuild/rules_jvm_external. Practical example: https://github.com/liucijus/rules-scala-proto-toolchain/blob/main/BUILD. 
Regarding the current doc. It needs to be improved to describe better where the deps come from. Maybe it can use Rules Jvm External names, so it's more obvious where these names come from.

PR #1331, opened 2022-01-12
scala_doc: Add support for building docs without transitive dependencies

rahuliyer95:
### Description
Add support to build scaladocs only for the provided targets.

This enables building scaladocs without building docs all transitive dependencies (i.e 3rdparty dependencies and other dependent targets). The aim is to provide a `transitive` arg that can be set to `true`/`false` (defaults to `true`) which decides whether to build scaladocs for transitive dependencies or not.

### Motivation
We are using bazel in a big mono-repo and building scaladocs for some of our targets take a long time because of the large dependency chain and sometimes long list of 3rdparty libraries. We only want to build scaladocs for the targets we specify in the `deps` field of the `scala_doc` rule. This change adds support for that.

### DIFF
--- a/scala/private/rules/scala_doc.bzl
+++ b/scala/private/rules/scala_doc.bzl
@@ -112,7 +112,7 @@ def make_scala_doc_rule(aspect = _scaladoc_transitive_aspect):
             ),
             "scalacopts": attr.string_list(),
             "_scaladoc": attr.label(
-                cfg = "host",
+                cfg = "exec",
                 executable = True,
                 default = Label("//src/scala/io/bazel/rules_scala/scaladoc_support:scaladoc_generator"),
             ),
--- a/scala/private/rules/scala_doc.bzl
+++ b/scala/private/rules/scala_doc.bzl
@@ -14,11 +14,14 @@ def _scaladoc_aspect_impl(target, ctx):
     # We really only care about visited targets with srcs, so only look at those.
     if hasattr(ctx.rule.attr, "srcs"):
         # Collect only Java and Scala sources enumerated in visited targets, including src_files in deps.
-        src_files = depset(
-            direct = [file for file in ctx.rule.files.srcs if file.extension.lower() in ["java", "scala"]],
-            transitive = [dep[_ScaladocAspectInfo].src_files for dep in ctx.rule.attr.deps if _ScaladocAspectInfo in dep],
-        )
+        direct_deps = [file for file in ctx.rule.files.srcs if file.extension.lower() in ["java", "scala"]]
+        transitive_deps = []
+        # Sometimes we only want to generate scaladocs for a single target and not all of its
+        # dependencies
+        if ctx.attr.transitive == "true":
+            transitive_deps = [dep[_ScaladocAspectInfo].src_files for dep in ctx.rule.attr.deps if _ScaladocAspectInfo in dep]
 
+        src_files = depset(direct = direct_deps, transitive = transitive_deps)
         # Collect compile_jars from visited targets' deps.
         compile_jars = depset(
             direct = [file for file in ctx.rule.files.deps],
@@ -43,6 +46,9 @@ def _scaladoc_aspect_impl(target, ctx):
 _scaladoc_aspect = aspect(
     implementation = _scaladoc_aspect_impl,
     attr_aspects = ["deps"],
+    attrs = {
+        "transitive": attr.string(default="true", values=["true", "false"]),
+    },
     required_aspect_providers = [
         [JavaInfo],
     ],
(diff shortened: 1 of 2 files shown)


scala/private/rules/scala_doc.bzl
```diff
@@ -43,6 +46,9 @@ def _scaladoc_aspect_impl(target, ctx):
 _scaladoc_aspect = aspect(
     implementation = _scaladoc_aspect_impl,
     attr_aspects = ["deps"],
+    attrs = {
+        "transitive": attr.string(default="true", values=["true", "false"]),
```
wisechengyi, line 50: Could this be boolean instead of string? ditto for the other.

wisechengyi (commented):
LGTM. thanks!

scala/private/rules/scala_doc.bzl
rahuliyer95: Setting this to `"transitive": attr.bool(default=True),` I get the following error,
```
	File "/private/var/tmp/_bazel_rahuli/7534b642ae61a45da9baaf1cc06fb325/external/rules_scala/scala/private/rules/scala_doc.bzl", line 46, column 26, in <toplevel>
		_scaladoc_aspect = aspect(
Error in aspect: Aspect parameter attribute 'transitive' must have type 'string' and use the 'values' restriction.
```
So I believe we need to keep it a `string` with restricted `values`. I am no expert on `bazel` so happy to include a better way to do this

simuons: According to [aspects doc](https://docs.bazel.build/versions/main/skylark/aspects.html) parameters must be of type string

> Public aspect attributes are of type string and are called parameters

simuons:
Hi, @rahuliyer95 ,

would you use `scala_doc` in transitive mode in your repo? Or you would always set `transitive = "false"`? If that's the case maybe other approach would be to use rule factory "pattern" and supply non transitive aspect for it. Something like `scala_doc = make_scala_doc(non_transitive_aspect)` and use that through your repo. Something like [this](https://github.com/bazelbuild/rules_scala/blob/master/scala_proto/private/scala_proto.bzl#L47)

scala/scala.bzl
```diff
@@ -50,7 +51,9 @@ def scala_specs2_junit_test(name, **kwargs):
 
 # Re-export private rules for public consumption
 scala_binary = _scala_binary
-scala_doc = _scala_doc
+make_scala_doc_rule = _make_scala_doc_rule
```
wisechengyi, line 54: is this to be expected to be used by user later? if not, can be removed I think.

or is it intended to be used with potentially custom aspect later?

scala/private/rules/scala_doc.bzl
```diff
@@ -8,16 +8,25 @@ _ScaladocAspectInfo = provider(fields = [
...
 
     # We really only care about visited targets with srcs, so only look at those.
     if hasattr(ctx.rule.attr, "srcs"):
         # Collect only Java and Scala sources enumerated in visited targets, including src_files in deps.
-        src_files = depset(
-            direct = [file for file in ctx.rule.files.srcs if file.extension.lower() in ["java", "scala"]],
-            transitive = [dep[_ScaladocAspectInfo].src_files for dep in ctx.rule.attr.deps if _ScaladocAspectInfo in dep],
-        )
+        direct_deps = [file for file in ctx.rule.files.srcs if file.extension.lower() in ["java", "scala"]]
+        transitive_deps = []
```
wisechengyi, line 22: nit: might be more clear to put this under "else"

wisechengyi (approved):
Thanks, Rahul. Just some minor comments.

scala/scala.bzl
rahuliyer95: we will probably need to use `make_scala_doc_rule` and the `scaladoc_intransitive_aspect` to build it internally. That's why they are exported.

wisechengyi: Thanks. Makes sense. If you don't mind, we can add a comment here, so it's more clear for folks later on. E.g. people may notice this is not used in the repo and it gets removed.

scala/private/rules/scala_doc.bzl
```diff
@@ -85,19 +103,20 @@ def _scala_doc_impl(ctx):
...
+def make_scala_doc_rule(aspect = _scaladoc_transitive_aspect):
+    return rule(
+        attrs = {
+            "deps": attr.label_list(
+                aspects = [aspect],
+                providers = [JavaInfo],
+            ),
+            "scalacopts": attr.string_list(),
+            "_scaladoc": attr.label(
+                cfg = "host",
```
liucijus, line 115: nit: `host` seems to be deprecated, and `exec` should be used instead. As you touch this code, I'd appreciate if you can also update it.

liucijus (changes requested):
LGTM, with a small nit I'd like to be addressed

PR #1336, opened 2022-01-18
use composition in runner to avoid double lazy initialization when ru…

mardaravicius:
when running some tests eg.:
```
import org.specs2.mutable.SpecWithJUnit

class MyTest extends SpecWithJUnit {
  new Throwable().printStackTrace()
  println("test initialized")

  "1" should {
    "equal 1" >> {
      println("test ran")
      1 must_== 1
    }
  }
}
```
sometimes the class init block runs twice and prints the `test initialized` message twice,
when inspecting the 2 stack traces:
https://www.diffchecker.com/kSHQiYj1

found that 2 runners are created and both run their lazy `JUnitRunner.specStructure` value which creates 2test classes.
changed the runner so it uses the parent runner as composition instead of inheriting it. this might fix the issue.
though could not reproduce with any test cases in this repo. maybe someone else has some other ideas on why this happens and how to reproduce in a test?

### DIFF
--- a/src/java/io/bazel/rulesscala/specs2/Specs2RunnerBuilder.scala
+++ b/src/java/io/bazel/rulesscala/specs2/Specs2RunnerBuilder.scala
@@ -48,16 +48,16 @@ class Specs2PrefixSuffixTestDiscoveringSuite(suite: Class[Any], runnerBuilder: R
 
 object Specs2FilteringRunnerBuilder {
   val f: FilteringRunnerBuilder = {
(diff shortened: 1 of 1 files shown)


mardaravicius:
tested on our repo when linking to this pr and the problem seems fixed.

liucijus:
@mardaravicius thanks for taking this! Please take look at failing checks

mardaravicius:
@googlebot I signed it!

liucijus (approved):

simuons (approved):

Issue #1337, opened 2022-01-25
Support proto3 optional


brandon-leapyear:
proto3 supports optionals by default after 0.15. This library does use a new enough version of protoc:
```
$ bazel run @com_google_protobuf//:protoc -- --version
libprotoc 3.18.0
```
But it seems like the version of scalapb doesn't support this?
```
error.proto: is a proto3 file that contains optional fields, but code generator protoc-gen-scala hasn't been updated to support optional fields in proto3. Please ask the owner of this code generator to support proto3 optional.--scala_out:
```

I currently see:
https://github.com/bazelbuild/rules_scala/blob/b85d1225d0ddc9c376963eb0be86d9d546f25a4a/third_party/repositories/scala_2_13.bzl#L133
Should this be updated to 0.11?
* https://github.com/scalapb/ScalaPB/issues/870#issuecomment-654320187

liucijus:
@brandon-leapyear if you are looking to use different version of scalapb in your projects, are you aware that you can use other versions by defining [scala_proto_deps_toolchain](https://github.com/bazelbuild/rules_scala/blob/master/docs/scala_proto_library.md)? Example of such toolchain configuration: https://github.com/liucijus/rules-scala-proto-toolchain (it uses old versions still, but should be trivial to update [here](https://github.com/liucijus/rules-scala-proto-toolchain/blob/main/WORKSPACE#L111-L160)). If you use Rules Scala in production projects eventually you'll need to define your own dependency configuration.

PR #1338, opened 2022-01-25
Bump scalapb version

brandon-leapyear:
Trying fix for #1337 

I'm new to Bazel/Scala, so there are certainly more changes (I'm not sure how to update the SHAs), but this is my first draft


liucijus:
> Trying fix for #1337
> 
> I'm new to Bazel/Scala, so there are certainly more changes (I'm not sure how to update the SHAs), but this is my first draft

@brandon-leapyear thanks for the contribution!

The easiest way to calculate hashes is to invalidate them buy changing one symbol (so that lookup to cache fails) and copy correct shas from the Bazel build output. You can even grep some from currently failing build on Rules Scala CI. The nasty part is that you'll have to do that for each supported Scala version.

I would also appreciate if you could contribute a small test target so that we can track features and regressions.

brandon-leapyear:
So actually, it seems like ScalaPB doesn't have artifacts for Scala 2.11, which means `scala_2_11.bzl` is stuck at the current version. I believe this means we can't upgrade ScalaPB for Scala 2.12 or 2.13, because then we'd have to update code to work for both ScalaPB 0.9 and 0.11, is that right?

scala_proto/BUILD
```diff
@@ -83,6 +83,7 @@ declare_deps_provider(
     deps = [
         "@com_google_protobuf//:protobuf_java",
         "@scala_proto_rules_protoc_bridge",
+        "@scala_proto_rules_protoc_gen",
```
simuons, line 86: Is this really needed? As I remember `rules_scala` doesn't use helpers from this jar ie `CodeGenApp`.

simuons:
Hi @brandon-leapyear, 

> we'd have to update code to work for both ScalaPB 0.9 and 0.11, is that right?

I'm not sure we need that (need to verify). As I remember `rules_scala` mainly depends on `protoc-bridge` and it's [protocbridge.{ProtocBridge, ProtocCodeGenerator}](https://github.com/bazelbuild/rules_scala/blob/master/src/scala/scripts/ScalaPBWorker.scala#L8) and maybe that didn't change between versions. If [ExtraProtobufGenerator](test/src/main/scala/scalarules/test/extra_protobuf_generator/ExtraProtobufGenerator.scala) causes issues probably we can simplify it be removing usages of `scalapb` implicits etc.

So I'd say let's try to upgrade `scalapb` for scala 2.12 and 2.13 and keep old version for 2.11

scala_proto/BUILD
brandon-leapyear: I'm sure i added this because something was giving an error. you can try removing it and seeing what errors

liucijus:
Closing as stale
> Hi @simuons thanks for the review! I think we were able to get around this for my work, and i haven't touched bazel since, so I probably won't be much more help on this PR

PR #1343, opened 2022-02-03
[upgrade policy to be discussed] Upgrade bazel to 5.0.0

wisechengyi:


liucijus:
@wisechengyi what is the motivation for this change? I would like to keep backwards compatibility with Bazel 4.1 as long as possible. I know users who haven't migrated to Bazel 5 yet.

wisechengyi:
Hi @liucijus

That's a good question. Currently Twitter is the process trying to upgrade to 5 in order to consume certain features, so would like to see how rules scala works under 5. That said, this is not urgent, and it also makes sense to keep 4.x compatibly for a while. I guess the question here is - what is the general policy on bazel version compatibility, e.g. would it make sense to be compatible to the latest 2-3 major releases?

liucijus:
Rules Scala should work with Bazel 5. If there's is a non compatible feature, it would be great to have a test for it.

Me and @simuons are currently going over existing Buildkite configuration for Rules Scala, I'll see if I can add Bazel 5 build to it.

We don't have any rules when to upgrade Bazel version for the project, but my rule of thumb is don't upgrade until it is really needed (or make sense). If there's a need to support multiple Bazel versions, we will need to introduce a way to support it, which is additional effort from maintainers.

wisechengyi:
> don't upgrade until it is really needed (or make sense)

Although there is definitely the need to be on stable builds and no urgency to upgrade here, I generally disagree because upgrade will eventually still have to happen. When it happens, it would be much easier to upgrade one version at a time compared to jumping through multiple major releases. I said this possibly due to our large surface area, as we do see some breakages going from 4 to 5 (not necessarily related to rules_scala and still investigating), so extent could vary for different orgs.

Assuming the above makes sense, then keeping (at least) 2 major releases compatible with rules scala might be a good idea. Whether we have the resources to do it is another story as this is definitely not trivial for the maintainers, but I'd like to see if we agree on the principles.

liucijus:
I think I'm not being precise with what I mean with upgrading. In Rules Scala bumping `.bazelversion` means dropping support for older Bazel versions. But it does not mean Rules Scala is not supported on newer Bazel version. Supporting Bazel LTS is priority for this ruleset, but version specified in `.bazelversion` is not about what our users are supposed to use in their projects.
Rules Scala has been supported on Bazel 4.1.0 and Bazel 5.0.0 for a long time already.

wisechengyi:
Oh I see. Thanks for clarifying. In that case there's no need for this PR. Thank you!

PR #1347, opened 2022-02-08
update default `scalafmt` phase order; place before `runfiles`

virusdave:
### Description
See comments in change for detailed motivation.  TL;DR: Allow the capture of formatted sources, original sources, and manifest to be passed via `runfiles` phase to downstream dependencies, such as a wrapping `sh_test`, allowing it to
actually access those during a `bazel test` invocation and fail a build if the format is incorrect.

### DIFF
--- a/scala/scalafmt/BUILD
+++ b/scala/scalafmt/BUILD
@@ -40,8 +40,11 @@ declare_deps_provider(
     visibility = ["//visibility:public"],
     deps = [
         "@com_geirsson_metaconfig_core",
+        "@org_scalameta_common",
         "@org_scalameta_parsers",
         "@org_scalameta_scalafmt_core",
+        "@org_scalameta_scalameta",
+        "@org_scalameta_trees",
     ],
 )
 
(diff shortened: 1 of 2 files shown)


liucijus (approved):

virusdave:
Lemme check why the tool deps are causing a test failure now whereas they weren't previously.  I might need to tweak something and will update the PR if so.

scala/scalafmt/BUILD
```diff
@@ -40,8 +40,11 @@ declare_deps_provider(
     visibility = ["//visibility:public"],
     deps = [
         "@com_geirsson_metaconfig_core",
+        "@org_scalameta_common",
         "@org_scalameta_parsers",
         "@org_scalameta_scalafmt_core",
+        "@org_scalameta_scalameta",
+        "@org_scalameta_trees",
```
virusdave, line 47: NB: These are already transitive dependencies of `scalafmt_core`, but with the new phase position, the test requiring explicitly-specified dependencies requires them to be listed explicitly.

simuons (approved):

PR #1360, opened 2022-02-24
set -e for ctx.actions.run_shell commands

simuons:
Actions might look like completed successfully even though one of the commands fail.
This might cause remote-cache poisoning.

### DIFF
--- a/scala/private/phases/phase_compile.bzl
+++ b/scala/private/phases/phase_compile.bzl
@@ -261,6 +261,7 @@ def _build_nosrc_jar(ctx):
     zipper_arg_path = ctx.actions.declare_file("%s_zipper_args" % ctx.label.name)
     ctx.actions.write(zipper_arg_path, resources)
     cmd = """
+set -e
 rm -f {jar_output}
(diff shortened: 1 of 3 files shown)


scala/private/phases/phase_compile.bzl
```diff
@@ -261,6 +261,7 @@ def _build_nosrc_jar(ctx):
     zipper_arg_path = ctx.actions.declare_file("%s_zipper_args" % ctx.label.name)
     ctx.actions.write(zipper_arg_path, resources)
     cmd = """
+set -e
```
eed3si9n, line 264:
```suggestion
set -o errexit
set -o nounset
set -o pipefail
```

Generally I prefer long options for scripting. If we're adding `-e`, might as well add `nounset` and `pipefail` as well? I'd be curious if `touch` always succeeds. It might not be rules_scala, but I sometimes see `touch` getting rejected. (Likely it already exists)

eed3si9n:
s/run_shall/run_shell/

liucijus (approved):

PR #1370, opened 2022-03-18
Justinass/dep analyzer

justinass-wix:
### Description
Minimal implementation of compiler utils for Scala 3 Dependency Analyzer tests.

### Motivation
Dependency Analyzer uses `TestUtil` to run compiler. For Scala 3 these utils have to be rewritten to work with new compiler.

--------
* Split `analyzer_test.bzl ` to 2 files for scala 2 & 3 respectively
* Passing `scala-library` and  `scala3-library_3`paths via JVM flags to satisfy compiler (for scala 2 `scala-library` and `scala-reflect` are passed)


third_party/utils/src/test/io/bazel/rulesscala/utils/Scala3CompilerUtils.scala
```diff
@@ -0,0 +1,52 @@
...
+object Scala3CompilerUtils {
+  def runCompiler(
+    code: String,
+    extraClasspath: List[String] = List.empty,
+  ): List[Diagnostic] = {
+    val reporter = new TestReporter()
+
+    implicit val context: FreshContext = (new ContextBase).initialCtx.fresh.setReporter(reporter)
+
+    val fullClassPath =
```
johnynek, line 19: can you comment if this order is intentional? if not, it is more efficient to build lists right to left:

i.e.
```scala
context.settings.classpath.value :: extraClasspath ::: builtinClasspaths.filterNot(_.isEmpty)
```

third_party/utils/src/test/io/bazel/rulesscala/utils/Scala3CompilerUtils.scala
```diff
@@ -0,0 +1,52 @@
...
+      builtinClasspaths.filterNot(_.isEmpty) ++
+        extraClasspath :+
+        context.settings.classpath.value
+
+    context.setSetting(context.settings.classpath, fullClassPath.mkString(":"))
+
+    val compiler = new Compiler()
+    val run = compiler.newRun
+    run.compileSources(List(SourceFile.virtual("code.scala", code, maybeIncomplete = false)))
```
johnynek, line 28: nit: `SourceFile.virtual("code.scala", code, maybeIncomplete = false) :: Nil` is more efficient than `List(...)` because `List` has to allocate an intermediate Seq of the args.

more important: `code.scala` looks a bit bland and I wonder if we will forget where it came from. I wonder if it should be passed in with a default of `scala_compiler_util_run_code.scala` or something.

third_party/utils/src/test/io/bazel/rulesscala/utils/Scala3CompilerUtils.scala
```diff
@@ -0,0 +1,52 @@
...
+  private def pathOf(jvmFlag: String) = {
+    val jar = System.getProperty(jvmFlag)
+    val libPath = Paths.get(baseDir, jar).toAbsolutePath
+    libPath.toString
+  }
+
+  private lazy val baseDir = System.getProperty("user.dir")
+}
+
+import dotty.tools.dotc.reporting.StoreReporter
```
johnynek, line 48: can we move this import to the top of the file?

third_party/utils/src/test/io/bazel/rulesscala/utils/Scala3CompilerUtils.scala
justinass-wix: Order does not matter so updated the line with your suggestion. Thanks!

third_party/utils/src/test/io/bazel/rulesscala/utils/Scala3CompilerUtils.scala
justinass-wix: Updated as per your suggestion.

third_party/utils/src/test/io/bazel/rulesscala/utils/Scala3CompilerUtils.scala
justinass-wix: Done

third_party/dependency_analyzer/src/test/analyzer_test_scala_3.bzl
```diff
@@ -0,0 +1,24 @@
+load("@io_bazel_rules_scala//scala:scala.bzl", "scala_test")
+
+def analyzer_tests_scala_3():
+    common_jvm_flags = [
+        "-Dplugin.jar.location=$(execpath //third_party/dependency_analyzer/src/main:dependency_analyzer)",
+        "-Dscala.library.location=$(rootpath @io_bazel_rules_scala_scala_library)",
+        "-Dscala.library2.location=$(rootpath @io_bazel_rules_scala_scala_library_2)",
```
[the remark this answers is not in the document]
justinass-wix: I don't actually know why Scala 2 library is required, I tried various things until I got the code to compile... One hint was this [line](https://github.com/justinass-wix/rules_scala/blob/9b85affa2e08a350a4315882b602eda55b262356/third_party/repositories/scala_3_1.bzl#L10) .

Without it compilation fails:
```
class dotty.tools.dotc.core.Symbols$NoSymbol$ cannot be cast to class dotty.tools.dotc.core.Symbols$ClassSymbol (dotty.tools.dotc.core.Symbols$NoSymbol$ and dotty.tools.dotc.core.Symbols$ClassSymbol are in unnamed module of loader 'app')
```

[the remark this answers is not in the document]
justinass-wix: I added a comment.

liucijus:
Thanks, @justinass-wix

simuons (approved):

PR #1372, opened 2022-03-19
Use java_common.stamp_jar to stamp scala_import jar

### LINKED ISSUE #1371: ijar hack revert?
Can this be reverted?
https://github.com/bazelbuild/rules_scala/blob/2437e40131072cadc1628726775ff00fa3941a4a/scala/scala_import.bzl#L7-L9

Looks like the referenced issue was fixed already.

liucijus:
Let's see if we can use ijar via java_common.stamp_jar

Fixes #1371

### DIFF
--- a/scala/scala_import.bzl
+++ b/scala/scala_import.bzl
@@ -2,7 +2,7 @@ load("@io_bazel_rules_scala//scala:jars_to_labels.bzl", "JarsToLabelsInfo")
 load("//scala/settings:stamp_settings.bzl", "StampScalaImport")
 
 def _stamp_jar(ctx, jar):
-    stamped_jar_filename = "%s.stamp/%s.jar" % (ctx.label.name, jar.basename.rstrip(".jar"))
+    stamped_jar_filename = "%s.stamp/%s" % (ctx.label.name, jar.basename)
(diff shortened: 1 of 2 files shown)


scala/scala_import.bzl
```diff
@@ -2,38 +2,16 @@ load("@io_bazel_rules_scala//scala:jars_to_labels.bzl", "JarsToLabelsInfo")
...
-        outputs = [stamped_file],
-        arguments = [
-            "--nostrip_jar",
-            "--target_label",
-            str(ctx.label),
-            jar.path,
-            stamped_file.path,
-        ],
-        mnemonic = "StampWithIjar",
+    stamped_jar_filename = "%s.stamp/%s.jar" % (ctx.label.name, jar.basename.rstrip(".jar"))
```
simuons, line 5: I guess there is no reason to `rstrip` jar suffix which is added back in `stamped_jar_filename`

simuons (approved):
Looks good. Just one minor comment.

scala/scala_import.bzl
liucijus: Thanks for catching this

PR #1376, opened 2022-04-01
Create jdeps output from classpath entries

liucijus:
### Description
jdeps files can be used by Intellij Bazel plugin to build project model without requiring special support for Scala rules.

This implementation is very simple: treats all classpath entries as explicit deps, and is written by ScalacWorker. I decided not to put it on dependency analyzer as it requires users to have it enabled, which can be undesired performance hit. Writing actual unused deps information from the analyzer can be implemented later.

### Motivation
Improving Scala support in Intellij Bazel plugin. This should help advance support for external source attachment. Though more work is required on the plugin side: 
- consuming full jars instead of ijar
- properly detecting source roots for Scala source artifacts


scala/private/rule_impls.bzl
```diff
@@ -79,6 +80,8 @@ def compile_scala(
     args.add("--Manifest", manifest)
     args.add("--PrintCompileTime", print_compile_time)
     args.add("--ExpectJavaOutput", expect_java_output)
+    if jdepsPath != None:
```
simuons, line 83: When `jdepsPath` can be `None`? Maybe there is no need for `None` and `null` checks?

src/java/io/bazel/rulesscala/jdeps/BUILD
```diff
@@ -0,0 +1,11 @@
+load("@rules_java//java:defs.bzl", "java_library")
+
+java_library(
+    name = "jdeps",
+    srcs = ["JdepsWriter.java"],
+    visibility = ["//visibility:public"],
+    deps = [
+        "@bazel_tools//src/main/protobuf:deps_java_proto",
```
simuons, line 8: Is it safe to depend on this? I do remember there were some issues with `@bazel_tools//src/main/protobuf:worker_java_proto` but I don't remember what was that and maybe this doesn't apply here.

src/java/io/bazel/rulesscala/jdeps/JdepsWriter.java
```diff
@@ -0,0 +1,30 @@
+package io.bazel.rulesscala.jdeps;
+
+import com.google.devtools.build.lib.view.proto.Deps;
+import java.io.BufferedOutputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+
+public class JdepsWriter {
```
simuons, line 9: I think according whole ceremony this class should be `final` with private constructor. But I'm fine with the way it is now.

simuons, line 9: I think that according to whole ceremony this class should be `final` with private constructor. But I'm fine with the way it is now.

src/java/io/bazel/rulesscala/jdeps/JdepsWriter.java
```diff
@@ -0,0 +1,30 @@
...
+
+    for (String jar : classpath) {
+      Deps.Dependency.Builder dependencyBuilder = Deps.Dependency.newBuilder();
+      dependencyBuilder.setKind(Deps.Dependency.Kind.EXPLICIT);
+      dependencyBuilder.setPath(jar);
+      builder.addDependency(dependencyBuilder.build());
+    }
+
+    try (OutputStream outputStream = new BufferedOutputStream(new FileOutputStream(jdpesPath))) {
+      outputStream.write(builder.build().toByteArray());
```
simuons, line 26: I think `builder.build().writeTo(outputStream)` should me more effective as it doesn't construct intermediary byte array.

simuons (approved):
LGTM. Just few minor comments.

src/java/io/bazel/rulesscala/jdeps/BUILD
liucijus: I had some quirks with scala toolchain resolution in other parts when I've tried it on dependecy analyzer, but that seems to be fixed in Bazel 5, and current setup seems to work fine with Bazel 4 (I've tested on a large code base)

simuons (approved):

PR #1377, opened 2022-04-04
Take --instrumentation_filter into account in coverage phase

fmeum:
### Description

Previously, all Scala code would be instrumented for coverage during a
bazel coverage run instead of only those source files that belong to
targets matching the instrumentation filter.

### Motivation

Fixes https://github.com/bazelbuild/bazel/issues/15167.

### DIFF
--- a/test/shell/test_coverage_scalatest_resources.sh
+++ b/test/shell/test_coverage_scalatest_resources.sh
@@ -5,7 +5,9 @@ dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 runner=$(get_test_runner "${1:-local}")
 
 test_coverage_succeeds_resource_call() {
-    bazel coverage //test/coverage_scalatest_resources/consumer:tests
+    bazel coverage \
+          --instrumentation_filter=^//test/coverage_scalatest_resources[:/] \
+          //test/coverage_scalatest_resources/consumer:tests
     diff test/coverage_scalatest_resources/expected-coverage.dat $(bazel info bazel-testlogs)/test/coverage_scalatest_resources/consumer/tests/coverage.dat
 }
 
--- a/scala/private/phases/phase_coverage.bzl
+++ b/scala/private/phases/phase_coverage.bzl
@@ -26,10 +26,20 @@ def _phase_coverage_default(ctx, p, _args = struct()):
     )
 
 def _phase_coverage(ctx, p, srcjars):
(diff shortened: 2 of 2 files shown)


fmeum:
@omerlh

test/shell/test_coverage_scalatest_resources.sh
```diff
@@ -5,7 +5,9 @@ dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 runner=$(get_test_runner "${1:-local}")
 
 test_coverage_succeeds_resource_call() {
-    bazel coverage //test/coverage_scalatest_resources/consumer:tests
+    bazel coverage \
```
simuons, line 8: @fmeum thanks for contribution! Why you had to change this test? Could add new one with `--instrumentation_filter` and leave this one as is. As I understand coverage should work without `--instrumentation_filter` as well.

simuons, line 8: @fmeum thanks for contribution! Why you had to change this test? Could add new one with `--instrumentation_filter` and leave this one as is. As I understand coverage should work without `--instrumentation_filter` as well.

fmeum: It generally does, but not in this situation: Bazel auto-detects the `--instrumentation_filter` as `^//test/coverage_scalatest_resources/consumer[:/]`, which is not sufficient to pass this test as instrumenting `//test/coverage_scalatest_resources/resource` is required to do so. 

This could also happen to users of this ruleset and is thus a mild version of a breaking change, but then `--instrumentation_filter` is only meant to be a best-effort guess in the first place. I can add information about this to the README or some other place in the docs if you prefer.

simuons: Ah wasn't aware of bazel's `auto-detect`. Sorry for the noise.

simuons (approved):

test/shell/test_coverage_scalatest_resources.sh
```diff
@@ -5,7 +5,9 @@ dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
...
+          --instrumentation_filter=^//test/coverage_scalatest_resources[:/] \
```
liucijus, line 9: I guess this is an example why `--instrumentation_filter` was not checked in the previous implementation, but I think let's have it as it is consistent with how Bazel coverage works

liucijus (approved):

liucijus:
Thanks, @fmeum!

PR #1314, opened 2021-11-04
Replace output_jar with output_sources_jar in java_common.pack_sources

comius:
### Description
This updates call to `java_common.pack_sources` according to https://github.com/bazelbuild/bazel/issues/12373

### Motivation
Support latest Bazel version.

PR #1378, opened 2022-04-04
Single-jar pack_sources should conform to Bazel behavior

wiwa:
### Description

Fix `_pack_source_jar` in `phase_compile.bzl` to return single input source jars if there are no source files and only one source jar.
This conforms to the expected behavior of `java_common.pack_sources` (at least as of Bazel 5.0.0).

The previous PR #1314 did not take into account the behavior of `pack_sources` with only one input jar.
The field `output_source_jar` is actually checked in `pack_sources`, which will only
output the input jar directly *if* `output_source_jar` is null.
We also remove any additional `.` in `output_source_jar`s, i.e. if they look like `my-jars.-src.jar`,
our change will instead name them `my-jars-src.jar`.

### Motivation

We have a macro which exposes jars with `.srcjar` extensions. These are in fact the input jars to `-src.jar` (bundled) output jars. However, we expect that singular input jars will maintain the `.srcjar` extension and *not* be bundled. 
This is the previous behavior of `rules_scala`, and we would like to maintain this for our codebase,
since it is also the behavior of Bazel's `java_common.pack_sources`.

It should be noted that #1314 does not correctly remove the entire extension, which we fix here.
i.e.: `my-jars.extension` will become `my-jars.-src.jar.`. Presumably, we should be naming it `my-jars-src.jar`, without the extra `.` before the `-src.jar` suffix. This is likely due to an off-by-one error in the code:
`source_jar_name = output_jar.basename[:-len(output_jar.extension)] + "-src.jar"`
Thus, we remove any trailing `.`s before appending `-src.jar`.


scala/private/phases/phase_compile.bzl
```diff
@@ -309,16 +309,26 @@ def _create_scala_compilation_provider(ctx, ijar, source_jar, deps_providers):
     )
 
 def _pack_source_jar(ctx, scala_srcs, in_srcjars):
```
wisechengyi, line 311: was confused a bit at the beginning as `in_srcjars` implies a bool, but looks like it's meant to be a list of src jars.

so nit possibly to rename it `scala_src_jars`. feel free to do so as a follow up.

wisechengyi (approved):
Thanks, Win!

scala/private/phases/phase_compile.bzl
wiwa: Agreed, though we should rename to `input_srcjars` imo (as followup).

wisechengyi: sgtm. thanks!

wiwa: Too lazy for followup. Renamed it and also changed the `.-src.jar` naming into `-src.jar`.

test/src/main/scala/scalarules/test/srcjars/pack_sources_test.bzl
```diff
@@ -0,0 +1,91 @@
...
+
+    target_under_test = analysistest.target_under_test(env)
+
+    srcjar_names = sets.make(
+        [j.basename for j in target_under_test[JavaInfo].source_jars],
+    )
+
+    # In line with Bazel's java_common.pack_sources,
+    # We return the initial .srcjar file since there are no source files
+    # Not sure where the second -src.jar comes from, maybe due to java rules
```
simuons, line 17: Could you check if adding `expect_java_output = False` to `:source_jar` target makes this go away?

scala/private/phases/phase_compile.bzl
```diff
@@ -308,17 +308,32 @@ def _create_scala_compilation_provider(ctx, ijar, source_jar, deps_providers):
...
+    # if so, it checks that output_source_jar is null
+    # passing that, it will return the input source jar directly
+    # However, pack_sources will FAIL if both output_source_jar and
+    # the deprecated output_jar field are BOTH null
+    # Therefore, we can return the single input jar ourselves
+    if not scala_srcs and len(input_srcjars) == 1:
+        return input_srcjars[0]
+    else:
+        output_jar = ctx.outputs.jar
+        without_ext = output_jar.basename[:-len(output_jar.extension)]
```
simuons, line 323: Not sure if this would make code look better and probably it's a matter of taste but `without_ext` could be calculated as `output_jar.basename[:-len(output_jar.extension) - 1] if output_jar.extension else output_jar.basename`
Please feel free to ignore this comment.

simuons (approved):
Thanks @wiwa . LGTM. Just a couple of minors comments which can be ignored.

scala/private/phases/phase_compile.bzl
wiwa: Good suggestion! I ended up with (ironically) a longer implementation. Your suggestion made me realize that we shouldn't expect that output_jar *has* an extension. Personally, I think that the ternary `if` is too long for one line, so a mutating `if` is fine.

test/src/main/scala/scalarules/test/srcjars/pack_sources_test.bzl
wiwa: It does! Thanks for the test case. I added one more to test multiple source jars, adding `SourceJar2.srcjar` to the test resources. Also refactored the test logic since it was insanely repetitive.

wisechengyi (approved):

liucijus (approved):

simuons (approved):
Thanks @wiwa

PR #1381, opened 2022-04-07
Update specs2 versions to 4.15.0 which has build for Scala 3

liucijus:
This change is part of Scala 3 support work.

5.0.0-RC-21 version used in Scala 3 repositories is not compatible with many features we use in Rules Scala tests.

Specs2 4.15.0 was released which has a build for Scala 3. Updating repositories for 2.12, 2.13 and 3.1 to this version. There is no release for 2.11 anymore.


liucijus:
There is a regression in test filtering since 4.12.2, closing.

PR #1384, opened 2022-04-12
Allow `scala_junit_tests` targets to specify test environment variables

wiwa:
### Description

Add the equivalent of `java_test`'s `env` attr to `scala_junit_tests` via `TestEnvironment` provider.
Allows targets to set their own environment variables during `bazel test` runs.

https://bazel.build/reference/be/common-definitions#test.env

### Motivation

We would like to test with environment variables different from the ones defaulted to by a `bazel test` invocation, on a target-specified basis. This also brings our test rules further in line with the functionality offered by native test rules.

### DIFF
--- a/scala/private/phases/phase_test_environment.bzl
+++ b/scala/private/phases/phase_test_environment.bzl
@@ -2,7 +2,7 @@
 # https://bazel.build/rules/lib/testing#TestEnvironment
 
 def phase_test_environment(ctx, p):
-    test_env = getattr(ctx.attr, "env", None)
+    test_env = ctx.attr.env
 
     if test_env:
         return struct(
--- a/scala/private/phases/phase_test_environment.bzl
+++ b/scala/private/phases/phase_test_environment.bzl
@@ -2,14 +2,13 @@
 # https://bazel.build/rules/lib/testing#TestEnvironment
 
 def phase_test_environment(ctx, p):
-
     test_env = ctx.attr.env if "env" in dir(ctx.attr) else None
 
     if test_env:
         return struct(
-          external_providers = {
-            "TestingEnvironment": testing.TestEnvironment(test_env)
-          }
+            external_providers = {
+                "TestingEnvironment": testing.TestEnvironment(test_env),
+            },
         )
 
     return struct()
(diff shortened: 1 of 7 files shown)


test/BUILD
```diff
@@ -777,6 +777,43 @@ scala_junit_test(
...
+)
+
+scala_junit_test(
+    name = "JunitNoTestEnvironment_test_runner",
+    size = "small",
+    runtime_jdk = "@bazel_tools//tools/jdk:remote_jdk11",
+    suffixes = ["Test"],
+    tests_from = [":JunitNoTestEnvironment"],
+    runtime_deps = [":JunitNoTestEnvironment"],
+)
```
wisechengyi, line 803: I wonder if this can be a `scala_junit_test` that directly declares `srcs` on `src/main/scala/scalarules/test/junit/JunitNoTestEnvironmentTest.scala`. ditto for the other, so the BUILD file can be simplified a little.

wisechengyi (approved):
Thanks, Win! This is awesome!

test/BUILD
wiwa: Good idea; don't want to get into it in this commit, but we can track it as an issue?

wisechengyi: It should be able to do that already - https://github.com/bazelbuild/rules_scala/blob/b85d1225d0ddc9c376963eb0be86d9d546f25a4a/test_expect_failure/scala_junit_test/BUILD#L10-L14, so should be minimal change?

wiwa: Nice catch! Simplified it.

wisechengyi:
@simuons @liucijus (no rush) would appreciate a look when you get a chance. Thank you!

liucijus (changes requested):
LGTM, please address my comment

scala/private/phases/phase_test_environment.bzl
```diff
@@ -0,0 +1,14 @@
+# Adds testing.TestEnvironment provider if "env" attr is specified
+# https://bazel.build/rules/lib/testing#TestEnvironment
+
+def phase_test_environment(ctx, p):
+    test_env = ctx.attr.env if "env" in dir(ctx.attr) else None
```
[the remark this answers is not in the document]
wiwa: `getattr` is awesome, thanks!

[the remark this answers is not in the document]
wiwa: Actually -- turns out I can directly assign it! Not sure why I thought I couldn't.

liucijus (approved):

wiwa:
@simuons is this mergeable?

simuons (approved):

PR #1385, opened 2022-04-16
Consistent argsfile parsing

wiwa:
### Description

`Worker`s should parse argsfiles/flagfiles regardless of being persistent or ephemeral.  
This doesn't change the behavior of only allowing argsfiles when called with a single argument.

### Motivation

Our functionality will be more in line with Bazel's `JavaBuilder`.  
This is also helpful for persistent workers outside of Bazel itself (e.g. on remote executors).  
Instead of expanding argsfiles, remote executors can pass in the argsfile as if it were an ephemeral invocation.

### DIFF
--- a/src/java/io/bazel/rulesscala/worker/Worker.java
+++ b/src/java/io/bazel/rulesscala/worker/Worker.java
@@ -81,7 +81,14 @@ public final class Worker {
           int code = 0;
 
           try {
-            workerInterface.work(stringListToArray(request.getArgumentsList()));
+            String[] workerArgs = stringListToArray(request.getArgumentsList());
+            String[] args;
+            if (workerArgs.length == 1) {
+              args = expandArgsfile(workerArgs[0]);
+            } else {
+              args = workerArgs;
+            }
+            workerInterface.work(args);
           } catch (ExitTrapped e) {
             code = e.code;
           } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


wiwa:
@liucijus @simuons ready to be reviewed, thanks!

src/java/io/bazel/rulesscala/worker/Worker.java
```diff
@@ -81,7 +81,14 @@ public void checkPermission(Permission permission) {
           int code = 0;
 
           try {
-            workerInterface.work(stringListToArray(request.getArgumentsList()));
+            String[] workerArgs = stringListToArray(request.getArgumentsList());
+            String[] args;
+            if (workerArgs.length == 1) {
```
johnynek, line 86: can we not unify this part as well? Seems it is duplicated below.

liucijus (changes requested):
Thanks, @wiwa! 

Please address duplication mentioned by @johnynek - extracting a method should do a trick. 

Other than this nit, LGTM!

src/java/io/bazel/rulesscala/worker/Worker.java
wiwa: Thanks for the catch!

johnynek (approved):

simuons (approved):

liucijus (approved):

PR #1388, opened 2022-04-28
Revert "Create jdeps output from classpath entries (#1376)"

liucijus:
Reverting jdeps implementation as it currently does more harm, than is useful. Intellij attach sources requires full jars and won't work with ijars for Scala code.

This reverts commit 3dd5d8110d56cfc19722532866cbfc039a6a9612.


wisechengyi (approved):

ianoc:
@liucijus When you mean does more harm than good, could you elaborate? in the sense that if we don't use ijar with a 3rdparty dep, would this solve the intellij navigate to source? What is the downside of keeping it?

liucijus:
It's a problem for 2nd party deps (deps which are source dependency, but not in Intellij project view). `scala_import` does not produce interface jar, so it's not a problem for 3rd party dep if its loader uses `scala_import` or just gives full jars.

Interface jars don't work with Scala attach sources (Intellij Scala plugin [issue](https://youtrack.jetbrains.com/issue/SCL-16975/Navigation-to-attached-Scala-sources-for-interface-jar-does-not-)). In Intellij `jdeps` is supported by Java plugin, and there is also Scala plugin which discovers deps from Intellij aspect. So, because of jdeps support, deps are attached multiple times during sync and Intellij indexing can get longer. In our (Wix) case it attaches interface jars next to full jars, because we have customized the Intellij aspect to return full jars for the sync to support sources attachment.

Also, having full jars is not enough to make source attachment to work in Intellij. Currently Bazel plugin does not look into language specific sync data - we did some changes to get it working for us in our internal plugin.

I'm still considering returning jdeps support later, but before I do it, I want to see if I can upstream some changes to Intellij plugin.

liucijus:
> Is this fully abandoned? What should be done in order to do it right then?

I think the main stopper is Intellij plugin. As such change will break attachment of jars, because jdeps contains links to ijars instead of full jars

liucijus:
> Or ijars should not strip SourceFile attribute, is it correct?

Did not test this, but could be the missing thing according to [the ticket](https://youtrack.jetbrains.com/issue/SCL-16975/Navigation-to-attached-Scala-sources-for-interface-jar-does-not-work#focus=Comments-27-3935151.0-0)

johnynek:
Side note: I'm actually not sure how much ijars are helping us with scala.

Ijar keeps the scala signature, but that could change enough to make the ijar almost as likely to invalidate cache as the jar.

I know scala puts the method size in the attributes for inlining information. Many (almost all?) changes to code will change the size of something.

It would be interesting to have some benchmark set of diffs and compare build after each diff applied using ijar vs the original jar and see what kind of win ijar is giving with scala.

Perhaps we could even find ways to make it more effective with such a benchmark suite.

PR #1389, opened 2022-05-03
Use a public java_compile_toolchain attr instead of _java_toolchain

wiwa:
### Description

Allow per-target selection of Java compilation toolchain.

Note: This change doesn't add similar functionality to `thrift_library`.

### Motivation

Helps with JDK migrations by allowing targets to incrementally update.  
Also allows targets to opt into newer Java syntax.

Co-authored-by: Shane Delmore <sdelmore@twitter.com>

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -485,13 +485,13 @@ scala_binary(
 # Make sure scala_library respects java_compile_toolchain during builds
 scala_junit_test(
     name = "CheckBytecodeMajorVersion",
+    srcs = ["CheckBytecodeMajorVersionTest.scala"],
+    suffixes = ["Test"],
     runtime_deps = [
-        "//test/src/main/resources/java_sources:CompiledWithJava8",
         "//test/src/main/resources/java_sources:CompiledWithJava11",
+        "//test/src/main/resources/java_sources:CompiledWithJava8",
     ],
     deps = ["@io_bazel_rules_scala_junit_junit"],
-    suffixes = ["Test"],
-    srcs = ["CheckBytecodeMajorVersionTest.scala"],
 )
 
 # Generate a file containing the rootpaths of a Scala binary.
--- a/test/BUILD
+++ b/test/BUILD
@@ -482,6 +482,18 @@ scala_binary(
     runtime_jdk = "@bazel_tools//tools/jdk:remote_jdk11",
 )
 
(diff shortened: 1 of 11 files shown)


test/BUILD
```diff
@@ -482,6 +482,18 @@ scala_binary(
     runtime_jdk = "@bazel_tools//tools/jdk:remote_jdk11",
 )
 
+# Make sure scala_library respects java_compile_toolchain during builds
+scala_junit_test(
+    name = "CheckBytecodeMajorVersion",
+    srcs = ["CheckBytecodeMajorVersionTest.scala"],
+    suffixes = ["Test"],
```
wisechengyi, line 489: for my learning, is `suffixes` significant?

wisechengyi (approved):
Thanks, Shane and Win!

test/CheckBytecodeMajorVersionTest.scala
```diff
@@ -0,0 +1,48 @@
...
+
+    val expectJava11 =  getBytecodeMajorVersion(
+      getClass
+        .getClassLoader
+        .getResourceAsStream("java_sources/SimpleJavaSourceFileB.class")
+    )
+      
+    Assert.assertTrue(
+      s"Expected java 8 (major version 52) but got major version $expectJava8",
+      majorVersionToJdkVersion(expectJava8) == 8
```
liucijus, line 40: nit: here and below using `assertEquals` is more idiomatic in junit. Eg.:
```
assertEquals(
  "Must be Java 8 (major version 52) bytecode", 
  8,
  majorVersionToJdkVersion(expectJava8)
)
```

test/BUILD
wiwa: Yes, this is necessary for `scala_junit_test` to discover test classes. Specifically one of `prefixes` or `suffixes` is needed.

test/CheckBytecodeMajorVersionTest.scala
wiwa: changed, thanks!

test/BUILD
wiwa: there isn't a default so we must manually specify 😞

simuons (approved):

liucijus (approved):

liucijus:
Thanks, @wiwa!

PR #1390, opened 2022-05-06
scala_doc: Use param file to build scaladocs

rahuliyer95:
### Description

Generate `scaladoc` by providing arguments via param file.

### Motivation

When we have a large set of dependencies or source files, `scaladoc` generation fails with the following error,
```
error=7, Argument list too long
``` 
Scaladoc supports reading arguments from an argfile which can help avoid the above problem,
```sh
$ scaladoc -help
...
  @<file>                              A text file containing compiler arguments (options and source files)
```
We can do this by using [param file](https://bazel.build/rules/lib/Args#use_param_file).

### DIFF
--- a/scala/private/rules/scala_doc.bzl
+++ b/scala/private/rules/scala_doc.bzl
@@ -83,6 +83,8 @@ def _scala_doc_impl(ctx):
     # Construct scaladoc args, which also include scalac args.
     # See `scaladoc -help` for more information.
     args = ctx.actions.args()
+    args.set_param_file_format("multiline")
+    args.use_param_file(param_file_arg = "@%s", use_always = True)
     args.add("-usejavacp")
     args.add("-nowarn")  # turn off warnings for now since they can obscure actual errors for large scala_doc targets
     args.add_all(ctx.attr.scalacopts)


scala/private/rules/scala_doc.bzl
```diff
@@ -82,18 +82,27 @@ def _scala_doc_impl(ctx):
...
+    )
     args = ctx.actions.args()
-    args.add("-usejavacp")
-    args.add("-nowarn")  # turn off warnings for now since they can obscure actual errors for large scala_doc targets
-    args.add_all(ctx.attr.scalacopts)
-    args.add("-d", output_path.path)
-    args.add_all(plugins, format_each = "-Xplugin:%s")
-    args.add_joined("-classpath", classpath, join_with = ctx.configuration.host_path_separator)
-    args.add_all(src_files)
+    args.add("@{}".format(argfile.path))
```
simuons, line 101: It think there should be enough just to add the following to the args
```
args.set_param_file_format("multiline")
args.use_param_file(param_file_arg = "@%s", use_always = True)
```
and there is no need to declare a file and write content to it. Bazel should write args file automatically when it's passed as an argument to the action.

Example is here https://github.com/bazelbuild/rules_scala/blob/master/scala/private/rule_impls.bzl#L74-L75

simuons (changes requested):
Hi @rahuliyer95 thanks for PR. Please take a look at my comment.

scala/private/rules/scala_doc.bzl
rahuliyer95: Thanks for the help, makes life much easier!

simuons (approved):

liucijus (approved):

liucijus:
Thanks, @rahuliyer95!

PR #1392, opened 2022-05-13
add package_name

exp0nge:
### Description

Support `package_name` option https://scalapb.github.io/docs/customizations/

```
package_name sets the Scala base package name, if this is not defined, then it falls back to the java_package option. If the java_package option is also not the found, then the package name from file's package statement is used.
```

### Motivation

Classpath collisions... need to rename the generated classes

### DIFF
--- a/scala_proto/scala_proto_toolchain.bzl
+++ b/scala_proto/scala_proto_toolchain.bzl
@@ -18,6 +18,8 @@ def _generators_opts(ctx):
         opts.append("grpc")
     if ctx.attr.with_flat_package:
         opts.append("flat_package")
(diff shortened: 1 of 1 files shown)


johnynek:
can you add a test for this?

simuons:
Hi @exp0nge,

as I understand this `package_name` is a [file-level-option](https://scalapb.github.io/docs/customizations/#file-level-options) which must be defined in a proto file. Toolchain is used to set [additional options to the generator](https://scalapb.github.io/docs/sbt-settings#additional-options-to-the-generator) and there there is no such options as `package_name`.

exp0nge:
😞

PR #1393, opened 2022-05-13
scala_doc: Include transitive runtime jars in classpath

rahuliyer95:
### Description

Some of our targets require additional jars on the classpath for scaladoc generation to be successful. This seems to be solved by including the `transitive_runtime_jars` on the classpath.

### Motivation

Fix compilation of some our targets which require additional jars on the classpath.

### DIFF
--- a/scala/private/rules/scala_doc.bzl
+++ b/scala/private/rules/scala_doc.bzl
@@ -34,6 +34,7 @@ def _scaladoc_aspect_impl(target, ctx, transitive = True):
             direct = [file for file in ctx.rule.files.deps],
             transitive = (
                 [dep[JavaInfo].compile_jars for dep in ctx.rule.attr.deps if JavaInfo in dep] +
+                [dep[JavaInfo].transitive_runtime_jars for dep in ctx.rule.attr.deps if JavaInfo in dep] +
                 [dep[_ScaladocAspectInfo].compile_jars for dep in ctx.rule.attr.deps if _ScaladocAspectInfo in dep]
             ),
         )
--- a/scala/private/rules/scala_doc.bzl
+++ b/scala/private/rules/scala_doc.bzl
@@ -10,9 +10,9 @@ _ScaladocAspectInfo = provider(fields = [
 
 def _scaladoc_intransitive_aspect_impl(target, ctx):
     """Build scaladocs only for the provided targets."""
-    return _scaladoc_aspect_impl(target, ctx, transitive = False)
+    return _scaladoc_aspect_impl(target, ctx, transitive = False, use_transitive_compile_jars = True)
 
-def _scaladoc_aspect_impl(target, ctx, transitive = True):
+def _scaladoc_aspect_impl(target, ctx, transitive = True, use_transitive_compile_jars = False):
     """Collect source files and compile_jars from JavaInfo-returning deps."""
 
     # We really only care about visited targets with srcs, so only look at those.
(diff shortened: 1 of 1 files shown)


simuons:
Hi @rahuliyer95 could you add a test for this? Also I'm wondering if `transitive_runtime_jars` is a superset of `compile_jars` if so then there is no reason to keep them both.

rahuliyer95:
> Hi @rahuliyer95 could you add a test for this? Also I'm wondering if `transitive_runtime_jars` is a superset of `compile_jars` if so then there is no reason to keep them both.

@simuons Looks like `transitive_runtime_jars` contains all the jars needed. I tested against some random targets in our internal repo and they seem to be working. Also, the `ScalaDoc` test is passing.

rahuliyer95:
Gentle ping for review @simuons @liucijus

johnynek:
could you try:

https://docs.bazel.build/versions/main/skylark/lib/JavaInfo.html#transitive_compile_time_jars

`transitive_compile_time_jars`?

The runtime jars are going to change on pretty much every compile, so if any dependency changes at all, it will invalidate everything downstream. By using compile jars, you are getting interface jars that will at least sometimes not change when the code changes, thus offering much better cache performance.

transitive_runtime_jars kind of has to work because it is all the jars that will be there at runtime, but I can't see why we would need that and you are harming caching for every user by making this change so I think we should definitely investigate other options and understand why it is that we need to do it.

+1 to finding an example test that shows the need for the change.

rahuliyer95:
`transitive_compile_time_jars` seems to be working for us. It's really hard to replicate this scenario in this repository as a test. The main reason being we have some custom rules like `target_union` which is an extension of [alias](https://bazel.build/reference/be/general#alias) supporting multiple dependencies and more. I believe the problem is happening because of multiple redirects happening via these custom rules across java/scala libraries. I tried a simple test with `alias` locally and it seems to be passing, so really out of ideas here on how to replicate the issue and add a test.

johnynek:
Are you sure this isn't something about your internal rules?

I wouldn't want to possibly slow down builds to work around a bug in someone's internal rules.

I wonder if an alternate approach would be to make a configuration on the scala toolchain, and then make the default as it is now, but allow pulling the transitive compile path as well.

rahuliyer95:
Thanks everyone for working with me on this. I like the idea of the alternative approach, just to make sure I understand it correctly, the idea is to extend the `_scaladoc_aspect_impl` to allow configuring the use of `transitive_compile_time_jars` by keeping the default to `compile_jars`. Something like
```py
def _scaladoc_aspect_impl(target, ctx, transitive = True, include_transitive_compile_time_jars = False):
  ...
```

rahuliyer95:
Does the newer implementation work for folks? @simuons @liucijus

rahuliyer95:
Another option would be to export `ScalaDocAspectInfo` and we write our own aspect implementation and use it with `make_scaladoc_rule` and not disturb the implementation here.

liucijus (approved):
I won't block this one, but I think we really need figure out if scala docs is more greedy for classpath entries than scalac. Ideally that should be demonstrated with a failing test.

rahuliyer95:
I am going to close this as I am not able to create a test. This seems to be very specific to our repository and setup, so I will create a separate PR to expose `ScalaDocAspectInfo` and write an implementation for our setup internally making use of the `make_scaladoc_rule`. Thanks for your time and help with this.

PR #1394, opened 2022-05-16
Expose `_source_jars` output group in `scala_import`

Duhemm:
### Description
Previously, `scala_import` rules would have no `OutputGroupInfo`, and
therefore would not expose their source jars. Java's counterpart,
`java_import`, does expose the source jars via an entry named
`_source_jars` in `OutputGroupInfo`.

This patch adds an `OutputGroupInfo` provider to `scala_import` rules
containing the source jars in an entry named `_source_jars`.

The source jars of a `scala_import` rule can now be requested by
passing `--output_groups=+_source_jars` to Bazel.

### Motivation
Querying Bazel for the source jars can be useful in a variety of situations, like IDE integration for instance.


liucijus (approved):
Thanks, @Duhemm!

simuons (approved):

PR #1397, opened 2022-05-25
Fix rules_scala JDK lookup when using --nolegacy_external_runfiles.

### LINKED ISSUE #1396: Regression for setting the JDK path when specifying --nolegacy_external_runfiles
Hi.

https://github.com/bazelbuild/rules_scala/pull/1373 seems to have broken resolving the JDK when --nolegacy_external_runfiles is specified. 

To reproduce, do the following:

* Run `bazel build test:ScalaBinaryInGenrule --nolegacy_external_runfiles` in this repository.

This will produce the following error

> /private/var/tmp/_bazel_sdoessing/95e60638f59d666df186023b663ca267/sandbox/darwin-sandbox/12/execroot/io_bazel_rules_scala/bazel-out/host/bin/test/ScalaBinary.runfiles/io_bazel_rules_scala/test/ScalaBinary_wrapper.sh: line 5: /private/var/tmp/_bazel_sdoessing/95e60638f59d666df186023b663ca267/sandbox/darwin-sandbox/12/execroot/io_bazel_rules_scala/bazel-out/host/bin/test/ScalaBinary.runfiles/io_bazel_rules_scala/external/remotejdk11_macos/bin/java

The root cause appears to be the switch from `java_executable_runfiles_path` to `java_executable_exec_path`. 

https://github.com/bazelbuild/rules_scala/pull/1373/files#diff-93df4d85f89773937a2c4a84d11313c329faa67e0388b87cc00b3c84fffe93a5

It is not clear to me from the Bazel documentation what the difference between these two variables is, but it seems like an issue like this has been fixed previously 

https://stackoverflow.com/a/49513476

Switching back to `java_executable_runfiles_path` appears to fix the issue, but I don't really have the context to know if the change was intentional.

I'm happy to open a PR to fix this, if this change sounds reasonable. Maybe the CI bazelrc should be updated to set `--nolegacy_external_runfiles ` as well?

srdo-humio:
java_executable_exec_path does not look up the right path when resolving
the JDK from the runfiles directory and legacy_external_runfiles is disabled.
java_executable_runfiles_path should be used instead.

Examples of these paths when running "bazel build test:ScalaBinaryInGenrule":

java_executable_exec_path: external/remotejdk11_macos/bin/java
java_executable_runfiles_path: ../remotejdk11_macos/bin/java

The working directory appears to be
ScalaBinary.runfiles/io_bazel_rules_scala.

When legacy_external_runfiles is enabled, the JDK is present in
ScalaBinary.runfiles/remotejdk11_macos but is also linked
in ScalaBinary.runfiles/io_bazel_rules_scala/external/remotejdk11_macos,
so looking up using either exec_path or runfiles_path is correct.

When legacy_external_runfiles is disabled, the JDK only goes in
ScalaBinary.runfiles/remotejdk11_macos, so only runfiles_path will work.

Fixes #1396

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -191,7 +191,7 @@ def java_bin(ctx):
         default_runtime = ctx.attr._java_runtime[java_common.JavaRuntimeInfo],
     )
 
-    java_path = str(java_runtime.java_executable_exec_path)
+    java_path = str(java_runtime.java_executable_runfiles_path)
 
     if paths.is_absolute(java_path):
         javabin = java_path
--- a/test_rules_scala.sh
+++ b/test_rules_scala.sh
@@ -24,7 +24,7 @@ $runner bazel test "--extra_toolchains=//scala:minimal_direct_source_deps -- tes
 $runner bazel build "test_expect_failure/missing_direct_deps/internal_deps/... --strict_java_deps=warn --extra_toolchains=//test/toolchains:high_level_transitive_deps_strict_deps_warn"
 $runner bazel build //test_expect_failure/proto_source_root/... --strict_proto_deps=off
 $runner bazel test //test/... --extra_toolchains="//test_expect_failure/plus_one_deps:plus_one_deps"
-$runner build test:ScalaBinaryInGenrule --nolegacy_external_runfiles
+$runner bazel build test:ScalaBinaryInGenrule --nolegacy_external_runfiles
 $runner bazel build //test_statsfile:Simple_statsfile
 $runner bazel build //test_statsfile:SimpleNoStatsFile_statsfile --extra_toolchains="//test/toolchains:enable_stats_file_disabled_toolchain"
 . "${test_dir}"/test_build_event_protocol.sh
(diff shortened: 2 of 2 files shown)


srdo-humio:
As a followup to this, I think it would be good to ensure that --nolegacy_external_runfiles is set when running tests.

liucijus:
@wiwa, @wisechengyi could you take a look at this one?

liucijus:
@srdo-humio, thanks for the contribution, would you be able to contribute a test for this change? Or do you think we should set `--nolegacy_external_runfiles` globally?

srdo-humio:
@liucijus It is my impression that --nolegacy_external_runfiles is supposed to become the default at some point, setting it globally might make sense?

If you would rather not set it globally, I think running (some of) the tests in both modes should be good enough to check that this doesn't regress. Let me know if you'd prefer handling it this way, and I'll add the extra check to test_rules_scala.sh.

Also apologies for the CLA check being red, I'm working on getting it signed internally.

pcj:
Also hit this while (still attempting) to upgrade rules_scala to latest.  In this repo, the rules_scala version was very old (circa 2020), but somehow was working with bazel 5.1.1.  After attempting to upgrade rules_scala to latest, hit this issue.

Confirming that this change fixed the issue:

```diff
-    java_path = str(java_runtime.java_executable_exec_path)
+    java_path = str(java_runtime.java_executable_runfiles_path)
```

In our case, the presence/absence of `--nolegacy_external_runfiles` did not seem to matter.

liucijus:
@srdo-humio will you sign CLA?

simuons (approved):
Hi @srdo-humio, thanks for contribution and sorry for delay.

To my understanding your change is correct. Since this `javabin` path is embedded into wrapper script it should point to runfiles.

Did a quick check with `java_binary` and it points to `runfiles` without `external` part.

If you could just add a single test for `bazel build test:ScalaBinaryInGenrule --nolegacy_external_runfiles` to catch such regression that would be awesome.

srdo-humio:
I've added the test. I'm still working on getting the corporate CLA signed, best guess is we'll get it done in the next few days.

srdo-humio:
Sorry for the delay, the CLA check is green now, so this might be ready to merge :)

liucijus (approved):

PR #1399, opened 2022-06-01
JacocoRunner script: update for Bazel 5.0+

gergelyfabian:
### Description

JacocoRunner builder script needs an update for Bazel 5.0+.

Upgraded jacoco package name for Bazel 5.0+.

Upgraded jacoco from 0.8.3 to 0.8.6.

Jacoco upgrade to 0.8.6 was made in Bazel in:
https://github.com/bazelbuild/bazel/issues/11674
https://github.com/bazelbuild/bazel/commit/cb7c1a2ec8de92828a018e2a3450a06f1d1c7efa

Removed options for workarounds where those have been fixed in the meantime:

Bazel's handling for branch coverage was fixed in:
https://github.com/bazelbuild/bazel/issues/12696

### Motivation

If you use a custom-built Jacoco version for Bazel <5.0, then when you upgrade Bazel to 5.0+ you'll need to rebuild that custom Jacoco version, otherwise your coverage executions would fail with an error similar to the following:

```
Cause: java.lang.ClassNotFoundException: org.jacoco.agent.rt.internal_b864216.Offline
```

### DIFF
--- a/scripts/build_jacocorunner/0001-Build-Jacoco-for-Bazel-5.0+.patch
+++ b/scripts/build_jacocorunner/0001-Build-Jacoco-for-Bazel-5.0+.patch
@@ -1,0 +1,25 @@
+From 712d62a8238f3a7fe51e1cf4cc2520b5f249e1d3 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Gergely=20F=C3=A1bi=C3=A1n?= <gergo.fb@gmail.com>
+Date: Fri, 18 Dec 2020 11:43:59 +0100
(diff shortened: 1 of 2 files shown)


gergelyfabian:
Added a change to have both versions of the script, pre-Bazel-5.0 and post-Bazel-5.0.

liucijus (approved):
Thanks, @gergelyfabian!

gergelyfabian:
Updated Bazel's tag to avoid an error reported in https://github.com/bazelbuild/rules_scala/issues/1400.

PR #1401, opened 2022-06-01
feat: added support for inherited_environment in scala_test rule

omerlh:
### Description
Added support for `inherited_environment` in `scala_test` rule

Follow up on this [slack thread](https://bazelbuild.slack.com/archives/CDCKJ2KFZ/p1653986701998679) I added support for the new `inherited_environment`  to `scala_test`.
This feature requires Bazel version 5.2 rc0 or higher.

### Motivation
While Bazel run tests in isolated environments, when running integration tests there is a need to pass env vars to the tested code - for example, credentials for external systems like databases. This PR added a new flag that is already supported by Bazel.

Note: I am a total noob. I got a lot of help on Slack in order to get to this point but I am not 100% understand everything I changed 🤦  I added a test so I know it is working, I will be helping to have some guidelines on how to run it only on Bazel 5.2:
* Passing a flag to `test_rules_scala` and let it know it should run this test?
* Add another shell file running only my test?
* Other option?

Also - where and how should I document it?

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -62,6 +62,18 @@ tasks:
     bazel: 4.1.0
     shell_commands:
       - "./test_coverage.sh"
+  test_bazel_52_linux:
+    name: "./test_bazel_5_2"
+    platform: ubuntu2004
+    bazel: 5.2.0rc2
+    shell_commands:
+      - "./test_bazel_5_2.sh"
+  test_bazel_52_macos:
+    name: "./test_bazel_5_2"
+    platform: macos
+    bazel: 5.2.0rc2
+    shell_commands:
+      - "./test_bazel_5_2.sh "
   test_reproducibility_linux:
     name: "./test_reproducibility.sh"
     platform: ubuntu1804
--- a/scala/private/phases/phase_test_environment.bzl
+++ b/scala/private/phases/phase_test_environment.bzl
@@ -3,12 +3,36 @@
 
 def phase_test_environment(ctx, p):
     test_env = ctx.attr.env
+    inherited_environment = ctx.attr.env_inherit
 
-    if test_env:
+    if inherited_environment and test_env:
         return struct(
             external_providers = {
-                "TestingEnvironment": testing.TestEnvironment(test_env),
+                "TestingEnvironment": testing.TestEnvironment(
+                    test_env,
+                    inherited_environment,
+                ),
             },
         )
 
-    return struct()
+    elif test_env:
+        return struct(
+            external_providers = {
+                "TestingEnvironment": testing.TestEnvironment(
+                    test_env,
+                ),
+            },
+        )
+
+    elif inherited_environment:
+        return struct(
+            external_providers = {
+                "TestingEnvironment": testing.TestEnvironment(
+                    {},
+                    inherited_environment,
+                ),
+            },
+        )
+
+    else:
+        return struct()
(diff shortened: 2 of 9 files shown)


scala/private/phases/phase_test_environment.bzl
```diff
@@ -1,14 +1,28 @@
...
 
+load("@bazel_skylib//lib:versions.bzl", "versions")
+
+
 def phase_test_environment(ctx, p):
     test_env = ctx.attr.env
+    inherited_environment = ctx.attr.inherited_environment
 
-    if test_env:
+    if inherited_environment:
```
omerlh, line 11: Note for reviewer: I removed the original if as `test_env` has default valued of `{}` so it is always set. Unless I am wrong, OFC :)

scala/private/rules/scala_test.bzl
```diff
@@ -75,6 +77,8 @@ _scala_test_attrs = {
     "_lcov_merger": attr.label(
         default = Label("@bazel_tools//tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator:Main"),
     ),
+    "env": attr.string_dict(default = {}),
+    "inherited_environment": attr.string_list(),
```
fmeum, line 81: Could be called `env_inherit` since it should behave just like the attr of that name on e.g. `java_test`.

scala/private/phases/phase_test_environment.bzl
```diff
@@ -1,14 +1,28 @@
...
         return struct(
             external_providers = {
-                "TestingEnvironment": testing.TestEnvironment(test_env),
+                "TestingEnvironment": testing.TestEnvironment(
+                        test_env,
+                        inherited_environment
```
fmeum, line 16: In order to retain partial compatibility with Bazel < 5.2.0, you could instantiate `TestingEnvironment` without `inherited_environment` if `env_inherit` has not been set by the user and set it in a separate branch.

fmeum (approved):

scala/private/phases/phase_test_environment.bzl
omerlh: Unless I am not following, I think this is what I just did?

fmeum: You are right, looks good!

scala/private/phases/phase_test_environment.bzl
fmeum: In Starlark, an empty dictionary evaluates to `False`: "A dictionary used in a Boolean context is considered true if it is non-empty."

scala/private/rules/scala_test.bzl
omerlh: Fixed, thanks!

scala/private/phases/phase_test_environment.bzl
omerlh: Ahhh right! Thanks :) This makes a lot more sense now - I assumed this is TypeScript, so it's true if it's not null...

.bazelci/presubmit.yml
```diff
@@ -62,6 +62,18 @@ tasks:
     bazel: 4.1.0
     shell_commands:
       - "./test_coverage.sh"
+  test_bazel_52_linux:
+    name: "./test_bazel_5_2"
+    platform: ubuntu2004
+    bazel: 5.2.0rc2
+    shell_commands:
+      - "./test_bazel_5_2.sh"
+  test_bazel_52_linux:
```
liucijus, line 71: Is this duplicated intentionally?

omerlh: OOps, one of them need to be macos - thanks!

liucijus (approved):
Thanks, @omerlh!

simuons (approved):
Thanks @omerlh and sorry for delay

PR #1403, opened 2022-06-03
phase_merge_jars: refactor to separate function that merges jars to o…

gergelyfabian:
…utput

### Description
Add a new function that is independent of runtime jars and the specific output.

### Motivation
This is a small refactoring as an intermediate step for possible extensions for creating
custom outputs. Not sure yet how to add those extensions, but such a refactoring here may be very useful.

An example how this can be changed further:

https://github.com/gergelyfabian/rules_scala/compare/refactor_phase_merge_jars...gergelyfabian:external-and-internal-deploy-jar?expand=1

### DIFF
--- a/scala/private/phases/phase_merge_jars.bzl
+++ b/scala/private/phases/phase_merge_jars.bzl
@@ -4,35 +4,38 @@
 # DOCUMENT THIS
 #
 
-def phase_merge_jars(ctx, p):
+def merge_jars_to_output(ctx, output, jars):
     """Calls Bazel's singlejar utility.
 
     For a full list of available command line options see:
     https://github.com/bazelbuild/bazel/blob/697d219526bffbecd29f29b402c9122ec5d9f2ee/src/java_tools/singlejar/java/com/google/devtools/build/singlejar/SingleJar.java#L337
     Use --compression to reduce size of deploy jars.
     """
-    deploy_jar = ctx.outputs.deploy_jar
-    runtime_jars = p.compile.rjars
     main_class = getattr(ctx.attr, "main_class", "")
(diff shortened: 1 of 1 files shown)


simuons (commented):
Thanks @gergelyfabian for the PR.

Have a separate function makes total sense. However this makes `merge_jars_to_output` essentially public (despite `private` word in its file path). This will become part of the api we will have to maintain. 

I would rather wait for your anticipated change and see from there if this is needed or not. Maybe other route will be taken.

Having said that I don't want to block this PR. I just want a second opinion on this. Maybe I'm totally wrong and missing something.

Maybe this needs to be moved into separate package like starlark-wrappers-for-bazel-tools. Ideally this could come from one of the bazel modules like `java_common.run_ijar`.

gergelyfabian:
Here is the intended usage: https://github.com/gergelyfabian/rules_scala/commit/0cd31de8bc1b6f815afb7debde76aae959f855c5

We had a usecase, that we wanted to generate fat jars with "internal" and "external" dependencies separately (to be able to package them separately for caching reasons). Wondering what route should be optimal for this. We were also considering to add custom phases from our own application (we could reuse part of this code then, I guess).

liucijus:
@gergelyfabian do you still need or use this change?

gergelyfabian:
Yes, we do all the time. We produce 3 jars for each scala_library we have and put them in 3 separate docker layers (i.e. we don't want external dependency jars - that change infrequently - land in the same layer as our internal deps - that change frequently).

simuons (approved):

liucijus (approved):

liucijus:
Sorry for taking so long, @gergelyfabian

PR #1412, opened 2022-07-28
Put -s in front of other parameters for the ScalaTest runner.

### LINKED ISSUE #1411: ScalaTest runner interpretation of --test_filter is problematic
The ScalaTest runner turns `--test_filter=className` into `-s className` before  handing over arguments to ScalaTest. This is causing me some issues in https://github.com/bazelbuild/intellij/pull/3744, where I'm trying to get Google's Intellij plugin to be able to run single ScalaTest test cases, rather than whole classes at a time.

The change I'm making there adds `-t testName` to the flags for ScalaTest, causing it to run whichever test the editor has selected. The issue I'm hitting is that ScalaTest cares about ordering between `-s` and `-t`. `-s -t` means "in the specified class, run the test with the specified name". `-t -s` means "run all tests with the specified name, and all tests in the specified class".

Would anyone object to me moving the `-s` flag added by `rules_scala` to the front of the argument list, rather than the back?

https://github.com/bazelbuild/rules_scala/blob/71d7135808750623b2c9340704a23384de0caefd/src/java/io/bazel/rulesscala/scala_test/Runner.java#L38

srdo-humio:
ScalaTest cares about the ordering between -s (class selection) and
-t (test name selection). "-s -t" will select the test in the specified class
with the specified name. "-t -s" will select all tests in the specified
class, and all tests with the specified name.

Putting the -s we derive from --test_filter at the front of the argument
list rather than at the back is more likely to do what people want when
they specify both --test_filter and other selector flags.

It is still possible to specify -s last by passing the arguments as
--test_arg=-t --test_arg=testName --test_arg=-s --test_arg=className
and leaving --test_filter out.

fixes #1411


johnynek:
I think I was originally responsible for this (or reviewing it) and I frankly didn't understand this. So, I don't think the prior behavior was intentional.

Put another way, I'm in support of merging.

johnynek (approved):

wiwa:
Looks good; could we get a test for this? i.e. an `shtest`

liucijus (changes requested):
Thanks, @srdo-humio. Please add a test to prevent future regressions.

srdo-humio:
:+1: Added tests.

test/scala_test_testfilter/BUILD
```diff
@@ -3,4 +3,5 @@ load("@io_bazel_rules_scala//scala:scala.bzl", "scala_test")
 scala_test(
     name = "tests",
     srcs = ["A.scala", "B.scala"],
+    tags = ["manual"]
```
liucijus: Please move this target under `test_expect_failure/` instead, where it's expected to be not passing

srdo-humio: Done. Should the shell script remain where it is, or where do I put that?

liucijus: Oh, I think this shell script does not get executed - you need to include it into `test_rules_scala.sh`

srdo-humio: Should be done now.

liucijus (approved):

liucijus:
Thanks @srdo-humio for contribution!

@johnynek and @wiwa thanks for reviewing!

PR #1418, opened 2022-09-08
Fix test_scala_proto_library to accept @-prefixed labels

Wyverald:
Needed to flip https://github.com/bazelbuild/bazel/issues/16196


simuons (approved):
Thanks @Wyverald

PR #1424, opened 2022-09-21
Allow scala_benchmark_jmh to specify a runtime_jdk version

wiwa:
### Description

Include `runtime_jdk` as an an `attr` on `jmh_support:benchmark_generator`.
`scala_benchmark_jmh` can now specify a Java runtime to build its `scala_binary` with.

To test whether our changes work for different specified JDKs, we include [rules_java@5.1.0](https://github.com/bazelbuild/rules_java/tree/5.1.0) as an `http_archive`.
We don't override our Bazel's (`@.bazelversion`) `rules_java` and instead pull in the newer `rules_java` as `@rules_java_extra`.  

This allows us to use `remote_jdk8` runtimes specified [here](https://github.com/bazelbuild/rules_java/blob/5.1.0/java/repositories.bzl). 

### Motivation

We want to be able to use JMH benchmarking on different Java runtimes. 

Co-authored-by: Shane Delmore <sdelmore@twitter.com>

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -138,6 +138,25 @@ go_rules_dependencies()
 
 go_register_toolchains()
 
+# Explicitly pull in a different (newer) version of rules_java for remote jdks
+rules_java_extra_version = "5.1.0"
+
+rules_java_extra_sha = "d974a2d6e1a534856d1b60ad6a15e57f3970d8596fbb0bb17b9ee26ca209332a"
+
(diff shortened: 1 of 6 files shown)


liucijus (approved):
Thanks, @wiwa!

wiwa:
@simuons

simuons (approved):
Thanks and sorry for the delay.

PR #1430, opened 2022-10-04
fix: docs pointed to invalid scala_xml dep

omerlh:
### Description

A small fix to the docs regarding custom tool chain

### Motivation

It failed when I tried it and I wanted to make the life of the next person easier

### DIFF
--- a/docs/scala_toolchain.md
+++ b/docs/scala_toolchain.md
@@ -78,7 +78,7 @@ scala_register_toolchains()
     declare_deps_provider(
         name = "my_scala_xml_provider",
         deps_id = "scala_xml",
-        deps = ["@scala_xml_dep"],
+        deps = ["@io_bazel_rules_scala_scala_xml"],
     )
     
     declare_deps_provider(


docs/scala_toolchain.md
```diff
@@ -78,7 +78,7 @@ scala_register_toolchains()
     declare_deps_provider(
         name = "my_scala_xml_provider",
         deps_id = "scala_xml",
-        deps = ["@scala_xml_dep"],
+        deps = ["@io_bazel_rules_scala_scala_xml"],
```
liucijus, line 81: `@io_bazel_rules_scala_scala_xml` is internal to Rules Scala codebase. The whole point of defining custom deps provider is to be able to use deps from your workspace. `@scala_xml_dep` is non existing name on purpose, it should come from users workspace and may have different names depending on which external deps loader it uses.

omerlh: Ah I see - I just couldn't find how to define it or where it is coming from... I will close the PR

Issue #1431, opened 2022-10-08
Enable to override `JAVABIN` in the wrapper script for `scala_binary`  rule


tanishiking:
## Problem
- The wrapper script (generated by the `scala_binary` rule) uses java binary specified by the `JAVABIN` environment variable.
- `JAVABIN` is kinda hardcoded in the script, and we can't modify it from outside of the script.
  - https://github.com/bazelbuild/rules_scala/blob/de3d3a772e2fd443bcd662c740f5f8e916e4b6f0/scala/private/phases/phase_write_executable.bzl#L106-L109
  - which points `${name}_wrapper.sh`
- This could be a problem when we build a docker image using `rules_docker`'s `container_image` rule.

for example

```python
# WORKSPACE
container_pull(
    name = "java_base_11",
    registry = "index.docker.io",
    repository = "library/eclipse-temurin",
    tag = "11-jre-alpine",
)

# BUILD
scala_binary(
    name = "main",
    main_class = "example.Main",
    srcs = glob(["*.scala"]),
)

container_run_and_commit(
  name = "alpine_with_bash",
  commands = ["apk add --no-cache bash"],
  image = "@java_base_11//image",
)

container_image(
    name = "server_temurin_jre",
    base = ":alpine_with_bash",
    files = [":main", ":main_deploy.jar"],
    entrypoint = ["./main", "--singlejar"]
)
```

this doesn't work because `JAVABIN` points to` server_bin_wrapper.sh` which doesn't exist in the docker image, and there's no way to include it AFAIK.

## Expected behavior

The wrapper script should be able to override `JAVABIN` by `JAVABIN=/path/to/java ./main`. Then we can use bin/java that exists in the docker image by `entrypoint = ["JAVABIN=...", "./main", "--singlejar"]` (not sure there's a way to add `ENV JAVABIN=...` to the Dockerfile via `rules_docker`... anyway). 

Actually, `java_binary` generates the wrapper script, and we can override JAVABIN for it.

```sh
# Set JAVABIN to the path to the JVM launcher.
JAVABIN=${JAVABIN:-${JAVA_RUNFILES}/local_jdk/bin/java}
```

## Workaround
- We can just use the `scala_image` rules provided by `rules_docker`.
  - but sometimes, we want to use lower-level API (`container_image` rule) for finer-grained control of the image.
- We can ignore the wrapper script and just run `java -jar name_deploy.jar`
  - but as the document says, using a wrapper script is recommended.
[...]

PR #1432, opened 2022-10-08
fix: enable to override JAVABIN of wrapper script

tanishiking:
fix https://github.com/bazelbuild/rules_scala/issues/1431

### Description

- This PR enables to override `JAVABIN` in the wrapper script generated by `scala_binary` rule
- This behavior aligns to the wrapper script generated by `java_binary` (sorry I couldn't find where the wrapper script for `java_binary` defined, but something like)
  - for `java_binary`: `JAVABIN=${JAVABIN:-${JAVA_RUNFILES}/local_jdk/bin/java}`

### Motivation

please see: https://github.com/bazelbuild/rules_scala/issues/1431

### DIFF
--- a/scala/private/phases/phase_write_executable.bzl
+++ b/scala/private/phases/phase_write_executable.bzl
@@ -103,7 +103,7 @@ def _write_executable_non_windows(ctx, executable, rjars, main_class, jvm_flags,
         [ctx.expand_location(f, ctx.attr.data) for f in jvm_flags],
     )
 
-    javabin = "export REAL_EXTERNAL_JAVA_BIN=${JAVABIN};JAVABIN=%s/%s" % (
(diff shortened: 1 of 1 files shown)


liucijus (approved):
Thanks, @tanishiking!

liucijus:
>     (sorry I couldn't find where the wrapper script for `java_binary` defined, but something like)
The Java rule logic ant the wrapper template seem to be located in this package: https://github.com/bazelbuild/bazel/tree/master/src/main/java/com/google/devtools/build/lib/bazel/rules/java

simuons (approved):
Thanks @tanishiking

tanishiking:
> The Java rule logic ant the wrapper template seem to be located in this package:  https://github.com/bazelbuild/bazel/tree/master/src/main/java/com/google/devtools/build/lib/bazel/rules/java

Thanks!
Oh, looks like `java_binary` uses `:=` instead of `:-` (where `:=` fall back to default value for the empty input `''`), but it's ok with `:-` assuming users won't set `JAVABIN=` (empty).

PR #1437, opened 2022-10-18
Allow overriding jmh_repositories artifact versions

bnavetta:
### Description

Pass along the `overriden_artifacts` argument from `jmh_repositories` so that JMH artifact versions can be customized.

The `jmh_repositories` macro, which sets up JMH benchmarking dependencies, has an `overriden_artifacts` argument for replacing dependencies. However, it doesn't pass the overrides to the underlying `repositories` call, so overrides are ignored.

### Motivation

I'd like to use a more recent version of JMH (1.35) than the current default of 1.20. Because of the benchmark code generation step, I can't do this without overriding the `jmh_repositories` versions.

### DIFF
--- a/jmh/jmh.bzl
+++ b/jmh/jmh.bzl
@@ -20,7 +20,7 @@ def jmh_repositories(
         ],
         fetch_sources = False,
         maven_servers = maven_servers,
-        overriden_artifacts = {},
+        overriden_artifacts = overriden_artifacts,
     )
 
     native.register_toolchains("@io_bazel_rules_scala//jmh:jmh_toolchain")


liucijus:
@bnavetta you should be able to achieve this by declaring a custom jmh toolchain and [pass your own dep providers](https://github.com/bazelbuild/rules_scala/blob/master/jmh/toolchain/toolchain.bzl#L13). Let me know if this works for you, or you have questions.

Regarding `overriden_artifacts` I would like to deprecate it in the future in favor to custom toolchains declaration.

bnavetta:
Using a custom toolchain worked! Is it still worth passing along the `overriden_artifacts` argument, or would you rather it stay a no-op to discourage use?

I can make a separate issue, but on the topic of JMH, would you be open to a patch upgrading the default version from 1.20 to the current version?

liucijus:
@bnavetta I plan to remove `overriden_artifacts` in favor to deps toolchains. I don't think Rules Scala should work as a Maven third party loader, instead users should be able to use deps from loaders used in their projects (eg. Rules Jvm External). 
If custom toolchains work, why do you want to update JMH version?

bnavetta:
Since I already had the latest versions of JMH and its dependencies handy, I figured it might be useful to other `rules_scala` users to update the defaults. If `rules_scala` is going to stop resolving Maven dependencies though, I agree it's probably not worth doing.

wisechengyi:
@bnavetta shall we close this one?

liucijus:
> Since I already had the latest versions of JMH and its dependencies handy, I figured it might be useful to other `rules_scala` users to update the defaults. 

@bnavetta feel free to update default versions. In some cases we still keep older dependency versions on rules scala to make sure rules work between Scala versions, but in case of JMH it shouldn't be a problem.

liucijus:

PR #1439, opened 2022-10-31
Simplify setup of scala_toolchain and its deps

liucijus:
### Description
Added `setup_scala_toolchain` macro to simplify scala toolchain setup.

### Motivation
To configuration of custom scala toolchain one needs to understand implementation details like dependency provider ids, which makes it a complicated task. We can encapsulate those details under a macro.

### DIFF
--- a/scala/private/macros/setup_scala_toolchain.bzl
+++ b/scala/private/macros/setup_scala_toolchain.bzl
@@ -16,8 +16,6 @@ def setup_scala_toolchain(
     scala_library_classpath_provider = "%s_scala_library_classpath_provider" % name
     scala_macro_classpath_provider = "%s_scala_macro_classpath_provider" % name
 
-    print("XML", scala_xml_provider)
-
     declare_deps_provider(
         name = scala_compile_classpath_provider,
         deps_id = "scala_compile_classpath",
--- a/scala/private/macros/setup_scala_toolchain.bzl
+++ b/scala/private/macros/setup_scala_toolchain.bzl
@@ -16,6 +16,8 @@ def setup_scala_toolchain(
     scala_library_classpath_provider = "%s_scala_library_classpath_provider" % name
     scala_macro_classpath_provider = "%s_scala_macro_classpath_provider" % name
 
+    print("XML", scala_xml_provider)
+
     declare_deps_provider(
         name = scala_compile_classpath_provider,
         deps_id = "scala_compile_classpath",
--- a/scala/private/macros/setup_scala_toolchain.bzl
+++ b/scala/private/macros/setup_scala_toolchain.bzl
@@ -3,11 +3,11 @@ load("//scala:providers.bzl", "declare_deps_provider")
 
 def setup_scala_toolchain(
         name,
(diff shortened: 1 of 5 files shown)


scala/private/macros/setup_scala_toolchain.bzl
```diff
@@ -0,0 +1,74 @@
...
+        scala_macro_classpath,
+        visibility = ["//visibility:public"],
+        **kwargs):
+    scala_xml_provider = "%s_scala_xml_provider" % name
+    parser_combinators_provider = "%s_parser_combinators_provider" % name
+    scala_compile_classpath_provider = "%s_scala_compile_classpath_provider" % name
+    scala_library_classpath_provider = "%s_scala_library_classpath_provider" % name
+    scala_macro_classpath_provider = "%s_scala_macro_classpath_provider" % name
+
+    print("XML", scala_xml_provider)
```
simuons, line 19: Is this really needed?

scala/private/macros/setup_scala_toolchain.bzl
```diff
@@ -0,0 +1,74 @@
+load("//scala:scala_toolchain.bzl", "scala_toolchain")
+load("//scala:providers.bzl", "declare_deps_provider")
+
+def setup_scala_toolchain(
+        name,
+        scala_xml_deps,
```
simuons, line 6: nit: why some attribute names ends with deps and others with classpath? What's the distinction?

scala/private/macros/setup_scala_toolchain.bzl
liucijus: Thanks, for catching, @simuons!

simuons (approved):

scala/private/macros/setup_scala_toolchain.bzl
liucijus: `classpath` is what historically was used on the toolchain, other names came when dep providers were introduced. How I see it: `classpath` is implicit deps on the classpath, and `deps` is tool related deps. But this naming assumption does not even hold with the current codebase (eg. compiler is tool, but has classpath naming). Not sure what to do about it. It's a good time to improve those names at least on the macro side. Naming suggestions are more than welcome. @simuons, WDYT?

simuons: I don't really have good suggestion. I wasn't aware of this distinction about `classpath` and `tools`. My initial thought was "all those deps go to the classpath at some point so why the distinction". The question probably is wether users should set 'internal' deps for things like `//scala/support:test_reporter` but this is out of scope for this PR. So I'm ok with state as it is now.

liucijus: After thinking a bit more, I'll change it to use default providers for tools deps. This should give us more room to do future refactorings. Eg. move scala_xml dep to testing toolchain.

liucijus:
@simuons I have made scala xml and parser combinators deps to be default if not specified

simuons (approved):

PR #1446, opened 2022-11-09
Revert "Declare non-windows executable with .sh extension"

ittaiz:
Reverts bazelbuild/rules_scala#1407
1407 was a breaking change.
The issue was that IJWB [assumes](https://github.com/bazelbuild/intellij/blob/a5522a0f1787b667a3a966e927da773cb4ec51a2/java/src/com/google/idea/blaze/java/run/fastbuild/FastBuildTestEnvironmentCreator.java#L90) the runfiles and executable are the same name as the target (java convention) and #1407 broke this convention/assumption by having the executable named `<target>.sh` and runfiles `<target>.sh.runfiles` directory.

### DIFF
--- a/docs/scala_repl.md
+++ b/docs/scala_repl.md
@@ -21,5 +21,5 @@ Instead, use `bazel build` to build the script, then run that script as normal t
 An example in this repo:
 ```
 bazel build test:HelloLibRepl
-bazel-bin/test/HelloLibRepl.sh
+bazel-bin/test/HelloLibRepl
 ```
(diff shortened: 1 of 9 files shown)


liucijus (approved):

ittaiz:
:) I wish I could take such a broad credit.
A temp project of mine is to get "fast test" working for scala inside IntelliJ.
One day the feature broke down completely so I went hunting for the cause :)

simuons (approved):

Issue #1450, opened 2022-11-18
Set up GitHub actions to automate Rules Scala releases


liucijus:
Minimally release should contain copy/paste setup instructions with matching archive **url** and **sha256**. Eg.:
```
# WORKSPACE
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

skylib_version = "1.0.3"
http_archive(
    name = "bazel_skylib",
    sha256 = "1c531376ac7e5a180e0237938a2536de0c54d93f5c278634818e0efc952dd56c",
    type = "tar.gz",
    url = "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/releases/download/{}/bazel-skylib-{}.tar.gz".format(skylib_version, skylib_version),
)

http_archive(
    name = "io_bazel_rules_scala",
    url = "https://github.com/bazelbuild/rules_scala/releases/download/20220201/rules_scala-20220201.zip",
    type = "zip",
    strip_prefix = "rules_scala-20220201",
    sha256 = "77a3b9308a8780fff3f10cdbbe36d55164b85a48123033f5e970fdae262e8eb2",
)

# Stores Scala version and other configuration
# 2.12 is a default version, other versions can be use by passing them explicitly:
# scala_config(scala_version = "2.11.12")
load("@io_bazel_rules_scala//:scala_config.bzl", "scala_config")
scala_config()

load("@io_bazel_rules_scala//scala:scala.bzl", "rules_scala_setup", "rules_scala_toolchain_deps_repositories")

# loads other rules Rules Scala depends on 
rules_scala_setup()

# loads Maven deps like Scala compiler and standard libs, on production projects you should consider 
# defining a custom deps toolchains to use your project libs instead 
rules_scala_toolchain_deps_repositories(fetch_sources = True)

load("@rules_proto//proto:repositories.bzl", "rules_proto_dependencies", "rules_proto_toolchains")
rules_proto_dependencies()
rules_proto_toolchains()

load("@io_bazel_rules_scala//scala:toolchains.bzl", "scala_register_toolchains")
scala_register_toolchains()
```

For the inspiration and possible implementations, we can look how other rules have implemented releases. Eg.: [rules_pythin](https://github.com/bazelbuild/rules_python/blob/main/.github/workflows/release.yml), [rules_js](https://github.com/aspect-build/rules_js/blob/main/.github/workflows/release.yml)

jvandew:
[...]

PR #1452, opened 2022-11-28
add tagged release workflow using github actions

jvandew:
### Description
This PR adds a new git tag based release workflow, as outlined in #1450, mirroring the setup that exists in other Bazel rulesets. Pushing a versioned git tag of the form `v1.2.3` will trigger a Github action to generate and document a release archive.

The setup here is largely identical to the release actions in [rules_python](https://github.com/bazelbuild/rules_python/blob/00545742ad2450863aeb82353d4275a1e5ed3f24/.github/workflows/release.yml) and [rules_js](https://github.com/aspect-build/rules_js/blob/fb1f84aca12627308422097e433fb2eb52b0cf23/.github/workflows/release.yml), though for simplicity it omits the test step as being redundant to the existing CI setup.

There are some sample runs in my fork [here](https://github.com/jvandew/rules_scala/actions/runs/3558745957), which generate releases such as [this](https://github.com/jvandew/rules_scala/releases/tag/v1.2.12).

### Motivation
Addresses #1450, originally motivated by a [slack request](https://bazelbuild.slack.com/archives/CDCKJ2KFZ/p1668601467490499).


liucijus:
Thanks @jvandew!
What would we potentially miss if we skip a testing step? Rules Scala test runs are very heavy, maybe we can rely on master builds? I don't have my opinion formed yet, would like to hear any pros/cons. 
If we have to test, then we need to run `./test_rules_scala.sh`

jvandew:
I'm not familiar with the existing build setup in rules_scala (looks like the master build you're referring to is scheduled daily on [Buildkite](https://buildkite.com/bazel/rules-scala-scala)?), but relying on the master builds seems reasonable to me... The main risk I think would be that two PRs individually pass CI but conflict when merged together, and a release is cut before the scheduled master build runs to detect the break, but that is probably a rare occurrence and it is probably easy enough to just wait for a green build on master before releasing.

The flip side I guess is that Github Actions appears to have fairly generous [resource limits](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration) for public repos, so there probably isn't an actual cost to running a heavy test suite beyond the time it takes to execute.

liucijus:
@jvandew maybe let's continue without testing step, if we see that testing is needed, we can re-add it later. Running without testing would simplify release a lot.

jvandew:
> @jvandew maybe let's continue without testing step, if we see that testing is needed, we can re-add it later. Running without testing would simplify release a lot.

updated! the release is [quite snappy](https://github.com/jvandew/rules_scala/actions/runs/3635653305) now

.github/workflows/release.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Prepare workspace snippet
+        run: .github/workflows/workspace_snippet.sh ${{ env.GITHUB_REF_NAME }} > release_notes.txt
```
simuons, line 18: Thanks @jvandew, this is really cool. Just a small question: does `${{ env.GITHUB_REF_NAME }}` needs to be passed here? Cannot find where it's taken in the script, looks like it's read from env.

jvandew: aha you appear to be correct, nice catch!

liucijus (approved):
Thanks, @jvandew!

Issue #1455, opened 2022-12-05
Same name resource files are overwritten by upstream deps


tanishiking:
Reproduce example: https://github.com/tanishiking/bazel-playground/tree/main/13-bazel-resources

- We have 2 targets, `foo` and `base`.
- Both targets have a resource `a.txt`
- `foo` depending on `base`

```
.
|-- base
|   |-- BUILD
|   |-- base.scala
|   `-- resources
|       `-- a.txt
`-- foo
    |-- BUILD
    |-- foo.scala
    `-- resources
        `-- a.txt
```

```python
# base/BUILD
load("@io_bazel_rules_scala//scala:scala.bzl", "scala_library")
scala_library(
    name = "base",
    srcs = ["base.scala"],
    resources = glob(["resources/**"]),
    visibility = ["//visibility:public"],
)

# foo/BUILD
load("@io_bazel_rules_scala//scala:scala.bzl", "scala_binary")
scala_binary(
    name = "foo",
    srcs = ["foo.scala"],
    main_class = "foo.Foo",
    deps = ["//scala/base"],
    resources = glob(["resources/**"]),
)
```

```sh
❯ cat base/resources/a.txt
I'm from a.txt under base

❯ cat foo/resources/a.txt
I'm from a.txt under foo
```

## Problem
When we build `bazel build //scala/foo` the generated jar file contains one `a.txt` in jar, that's ok

```sh
❯ jar tf bazel-bin/scala/foo/foo.jar
META-INF/
META-INF/MANIFEST.MF
a.txt
foo/
foo/Foo$.class
foo/Foo.class
```

The problem is it's from `base`, instead of `foo`

```
❯ bazel-bin/scala/foo/foo
I'm from a.txt under base
```

It looks like the upstream resources are prioritized when the resource name crashes.

## Expectation
```
❯ bazel-bin/scala/foo/foo
I'm from a.txt under foo
```

Or is there a workaround for this?

tanishiking:
Another example (comparing to sbt) with typesafe config https://github.com/tanishiking/typesafeconfig-sbt-bazel

liucijus:
@tanishiking could you clarify what would be an expected behavior for you?

tanishiking:
@liucijus Sorry, I was struggling to explain the expected behavior 😅 

Let's say we have targets `A` and `B` (where `A` depends on `B`), and both targets have the same name resource `a.txt`.
[...]

PR #1456, opened 2022-12-07
Add failing test for resource name crash

tanishiking:
### Description

This PR adds a failing test for https://github.com/bazelbuild/rules_scala/issues/1455

@liucijus mentioned this issue occurs around here https://github.com/bazelbuild/rules_scala/blob/master/src/java/io/bazel/rulesscala/scalac/ScalacWorker.java#L65-L71 but re-ordering those lines doesn't work (I guess because re-ordering them control the priority of `resource`, `resource_jars`, and `classpath_resources`. It's not about resource name crash with things in JAR files).

My hunch is, this is caused because the executable file has a "wrong" order of CLASSPATH.
For example, `bazel-bin/test/src/main/scala/scalarules/test/duplicated_resources/child/child` (generated by`$ bazel build //test/src/main/scala/scalarules/test/duplicated_resources/child`) contains the following `CLASSPATH` order.

```
CLASSPATH="
  ${RUNPATH}test/src/main/scala/scalarules/test/duplicated_resources/parent/parent.jar:
  ${RUNPATH}../io_bazel_rules_scala_scala_library/scala-library-2.12.14.jar:
  ...
  ${RUNPATH}../bazel_tools/tools/java/runfiles/librunfiles.jar
  ${RUNPATH}test/src/main/scala/scalarules/test/duplicated_resources/child/child.jar:
  "
```

`parent.jar` is first, and `child.jar` is last.

I think we can fix this by moving `child.jar` to the first of `CLASSPATH` is the fix. Actually, moving `child.jar` to the place before `parent.jar` fixed the issue.

What do you think?

### Motivation

see https://github.com/bazelbuild/rules_scala/issues/1455

### DIFF
--- a/test/src/main/scala/scalarules/test/duplicated_resources/child/BUILD
+++ b/test/src/main/scala/scalarules/test/duplicated_resources/child/BUILD
@@ -1,0 +1,12 @@
+load("//scala:scala.bzl", "scala_test")
+
+# https://github.com/bazelbuild/rules_scala/issues/1455
+scala_test(
+    name = "child",
+    size = "small",
+    srcs = ["ScalaLibResourcesDuplicatedTest.scala"],
+    resources = ["resource.txt"],
+    unused_dependency_checker_mode = "off",
+    resource_strip_prefix = "test/src/main/scala/scalarules/test/duplicated_resources/child/",
+    deps = ["//test/src/main/scala/scalarules/test/duplicated_resources/parent"],
+)
--- a/test/src/main/scala/scalarules/test/duplicated_resources/child/ScalaLibResourcesDuplicatedTest.scala
+++ b/test/src/main/scala/scalarules/test/duplicated_resources/child/ScalaLibResourcesDuplicatedTest.scala
@@ -1,0 +1,14 @@
+package scalarules.test.duplicated
+
+import org.scalatest.funsuite._
+
+class ScalaLibResourcesDuplicatedTest extends AnyFunSuite {
+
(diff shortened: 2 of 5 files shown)


liucijus:
@tanishiking indeed you're right, this is runtime behavior. I would look for problem in [collect jars phase](https://github.com/bazelbuild/rules_scala/blob/master/scala/private/phases/phase_collect_jars.bzl).

tanishiking:
It looks like the problem is around [here in phase_compile](https://github.com/tanishiking/rules_scala/blob/2cc2d56fdf23b3738c7f27bfe49bfcd2d19aecdc/scala/private/phases/phase_compile.bzl#L152) 

```starlark
rjars = depset(out.full_jars, transitive = [rjars]),
```

where `out.full_jars` is `child.jar` and `rjars` contains the rest of dependent jars.
I confirmed `depset(out.full_jars, transitive = [rjars], order = "preorder")` fixes the issue, but I'm not really understand the [depset](https://bazel.build/rules/lib/depset), so not sure it's good solution 😅

liucijus (approved):
Thanks, @tanishiking!

tanishiking:
(Just force-pushed the change on test description) https://github.com/bazelbuild/rules_scala/pull/1456/files#diff-96060316d4c081f3a4d50a995f83805f403e50c478da6c2bd14c26068f84666cR7

simuons (approved):
Thanks @tanishiking

PR #1457, opened 2022-12-13
add 3.2 support

nolen777:


liucijus:
@nolen777 will you sign Google CLA? 

In addition, we need a test for such change. This can be done by adding one more run for Scala 3 example with `--repo_env=3.2.1` here: https://github.com/bazelbuild/rules_scala/blob/master/test/shell/test_examples.sh#L19

simuons:
Duplicates https://github.com/bazelbuild/rules_scala/pull/1465

Issue #1483, opened 2023-03-20
[Question] Unable to build custom JacocoRunner using the script


tanishiking:
Related https://github.com/bazelbuild/rules_scala/issues/1453

## Environment

```
❯ sysctl machdep.cpu.brand_string
machdep.cpu.brand_string: Apple M2

❯ bazel version
Bazelisk version: development
WARNING: Invoking Bazel in batch mode since it is not invoked from within a workspace (below a directory having a WORKSPACE file).
Build label: 6.1.1
Build target: bazel-out/darwin_arm64-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Wed Mar 15 15:45:56 2023 (1678895156)
Build timestamp: 1678895156
Build timestamp as int: 1678895156
```

## Background

- Code coverage taken from standard Jacoco takes into account the synthetic methods in case classes
  - BTW, why they aren't ignored, although the filters should be applied unconditionally 🤔 https://github.com/jacoco/jacoco/wiki/FilteringOptions
- As commented in `build_jacocorunner_bazel` script, it seems we need to build the custom Jacoco runner patched in https://github.com/gergelyfabian/jacoco
  - https://github.com/bazelbuild/rules_scala/blob/6c1f034d0583f0f39f54d5631b8efae430df0522/scripts/build_jacocorunner/build_jacocorunner_bazel_5.0%2B.sh#L7-L9

<img width="1329" alt="Screenshot 2023-03-20 at 14 32 59" src="https://user-images.githubusercontent.com/9353584/226254873-6627cad0-7b79-4f73-bf42-a62e9a3a8ade.png">

## Problem

I tried to build the custom Jacoco runner with the following patch against [current master](https://github.com/bazelbuild/rules_scala/commit/6c1f034d0583f0f39f54d5631b8efae430df0522)  + https://github.com/tanishiking/rules_scala/commit/9fdb2ba8c8f1e1ee8260ab26f2f0ad142421c509

(Ignoring groovy validation because it fails to build in my local environment, it should be ok because it's test, right?)

when I run `./scripts/build_jacocorunner/build_jacocorunner_bazel_5.0+.sh`, it fails with the following error. 

```
[...]

PR #1485, opened 2023-03-21
Fixing JacocoRunner building script for Jacoco

gergelyfabian:
- Latest Bazel uses Jacoco 0.8.7
- Ensure that for building JacocoRunner we use a compatible Bazel version
- Created Bazel branch that adds interface extensions (interfaceNames) to Bazel's Jacoco implementation
- Java 8 needs to be used for building Jacoco

### Description
This fixes the build issues reported in #1483.


scripts/build_jacocorunner/build_jacocorunner_bazel_5.0+.sh
```diff
@@ -36,6 +36,10 @@
 
 set -e
 
+# Ensure Java 8 is used for building Jacoco (experienced issue when using e.g. Java 17).
+# You may need to change this on your system.
+export JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
+export PATH=$JAVA_HOME/bin:$PATH
```
tanishiking, line 42: I think these settings are too specific (I have different place for JAVA_HOME, and I can switch java versions using jenv in my local environment).

What do you think to add a sentence "Ensure Java 8 is used for building Jacoco" in `docs/coverage.md` and remove these `export`s?

scripts/build_jacocorunner/build_jacocorunner_bazel_5.0+.sh
```diff
@@ -108,7 +114,9 @@ mvn clean install
 (
 cd $bazel_repo
 git remote update
-git checkout tags/$bazel_tag
+git checkout $bazel_branch
+
+echo "$bazel_version" > .bazelversion
```
tanishiking, line 119: :+1:

tanishiking (approved):
Hey, @gergelyfabian thank you so much for fixing both this script and the branch!
Looks like we still miss some synthetic methods generated by case class, we can build the runner with the script, and it looks better :tada:

Screenshot 2023-03-22 at 13 34 12
tested on https://github.com/tanishiking/bazel-playground/tree/main/17-scala-coverage

scripts/build_jacocorunner/build_jacocorunner_bazel_5.0+.sh
gergelyfabian: Whatever solution works for you. Should I comment these lines and leave a fixme?

scripts/build_jacocorunner/build_jacocorunner_bazel_5.0+.sh
gergelyfabian: This is fixing Bazel autodownloading version 6+ when building JacocoRunner, that breaks that build...

gergelyfabian:
@liucijus I have no idea why the automatic CI build fails. Do you have any idea? I just changed a script that should not be used in the CI build...

scripts/build_jacocorunner/build_jacocorunner_bazel_5.0+.sh
gergelyfabian: Updated my commit. Now, you need to manually ensure, and the script will error out if the wrong Java version is set.

liucijus (approved):
Thanks a lot, @gergelyfabian!

PR #1489, opened 2023-03-28
Catch macro expansion exceptions in ScalacWorker

### LINKED ISSUE #1488: When ScalacWorker fails during macro evaluation, errors are not made available in build log
When compiling a scala_library that requires a macro expansion, and there is a user error in that macro, the entire scalac process will crash, resulting in a Bazel error like this:
```
ERROR: /Users/tkaplan/<redacted>:3:14: scala @//<redacted> failed: Worker process did not return a WorkResponse:

---8<---8<--- Start of log snippet, file at /private/var/tmp/_bazel_tkaplan/fa1bc4e61b191b698b582e9d142cb714/bazel-workers/worker-6-Scalac.log ---8<---8<---
[... truncated ...]
cros$MacroExpander.apply(Macros.scala:609)
	at scala.tools.nsc.typechecker.Macros.standardMacroExpand(Macros.scala:784)
	at scala.tools.nsc.typechecker.Macros.standardMacroExpand$(Macros.scala:782)
	at scala.tools.nsc.Global$$anon$5.standardMacroExpand(Global.scala:482)
	at scala.tools.nsc.typechecker.AnalyzerPlugins$$anon$10.default(AnalyzerPlugins.scala:457)
	at scala.tools.nsc.typechecker.AnalyzerPlugins$$anon$10.default(AnalyzerPlugins.scala:454)
	at scala.tools.nsc.typechecker.AnalyzerPlugins.invoke(AnalyzerPlugins.scala:411)
	at scala.tools.nsc.typechecker.AnalyzerPlugins.pluginsMacroExpand(AnalyzerPlugins.scala:454)
	at scala.tools.nsc.typechecker.AnalyzerPlugins.pluginsMacroExpand$(AnalyzerPlugins.scala:454)
	at scala.tools.nsc.Global$$anon$5.pluginsMacroExpand(Global.scala:482)
```

There are several reasons why this isn't very user-friendly:
1) The actual reason for the error is in the truncated part of the stack trace (in this case, it was a NoSuchMethodError caused by mixing Scala 2.11 and 2.12 libs)
2) If the build is happening on CI, getting the worker-n-Scalac.log file can be difficult
3) The error won't be sent as part of the BES to any BES viewers like BuildBuddy

Maybe the ScalacWorker implementation should try to catch unchecked exceptions and determine if they are the result of macro expansion, rather than crashing.

thirtyseven:
### Description

Catch scalac exceptions that occur during expansion of user macros so that the entire worker doesn't crash.

### Motivation

Fixes #1488

### DIFF
--- a/src/java/io/bazel/rulesscala/scalac/ScalacWorker.java
+++ b/src/java/io/bazel/rulesscala/scalac/ScalacWorker.java
@@ -248,6 +248,15 @@ class ScalacWorker implements Worker.Interface {
     return pluginParams.toArray(new String[pluginParams.size()]);
   }
 
+  private static boolean isMacroException(Throwable ex) {
+    for (StackTraceElement elem : ex.getStackTrace()) {
+      if (elem.getMethodName().equals("macroExpand")) {
+        return true;
+      }
+    }
+    return false;
+  }
+
   private static void compileScalaSources(CompileOptions ops, String[] scalaSources, Path classes)
       throws IOException {
 
@@ -269,6 +278,8 @@ class ScalacWorker implements Worker.Interface {
     } catch (Throwable ex) {
       if (ex.toString().contains("scala.reflect.internal.Types$TypeError")) {
         throw new RuntimeException("Build failure with type error", ex);
+      } else if (isMacroException(ex)) {
+        throw new RuntimeException("Build failure during macro expansion", ex);
       } else {
         throw ex;
       }
--- a/test/shell/test_persistent_worker.sh
+++ b/test/shell/test_persistent_worker.sh
@@ -5,13 +5,29 @@ dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 . "${dir}"/test_helper.sh
 runner=$(get_test_runner "${1:-local}")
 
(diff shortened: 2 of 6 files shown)


liucijus (changes requested):
Thanks, @thirtyseven! Would you be able to contribute an end to end test?

thirtyseven:
> Thanks, @thirtyseven! Would you be able to contribute an end to end test?

I can take a stab at it, are there any existing e2e tests for the existing case of handling `scala.reflect.internal.Types$TypeError` that I can base it off of?

thirtyseven:
It doesn't appear that there are any, I'll try to add e2e tests for both cases.

thirtyseven:
@liucijus Actually, I couldn't figure out how to reproduce the previous `scala.reflect.internal.Types$TypeError` case, but I did manage to add a test for the new case. I've confirmed that it fails without the change from this PR.

liucijus (approved):

simuons (approved):

thirtyseven:
Thanks for the review! FYI I do not have commit access so will need someone else to merge.

PR #1491, opened 2023-04-06
Handle StackOverflowError in macros

thirtyseven:
### Description
In a similar fix to #1489, catch StackOverflowErrors in ScalacWorker so that buggy macros don't crash the persistent worker resulting in unhelpful logs.

### Motivation

Fix a similar class of issue as #1488


liucijus (approved):
Thanks, @thirtyseven!

thirtyseven:
Requesting a merge on this PR from any eligible committers

simuons (approved):

PR #1500, opened 2023-05-20
Replace relative paths to ../ with external/

prebeta:
### Description
Java throws an IllegalArgumentException when argument paths are not normalized. This error happens when running tests in external repositories i.e @external_scala_repo//path/to/target because short_path returns ../path/to/target instead of external/path/to/target.

### Motivation
This fixes cases where we want to run tests in external repositories. The stack trace of this type of error looks like:

```
Exception in thread "main" java.lang.IllegalArgumentException: path is not normalized: "__main__/../path/to/target/tests.jar"
	at com.google.devtools.build.runfiles.Util.checkArgument(Util.java:45)
	at com.google.devtools.build.runfiles.Runfiles.rlocation(Runfiles.java:134)
	at io.bazel.rulesscala.scala_test.Runner.rlocateRunpathValue(Runner.java:95)
	at io.bazel.rulesscala.scala_test.Runner.extendFromFileArgs(Runner.java:59)
	at io.bazel.rulesscala.scala_test.Runner.extendArgs(Runner.java:37)
	at io.bazel.rulesscala.scala_test.Runner.main(Runner.java:33)
```


simuons:
Hi, @prebeta,

thanks for pr. Could you add a test case for this?

prebeta:
> Hi, @prebeta,
> 
> thanks for pr. Could you add a test case for this?

Sounds good! I reused `strip_resource_external_workspace` and renamed it to `example_external_workspace` so we don't need to add another `local_repository` in the WORKSPACE file.

simuons (approved):

PR #1465, opened 2023-01-23
Add support for Scala 3.2

meisam:
### Description
Add support of Scala 3.2

- Add the build and repository definitions for Scala 3.2.1
- Add an extra set of tests for Scala 3.2.1

### Motivation
- To stay as close as possible to the head of Scala.
-  To make new features in Scala 3.2 available to developers.

Fixes #1464

PR #1501, opened 2023-05-31
Add support for Scala 3.3 LTS

csubj:
### Description
Add support for Scala 3.3.0

- Add the build and repository definitions for Scala 3.3.0
- Add an extra set of tests for Scala 3.3.0

### Motivation
- To stay as close as possible to the head of Scala and to provide a base for LTS. 
- To make new features in Scala 3.3 available to developers.

Followed the steps of #1465 and the changes pass those tests but I think it's probably not exhaustive nor complete. I computed the dep SHA256 with a tool I hacked together and updated the values. If people are interested I can make my tool a bit more complete for managing and updating the artifact dict and publish it somewhere.


csubj:
I'm not sure I have the ability to re-run the build but I believe the failure is unrelated to changes, perhaps someone can give it a kick for me

third_party/repositories/scala_3_3.bzl
```diff
@@ -0,0 +1,536 @@
+scala_version = "3.3.0"
+scalatest_version = "3.2.16"
+netty_version = "4.1.32.Final"
+grpc_version = "1.24.0"
+specs2_version = "5.0.0-RC-21"
```
liucijus, line 5: Could you please inline version vars, except `scala_version` which is referred in other macros?  I'd like to keep this file easy to grep

liucijus (changes requested):
Thanks, @csubj. Please address my comment

third_party/repositories/scala_3_3.bzl
csubj: o7 np, thought i was being clever

liucijus (approved):

erinmez:
poke, bump :) would also love to see this find its way into the master

simuons (approved):

aaylward:
@csubj Thanks for adding this! Does scalatest-compatible work for you with scala 3.3? I get a sha256 mismatch: 
```
Checksum was 9283e684d401d821a4cbb2646f9611cbbcd7828d2499483d13a4b507775a4cd7 but 
wanted 7e5f1193af2fd88c432c4b80ce3641e4b1d062f421d8a0fcc43af9a19bb7c2eb
```

The `928...` checksum is what I see when I do
```
wget https://maven-central.storage-download.googleapis.com/maven2/org/scalatest/scalatest-compatible/3.2.16/scalatest-compatible-3.2.16.jar
shasum -a 256 scalatest-compatible-3.2.16.jar
```

PR #1503, opened 2023-06-24
Make DepsTrackingReporter handle Windows path separators

### LINKED ISSUE #1487: New dep tracking not working on Windows due to 'unused dep' errors
The new dep tracking developed in #1459 is not working on Windows, and is causing erroneous "unused dep" errors.

The issue is that in the patch files the "ArchivePath" variable has a windows path (with backslash '\\' instead of '/').  When the DepsTrackingReporter uses this to compare, it fails due to the slashes being different. 

Easy fix with a .replace('\\\\', '/'), but not sure if its best to put change in the DepsTrackingReporter or in each of the patch files. 
@liucijus can you advise?

crt-31:
### Description
This fixes issue where the ‘compiler-dependency-analyzer’ was not handling Windows path separator and failed with false negatives. 
* Added one-liner in DepsTrackingReporter.java that replaces ‘\\’ with ‘/’ in the paths it receives from the scalac compiler.

### Motivation
Makes the new compiler-dependency-analyzer work on Windows. 

Before this PR, on Windows, compiling would report a bunch of erroneous 'unused deps' errors.

This fixes #1487

### DIFF
--- a/src/java/io/bazel/rulesscala/scalac/reporter/DepsTrackingReporter.java
+++ b/src/java/io/bazel/rulesscala/scalac/reporter/DepsTrackingReporter.java
@@ -17,9 +17,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.jar.JarFile;
 import java.util.stream.Collectors;
-
-import javax.print.attribute.standard.Severity;
-
 import scala.reflect.internal.util.NoPosition$;
 import scala.reflect.internal.util.Position;
 import scala.tools.nsc.Settings;
--- a/src/java/io/bazel/rulesscala/scalac/reporter/DepsTrackingReporter.java
+++ b/src/java/io/bazel/rulesscala/scalac/reporter/DepsTrackingReporter.java
@@ -17,6 +17,9 @@ import java.util.Map;
 import java.util.Set;
 import java.util.jar.JarFile;
 import java.util.stream.Collectors;
+
+import javax.print.attribute.standard.Severity;
+
 import scala.reflect.internal.util.NoPosition$;
 import scala.reflect.internal.util.Position;
 import scala.tools.nsc.Settings;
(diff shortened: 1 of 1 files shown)


scala/private/phases/phase_write_executable.bzl
```diff
@@ -85,12 +85,16 @@ def _write_executable_windows(ctx, executable, rjars, main_class, jvm_flags, wra
     cpfile = ctx.actions.declare_file("%s.classpath" % ctx.label.name)
     ctx.actions.write(cpfile, classpath)
 
+    specifiedEnv = {}
```
liucijus, line 88: Do we need the same change in both PRs? https://github.com/bazelbuild/rules_scala/pull/1502#discussion_r1242565719

liucijus (changes requested):
Thanks @crt-31, LGTM, please address the comment

scala/private/phases/phase_write_executable.bzl
crt-31: Both PRs need the change to work (I was trying to make the PRs standalone). But I could put in only one if thats what you want, probably not big deal if you are gonna commit both PRs soon. 
Please advise.

liucijus: @crt-31 we can merge PRs in order, #1502 and then #1503

crt-31: OK. I removed this 'specifiedEnv' change from this PR.
@liucijus  Ready for your review.

src/java/io/bazel/rulesscala/scalac/reporter/DepsTrackingReporter.java
```diff
@@ -17,6 +17,9 @@
 import java.util.Set;
 import java.util.jar.JarFile;
 import java.util.stream.Collectors;
+
+import javax.print.attribute.standard.Severity;
```
liucijus, line 21: Looks to be unused import

liucijus (changes requested):

src/java/io/bazel/rulesscala/scalac/reporter/DepsTrackingReporter.java
crt-31: @liucijus:  ok. That unused import is removed.
Fixed.

liucijus (approved):

simuons (approved):

crt-31:
@liucijus, I'm not sure why the CI tests failed on this merge. The tests ran fine on my linux vm.
Any chance you could re-run the CI tests on this merge?

PR #1504, opened 2023-07-11
call close() on compiler in scalacWorker after each job

### LINKED ISSUE #1486: ScalacWorker not releasing all file handles after job - causing errors
ScalacWorker is not releasing all the file handles (i.e. for the jar files) after it is done with a job.  Happening in rules_scala 5.0.0 as well as latest master code.

One error caused by this happens when editing code that causes an ijar to change. The ijar file handle is still open in the ScalacWorker from a previous compile, and the JavaIjar action cannot write to it and gives error: 

> ERROR: ... Extracting interface for jar xxxx.jar failed: failed to delete output files before executing action

I can see that the xxx-ijar.jar file (among many other jar files) is still open by the ScalacWorker process. Calling bazel shutdown will solve the problem since it closes the ScalacWorker processes which in turn closes the file handles. 

I am kind of surprised this hasn’t come up before, but it is pretty annoying when compiling after code edits. Note I am using **Windows**, so there’s a chance this only causes the error on windows but not linux.

I didn’t see too much wrong with the ScalacWorker code, so I suspect it is either the ScalacWorker is not closing the scalac objects correctly at the end of the job, OR there is a bug in the scalac code itself. Someone smarter than me probably should investigate further.

I would categorize this issue as **critical**, since it is causing unnecessary errors and it is nearly unusable when editing code.

crt-31:
### Description
This PR fixes the scalacWorker so that it calls close() on the compiler object (nsc.Global) after each job is done. This allows file resources (i.e. dependent jar files, source files, etc) that the compiler uses to be properly released after each job.

Note that nsc.Global did not add the close() functionality until v2.12. This is handled in this PR.

### Motivation
This issue was noticed more on Windows since Windows will lock the file (and not allow edits) until it is released. And with scalacWorker being a persistent process, it would lock the files until you called “bazel shutdown”. 

This should resolve #1486, where this issue was causing *-ijar.jar handles not to be released.

_Aside: There is still an issue where the jars for plugins are not being released (because calling nsc.Global.close() does not release the plugins). This is probably more of a Bazel core issue where Bazel needs to stop the workers when any plugins (or other tooling used by the worker) need to be recompiled. The related issue is Bazel issue #[10498](https://github.com/bazelbuild/bazel/issues/10498)._

### DIFF
--- a/src/java/io/bazel/rulesscala/scalac/ReportableMainClass.java
+++ b/src/java/io/bazel/rulesscala/scalac/ReportableMainClass.java
@@ -16,20 +16,20 @@ import java.lang.AutoCloseable;
 public class ReportableMainClass extends MainClass {
   private Reporter reporter;
   private final CompileOptions ops;
-  private Global _compiler = null;
+  private Global compiler = null;
 
   public ReportableMainClass(CompileOptions ops) {
     this.ops = ops;
   }
 
-  public void Close() throws Exception{
-    if(_compiler != null){
+  public void close() throws Exception{
+    if(compiler != null){
 
       //nsc.Global didn't inherit from Closeable until 2.12.9.
-      if(_compiler instanceof AutoCloseable){
-        ((AutoCloseable)_compiler).close();
+      if(compiler instanceof AutoCloseable){
+        ((AutoCloseable)compiler).close();
       }
-      _compiler = null;
+      compiler = null;
     }
   }
 
@@ -44,8 +44,8 @@ public class ReportableMainClass extends MainClass {
 
     reporter = new DepsTrackingReporter(settings, ops, reporter);
 
-    _compiler = new Global(settings, reporter);
-    return _compiler;
+    compiler = new Global(settings, reporter);
+    return compiler;
   }
 
   private void createDiagnosticsFile() {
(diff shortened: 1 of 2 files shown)


src/java/io/bazel/rulesscala/scalac/ReportableMainClass.java
```diff
@@ -11,15 +11,28 @@
 import scala.tools.nsc.MainClass;
 import scala.tools.nsc.Settings;
 import scala.tools.nsc.reporters.Reporter;
+import java.lang.AutoCloseable;
 
 public class ReportableMainClass extends MainClass {
   private Reporter reporter;
   private final CompileOptions ops;
+  private Global _compiler = null;
```
liucijus, line 19: nit: uncommon in Java style to prefix private vars with underscore - please rename to `compiler`

src/java/io/bazel/rulesscala/scalac/ReportableMainClass.java
```diff
@@ -11,15 +11,28 @@
...
 
   public ReportableMainClass(CompileOptions ops) {
     this.ops = ops;
   }
 
+  public void Close() throws Exception{
```
liucijus, line 25: style nit: don't start method names with a capital letter. Please rename to `close()`

liucijus (changes requested):
Thanks, @crt-31! Please address style related comments

src/java/io/bazel/rulesscala/scalac/ReportableMainClass.java
crt-31: done.

src/java/io/bazel/rulesscala/scalac/ReportableMainClass.java
crt-31: done

liucijus (approved):

PR #1505, opened 2023-07-14
Register java toolchain inside rules_scala

kotlaja:
### Description
Any rule which uses java_common should register java toolchain_type.


liucijus (approved):
Thanks, @kotlaja!

PR #1506, opened 2023-07-24
Update minimal supported Bazel version to 6.3.0

liucijus:
Bump supported Bazel to `6.3.0`

We need to bump Bazel version to ease dealing with backwards incompatible changes from 6 and future version.

Motivation:
- Update dependency tracking patterns, which got affected with the release of 6
- Prepare for bzlmod changes, in case someone volunteers to work on it


simuons (approved):
👍

PR #1509, opened 2023-08-25
Add support for netrc in jvm_maven_import_external

dmivankov:
### Description

`auth=` parameter for `repository_ctx.download` is both required to fetch from private registries and is not well-documented https://github.com/bazelbuild/bazel/issues/13709#issuecomment-1336699672

With this change
```python
rules_scala_toolchain_deps_repositories(
    maven_servers = ["some_private_artifactory_url"]
)
```
would be able to authenticate using default .netrc (like `~/.netrc`)

### Motivation
Private maven registry could be useful to host patched versions of libraries or to ensure that there's a private persistent mirror for those artifacts


liucijus (approved):
Thanks, @dmivankov!

Issue #1514, opened 2023-09-22
[Scala 2.13.12] Issue with scala.tools.nsc.reporters.Reporter when compiling


scoquelin:
Seems like this [change](https://github.com/scala/scala/pull/10406/files#diff-18816e2aeb44e75b443a1aa106af41dba1cc020090b093d1c0800618dd9581ccR117) in the Scala codebase has broken compilation of `DepsTrackingReporter.java` since it introduces a new parameter (with a default value) to existing error/warning methods that Java cannot interpret seamlessly.

Here is the stacktrace I get when compiling with Scala `2.13.12`  (builds fine with `2.13.11` btw)

```
ERROR: /private/var/tmp/_bazel_sebastien/beeef587c4e635c319809c3cfc50320b/external/io_bazel_rules_scala/src/java/io/bazel/rulesscala/scalac/reporter/BUILD:5:13: Building external/io_bazel_rules_scala/src/java/io/bazel/rulesscala/scalac/reporter/libreporter.jar (2 source files) [for tool] failed: (Exit 1): java failed: error executing command (from target @io_bazel_rules_scala//src/java/io/bazel/rulesscala/scalac/reporter:reporter) external/remotejdk11_macos_aarch64/bin/java -XX:-CompactStrings '--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED' '--add-exports=jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED' ... (remaining 18 arguments skipped)
warning: [path] bad path element "/private/var/tmp/_bazel_sebastien/beeef587c4e635c319809c3cfc50320b/execroot/platform/bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/external/io_bazel_rules_scala_scala_compiler/io_bazel_rules_scala_scala_compiler.stamp/scala-reflect.jar": no such file or directory
warning: [path] bad path element "/private/var/tmp/_bazel_sebastien/beeef587c4e635c319809c3cfc50320b/execroot/platform/bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/external/io_bazel_rules_scala_scala_compiler/io_bazel_rules_scala_scala_compiler.stamp/scala-library.jar": no such file or directory
external/io_bazel_rules_scala/src/java/io/bazel/rulesscala/scalac/reporter/DepsTrackingReporter.java:220: error: method warning in class Reporter cannot be applied to given types;
[...]

PR #1515, opened 2023-09-28
Upgrade scala 2.12.x and 2.13.x versions

scoquelin:
### Description
Bumping supported Scala 2.12 and 2.13 minor versions

Found a compilation issue related to `kind_projector` so updated to latest artifacts as well for all Scala major versions.

### Motivation
I wanted to level up minor Scala versions before trying to tackle #1514 that has breaking changes related to Scala `2.13.12`

### DIFF
--- a/test/coverage_scalatest/expected-coverage.dat
+++ b/test/coverage_scalatest/expected-coverage.dat
@@ -9,15 +9,15 @@ FNDA:1,coverage_scalatest/A1$::a1 (Z)Lcoverage_scalatest/B1$;
 FNDA:0,coverage_scalatest/A1::a1 (Z)Lcoverage_scalatest/B1$;
 FNF:4
 FNH:3
-BRDA:5,0,0,0
-BRDA:5,0,1,1
+BRDA:5,0,0,1
+BRDA:5,0,1,0
 BRF:2
 BRH:1
 DA:3,1
 DA:5,1
-DA:6,1
+DA:6,0
 DA:7,1
-LH:4
+LH:3
 LF:4
 end_of_record
 SF:test/coverage_scalatest/A2.scala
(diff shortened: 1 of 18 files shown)


test/src/main/scala/scalarules/test/compiler_plugin/KindProjected.scala
```diff
@@ -3,4 +3,4 @@ package scalarules.test.compiler_plugin
 import scala.language.higherKinds
 
 class HKT[F[_]]
-class KKTImpl extends HKT[Either[String, ?]]
\ No newline at end of file
+class KKTImpl extends HKT[Either[String, *]]
```
scoquelin, line 6: This [commit](https://github.com/scala/scala/pull/10005/files) has broken compilation for this file since Scala `2.12.16` where `?` now has a different purpose in Scala 3. Found out that reverting to `*` was the simplest workaround to avoid bringing `-Xsource:3` compiler flags just yet.

See https://docs.scala-lang.org/scala3/guides/migration/plugin-kind-projector.html

test/coverage_specs2_with_junit/expected-coverage.dat
```diff
@@ -9,15 +9,15 @@ FNDA:1,coverage_specs2_with_junit/A1$::a1 (Z)Lcoverage_specs2_with_junit/B1$;
...
+BRDA:5,0,1,0
 BRF:2
 BRH:1
 DA:3,1
 DA:5,1
-DA:6,1
+DA:6,0
 DA:7,1
-LH:4
+LH:3
```
scoquelin, line 20: Not sure where the changes in the coverage files comes from but after generating and inspecting coverage reports new changes are definitely more accurate as the failure case for `A1.a1(false)` was actually not covered and is now depicted as a missing branch in the report.

Before : 

expected-coverage-scalatest

After :

test/coverage_scalatest/expected-coverage.dat
```diff
@@ -9,15 +9,15 @@ FNDA:1,coverage_scalatest/A1$::a1 (Z)Lcoverage_scalatest/B1$;
...
+BRDA:5,0,1,0
 BRF:2
 BRH:1
 DA:3,1
 DA:5,1
-DA:6,1
+DA:6,0
 DA:7,1
-LH:4
+LH:3
```
scoquelin, line 20: Not sure where the changes in the coverage files comes from but after generating and inspecting coverage reports new changes are definitely more accurate as the failure case for `A1.a1(false)` was actually not covered and is now depicted as a missing branch in the report.

Before : 

expected-coverage-scalatest

After :

actual-coverage-scalatest

simuons (approved):
Thanks @scoquelin

liucijus (approved):

PR #1516, opened 2023-09-28
JacocoRunner script: update for Bazel 6.0+

gergelyfabian:
Added a parameter to specify the Bazel major version (to have a single script for Bazel 5 and 6).
Manually tested on a repo with both Bazel versions.

Had to add a new patch file for Bazel 6.
Added new Bazel branch with the Jacoco interface implementation, updated for Bazel 6.3.2.

Updated manual test_coverage_with_local_jacocorunner to run the script with Bazel 6 provided as a parameter (it was failing after the repo's upgrade to Bazel 6).


manual_test/coverage_local_jacocorunner/test.sh
```diff
@@ -24,7 +24,8 @@ test_coverage_with_local_jacocorunner() {
 }
 
 build_local_jacocorunner() {
-    $root_dir/scripts/build_jacocorunner/build_jacocorunner_bazel_5.0+.sh
+    # Build for Bazel 6 (as this is the default Bazel version in the repo).
+    $root_dir/scripts/build_jacocorunner/build_jacocorunner_bazel_5.0+.sh 6
```
simuons, line 28: Sine this script now takes a bazel version as a parameter, maybe you could remove version from file name? I think it would be a bit cleaner.

simuons (changes requested):
LGTM just one minor comment. Thanks @gergelyfabian

manual_test/coverage_local_jacocorunner/test.sh
gergelyfabian: I mean, there are two scripts. One for Bazel 4.* and another for Bazel 5+ (5 and 6).
Should I remove the Bazel 4.* script maybe? (or rename it to the 4.0 suffix)

gergelyfabian: And then this new script can have a name without the version suffix.

gergelyfabian: The issue was that the script for Bazel 4.* is significantly different (including comments), so I was hesitant to unite the script for the three Bazel versions supported.

gergelyfabian: I'll rename the script for Bazel 4.* to have a version suffix in the filename, and rename the Bazel 5.0+6.0 script to be the "default" (without version suffix). It should be more obvious this way (especially as Bazel 6.0 is the minimum supported version in rules_scala currently).

gergelyfabian: Done.

gergelyfabian:
> LGTM just one minor comment. Thanks @gergelyfabian

I think the issue is now fixed.

liucijus (approved):

PR #1524, opened 2023-11-03
Bazel 7.0 rc2 support

gergelyfabian:
Added `@bazel_tools//tools/jdk:toolchain_type` toolchain for scala_junit_test: https://github.com/bazelbuild/bazel/issues/18970

Changed deprecated transitive_runtime_deps to transitive_runtime_jars in jmh.bzl:
https://bazel.build/versions/6.4.0/rules/lib/JavaInfo#transitive_runtime_deps

With these changes rules_scala (for the cases I checked) started working with Bazel 7.0 rc2.

### DIFF
--- a/jmh/jmh.bzl
+++ b/jmh/jmh.bzl
@@ -39,7 +39,7 @@ def _scala_generate_benchmark(ctx):
 
     # just try to take the first one and see if that works
     class_jar = outs[0].class_jar
-    classpath = info.transitive_runtime_deps
+    classpath = info.transitive_runtime_jars
     ctx.actions.run(
         outputs = [ctx.outputs.src_jar, ctx.outputs.resource_jar],
         inputs = classpath,
--- a/scala/private/rules/scala_junit_test.bzl
+++ b/scala/private/rules/scala_junit_test.bzl
@@ -136,7 +136,10 @@ def make_scala_junit_test(*extras):
             *[extra["outputs"] for extra in extras if "outputs" in extra]
         ),
         test = True,
-        toolchains = ["@io_bazel_rules_scala//scala:toolchain_type"],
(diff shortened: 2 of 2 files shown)


liucijus (approved):
Thanks, @gergelyfabian!

gergelyfabian:
There is still one thing that doesn't work for Bazel 7.0rc2. When running a target like this:

```
bazel run //utils/smth:library.format-test
```

I receive:

```
utils/smth/src/main/scala/com/dummy/smth/Smth.scala
diff: /home/gege/.cache/bazel/_bazel_user/7f83a0b1f0940f88aec1f4530f927169/execroot/repo/bazel-out/k8-fastbuild/bin/utils/smth/utils/smth/src/main/scala/com/dummy/smth/Smth.scala.fmt.output: No such file or directory
```

It seems the action responsible for creating this file executes (I have tested disabling writing the output there, but then obviously it fails due to not generating an output), but at a later step the action running scalafmt from the manifest cannot see the scalafmt outputs.

gergelyfabian:
One difference I can see in the bazel-bin direction between Bazel 6.4.0 and Bazel 7.0rc2 (after running `bazel run //utils/smth:library.format-test`):

```
# 6.4.0:
ls bazel-bin/utils/smth/utils/smth/src/main/scala/com/dummy/smth/
Smth.scala.fmt.output
# 7.0rc2:
ls bazel-bin/utils/smth/utils/smth/src/main/scala/com/dummy/smth/
[empty]
```

simuons (approved):

gergelyfabian:
Opened an issue for the scalafmt error:

https://github.com/bazelbuild/rules_scala/issues/1525

PR #1532, opened 2023-12-02
Fix diagnostic outputs for Scala 2.12.13

aishfenton:
### Description
Adds the correct overrides in `DepsReporter` to keep diagnostic reporting working in Scala `2.12.13 <= x < 2.13.12`.

This is due to [SemanticdbReporter](https://github.com/scalameta/scalameta/blob/9712a5b217351323b06ea57b9dc1ce8b79a812f5/semanticdb/scalac/library/src/main/scala-2.12/scala/meta/internal/semanticdb/scalac/SemanticdbReporter.scala#L7) calling `doReport` on it's underlying reporter, rather `info0`.  

### Motivation
Currently if you have SemanticDB enabled, and are using Scala `2.12.13 <= x < 2.13.12`, you get empty diagnostic report protos.  

Related to this issue: https://github.com/bazelbuild/rules_scala/issues/1484


eed3si9n:
Is the current test coverage (https://github.com/bazelbuild/rules_scala/tree/master/test/shell has a bunch of shell scripts) missing the test for Scala 2.12.x on this issue?

aishfenton:
@eed3si9n I think it doesn't test the combination of Semanticdb +  Diagnostics. I can add something in to test that combo. 

Harder to test is the combinations of < 2.12.13 and >= 2.12.13, as the repo's default dep_providers are for 2.12.18.

aishfenton:
And  cc @scoquelin, since this is a follow up to  https://github.com/bazelbuild/rules_scala/pull/1522

src/java/io/bazel/rulesscala/scalac/deps_tracking_reporter/BUILD
```diff
@@ -3,8 +3,10 @@ load("@io_bazel_rules_scala_config//:config.bzl", "SCALA_MAJOR_VERSION", "SCALA_
 filegroup(
     name = "deps_tracking_reporter",
     srcs = [
-        "before_2_13_12/DepsTrackingReporter.java",
-    ] if SCALA_MAJOR_VERSION.startswith("2.11") or SCALA_MAJOR_VERSION.startswith("2.12") or (SCALA_MAJOR_VERSION.startswith("2.13") and int(SCALA_MINOR_VERSION) < 12) else [
+        "before_2_12_13/DepsTrackingReporter.java",
```
liucijus, line 6: Please update `startswith` with `==` as full check is required here

aishfenton:
@liucijus I'm adding tests for the cross-product of versions having semanticdb + diagnostics enabled. But I'm finding that having semanticdb enabled also changes the lines numbers that are output (I have no idea how... they're still correct just more similar to 2.13's way of doing it). 

The cross product of (version X semanticdb) is getting a bit nuts to test for. I'm tempted to remove the lines numbers from `DiagnosticsReporterTest` and instead test of the message contents. I'm thinking that the specific lines numbers output are just too unstable between plugins/versions to test against. How would you feel about that?

aishfenton:
Added that change in to support cross-testing @liucijus, but let me know if you want to change it.

liucijus (approved):

simuons (approved):

PR #1537, opened 2024-01-15
Remove binary artifacts due to security policy violation

mateuszkuta256:
### Description
This PR removes .jar files that trigger security violation alert

### Motivation
Resolves issue created by Allstar bot https://github.com/bazelbuild/rules_scala/issues/1458


test/src/main/scala/scalarules/test/scala_import/BUILD
```diff
@@ -36,14 +28,6 @@ scala_specs2_junit_test(
...
 )
 
-scala_specs2_junit_test(
-    name = "scala_import_exposes_file_jars",
-    size = "small",
-    srcs = ["ScalaImportExposesFileJarsTest.scala"],
-    suffixes = ["Test"],
-    deps = [":relate"],
-)
-
```
mateuszkuta256, line 46: this is about testing core bazel functionality introduced [here](https://github.com/mateuszkuta256/rules_scala/commit/efffa526a31724803532b46e659a7ea3240b9509)
thus I think it's not crucial to have such a test. Dropping it allows us to get rid of .jar file

mateuszkuta256:
I close due to failed cla check

PR #1538, opened 2024-01-16
Remove binary artifacts due to security policy violation

mateuszkuta256:
Description
This PR removes .jar files that trigger security violation alert

Motivation
Resolves issue created by Allstar bot https://github.com/bazelbuild/rules_scala/issues/1458


test/src/main/scala/scalarules/test/scala_import/BUILD
```diff
@@ -36,14 +28,6 @@ scala_specs2_junit_test(
...
 )
 
-scala_specs2_junit_test(
-    name = "scala_import_exposes_file_jars",
-    size = "small",
-    srcs = ["ScalaImportExposesFileJarsTest.scala"],
-    suffixes = ["Test"],
-    deps = [":relate"],
-)
-
```
mateuszkuta256, line 46: this is about testing core bazel functionality introduced [here](https://github.com/mateuszkuta256/rules_scala/commit/efffa526a31724803532b46e659a7ea3240b9509)
thus I think it's not crucial to have such a test. Dropping it allows us to get rid of .jar file

liucijus: @mateuszkuta256 this tests if `scala_import` is working correctly, why do you think it is not needed?

mateuszkuta256: this test uses the "relate_2.11-2.1.1.jar" file that I want to get rid of because of security alerts
I replaced jar files with "scala_import" wherever possible, but for this case it's not possible, because the test is exactly about importing .jar file
according to to git history it was introduced along with [this](https://github.com/mateuszkuta256/rules_scala/commit/efffa526a31724803532b46e659a7ea3240b9509) single line change: `"jars": attr.label_list(allow_files=True)`
looks to me like testing core bazel functionality rather than scala import rule itself, so I think we can drop the test so we could drop  "relate_2.11-2.1.1.jar" too

liucijus: what will happen if someone accidentally removes `allow_files=True`from `scala_import` implementation?

mateuszkuta256: in such a case the project won't build anymore
[repository def](https://github.com/mateuszkuta256/rules_scala/blob/12d60d203591d92572c812f345b45babff688230/third_party/repositories/repositories.bzl#L59) leads to [this](https://github.com/mateuszkuta256/rules_scala/blob/0391ef4e43fecac5b63518082294e2151f74aeb4/scala/scala_maven_import_external.bzl#L202) clause which generates BUILD files:
```
# DO NOT EDIT: generated by jvm_import_external()

load("@io_bazel_rules_scala//scala:scala_import.bzl", "scala_import")

licenses(["notice"])

scala_import(
    name = "io_bazel_rules_scala_scala_xml",
    jars = ["scala-xml_2.12-1.2.0.jar"],
    srcjar = "scala-xml_2.12-1.2.0-src.jar",
)
```
`jars = ["scala-xml_2.12-1.2.0.jar"]` does not compile then
of course would be great to have an explicit test... but I lack an idea how to do that without putting .jar into the codebase

liucijus: such test may look redundant because we already implicitly test that code in Rules Scala by depending on that functionality in our codebase ourselves, but:
- it's easier to develop such features when there's a small isolated test case ava
- at some point we may change internal dependency on `scala_import` and it won't be implicitly tested anymore

liucijus (changes requested):
@mateuszkuta256 thanks for taking this!

Please check if using implicit jar output can be used in places where original external jar was imported, like in this [example](https://github.com/bazelbuild/rules_scala/pull/1538#discussion_r1469506990)

test/src/main/scala/scalarules/test/scala_import/BUILD
mateuszkuta256: thx! your solution works well. I also removed the very last `thirft3_scrooge.jar`

liucijus: @mateuszkuta256 please push your changes

mateuszkuta256: ah, pushed to another branch

test/example_jars/expected_service_manifest.txt
```diff
@@ -1,2 +1,2 @@
 org.apache.beam.sdk.extensions.gcp.storage.GcsFileSystemRegistrar
-org.apache.beam.sdk.io.LocalFileSystemRegistrar
+org.apache.beam.sdk.extensions.gcp.storage.GcsFileSystemRegistrar
```
liucijus, line 2: Please leave the original behavior with two different services

test/src/main/scala/scalarules/test/resource_jars/BUILD
```diff
@@ -1,7 +1,19 @@
...
+        "//test/src/main/resources/scalarules/test:more-hellos",
+    ],
+    visibility = ["//visibility:public"],
+)
+
 scala_test(
     name = "resource_jars",
     srcs = ["TestResourceJars.scala"],
-    resource_jars = ["//test/src/main/resources/scalarules/test:hellos-and-byes.jar"],
+    resource_jars = [":jarWithResources"],
```
liucijus, line 18: please refer explicitly to `.jar` output - it will be easier to follow test meaning

test/src/main/scala/scalarules/test/scala_import/nl/BUILD.bazel
```diff
@@ -1,10 +1,17 @@
...
+    name = "ScalaImportNeverLink",
+    srcs = ["ScalaImportNeverLink.java"],
+    visibility = ["//visibility:public"],
+)
 
 scala_import(
     name = "scala_import_never_link",
     jars = [
-        "scala_import_never_link.jar",
+        ":ScalaImportNeverLink",
```
liucijus, line 14: please explicitly depend on `.jar` output

test/src/main/scala/scalarules/test/twitter_scrooge/thrift/bare_jar_thrifts/BUILD
```diff
@@ -1,17 +1,8 @@
...
-    jars = ["bare-thrift.jar"],
-    visibility = ["//visibility:public"],
-)
-
 thrift_library(
     name = "bare_jar_thrifts",
-    external_jars = [
-        ":barejar",
-    ],
+    srcs = ["//test/src/main/resources/scalarules/test:Foo.thrift"],
```
liucijus, line 5: I think this changes the meaning of the test (test *bare* `external_jar` jar)

test/src/main/scala/scalarules/test/twitter_scrooge/thrift/bare_jar_thrifts/bare_jar_1/BUILD
```diff
@@ -2,8 +2,6 @@ load("//thrift:thrift.bzl", "thrift_library")
 
 thrift_library(
     name = "bare_jar_1",
-    external_jars = [
-        "bare-thrift-1.jar",
-    ],
+    srcs = ["//test/src/main/resources/scalarules/test:Bar.thrift"],
```
liucijus, line 5: same as above. Do you know why the second test is needed?

test/src/main/scala/scalarules/test/twitter_scrooge/thrift/bare_jar_thrifts/bare_jar_2/BUILD
```diff
@@ -2,8 +2,6 @@ load("//thrift:thrift.bzl", "thrift_library")
 
 thrift_library(
     name = "bare_jar_2",
-    external_jars = [
-        "bare-thrift-2.jar",
-    ],
+    srcs = ["//test/src/main/resources/scalarules/test:Baz.thrift"],
```
liucijus, line 5: same as above

test/src/main/scala/scalarules/test/twitter_scrooge/thrift/thrift2/thrift3/BUILD
```diff
@@ -9,7 +9,7 @@ thrift_library(
 
 scrooge_scala_import(
     name = "thrift3_import",
-    scala_jars = ["thrift3_scrooge.jar"],
+    scala_jars = ["//test/src/main/scala/scalarules/test/twitter_scrooge:scrooge3"],
```
liucijus, line 12: I'd like a comment (in the code) here about what's going on. Just by looking at the test is very hard to understand what's going on and how is it being tested.

liucijus (changes requested):
@mateuszkuta256 I think we are very close to get this merged, please address the comments. Thanks for the patience.

test/src/main/scala/scalarules/test/twitter_scrooge/thrift/bare_jar_thrifts/bare_jar_1/BUILD
mateuszkuta256: now I reference .jar directly
the second test is meant for test_version.sh, apparently every test running against various versions has its copy there

test/src/main/scala/scalarules/test/twitter_scrooge/thrift/thrift2/thrift3/BUILD
mateuszkuta256: kk, hope the comment I added is useful
thx for your patience too ;D

liucijus (approved):
🥇

simuons (approved):

Issue #1521, opened 2023-10-26
rules_scala support for JDK 21


gergelyfabian:
When running rules_scala (trying to compile some Scala code) with JDK 21, I receive the following exception:

```
Exception in thread "main" java.lang.UnsupportedOperationException: The Security Manager is deprecated and will be removed in a future release
	at java.base/java.lang.System.setSecurityManager(System.java:429)
	at io.bazel.rulesscala.worker.Worker.persistentWorkerMain(Worker.java:47)
	at io.bazel.rulesscala.worker.Worker.workerMain(Worker.java:39)
	at io.bazel.rulesscala.scalac.ScalacWorker.main(ScalacWorker.java:36)
```

Reproduced on Linux with:

```
# WORKSPACE:
# Idea from: https://gist.github.com/ghostbear/b0aa3bf1a323b38be98bb1a66db65b41
# Link can be found at: https://www.azul.com/downloads/?package=jdk#zulu
remote_java_repository(
    name = "java_21_repository",
    prefix = "zulujdk",
    sha256 = "bf4842dd3a17cfe85523be5848b5ec3bc3d811afc74feab791befa4c895c4449",
    strip_prefix = "zulu21.30.15-ca-jdk21.0.1-linux_x64",
    target_compatible_with = ["@platforms//os:linux"],
    urls = ["https://cdn.azul.com/zulu/bin/zulu21.30.15-ca-jdk21.0.1-linux_x64.tar.gz"],
    version = "21",
)

register_toolchains("//tools/jdk:java21_toolchain_definition")

# //tools/jdk:BUILD

default_java_toolchain(
    name = "java21_toolchain",
    configuration = DEFAULT_TOOLCHAIN_CONFIGURATION,
    java_runtime = "@java_21_repository//:jdk",
    javacopts = DEFAULT_JAVACOPTS,
    # jvm_opts needs to stay this one also for Java 11, it seems to be common for Java 9+ versions.
    # It includes e.g. Java module system options.
    jvm_opts = BASE_JDK9_JVM_OPTS,
    source_version = "21",
    target_version = "21",
    visibility = ["//visibility:public"],
)
```

gergelyfabian:
Seems this exception is starting with JDK 18 (before it was only a warning):

https://bz.apache.org/bugzilla/show_bug.cgi?id=65381

gergelyfabian:
Trying to localize this, it seems to occur when trying to compile any Scala code (tried `scala_macro_library`).

[...]

PR #1539, opened 2024-01-17
Draft: Enable security manager for ScalacWorker to allow using JDK 18+

thirtyseven:
### Description

ScalacWorker.java uses the Java Security Manager to trap calls to System.exit(). In JDK 18+, enabling the security manager is disallowed by default, so a flag must be passed to allow it.

### Motivation

When building with `--tool_java_runtime=remotejdk_21`, I found that this flag was needed when using Scalac persistent workers otherwise they would crash with an `UnsupportedOperationException`.


thirtyseven:
Hm, looks like this flag is not backwards compatible with earlier JVMs, maybe some kind of version detection is needed.

srdo:
@thirtyseven Thanks for looking into this. 

Just to help you along, we've been discussing this issue over [here](https://github.com/bazelbuild/rules_scala/issues/1521#issuecomment-1789002782).

I think the version detection you mention has become available in `JavaRuntimeInfo`, there's an example here of how to interact with it https://github.com/bazelbuild/bazel/commit/7556e1107b666d10b660470a571631463c7eb4ec#diff-ece537407bf4ad71e14df88b62cc07ad0666e3938a3422d91e8714962c004d48R57.

simuons:
Hi, @thirtyseven,  have you looked at what @srdo suggested? Maybe that would work?

But in general, since jdk does not provide any alternative for exit code trapping is see two options how we could go here:

- remove security manager as it will go away in the future. Also not sure this exit code trapping saves us from anything (need more opinions on this)
- configure `-Djava.security.manager=allow` on java toolchain level. I'll try to come up with an example and probably note in a readme (this means a little bit of inconvenience for users since preconfigured java toolchains doesn't work by default)

Frankly I'm not a big fan of ifs by java version which might break in future java releases.

WDYT would extra java toolchain configuration be a big hurdle?

srdo:
@simuons I don't understand why adjusting the Java toolchain would be the only option other than removing the exit trap? Can't we detect the Java version from Starlark in the way I suggested, and add `-Djava.security.manager=allow` to worker/other command invocations based on that (for example set the flag if the Java version `>= 12`?

I think that would keep basic toolchains working, and not cause any problems for existing users. It should keep working until the SM is actually removed from the JDK, and at that point there will hopefully be a replacement.

My understanding is that not trapping exits means that any user code (e.g. tests) that invokes `System.exit` will also cause the Scala worker to exit. Not sure what the consequences are of that?

simuons:
This was resolved by https://github.com/bazelbuild/rules_scala/pull/1556

PR #1543, opened 2024-01-26
Bump scalafmt version

mateuszkuta256:
### Description
Bump scalafmt version

### Motivation
With older version formatting didn't work for scala3

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,8 +1,8 @@
 align.openParenCallSite = false
 align.openParenDefnSite = false
 continuationIndent.defnSite = 2
-danglingParentheses = true
-docstrings = JavaDoc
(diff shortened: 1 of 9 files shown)


.scalafmt.conf
```diff
@@ -1,8 +1,8 @@
 align.openParenCallSite = false
 align.openParenDefnSite = false
 continuationIndent.defnSite = 2
-danglingParentheses = true
-docstrings = JavaDoc
+danglingParentheses.preset = true
+docstrings.style = Asterisk
```
mateuszkuta256, line 5: not really sure if  release note or 'incompatible change' action is necessary
in newer versions of scalafmt properties are a bit different

eed3si9n (approved):
Thanks!

PR #1529, opened 2023-11-30
Make tests pass on Windows

crt-31:
### Description

Bunch of changes. Here's some highlights:

Code Changes
- added lineEndings=preserve to scalafmt.conf
- added .bazelrc to enable needed flags for Windows
- Refactored dependency-analyzer plugin's javaargs to use ‘;’ as a delimiter instead of ‘:’. (This made things simpler and easier since ':' as drive separator on windows and ':' is normal in a Bazel label.)
- Refactored ScalacWorker to abstract common code between Scala 2 & 3. (This consolidates some duplicated code that needed to be changed). 
- Renamed a couple function names containing rpathlocation to rlocationpath

Test Changes
- disable running tests for features not-supported on Windows
- make tests handle fact that sh_binary outputs .exe on Windows
- Handle other Windows specifics where needed (i.e. backslash as pathseparator, crlf linendings, dependence on bash)

Other notes:
- It expects git's core.autocrlf=true on windows. 

### Motivation
Keeping the tests working on Windows greatly helps ensure continuous compatibility with Windows OS.

PR #1545, opened 2024-02-12
Fix to close the compiler object in scalacinvoker 

crt-31:
### Description
This fixes functionality that was accidently broken in PR #1529   (my bad).

The fix is to call close on the compiler within the scalaCWorker (originally added in PR #1504)

### DIFF
--- a/src/java/io/bazel/rulesscala/scalac/ScalacInvoker.java
+++ b/src/java/io/bazel/rulesscala/scalac/ScalacInvoker.java
@@ -32,7 +32,7 @@ class ScalacInvoker{
       } else {
         throw ex;
       }
-    }finally{
+    } finally {
       comp.close();
     }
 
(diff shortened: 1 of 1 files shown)


simuons (changes requested):
Please fix code formatting. Also wondering maybe same changes needs to be done in `ScalacInvoker3`?

crt-31:
@simuons 
> Please fix code formatting. 

Done

> Also wondering maybe same changes needs to be done in `ScalacInvoker3`?

I looked at dotty's sourcecode, and they don't have a close() mechanism like Scala2 does yet.  There is a PR written up for it  https://github.com/lampepfl/dotty/issues/5412.

Suggest we push this PR through, and deal with Scala3 when dotty implements that PR.

liucijus (approved):

simuons (approved):

Issue #1290, opened 2021-09-22
FR: Support multiple scala language versions in one workspace


virusdave:
Not a new desire or request, of course.  See this [prior issue](https://github.com/bazelbuild/rules_scala/issues/962) as an example.

We'd really like to be able to build different targets with different versions of Scala to make language version upgrades sane & realistic in large codebases.  Atomically changing everything at once gets really difficult as codebases scale, especially with large external dependencies like Spark.

liucijus:
@virusdave do you have any ideas how should multiple version support look from a user side? I think having Scala targets that build for different versions is involving, but not hard to achieve (fix compiler bootstrapping, toolchains and parametrize rules), what is hard how do we configure and provide deps for different Scala versions. I see a few challenges related to deps:
- how to integrate with loaders like rules_jvm_external and others, which assume single Maven artifact for Scala deps
- not breaking tools and IDEs compatibility

lukaszwawrzyk:
@liucijus We want to work on supporting multiple scala versions in one build. I was looking into rules_python implementation and I hope that this approach can be implemented into rules_scala. It has some utility macros which allows to specify which versions you want to use, which version is the default and it generates a repository for each version. The rules are wrapped in transition that changes python version. User can import py_* rules from repository with a specific version.
This approach would at least enable us to create separate, unrelated targets that build for different version of scala in one build. In this basic approach, I believe users could specify each library with required version in rules_jvm_external and just depend on it. For example cats-core_2.13 and cats-core_3 and explicitly depend on right version when specifying targets.
Assuming shared targets that can be built with multiple versions of scala, this could be archieved with macros. For example `scala_cross_library(name = "a", deps = ["@maven//:org_typelevel_cats_core_{scala_version}"])`. Similarly a macro could help with adding 3rd party deps for multiple scala versions.
As for IDE support, I think that during import we could force to only use targets with default scala version and it should be good enough to work with existing tooling. I think that having separate targets with different scala versions that are not crossbuilt could be supported in the IDE. Targets with shared sources could be messy but I think we could somehow handle it.
Maybe more involved idea would be to only do target transition on _binary and _test targets and _library targets would be transitioned through the deps field of _binary/_test targets. The library targets would use `select` to configure themselves based on version that they are used with. It could also be wrapped in a convenience macros.
It still should play well with the IDE if we do import using default scala version.

Also, the single scala version in the repo would look as usual, using rules without transitions, so the simple use case would not break.

[...]

PR #1546, opened 2024-02-13
Cross build support

mateuszkuta256:
### Description
This PR allows the registration of multiple toolchains for various scala versions. 
Extends both `scala_library` and `scala_binary` with `scala_version` parameter which allows to pick proper toolchain
Detailed explanation [here](https://github.com/bazelbuild/rules_scala/issues/1290#issuecomment-1924182988)

I am already researching the next steps:
- support scalafmt for multiple scala versions
- fix glitches with BSP
- extend scala_test rule too

Gonna add some documentation when the API is approved

### Motivation
https://github.com/bazelbuild/rules_scala/issues/1290

### DIFF
--- a/scala/private/common_attributes.bzl
+++ b/scala/private/common_attributes.bzl
@@ -90,6 +90,30 @@ implicit_deps = {
         default = Label("@io_bazel_rules_scala//src/java/io/bazel/rulesscala/scalac"),
         allow_files = True,
     ),
+    "_scalac_before_2_12_13": attr.label(
+        executable = True,
+        cfg = "exec",
+        default = Label("@io_bazel_rules_scala//src/java/io/bazel/rulesscala/scalac:scalac_before_2_12_13"),
+        allow_files = True,
+    ),
+    "_scalac_after_2_12_13_and_before_2_13_12": attr.label(
+        executable = True,
+        cfg = "exec",
+        default = Label("@io_bazel_rules_scala//src/java/io/bazel/rulesscala/scalac:scalac_after_2_12_13_and_before_2_13_12"),
+        allow_files = True,
+    ),
+    "_scalac_after_2_13_12": attr.label(
+        executable = True,
+        cfg = "exec",
+        default = Label("@io_bazel_rules_scala//src/java/io/bazel/rulesscala/scalac:scalac_after_2_13_12"),
+        allow_files = True,
+    ),
+    "_scalac_3": attr.label(
+        executable = True,
+        cfg = "exec",
+        default = Label("@io_bazel_rules_scala//src/java/io/bazel/rulesscala/scalac:scalac_3"),
+        allow_files = True,
+    ),
     "_exe": attr.label(
         executable = True,
         cfg = "exec",
--- a/scala/private/phases/phase_compile.bzl
+++ b/scala/private/phases/phase_compile.bzl
@@ -18,6 +18,7 @@ load(
 )
 load(":resources.bzl", _resource_paths = "paths")
 load("@io_bazel_rules_scala_config//:config.bzl", "SCALA_VERSION")
+load("@io_bazel_rules_scala//scala:scala_cross_version.bzl", "extract_major_version", "extract_minor_version")
 
 buildijar_default_value = True if SCALA_VERSION.startswith("2.") else False
 
@@ -198,7 +199,6 @@ def _compile_or_empty(
         scala_srcs = _get_files_with_extension(ctx, _scala_extension)
         in_srcjars = _get_files_with_extension(ctx, _srcjar_extension)
         all_srcjars = depset(in_srcjars, transitive = [srcjars])
-
         sources = scala_srcs + java_srcs
         _compile_scala(
             ctx,
@@ -221,7 +221,7 @@ def _compile_or_empty(
             ctx.attr.expect_java_output,
             ctx.attr.scalac_jvm_flags,
             scalacopts,
-            ctx.executable._scalac,
+            _select_scalac(ctx),
             dependency_info,
             unused_dependency_checker_ignored_targets,
             additional_outputs,
@@ -390,3 +390,18 @@ def _interim_java_provider_for_java_compilation(scala_output):
         compile_jar = scala_output,
         neverlink = True,
     )
+
+def _select_scalac(ctx):
+    scala_version = ctx.toolchains["@io_bazel_rules_scala//scala:toolchain_type"].scala_version
+    if scala_version:
+        major_version = extract_major_version(scala_version)
+        minor_version = extract_minor_version(scala_version)
+        if major_version.startswith("2.11") or (major_version.startswith("2.12") and int(minor_version) < 13):
+            return ctx.executable._scalac_before_2_12_13
+        if ((major_version.startswith("2.12") and int(minor_version) >= 13) or (major_version.startswith("2.13") and int(minor_version) < 12)):
+            return ctx.executable._scalac_after_2_12_13_and_before_2_13_12
+        if (major_version.startswith("2.13") and int(minor_version) >= 12):
+            return ctx.executable._scalac_after_2_13_12
+        if (major_version.startswith("3")):
+            return ctx.executable._scalac_3
+    return ctx.executable._scalac
--- a/examples/crossbuild/BUILD
+++ b/examples/crossbuild/BUILD
@@ -2,14 +2,14 @@ load("@io_bazel_rules_scala//scala:scala.bzl", "scala_binary", "scala_library")
 
 scala_library(
     name = "hello211",
-    scala_version = "2.11.12",
     srcs = ["Hello211.scala"],
+    scala_version = "2.11.12",
 )
 
 scala_library(
     name = "hello213",
-    scala_version = "2.13.12",
     srcs = ["Hello213.scala"],
+    scala_version = "2.13.12",
 )
 
 # scala 3.3.1 is default for this workspace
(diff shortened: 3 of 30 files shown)


scala/private/common_attributes.bzl
```diff
@@ -90,6 +90,30 @@ implicit_deps = {
         default = Label("@io_bazel_rules_scala//src/java/io/bazel/rulesscala/scalac"),
         allow_files = True,
     ),
+    "_scalac_before_2_12_13": attr.label(
```
simuons, line 93: Hi, @mateuszkuta256, thanks for PR. Does that mean than every scala version needs to be setup in workspace for this toolchain to work? ie all flavours of scalac worker needs to be compiled?

mateuszkuta256: hey, let me explain this change in detail, as I am searching for an alternative solution anyway...
sources for `scalac` are defined [here](https://github.com/mateuszkuta256/rules_scala/blob/a02f738f84af2cfbf66520ceb437bc2f6f95dd6b/src/java/io/bazel/rulesscala/scalac/BUILD#L104)
these are conditional upon: `if SCALA_MAJOR_VERSION.startswith("2") `
so I simply made a few `scalac_...` targets, separate for 2.11, 2 and 3
would love to keep a single `scalac`, but missing an idea of how to achieve that...
if we don't specify e.g. scala 3.3.1 in a config setting then I believe _scalac3 is skipped (?)
at least the build is successful without complaining that scala3 deps are missing

simuons: Right, I see. At this point it's just a label and magic happens on action when you select `scalac`.

As for alternatives I was playing with toolchain setup where each scala version is encapsulated/scoped in it's own repository and compiles it's own `scalac` (similar to `remote_java_repository`). https://github.com/simuons/rules_scala/commit/919471ae6d04811b3b07a5ed134278f5681c71e9 Still not sure where this leads.

scala/private/phases/phase_compile.bzl
```diff
@@ -390,3 +390,18 @@ def _interim_java_provider_for_java_compilation(scala_output):
         compile_jar = scala_output,
         neverlink = True,
     )
+
+def _select_scalac(ctx):
```
simuons, line 394: Could this `scalac` selection be done on toolchain? That would keep implicit toolchain attributes and selection by version closer to each other and would simplify `phase_compile`.

scala/private/common_attributes.bzl
mateuszkuta256: I see you got rid of the circular dependency between toolchain and `scalac` caused by `toolchain_deps:scala_compile_classpath` dep. It is something that I was missing in my first drafts and allows to apply nice solutions 👍 
in the meantime, I tested something like this https://github.com/mateuszkuta256/rules_scala/commit/f247cc8a8d0ad249d78743f2059106a91a10a4c9
`_scalac` is removed from implicit attributes and attached to `scala_library` and other rules. Then we can make a version specific implementations, e.g.
`scala_3_3_binary = make_scala_binary(scala_version = "3.3.1")`
since you found a way to remove circular dependencies, the same thing can be applied directly on the toolchain [here](https://github.com/mateuszkuta256/rules_scala/blob/d8259b1907b0429cf4d4161088d5cbfbc2600251/scala/scala_toolchain.bzl#L177)
from the user perspective, it's even nicer, because can use regular scala_library without making version specific one

mateuszkuta256:
A brief summary:

- in this PR I define a few `_scalac` targets that are included in `common_attributes` and selected while 'compile phase' depending on `scala_version`:
```
def _select_scalac(ctx):
    scala_version = ctx.toolchains["@io_bazel_rules_scala//scala:toolchain_type"].scala_version
    if scala_version:
        major_version = extract_major_version(scala_version)
        minor_version = extract_minor_version(scala_version)
        if major_version.startswith("2.11") or (major_version.startswith("2.12") and int(minor_version) < 13):
            return ctx.executable._scalac_before_2_12_13
...
```
- the other solution is to build separate rules `scala_library`, `scala_binary`, etc. that encapsulate corresponding `_scalac`: https://github.com/mateuszkuta256/rules_scala/commit/f247cc8a8d0ad249d78743f2059106a91a10a4c9
```
scala_2_13_binary = make_scala_binary(scala_version = "2.13.12")
scala_3_1_binary = make_scala_binary(scala_version = "3.1.0")
```
this one works with BSP perfectly
- an update of https://github.com/bazelbuild/rules_scala/pull/1546#discussion_r1490748804:
https://github.com/mateuszkuta256/rules_scala/commit/674453ecf953c172cef9bf782a8b6ff082b3f21f
`scalac` targets are pushed into new repositories, so we get rid of circular dependencies and can add those targets directly to the toolchain. Then 'compile phase' does:
`ctx.toolchains["@io_bazel_rules_scala//scala:toolchain_type"].scalac`
instead of `ctx.executable._scalac`

@simuons @lukaszwawrzyk feel free to test and share your thoughts on which of these is superior

simuons:
Hi, @mateuszkuta256, sorry for delays. I'll try to take a deeper look on weekend.

mateuszkuta256:
> Hi, @mateuszkuta256, sorry for delays. I'll try to take a deeper look on weekend.

thx! FYI I am checking the remaining rules in the context of cross-build
there's a same problem for e.g. `scala_test` - parameter `_scalatest_reporter` is built differently, depending on `SCALA_VERSION`
what I propose is to make it a label_list, instead of passing few different attributes for each version https://github.com/mateuszkuta256/rules_scala/blob/356c088ccc8d0c0f98113d73afee25180f26f6e2/scala/private/rules/scala_test.bzl#L81
then, proper version can be selected like this: https://github.com/mateuszkuta256/rules_scala/blob/356c088ccc8d0c0f98113d73afee25180f26f6e2/scala/private/phases/phase_collect_jars.bzl#L22
and we don't define more targets than required, only one per scala_version requested by the user https://github.com/mateuszkuta256/rules_scala/blob/356c088ccc8d0c0f98113d73afee25180f26f6e2/scala/support/support.bzl#L98
we can consider something for `scalac` too

lukaszwawrzyk:
@simuons Can we do anything to help with the review? Or maybe someone else could also take a look?

At this point it would be good to take a look at the overall design and pick concrete solutions. If you'd feel that it is needed, we can write some doc to explain what happens here in more detail. Let us know if you'd like it or just the comments here and code is enough.

Once we are settled on the overall idea, we can try to split these changes into smaller PRs if it would be helpful.

simuons (changes requested):
This looks really promising and exciting. Probably transitions is the way to go (just need to be careful with [memory and performance considerations](https://bazel.build/extending/config#memory-performance-considerations))

However I think this change is quite big and should be split in few steps:
* As a first step I think we should move to toolchains and get rid of branching on `SCALA_VERSION` constant (I feel this needed for bzlmod as well)
* Then probably move some of the toolchains properties to build_settings meaning toolchains will contain only scala version specific stuff. Otherwise we will have a combinatorial explosion of scala versions and properties like `strict_deps`
* Then add transitions which should be considerably easier when having scala version selection by toolchain or build_setting (instead of env var)

simuons:
@lukaszwawrzyk I think doc makes a lot of sense because change is substantial.

> Once we are settled on the overall idea, we can try to split these changes into smaller PRs if it would be helpful.

Thanks for this comment. I was unsure do you consider this to be merged or exploring design.

mateuszkuta256:
thx @simuons, this PR will be split into a few smaller ones. I already prepared the first part: https://github.com/mateuszkuta256/rules_scala/commits/multiple_scala_versions/
> get rid of branching on SCALA_VERSION

In there I introduced `scala_versions` config property with the semantics you suggested:
- `scala_version` - signifies the default version
- `scala_versions` -  contains all versions required by the repository to build. Default to `[scala_version]`

there is no more statement like 'if SCALA_VERSION' - instead I rely on elements of `SCALA_VERSIONS`
for example, the configuration like this: `scala_version = "2.13.12",  scala_versions = ["3.3.1"]`
will register two targets: `scalac_2_13_12, scalac_3_3_1`
instead of single target : `scalac -> srcs = [...] if SCALA_VERSION...`

please take a brief look and confirm this is what you expected, in the meantime I'm gonna continue with "move some of the toolchains properties to build_settings"

mateuszkuta256:
one more question @simuons 
>move some of the toolchains properties to build_settings meaning toolchains will contain only scala version specific stuff. Otherwise we will have a combinatorial explosion of scala versions and properties like strict_deps

you mean to introduce same settings as for `scala_version`? like:
```
config_setting(
    name = "strict_deps",
    flag_values = ...
)
```
such a change wouldn't be backward-compatible, or?
also, one could argue that global settings for each toolchain are not 'elastic' enough (sounds ok for me, just wanna assure we are on the same page)

simuons:
Hi, @mateuszkuta256, I think I've introduced more confusion by mentioning build_settings. I thought about it and looks like it's not relevant now. So let's not concentrate on that (I'll share latter what I had in mind). Looks like your multiple_scala_version branch attempts to solve smaller scope of this issue. I took a glance at it and think maybe you should open a PR with that and we could move discussion there. After this will be sorted out you will add transitions. wdyt?

mateuszkuta256:
I extracted the first part of this PR into a smaller one https://github.com/bazelbuild/rules_scala/pull/1552
another PR will add a transition for toolchain selection
[poc](https://github.com/mateuszkuta256/rules_scala/commits/multiple_scala_versions_all/) for the whole feature

simuons:
Closing as it was completed as part of https://github.com/bazelbuild/rules_scala/issues/1290

PR #1548, opened 2024-02-26
Don't output unused dependency messages when compilation fails

laurynaslubys:
### Description
When a scala source is malformed, the AST parsing dependency checking step comes back with no/very little dependencies used. It then prints out a bunch of these warnings which are not necessarily true. As a workaround, let's not print the unused dependency warnings if the build has already failed before hand.

### Motivation
The printed warnings are potentially misleading and therefore should not be printed.


liucijus:
Thanks, @laurynaslubys! I think build status also needs to be reflected in produced `sdeps` files. Maybe the file should be produced without dependency entries and have a status field. Status should reflect that dependency analysis has not be done or is incomplete.

laurynaslubys:
Good point. I've updated this so we pretend that there are no unused dependencies if the reporter has errors already.

simuons (approved):

PR #1551, opened 2024-03-07
Add support for add_exports/add_opens

### LINKED ISSUE #1523: Add add_opens and add_exports support
JDK 17 [strongly encapsulates JDK internals](https://openjdk.org/jeps/403), so many libraries need `--add-opens=` and `--add-exports=` JVM flags. Bazel's Java rules support adding those flags through `java_library(add_opens = [...], add_exports = [...])`, and these fields are read in Bazel's `java_binary` rule:

https://github.com/bazelbuild/bazel/blob/a2d3f20b577b8b5c7c5027a8a00f8c24a66228b9/src/main/starlark/builtins_bzl/common/java/java_binary.bzl#L173-L182

However it appears that the rules_scala doesn't read those fields when filling out the executable template:

https://github.com/bazelbuild/rules_scala/blob/421a9e8c9ea896d37949edf8d05e1e21aca03a94/scala/private/phases/phase_write_executable.bzl#L106

Additionally, it would be nice if `scala_library`, `scala_binary`, and `scala_import` support `add_opens` and `add_exports` attributes also, but it's not as urgent as the first issue.

timothyg-stripe:
I'm happy to write some tests for this, but existing tests appear to be broken in CI, and much of the functionality requires Bazel 7.0.0 to work anyway (while CI is still on Bazel 6.3).

----

### Description
* In `scala_binary`, `scala_test`, and friends, read `java_info.module_flags_info` and include those `--add-exports` and `--add-opens` JVM flags in the launcher script.
* On Bazel 7.0.0+, add `add_exports` and `add_opens` attributes to all rules (including `scala_import`) and propagate the flags through `JavaInfo`.
    * Bazel's `JavaInfo` constructor only added this functionality in 7.0.0; see https://github.com/bazelbuild/bazel/issues/20033.

### Motivation
JDK 17 [strongly encapsulates JDK internals](https://openjdk.org/jeps/403), so many libraries need `--add-exports=` and `--add-opens=` JVM flags to run. Bazel's Java rules already support adding those flags through `java_library(add_opens = [...], add_exports = [...])`, and these fields are read in Bazel's `java_binary` rule. This PR adds the same functionality to Scala rules.

Fixes #1523

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -15,6 +15,17 @@ load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
 
 bazel_skylib_workspace()
 
+http_archive(
+    name = "bazel_features",
+    sha256 = "70d355d5e34c3fe453f5556d6c0f02ffed0eb2c7ce4c8ee016d94d654bc6a014",
+    strip_prefix = "bazel_features-1.8.0",
+    url = "https://github.com/bazel-contrib/bazel_features/releases/download/v1.8.0/bazel_features-v1.8.0.tar.gz",
+)
+
+load("@bazel_features//:deps.bzl", "bazel_features_deps")
+
+bazel_features_deps()
+
 _build_tools_release = "5.1.0"
 
 http_archive(
--- a/scala/private/common_attributes.bzl
+++ b/scala/private/common_attributes.bzl
@@ -1,5 +1,7 @@
 """Shared attributes for rules"""
 
+load("@bazel_features_version//:version.bzl", _bazel_version = "version")
+load("@bazel_skylib//lib:versions.bzl", "versions")
 load(
     "@io_bazel_rules_scala//scala/private:coverage_replacements_provider.bzl",
     _coverage_replacements_provider = "coverage_replacements_provider",
@@ -48,7 +50,10 @@ common_attrs_for_plugin_bootstrapping = {
         default = False,
         mandatory = False,
     ),
-}
+} | ({
+    "add_exports": attr.string_list(),
+    "add_opens": attr.string_list(),
+} if versions.is_at_least("7.0.0", _bazel_version) else {})
 
 common_attrs = {}
 
--- a/scala/private/phases/phase_compile.bzl
+++ b/scala/private/phases/phase_compile.bzl
@@ -3,6 +3,8 @@
 #
 # DOCUMENT THIS
 #
+load("@bazel_features_version//:version.bzl", _bazel_version = "version")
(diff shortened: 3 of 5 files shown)


timothyg-stripe:
I managed to get a passing build with this PR, but it's a bit unfortunate since it would force users to add some additional lines to their WORKSPACE file.

1. I'm currently using [`bazel_features`](https://github.com/bazel-contrib/bazel_features) to detect Bazel version for support for `add_opens`/`add_exports`.
2. The project currently recommends that folks run `load("@io_bazel_rules_scala//scala:scala.bzl", "rules_scala_setup", "rules_scala_toolchain_deps_repositories")` in WORKSPACE. But loading `@io_bazel_rules_scala//scala:scala.bzl` also loads Bazel rules like `scala_library`, which requires `@bazel_features`.
3. For `load("@bazel_features//...")` to work, the WORKSPACE file needs to run `bazel_features_deps()` _before_ the attempted load.

The sum effect of this is that users must have

```bazel
http_archive(
    name = "bazel_features",
    sha256 = "d7787da289a7fb497352211ad200ec9f698822a9e0757a4976fd9f713ff372b3",
    strip_prefix = "bazel_features-1.9.1",
    url = "https://github.com/bazel-contrib/bazel_features/releases/download/v1.9.1/bazel_features-v1.9.1.tar.gz",
)

load("@bazel_features//:deps.bzl", "bazel_features_deps")
bazel_features_deps()
```

in the WORKSPACE file _before_ loading anything from `rules_scala` – and specifically, we cannot call `bazel_features_deps()` for them in `rules_scala_toolchain_deps_repositories()`.

----

I think there are three options:

1. Bite the bullet and force every user to install `bazel_features`. (That's what this PR would do, in its current shape.)
2. Abandon support for Bazel 6.x. (Bazel 7.0.0+ has proper support for this feature.)
3. Change the `.bzl` file structure so that files that should be loaded from BUILD files are separate from those that should be loaded from WORKSPACE files. In other words, lift `rules_scala_setup` and `rules_scala_toolchain_deps_repositories` out of `@io_bazel_rules_scala//scala:scala.bzl`, and into something like `@io_bazel_rules_scala//scala:workspace.bzl`.

    This way, users still have to load `bazel_features` – but at least we can call `bazel_features_deps()` for them.

I think option 2 seems most straightforward, though I'll leave it up to the maintainers to decide.

@liucijus @simuons

timothyg-stripe:
@simuons unfortunately, skylib versions isn't sufficient since `versions.get` does not work in a BUILD file. See https://github.com/bazelbuild/bazel/issues/4566.

PR #1556, opened 2024-03-21
Build with JDK 21

simuons:
Codebase doesn't build with jdk 21 with default settings.

Issue is that workers and test runners use SecurityManager which fails at runtime.

Fix is to conditionally add -Djava.security.manager=allow jvm flag to:
- tools executed via ctx.actions.run if host jdk is >= 17
- generated test executables if target jdk is >= 17

This change involves some kind of guessing with which jdk version worker will be run. 

JDK is linked into worker/executable runfiles and it matches host jdk (hence added _java_host_runtime attribute to all relevant rules)

Added separate [build](https://github.com/bazelbuild/rules_scala/pull/1556/files#diff-b03048100ab56455b5b377a5c20ffd3d17f80c719befca45253af622eb9395efR93) for jdk 21

### DIFF
--- a/test/BUILD
+++ b/test/BUILD
@@ -776,22 +776,22 @@ scala_junit_test(
     runtime_deps = [":JunitMixedSeparateTarget"],
 )
 
-scala_library(
-    name = "JunitRuntimePlatform",
-    srcs = [
-        "src/main/scala/scalarules/test/junit/runtime_platform/JunitRuntimePlatformTest.java",
-    ],
-    deps = ["@io_bazel_rules_scala_junit_junit"],
-)
-
-scala_junit_test(
-    name = "JunitRuntimePlatform_test_runner",
-    size = "small",
-    runtime_jdk = "@bazel_tools//tools/jdk:remote_jdk11",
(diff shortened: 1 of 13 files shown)


test/BUILD
```diff
@@ -776,22 +776,22 @@ scala_junit_test(
...
+#)
+#
+#scala_junit_test(
+#    name = "JunitRuntimePlatform_test_runner",
+#    size = "small",
+#    runtime_jdk = "@bazel_tools//tools/jdk:remote_jdk11",
+#    suffixes = ["Test"],
+#    tests_from = [":JunitRuntimePlatform"],
+#    runtime_deps = [":JunitRuntimePlatform"],
+#)
```
liucijus, line 794: @simuons are these accidentally commented out, or is there a reason? If there's a reason, please leave a comment in the code.

simuons: @liucijus thanks for noting. I forgot about it.

simuons: @liucijus fixed test for jdk > 11. Please take a look.

liucijus (approved):

PR #1552, opened 2024-03-08
Introduce 'SCALA_VERSIONS' config setting

mateuszkuta256:
### Description
Extend `scala_config` repo with a `scala_versions` property

This is the first step for cross-compilation feature. User should be able to define multiple scala versions supported by the repository:
```
scala_config(
    scala_version = "3.2.1",
    scala_versions = [
        "2.13.12",
        "3.2.1",
    ],
)
```
the semantics will be:
- `scala_version` - default version for repository. Will always fall back to this value when a rule does not specify the version explicitly
-  `scala_versions` - a list of all the versions supported by the repository

This change is backward-compatible. `scala_versions` is optional and default to `[scala_version]`

This PR also registers toolchain-related repositories with a versions suffix. It is needed to avoid collisions, e.g. when `2.13.12` and `3.3.1` are set, we will register two `io_bazel_rules_scala_scala_compiler`:
`io_bazel_rules_scala_scala_compiler_2_13_12`
`io_bazel_rules_scala_scala_compiler_3_3_1`

In the follow-up PR  I'm gonna add a transition, that will select proper toolchains based on scala_version config

### Motivation
Extracted from https://github.com/bazelbuild/rules_scala/pull/1546

PR #1557, opened 2024-03-22
Introduce `SCALA_VERSIONS`

aszady:
### Description
A variable to hold all (possibly more than one in the future) configured versions of Scala.

### Motivation
Originally #1290.
#1552 may be still too big to review. Let's start with some manageable, smaller chunks.

Plan for following PRs: https://github.com/bazelbuild/rules_scala/compare/master...aszady:rules_scala:cross-compilation?expand=1

### DIFF
--- a/cross-compilation-doc.md
+++ b/cross-compilation-doc.md
@@ -1,0 +1,10 @@
+# Cross compilation support
+
+The support for cross-compilation is currently under development.
+
+## Version configuration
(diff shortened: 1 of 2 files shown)


simuons (approved):
Thanks, @aszady!

Really appreciate for going with small steps.

liucijus (approved):

PR #1558, opened 2024-03-25
Introduce build setting for current Scala version

aszady:
### Description
Build setting & config setting to transition and query about the Scala version.

A small change compared to #1552 is that I don't allow an empty string value for the build setting.

This shouldn't change any behavior.

### Motivation
Originally #1290.
Partitioned from #1552.

### DIFF
--- a/scala_config.bzl
+++ b/scala_config.bzl
@@ -1,4 +1,4 @@
-load("//scala:scala_cross_version.bzl", "extract_major_version", "extract_minor_version")
+load("//scala:scala_cross_version.bzl", "extract_major_version", "extract_minor_version", "version_suffix")
 
 def _default_scala_version():
     """return the scala version for use in maven coordinates"""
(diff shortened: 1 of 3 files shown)


scala_config.bzl
```diff
@@ -10,6 +10,17 @@ def _validate_supported_scala_version(scala_major_version, scala_minor_version):
     if scala_major_version == "2.12" and int(scala_minor_version) < 1:
         fail("Scala version must be newer or equal to 2.12.1 to use compiler dependency tracking.")
 
+def _config_setting(scala_version):
+    return "\n".join([
```
simuons, line 14: Maybe it would be cleaner if you format a multiline string instead of joining on newline.

scala_config.bzl
```diff
@@ -39,8 +50,19 @@ def _store_config(repository_ctx):
         "ENABLE_COMPILER_DEPENDENCY_TRACKING=" + enable_compiler_dependency_tracking,
     ])
 
+    build_file_content = "\n".join([
```
simuons, line 53: Same as above. I think multiline string would be cleaner (at least this is what is used in other places)

simuons (changes requested):
Thanks. Overall looks good. Please check minor comments.

simuons:
Also big thanks for

> don't allow an empty string value for the build setting.

that was puzzling me in previous pr.

scala_config.bzl
```diff
@@ -39,8 +50,19 @@ def _store_config(repository_ctx):
...
+        'load("@bazel_skylib//rules:common_settings.bzl", "string_flag")',
+        "string_flag(",
```
simuons, line 55: Maybe we should keep this as `string_setting` for now and do not allow to change this via command line? wdyt?

aszady: Indeed, I see no use case for a flag at the moment.

simuons (approved):

aszady:
Updated the name of config setting as suggested in https://github.com/bazelbuild/rules_scala/pull/1559#discussion_r1540591226

liucijus (approved):

PR #1560, opened 2024-03-26
Use version-aware naming for `scala_compiler_source` repository

aszady:
### Description
A small preparation for multiple `SCALA_VERSIONS`.

In theory breaking, as removes `@scala_compiler_source` repository.
But I guess it's too internal to anyone depend on it.

### Motivation
Originally #1290.
Partitioned from #1552.

### DIFF
--- a/scala/private/macros/scala_repositories.bzl
+++ b/scala/private/macros/scala_repositories.bzl
@@ -1,14 +1,16 @@
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load(
     "@io_bazel_rules_scala//scala:scala_cross_version.bzl",
+    "extract_major_version",
+    "extract_minor_version",
+    "version_suffix",
     _default_maven_server_urls = "default_maven_server_urls",
 )
 load("//third_party/repositories:repositories.bzl", "repositories")
 load(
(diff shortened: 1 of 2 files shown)


aszady:
Hi @simuons, @liucijus,
can we do anything to help you reviewing these PRs regarding cross-build support?
Note that [all open PRs](https://github.com/bazelbuild/rules_scala/pulls?q=is%3Aopen+is%3Apr+author%3Aaszady+-is%3Adraft+%231290) are independent of each other and could be reviewed/merged in any order.

simuons:
Hi @aszady, sorry last week I was on vacation and could get over your PRs. I'm looking at them now and will answer today/tomorrow.

simuons:
Hi, @aszady, could you confirm this is still relevant? I mean in another PR you mentioned that direction you are taking is to have single label with select. This doesn't apply here?

aszady:
Yes; I think the general convention should be like:
* for repositories, that actually need to download something (artifacts, see: #1562; or here the compiler source) – use individual repository names with the suffix;
* for targets, that are usually built (e.g. majority of #1564) – distinguish them by configurations + select.

Toolchains with config transitions seem to nicely join these two worlds together.

liucijus (approved):

PR #1561, opened 2024-03-26
Move default toolchain dependencies logic to .bzl file

aszady:
### Description

Purely refactoring change.
Encapsulates the logic of selecting dependencies into a function. Should come in handy in future changes.

### Motivation
Originally #1290.
Partitioned from / inspired by #1552.

### DIFF
--- a/scala/BUILD
+++ b/scala/BUILD
@@ -1,65 +1,31 @@
 load("@rules_java//java:defs.bzl", "java_import", "java_library")
 load("//scala:providers.bzl", "declare_deps_provider")
-load("//scala:scala.bzl", "setup_scala_toolchain")
-load("@io_bazel_rules_scala_config//:config.bzl", "SCALA_MAJOR_VERSION")
+load("//scala/private:macros/setup_scala_toolchain.bzl", "default_deps", "setup_scala_toolchain_with_default_classpaths")
+load("@io_bazel_rules_scala_config//:config.bzl", "SCALA_VERSION")
 
 toolchain_type(
     name = "toolchain_type",
     visibility = ["//visibility:public"],
 )
 
-_SCALA_COMPILE_CLASSPATH_DEPS = [
-    "@io_bazel_rules_scala_scala_compiler",
-    "@io_bazel_rules_scala_scala_library",
-] + (["@io_bazel_rules_scala_scala_reflect"] if SCALA_MAJOR_VERSION.startswith("2") else [
-    "@io_bazel_rules_scala_scala_interfaces",
-    "@io_bazel_rules_scala_scala_tasty_core",
-    "@io_bazel_rules_scala_scala_asm",
-    "@io_bazel_rules_scala_scala_library_2",
-])
-
-_SCALA_LIBRARY_CLASSPATH_DEPS = [
-    "@io_bazel_rules_scala_scala_library",
-] + (["@io_bazel_rules_scala_scala_reflect"] if SCALA_MAJOR_VERSION.startswith("2") else [
-    "@io_bazel_rules_scala_scala_library_2",
-])
-
-_SCALA_MACRO_CLASSPATH_DEPS = [
-    "@io_bazel_rules_scala_scala_library",
-] + (["@io_bazel_rules_scala_scala_reflect"] if SCALA_MAJOR_VERSION.startswith("2") else [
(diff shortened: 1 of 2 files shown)


scala/private/macros/setup_scala_toolchain.bzl
```diff
@@ -95,3 +95,69 @@ def setup_scala_toolchain(
         toolchain_type = "@io_bazel_rules_scala//scala:toolchain_type",
         visibility = visibility,
     )
+
+_DEFAULT_DEPS = {
+    "scala_compile_classpath": {
+        "any": [
```
simuons, line 101: Was this "any" introduced to reduce duplication? I'm afraid this assumption that labels match between scala version might not always hold, leaving "any" obsolete.

I think some duplication here is fine. I would just list all major scala versions and deps lables for each dep_id. 

imho that would look simpler. wdyt?

scala/BUILD
```diff
@@ -1,65 +1,31 @@
...
-])
-
-_PARSER_COMBINATORS_DEPS = ["@io_bazel_rules_scala_scala_parser_combinators"]
-
-_SCALA_XML_DEPS = ["@io_bazel_rules_scala_scala_xml"]
-
-_SEMANTICDB_DEPS = ["@org_scalameta_semanticdb_scalac"] if SCALA_MAJOR_VERSION.startswith("2") else []
-
-setup_scala_toolchain(
+setup_scala_toolchain_with_default_classpaths(
```
simuons, line 11: Is this new macro really needed? I mean `scala_version` attribute eventually should be passed to `setup_scala_toolchain` because it will be needed in `target_setting`. So maybe default classpaths should be added there if None was passed. 

I think it is nice to have deps_id mappings extracted into separate bzl file with lookup exposed function. But for now I would avoid adding new macro and change `setup_scala_toolchain` to
```
setup_scala_toolchain(
    name = "default_toolchain",
    scala_compile_classpath = default_deps("scala_compile_classpath", SCALA_VERSION),
    scala_library_classpath = default_deps("scala_library_classpath", SCALA_VERSION),
    scala_macro_classpath = default_deps("scala_macro_classpath", SCALA_VERSION),
    use_argument_file_in_runner = True,
)
```

Similarly to what you did for `declare_deps_provider`

simuons (changes requested):
Please take a look at comments.

scala/BUILD
aszady: This macro is not necessary.

So since we have #1566 already merged (`scala_version` is now passed to `setup_scala_toolchain`), we can already proceed with default classpaths if `None` is provided.

I pushed the relevant revision.

scala/private/macros/setup_scala_toolchain.bzl
aszady: Indeed, `any` is here just to avoid duplication.

My goal was to keep the change minimal (here: not changing the logic of producing the relevant labels). This deduplication is not a new concept, as it expresses the same intention as the old version in the BUILD file, e.g.:

```
_SCALA_COMPILE_CLASSPATH_DEPS = [
    "@io_bazel_rules_scala_scala_compiler",
] + (["@io_bazel_rules_scala_scala_reflect"] if SCALA_MAJOR_VERSION.startswith("2") else [
    "@io_bazel_rules_scala_scala_interfaces",
    "@io_bazel_rules_scala_scala_tasty_core",
    "@io_bazel_rules_scala_scala_asm",
    "@io_bazel_rules_scala_scala_library_2",
])
```

translated into:

```
    "scala_compile_classpath": {
        "any": [
            "@io_bazel_rules_scala_scala_compiler",
        ],
        "2": [
        ],
        "3": [
            "@io_bazel_rules_scala_scala_interfaces",
            "@io_bazel_rules_scala_scala_tasty_core",
            "@io_bazel_rules_scala_scala_asm",
            "@io_bazel_rules_scala_scala_library_2",
        ],
    },
```

simuons: I understand your intent. Probably it's just a matter of personal taste. Since it's private and encapsulated in a function I'm totally fine with that.

simuons (approved):

liucijus (approved):
Thanks, @aszady!

PR #1564, opened 2024-03-28
Don't rely on global `SCALA_VERSION` in BUILD files, use select instead

aszady:
### Description

The second part (after #1559) of eliminating dependencies on `SCALA_VERSION`. This time within target attributes. Together with #1560 we should be all done.

This change extensively uses the new `select_for_scala_version` macro (#1563).
I tried my best to emerge the most correct intention, but in many cases the choice is rather arbitrary (e.g. `any_2` + `any_3` vs. `before_3` + `since_3` vs. `any_2` + `default`) 

### Motivation
#1290.


simuons (approved):

liucijus (approved):

liucijus:
Thanks, @aszady!

PR #1566, opened 2024-04-02
Bind toolchains to Scala version

aszady:
### Description
Use `target_settings` to indicate which Scala version is supported by the toolchain. This will become useful when multiple toolchains for each Scala version are defined.

The choice of toolchains to cover by this (first) migration is quite arbitrary, but should be enough to enable the basic cross-build use cases.

No change in behavior if the default Scala version is used.

### Motivation
Originally #1290.
Partitioned from #1552.

### DIFF
--- a/cross-compilation-doc.md
+++ b/cross-compilation-doc.md
@@ -53,3 +53,31 @@ def _rule_impl(ctx):
 
 ### From config setting
 TODO
+
+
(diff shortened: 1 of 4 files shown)


simuons (approved):

liucijus (approved):
Thanks, @aszady!

PR #1567, opened 2024-04-09
JacocoRunner script: update for Jacoco 0.8.11 and Bazel 7.0.2

gergelyfabian:
rules_scala has issues with coverage on JDK 21.

Bazel has solved this in https://github.com/bazelbuild/bazel/issues/20845 by upgrading Jacoco and ASM.

Added option to build JacocoRunner for Bazel 7, the version with fix will be used with Jacoco interface changes.
Jacoco upgraded to 0.8.11 (including the Jacoco/Bazel branches used for that).


simuons (approved):

gergelyfabian:
Will update this PR with latest learnings. Bazel 7.1.0 is the first version that includes the Bazel fix that I branched off from, so better to use a release (will use 7.1.1).

gergelyfabian:
Even though Bazel 7.1.0 seems to solve the original issue that motivated this upgrade, that release does not have the Jacoco and ASM upgrade when using it for building the jacocorunner.

gergelyfabian:
> Even though Bazel 7.1.0 seems to solve the original issue that motivated this upgrade, that release does not have the Jacoco and ASM upgrade when using it for building the jacocorunner.

So I'll leave the Bazel branch as is.

liucijus (approved):
Thanks, @gergelyfabian!

PR #1568, opened 2024-04-15
Propagate maven_coordinates tag to generated scala_import so pom generation includes the dependency

vinnybod:
### Description

Add a `maven_coordinates=` tag to the generated scala maven dependency targets.

### Motivation

`rules_jvm_external` pom generation uses the `maven_coordinates=x:y:z` tag to determine the formatting of a dependency in the pom. Because this tag is missing, the scala dependencies are excluded from the generated pom.
I need for example `@io_bazel_rules_scala_scala_library//jar` to appear in my generated pom as
```xml
<dependency>
    <groupId>org.scala-lang</groupId>
    <artifactId>scala-library</artifactId>
    <version>2.13.12</version>
</dependency>
```

### DIFF
--- a/scala/scala_maven_import_external.bzl
+++ b/scala/scala_maven_import_external.bzl
@@ -204,8 +204,10 @@ def _serialize_given_rule_import(
         "%s(" % rule_name,
         "    name = %s," % repr(name),
         "    jars = [%s]," % repr(path),
-        "    tags = [\"maven_coordinates=%s\"]," % coordinates,
     ]
+
+    if coordinates:
+        lines.append("    tags = [\"maven_coordinates=%s\"]," % coordinates)
     if srcpath:
         lines.append("    srcjar = %s," % repr(srcpath))
     for prop in props:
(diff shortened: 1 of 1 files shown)


scala/scala_maven_import_external.bzl
```diff
@@ -193,13 +196,15 @@ def _serialize_given_rule_import(
...
         srcpath,
+        coordinates,
         attrs,
         props,
         additional_rule_attrs):
     lines = [
         "%s(" % rule_name,
         "    name = %s," % repr(name),
         "    jars = [%s]," % repr(path),
+        "    tags = [\"maven_coordinates=%s\"]," % coordinates,
```
simuons, line 207: Since `coordinates` is non mandatory attribute I think tags should be added conditionally rather than having empty `maven_coordinates=`. wdyt?

simuons (approved):
Thanks @vinnybod

vinnybod:
@liucijus CLA was signed and submitted yesterday. It was a corporate CLA, so according to [opensource.google/documentation/reference/cla#cla-accepted](https://opensource.google/documentation/reference/cla#cla-accepted) , it might take a couple days for it to process.

vinnybod:
CLA is approved

PR #1569, opened 2024-04-17
Allow transitioning Scala version with `scala_version` attribute of rules

aszady:
### Description
Implements transition and exposes it via the `scala_version` attribute for a number of common rules.

No effect at the moment, as for now we have only one valid setting.
After allowing more Scala versions when configuring rules repository, this will become the central part of the cross-build.

### Motivation
Originally #1290.
Partitioned from #1552.


simuons (approved):
Thanks @aszady

PR #1573, opened 2024-05-08
Use version-aware naming for artifact repositories

aszady:
### Description
It enables the logic implemented in #1562. The choice of toolchains to cover is the same as in #1566.

Now, e.g., instead of referring to the artifact's repository `@io_bazel_rules_scala_scala_compiler`, we ~need to~ may use the version-aware name, like `@io_bazel_rules_scala_scala_compiler_3_3_0`.

**In future** this may require an update in client's code. In particular in cases they don't use the default toolchains provided by the rules, and construct their own dependency providers. One would need to manually add the version suffix to the repository names, or use a construction like `+ version_suffix(SCALA_VERSION)` in their code.

**For now** a backward compatibility layer is provided in a shape of repository with aliases, pointing to artifacts for the default Scala version.

### Motivation
#1290

### DIFF
--- a/examples/testing/multi_frameworks_toolchain/BUILD
+++ b/examples/testing/multi_frameworks_toolchain/BUILD
@@ -1,4 +1,6 @@
 load("@io_bazel_rules_scala//scala:scala.bzl", "setup_scala_testing_toolchain")
+load("@io_bazel_rules_scala//scala:scala_cross_version.bzl", "version_suffix")
+load("@io_bazel_rules_scala_config//:config.bzl", "SCALA_VERSION")
 
 setup_scala_testing_toolchain(
     name = "testing_toolchain",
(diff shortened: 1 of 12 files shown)


simuons (changes requested):
Thanks, @aszady,

Could you please add aliases for backwards compatibility as you proposed
> providing some kind of an alias that points from non-suffixed repository to the one with default Scala version 

Not sure how complicated that would be, but definitely better than adding ifs on default scala version all over the place. Later we could deprecate those aliases and allow people to migrate.

aszady:
Added. Out of possible solutions, two aliases worked the best for me.
I also tried sth like using `scala_library` with single dependency, though it didn't work well with dependency checker (suggesting unused dependency).
@simuons, please take a look at the revised version.

Now this PR consists of three commits:
1. Preparation: add alias repo for backward compatibility.
2. Do the actual change (this state demonstrates that aliases work).
3. Cleanup: migrate all the internal rules' deps to suffixed repos.

Do you prefer to have them merged together or separately?

simuons (approved):
Thanks, @aszady

liucijus (approved):

PR #1577, opened 2024-05-14
Fix the remaining default providers in `setup_scala_toolchain`

aszady:
### Description
These should also point to Scala version-specific artifacts.
See previous PR: #1561

### Motivation
#1290


simuons (approved):

liucijus (approved):
Thanks, @aszady!

PR #1579, opened 2024-05-15
Allow configuring multiple Scala versions

aszady:
### Description
In my belief, all the changes required in order to enable this feature are done.
We can flip the switch!

Please find the following draft PRs as a support for this change:
- [ ] test: #1586
- [ ] example: #1587

### Motivation
#1290

### DIFF
--- a/scala_config.bzl
+++ b/scala_config.bzl
@@ -28,7 +28,11 @@ def _store_config(repository_ctx):
     )
 
     # All versions supported
-    scala_versions = [scala_version]
+    scala_versions = repository_ctx.attr.scala_versions
+    if not scala_versions:
+        scala_versions = [scala_version]
+    elif scala_version not in scala_versions:
+        fail("You have to include the default Scala version (%s) in the `scala_versions` list." % scala_version)
 
     enable_compiler_dependency_tracking = repository_ctx.os.environ.get(
         "ENABLE_COMPILER_DEPENDENCY_TRACKING",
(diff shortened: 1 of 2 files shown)


scala_config.bzl
```diff
@@ -28,7 +28,7 @@ def _store_config(repository_ctx):
     )
 
     # All versions supported
-    scala_versions = [scala_version]
+    scala_versions = [scala_version] + [v for v in repository_ctx.attr.scala_versions if v != scala_version]
```
simuons, line 31: Hi, @aszady, all looks good to me. Just one question. Shouldn't there be a check that `scala_version` (default) is specified in `scala_versions` (all supported)?
Now `scala_config(scala_version = "2.12.18", scala_versions = ["3.3.1"])` would result in a default version `"2.12.18"` and all configured `["2.12.18", "3.3.1"]` I think this is different from what doc says. What do you think? Am I missing something?

aszady: The doc currently says only about the variables `SCALA_VERSION`, `SCALA_VERSIONS` in the generated file `@io_bazel_rules_scala_config//:config.bzl`.

We can of course make them more consistent with the args to `scala_config` if it is too confusing.
I pushed a revision that enforces containing `scala_version` in `scala_versions` – instead of silently fixing the user's oversight.
One can also still not provide the `scala_versions` at all and it will default to `[scala_version]` (compatibility).

simuons: Right that use-case without `scala_versions` :) somehow managed to forget.

simuons (approved):
Thanks

PR #1580, opened 2024-05-18
Semanticdb clear dir

crt-31:
### Description
This change clears out and deletes the _semanticdb folder before compiling. This is similar to how ScalacWorker was already clearing out the _classes directory.

### Motivation
I ran into a bug where, when you delete a source *.scala file, it's associated semanticdb file remained and was read-only which caused permission errors when the compiler tried to clear the semanticdb output directory.


src/java/io/bazel/rulesscala/scalac/ScalacWorker.java
```diff
@@ -95,6 +98,15 @@ private static Path ensureEmptyWorkDirectory(Path output, String label) throws I
 
     return Files.createDirectories(dir);
   }
+  
+  private static void ensureEmptySemanticDBDirectory(Path output, String label) throws IOException {
```
simuons, line 102: Would it be possible to refactor `ensureEmptyWorkDirectory` instead of adding this method wich looks almost identical and serves same purpose?

simuons (changes requested):
Thanks, @crt-31, for taking this. Please take a look at comment.

src/java/io/bazel/rulesscala/scalac/ScalacWorker.java
crt-31: done.

(But the tests seem flaky... they have failed in different places (unrelated to this change) on last 2 runs).

simuons (approved):

liucijus (approved):
Thanks, @crt-31!

PR #1582, opened 2024-05-20
Allow users to configure `buildijar` value when building a scala library

ladinu:
### Description
Allow users to configure `buildijar` value when building a library

### Motivation
I'm generating some Scala code and compiling them as a library. But IntelliJ is failing to detect the sources of the library. When I turn off `build_ijar`, I'm able to view sources.

### DIFF
--- a/scala/private/phases/phase_compile.bzl
+++ b/scala/private/phases/phase_compile.bzl
@@ -32,6 +32,7 @@ def phase_compile_binary(ctx, p):
 def phase_compile_library(ctx, p):
     args = struct(
         srcjars = p.collect_srcjars,
+        buildijar = ctx.attr.build_ijar if hasattr(ctx.attr, "build_ijar") else _ijar_default_value(ctx),
         unused_dependency_checker_ignored_targets = [
             target.label
             for target in p.scalac_provider.default_classpath + ctx.attr.exports +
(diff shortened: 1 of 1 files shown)


crt-31:
I'm not opposed to the setting necessarily, but for the use case the OP describes: 
My understanding is that IntelliJ is using [ bazel-bsp](https://github.com/JetBrains/bazel-bsp) to incorperate bazel into their IDE. I would think it would be more proper for bazel-bsp to be the ones that do the source-jar / ijar mapping, they should have or be able to get the information.  If you do it in scala_library, you would lose some of the benefits of the ijar.

ladinu:
@crt-31 I agree that it makes more sense for bazel-bsp to handle ijar -> source-jar mapping. I don't know what it would take to get it working there though.  Since I'm only enabling this for generated scala code, I don't _think_ I'll miss out on any of the benefits provided by the ijar.

But I think having this setting is reasonable. It should ultimately be up to the user as to how they want to configure their scala library.

ladinu:
@liucijus @simuons Any opposition to this change?

liucijus:
@ladinu It was fixed on [intellij plugin](https://github.com/bazelbuild/intellij/commit/463b16955ade5270f9e4366381f892061460d8e7). Are you referring to this plugin, or something else?

ladinu:
@liucijus I'm referring to this PR where users can choose to set `build_ijar = True` for scala library (default is `False`)

liucijus:
> I'm generating some Scala code and compiling them as a library. But IntelliJ is failing to detect the sources of the library.

@ladinu would you be able to make a small repro repository for me to debug it? I think it needs to be fixed on the Intellij plugin side.

PR #1584, opened 2024-05-23
Make scaladoc rule handle transient deps better (and add some tests)

crt-31:
### Description
The main goal of this PR is to fix issue where transient dependencies weren't being handled correctly. This issue manifested itself when there was an external library (i.e. scala_import) that was a transient dependency. 

Also added tests for the scaladoc rule.

### Motivation
This is needed for the scaladoc rule to work correctly.

I believe this also addresses the issue that #1331 was trying to address.


simuons (approved):
Thanks, @crt-31

liucijus (approved):

PR #1588, opened 2024-06-11
Documentation update after development of the cross-build

aszady:
### Description
Adding the very basic usage of the rules with multiple Scala versions.
Moving the doc file to the `docs/` folder & referring it from the main Readme.

Please review the docs as a whole, maybe there's sth else that should be included? 

### Motivation
#1290


simuons (approved):
Thanks, @aszady

liucijus (approved):

PR #1597, opened 2024-08-12
Updated artifact versions with sha sums and scala versions

lm1nrt:
### Description

-  Updated artifact SHA sums
-  Updated scala versions
-  Updated tests that are dependent on the chosen scala version
-  Updated shell command for ubuntu presubmit due to fatal error detected by the Java Runtime Environment

### Motivation
There are newer versions of Scala that can be applied to certain artifacts.


.bazelci/presubmit.yml
```diff
@@ -27,15 +27,15 @@ tasks:
     platform: ubuntu2004
     shell_commands:
     # Install xmllint
-    - sudo apt update && sudo apt install -y libxml2-utils
+    - sudo apt update && sudo apt install --reinstall libxml2-utils -y
```
lm1nrt, line 30: Due to https://buildkite.com/bazel/rules-scala-scala/builds/4919#01915033-1744-4437-9781-756df196c887/122-496

lukaszwawrzyk:
@liucijus @simuons could any of you take a look? This just tries to keep scala versions up to date. We are looking into adding scala 3.4 in another PR as it doesn't work for some reason. Also we hope to make these bumps more automated.

liucijus (approved):
Thanks, @lm1nrt

.bazelci/presubmit.yml
lm1nrt, line 30: Due to: 

```
A fatal error has been detected by the Java Runtime Environment:
Problematic frame:
# C  [libc.so.6+0x18bd71]
```

simuons (approved):
Thanks @lm1nrt

PR #1603, opened 2024-08-26
Scala 3.4.3 support

lm1nrt:
### Description

-  Added artifacts for Scala 3.4.3 version
-  Updated tests that are dependent on the chosen scala version
-  Fixed `java.lang.NoClassDefFoundError: xsbti/UseScope` for Scala versions >= 3.4.0

### Motivation
There are newer versions of Scala that can be supported.

### DIFF
--- a/scala/private/macros/scala_repositories.bzl
+++ b/scala/private/macros/scala_repositories.bzl
@@ -140,6 +140,7 @@ def _artifact_ids(scala_version):
         "io_bazel_rules_scala_scala_xml",
         "io_bazel_rules_scala_scala_parser_combinators",
         "io_bazel_rules_scala_scala_library_2",
+        "org_scala_sbt_compiler_interface",
     ]
 
 def rules_scala_toolchain_deps_repositories(
--- a/third_party/repositories/scala_3_1.bzl
+++ b/third_party/repositories/scala_3_1.bzl
@@ -99,6 +99,10 @@ artifacts = {
             "@io_bazel_rules_scala_scala_library",
         ],
     },
+    "org_scala_sbt_compiler_interface": {
+        "artifact": "org.scala-sbt:compiler-interface:1.10.1",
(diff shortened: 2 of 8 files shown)


scala/private/macros/scala_repositories.bzl
```diff
@@ -140,6 +140,7 @@ def _artifact_ids(scala_version):
         "io_bazel_rules_scala_scala_xml",
         "io_bazel_rules_scala_scala_parser_combinators",
         "io_bazel_rules_scala_scala_library_2",
+        "org_scala_sbt_compiler_interface",
```
lm1nrt, line 143: Added due to `java.lang.NoClassDefFoundError: xsbti/UseScope` for `test_examples.sh`

third_party/repositories/scala_3_1.bzl
```diff
@@ -99,6 +99,10 @@ artifacts = {
             "@io_bazel_rules_scala_scala_library",
         ],
     },
+    "org_scala_sbt_compiler_interface": {
```
lm1nrt, line 102: What should be better in your opinion - add this artifact for versions that don't need that (3.1.0 =< version < 3.4.0) to stay concise with [artifact_ids](https://github.com/bazelbuild/rules_scala/pull/1603/files#diff-219f92bb35f88485a21a0615e73a715488fa38dbeeda0eb7b9abd251c18573d9L126) or modify `def _artifact_ids(scala_version):`? If modifying, I can do that in next PR.

simuons: I'd say keep it as is and do not complicate  `def _artifact_ids(scala_version)` any further. I guess those who use scala 3 will quickly move to never version and older ones will become obsolete.

simuons (approved):
Thanks @lm1nrt

liucijus (approved):
Thanks, @lm1nrt!

PR #1606, opened 2024-08-30
Handle incorrect Scalac options and prevent printing ScalacWorker stacktraces

### LINKED ISSUE #1605: scala compilation error with cross compilation produces additional Null Pointer Exception
Detecting errors for cross compilation is somehow broken, and produces an additional Null Pointer Exception.

How to reproduce:

Apply this patch in latest rules_scala code:

```
diff --git a/examples/crossbuild/3_select/BUILD b/examples/crossbuild/3_select/BUILD
index 2e58b6f..e5f8b7b 100644
--- a/examples/crossbuild/3_select/BUILD
+++ b/examples/crossbuild/3_select/BUILD
@@ -21,6 +21,7 @@ scala_library(
 scala_binary(
     name = "bin2",
     srcs = ["bin.scala"],
+    scalacopts = ["-Ywarn-unused-import"],
     main_class = "B",
     scala_version = "2.13.12",
     deps = [":lib"],
```

Then run:
```
cd examples/crossbuild/3_select
bazel build //3_select:bin2
```

This produces:

```
INFO: Invocation ID: 6bdd906c-b4f3-4c6e-9bf6-f9ace6cba6e8
INFO: Analyzed target //3_select:bin2 (1 packages loaded, 4 targets configured).
ERROR: /home/user/opt/rules_scala/examples/crossbuild/3_select/BUILD:21:13: scala @@//3_select:bin2 failed: (Exit 1): scalac failed: error executing Scalac command (from target //3_select:bin2) bazel-out/k8-opt-exec-ST-1e9cf9510058/bin/external/io_bazel_rules_scala/src/java/io/bazel/rulesscala/scalac/scalac @bazel-out/k8-fastbuild-ST-9386492f403d/bin/3_select/bin2.jar-0.params
scalac error: bad option: '-Ywarn-unused-import'
  scalac -help  gives more information
null
java.lang.NullPointerException
	at io.bazel.rulesscala.scalac.ScalacInvoker.invokeCompiler(ScalacInvoker.java:53)
	at io.bazel.rulesscala.scalac.ScalacWorker.compileScalaSources(ScalacWorker.java:257)
	at io.bazel.rulesscala.scalac.ScalacWorker.work(ScalacWorker.java:73)
	at io.bazel.rulesscala.worker.Worker.persistentWorkerMain(Worker.java:86)
	at io.bazel.rulesscala.worker.Worker.workerMain(Worker.java:39)
	at io.bazel.rulesscala.scalac.ScalacWorker.main(ScalacWorker.java:33)
Target //3_select:bin2 failed to build
INFO: Elapsed time: 0.435s, Critical Path: 0.04s
INFO: 2 processes: 2 internal.
```

In my own project I have seen:

```
ERROR: /home/user/myprojest/mypackage/BUILD:42:20: scala @//myprojest/mypackage:library failed: (Exit 1): scalac failed: error executing Scalac command (from target //myprojest/mypackage:library) bazel-out/k8-opt-exec-ST-bb79e0899c0a/bin/external/io_bazel_rules_scala/src/java/io/bazel/rulesscala/scalac/scalac '--jvm_flag=-Djava.security.manager=allow' ... (remaining 1 argument skipped)
scalac error: bad option: '-Ywarn-unused-import'
  scalac -help  gives more information
Cannot invoke "scala.tools.nsc.reporters.ConsoleReporter.hasErrors()" because "reporter" is null
java.lang.NullPointerException: Cannot invoke "scala.tools.nsc.reporters.ConsoleReporter.hasErrors()" because "reporter" is null
	at io.bazel.rulesscala.scalac.ScalacInvoker.invokeCompiler(ScalacInvoker.java:53)
	at io.bazel.rulesscala.scalac.ScalacWorker.compileScalaSources(ScalacWorker.java:257)
	at io.bazel.rulesscala.scalac.ScalacWorker.work(ScalacWorker.java:73)
	at io.bazel.rulesscala.worker.Worker.persistentWorkerMain(Worker.java:86)
	at io.bazel.rulesscala.worker.Worker.workerMain(Worker.java:39)
	at io.bazel.rulesscala.scalac.ScalacWorker.main(ScalacWorker.java:33)
```

I suppose there may be a bug in the error handling for compiler errors.

WojciechMazur:
### Description
- Handle cases when passing invalid scalac options 
- Prevent printing internal stack traces of ScalaWorker on compilation failing in an expected way

### Motivation
Fixes #1605 
When passing invalid scalac options, eg. `-source:not-existing` the Scala compiler would either: 
 - return `None` form `Driver.setup` leading to runtime exception in `None.get` (Scala 3) 
 - would not create an instance in `Driver.newCompiler` required to create `ProtoReporter` or `DepsTrackingReporter` leading to `reporter` being not initialized (Scala 2) 
 This fix handles both of this cases by termination executing earlier with a known exception. 
 
 In each of cases Scalac compiler would always show error message about which setting was invalid. This is expected behaviour. Because of that, there is no reason to print stack traces on invalid scalac setting  from `ScalaInvoker` exposing internal implementaiton to the user. It can be handled by either not filling stack traces or introducing a common subtype handled in the worker. 
 The later option was chosen and additionally extended to handle expected compiler errors. This should reduce amount of boilerplate logs.

### DIFF
--- a/src/java/io/bazel/rulesscala/worker/Worker.java
+++ b/src/java/io/bazel/rulesscala/worker/Worker.java
@@ -33,6 +33,9 @@ public final class Worker {
 			public WorkerException(String message) {
 				super(message);
 			}
+      public WorkerException(String message, Throwable cause) {
+				super(message, cause);
+			}
 		}
   }
 
--- a/src/java/io/bazel/rulesscala/worker/Worker.java
+++ b/src/java/io/bazel/rulesscala/worker/Worker.java
@@ -97,8 +97,7 @@ public final class Worker {
           } catch (ExitTrapped e) {
             code = e.code;
           } catch (Exception e) {
-            System.err.println(e.getMessage());
-            if (e instanceof Interface.WorkerException) {}
+            if (e instanceof Interface.WorkerException) System.err.println(e.getMessage());
             else e.printStackTrace();
             code = 1;
           }
--- a/src/java/io/bazel/rulesscala/worker/Worker.java
+++ b/src/java/io/bazel/rulesscala/worker/Worker.java
@@ -27,6 +27,13 @@ public final class Worker {
 
   public static interface Interface {
     public void work(String[] args) throws Exception;
+
+
+    public abstract class WorkerException extends RuntimeException {
+			public WorkerException(String message) {
+				super(message);
+			}
+		}
   }
 
   /**
@@ -88,7 +95,8 @@ public final class Worker {
             code = e.code;
           } catch (Exception e) {
             System.err.println(e.getMessage());
-            e.printStackTrace();
+            if (e instanceof Interface.WorkerException) {}
+            else e.printStackTrace();
             code = 1;
           }
 
--- a/src/java/io/bazel/rulesscala/worker/Worker.java
+++ b/src/java/io/bazel/rulesscala/worker/Worker.java
@@ -30,13 +30,13 @@ public final class Worker {
 
 
     public abstract class WorkerException extends RuntimeException {
-			public WorkerException(String message) {
-				super(message);
-			}
+      public WorkerException(String message) {
+        super(message);
+      }
       public WorkerException(String message, Throwable cause) {
-				super(message, cause);
-			}
-		}
+        super(message, cause);
+      }
+    }
   }
 
   /**
--- a/test/shell/test_invalid_scalacopts.sh
+++ b/test/shell/test_invalid_scalacopts.sh
@@ -1,0 +1,43 @@
+# shellcheck source=./test_runner.sh
+
+dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+. "${dir}"/test_runner.sh
+. "${dir}"/test_helper.sh
+runner=$(get_test_runner "${1:-local}")
+
+test_logs_contains() {
+  scalaVersion=$1
+  expected=$2
+  
+  bazel build \
+   --repo_env=SCALA_VERSION=${scalaVersion} \
+   //test_expect_failure/scalacopts_invalid:empty \
+   2>&1 | grep "$expected"
+}
+
+test_logs_not_contains() {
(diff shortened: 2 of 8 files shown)


test/shell/test_invalid_scalacopts.sh
```diff
@@ -0,0 +1,43 @@
...
+  scalaVersion=$1
+  expected=$2
+
+  bazel build \
+   --repo_env=SCALA_VERSION=${scalaVersion} \
+   //test_expect_failure/scalacopts_invalid:empty \
+   2>&1 | grep -v "$expected"
+}
+
+test_logs_contains_no_stacktrace() {
```
simuons, line 28: Is this used anywhere?

src/java/io/bazel/rulesscala/worker/Worker.java
```diff
@@ -88,7 +98,8 @@ public void checkPermission(Permission permission) {
             code = e.code;
           } catch (Exception e) {
             System.err.println(e.getMessage());
-            e.printStackTrace();
+            if (e instanceof Interface.WorkerException) {}
```
simuons, line 101: Maybe here instead of empty block you could do `System.err.println(e.getMessage());`?
I think message is printed with `e.printStackTrace();` in an else statement as well. If not then leave this as is.

simuons (changes requested):
Thanks, @WojciechMazur, overall looks good to me. Could you please go over my comments/questions and also please format code (in some places it's off).

src/java/io/bazel/rulesscala/worker/Worker.java
WojciechMazur: Sure, the `e.printStackTrace` would print the `e.getMessage` so previously there have been exception message printed twice. Let's only print either only exception message or stack trace

simuons (approved):

liucijus (approved):
Thanks, @WojciechMazur!

PR #1609, opened 2024-09-04
Add netrc support for downloading sources

grepwood:
### Description
https://github.com/bazelbuild/rules_scala/pull/1509 missed a spot - netrc authentication also needs to be added to downloading sources for this to work in an environment, where we only use private mirrors, and these private mirrors will not even respond without authentication.

### Motivation
I found myself needing this kind of change at work. I would like to contribute this change so that I don't have to patch this into my Bazel workspace.


liucijus (approved):
Thanks. @grepwood!

simuons (approved):

PR #1611, opened 2024-09-04
Collect diagnostics in Scala 3

### LINKED ISSUE #1602: Compiler diagnostics for scala 3 are missing.
The [current Scala 3 diagnostics reporter](https://github.com/bazelbuild/rules_scala/blob/master/src/java/io/bazel/rulesscala/scalac/reporter/PlaceholderForEmptyScala3Lib.java) looks like it is a noop. 

It looks like the scala 3 compiler has a `ConsoleReporter` [here](https://github.com/scala/scala3/blob/5101daf9f454c23898ec205894a1c97025f39515/compiler/src/dotty/tools/dotc/reporting/ConsoleReporter.scala#L13). Are there any obstacles for implementing it like the scala 2 variant?

WojciechMazur:
### Description
* Port `ProtoReporter` and `DepsTrackingReporter` to Scala 3 
* Adapt ScalacInvoker to use new reports in Scala 3 
* Enable reports tests for Scala 3  (test_scala_version). Mitigate differences in warnings/errors/info reported by Scala 3 and Scala 2 by introducing a dedicated test cases

### Additional info: 
Would conflict with #1606 in ScalacInvoker for Scala 3
Required workaround for https://github.com/scala/scala3/issues/21533 - we cannot extend `ConsoleReporter` in Scala 3 directly. I've decided to port exact behaviour of scalac ConsoleReporter to allow for the most identical semantics across different build tools.
I'm also going to remove tech-debt between all 4 implementations of ProtoReporter/ProtocReporter by introducing a common abstraction in the follow-up PR

### Motivation
Fixes #1602


simuons (changes requested):
Hi, @WojciechMazur, thanks for this PR, looks good to me, sorry for delay. Could you please rebase this PR on master? Then I'll merge it once green.

WojciechMazur:
Thank you for taking care of that, I've made a rebase. 
I'm also preparing a port of dependency_analyzer compiler plugin that is missing for Scala 3. It should finally introduce parity for most of Scala 2 and 3 features and would take advantage of DepsTrackingReporter introduced in this PR.

simuons (approved):

Issue #1550, opened 2024-03-06
javacopts order differs from java_library


patrick-premont:
The order of javacopts parameters passed to javac differs from the order seen in java_library.

This can result in different options taking precedence within the same toolchain based on whether it is used from scala_library or java_library.

In the example below `--release` and `-source/-target` are being reordered leading to a java_library with Java 8 classes and a scala_library with Java 21 classes.

## BUILD.bazel
```
load("@io_bazel_rules_scala//scala:scala.bzl", "scala_library")
load("@bazel_tools//tools/jdk:default_java_toolchain.bzl", "default_java_toolchain")

default_java_toolchain(
    name = "custom_toolchain",
    package_configuration = [":custom_package_config"],
    source_version = "21",
    target_version = "21",
)

java_package_configuration(
    name = "custom_package_config",
    javacopts = ["--release 8"],
    packages = [":this_package"],
)

package_group(
    name = "this_package",
    packages = ["//"],
)

genrule(
    name = "generate_example_java",
    outs = ["Example.java"],
    cmd = "echo 'public class Example {}' > $@",
)

scala_library(
    name = "scala_library",
    srcs = ["Example.java"],
)

java_library(
    name = "java_library",
    srcs = ["Example.java"],
)
``` 

## WORKSPACE.bazel
```
# WORKSPACE
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

register_toolchains("//:custom_toolchain_definition")

# The rest of this file is taken from the README.md of rules_scala, with version 6.4.0
http_archive(
    name = "bazel_skylib",
    sha256 = "b8a1527901774180afc798aeb28c4634bdccf19c4d98e7bdd1ce79d1fe9aaad7",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/releases/download/1.4.1/bazel-skylib-1.4.1.tar.gz",
        "https://github.com/bazelbuild/bazel-skylib/releases/download/1.4.1/bazel-skylib-1.4.1.tar.gz",
    ],
)

# See https://github.com/bazelbuild/rules_scala/releases for up to date version information.
http_archive(
    name = "io_bazel_rules_scala",
    sha256 = "9a23058a36183a556a9ba7229b4f204d3e68c8c6eb7b28260521016b38ef4e00",
    strip_prefix = "rules_scala-6.4.0",
[...]

PR #1613, opened 2024-09-09
extra javac opts should override default ones

thomasbao12:
### Description

Fixes https://github.com/bazelbuild/rules_scala/issues/1550

Currently the default javac flags will be appended last, overriding the manual ones.

### DIFF
--- a/scala/private/rule_impls.bzl
+++ b/scala/private/rule_impls.bzl
@@ -157,10 +157,9 @@ def compile_java(ctx, source_jars, source_files, output, extra_javac_opts, provi
         output = output,
         javac_opts = expand_location(
             ctx,
-            extra_javac_opts +
             java_common.default_javac_opts(
                 java_toolchain = java_toolchain,
-            ),
+            ) + extra_javac_opts,
         ),
         deps = providers_of_dependencies,
         #exports can be empty since the manually created provider exposes exports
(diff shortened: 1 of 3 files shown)


liucijus:
@thomasbao12 could you add a test to prevent regressions?

thomasbao12:
@liucijus I added a test and verified it fails on master https://github.com/bazelbuild/rules_scala/pull/1615.

I tried to follow the pattern established in other javac tests.

One issue though is I assume the default java toolchain sets a -target flag.  

3 potential options:
0. merge as is
1. I could rename the test to clarify that I actually am only testing that I want to be able to override the java target version.  This is actually the behavior I want to fix.
2. I could try a more elaborate unit-test that verifies that two javac flags exist, one from the default toolchain and one from javacopts and that the javacopt one is ordered after the default toolchain.

I'm okay with 0 & 1.  I worry 2 is not worth doing and overly specific.

liucijus:
> 3 potential options:
> 
>     0. merge as is
> 
>     1. I could rename the test to clarify that I actually am only testing that I want to be able to override the java target version.  This is actually the behavior I want to fix.
> 
>     2. I could try a more elaborate unit-test that verifies that two javac flags exist, one from the default toolchain and one from javacopts and that the javacopt one is ordered after the default toolchain.
> 

@thomasbao12 if I haven't missed anything, option 0 seems the best to me - it communicates  your need plus the name implies generic expectation

liucijus (approved):
Thanks, @thomasbao12!

thomasbao12:
@liucijus Can you merge?

simuons (approved):

PR #1622, opened 2024-10-15
Update test scripts for Bzlmod compatibility

mbland:
### Description

These are mostly small changes to make test assertions more flexible between `WORKSPACE` and Bzlmod runs.

For Bzlmod runs:

- Fixed `test_scala_config_content` from `test_scala_config.sh` by changing a path from `external/io_bazel_rules_scala_config` to `external/*io_bazel_rules_scala_config`.

- Fixed a number of tests by updating expected output messages to allow them to start with either `@//` or `@@//`.

- Fixed `test_stamped_target_label_loading` from `test/shell/test_strict_dependency.sh` by accommodating the canonical `io_bazel_rules_scala_guava` repo name. Also allows for the optional current Scala version suffix.

Also made these other important changes:

- Updated all the assertions in `test_helper.sh` to use Bash builtin regex matching via `_expect_failure_with_messages` instead of `grep`. This allows the expected message patterns to use full regular expressions while avoiding forking a new process. This new function helped reduce duplication in that file at the same time.

- Added `--repo_env="SCALA_VERSION=..."` to each test script called from `./test_coverage.sh`, and set `SCALA_VERSION` to 2.12.19 in each of these files. Using other Scala versions technically works, but the output is slightly different, causing the `diff` commands in the test cases to fail.

- Updated `test_version.sh` to copy the top level `.bazelversion` file into its test repo.

- Changed how `test_version.sh` handles injecting `twitter_scrooge` repos into the `WORKSPACE` file. This will make it easy to do the equivalent for `MODULE.bazel` when the time comes.

Also includes a few minor, opportunistic formatting cleanups.

### Motivation

Part of adding Bzlmod support per #1482.

I'm breaking this out to make sure the current default build configuration continues to pass as I carve off more of [my fork's Bzlmod working branch](https://github.com/mbland/rules_scala/commits/bzlmod/). At the same time, the later changes can land without including so many test changes.

cc: @BillyAutrey @jayconrod @benjaminp @TheGrizzlyDev

### DIFF
--- a/test/shell/test_coverage_scalatest.sh
+++ b/test/shell/test_coverage_scalatest.sh
@@ -4,15 +4,19 @@ dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 . "${dir}"/test_helper.sh
 runner=$(get_test_runner "${1:-local}")
 
+SCALA_VERSION="${SCALA_VERSION:-2.12.19}"
+
 test_coverage_on() {
-    bazel coverage //test/coverage_scalatest:test-scalatest
+    bazel coverage \
+        --repo_env="SCALA_VERSION=${SCALA_VERSION}" \
+        //test/coverage_scalatest:test-scalatest
     diff test/coverage_scalatest/expected-coverage.dat $(bazel info bazel-testlogs)/test/coverage_scalatest/test-scalatest/coverage.dat
 }
 
 test_coverage_includes_test_targets() {
     bazel coverage \
-          --instrument_test_targets=True \
-          //test/coverage_scalatest:test-scalatest
+        --instrument_test_targets=True \
+        //test/coverage_scalatest:test-scalatest
     grep -q "SF:test/coverage_scalatest/TestWithScalaTest.scala" $(bazel info bazel-testlogs)/test/coverage_scalatest/test-scalatest/coverage.dat
 }
 
--- a/test/shell/test_coverage_scalatest.sh
+++ b/test/shell/test_coverage_scalatest.sh
@@ -4,6 +4,7 @@ dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 . "${dir}"/test_helper.sh
 runner=$(get_test_runner "${1:-local}")
 
+# Default to 2.12.19 for `diff` tests because other versions change the output.
 SCALA_VERSION="${SCALA_VERSION:-2.12.19}"
 
 test_coverage_on() {
--- a/test/shell/test_coverage_equals_in_target.sh
+++ b/test/shell/test_coverage_equals_in_target.sh
@@ -4,8 +4,12 @@ dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 . "${dir}"/test_helper.sh
 runner=$(get_test_runner "${1:-local}")
 
+SCALA_VERSION="${SCALA_VERSION:-2.12.19}"
+
 test_coverage_target_name_contains_equals_sign() {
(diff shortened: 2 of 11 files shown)


test/shell/test_coverage_scalatest.sh
```diff
@@ -4,15 +4,19 @@ dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
...
+        --repo_env="SCALA_VERSION=${SCALA_VERSION}" \
+        //test/coverage_scalatest:test-scalatest
     diff test/coverage_scalatest/expected-coverage.dat $(bazel info bazel-testlogs)/test/coverage_scalatest/test-scalatest/coverage.dat
 }
 
 test_coverage_includes_test_targets() {
     bazel coverage \
-          --instrument_test_targets=True \
-          //test/coverage_scalatest:test-scalatest
+        --instrument_test_targets=True \
```
simuons, line 18: Shouldn't `--repo_env="SCALA_VERSION=${SCALA_VERSION}"` be passed here as well?

simuons (changes requested):
Thanks, @mbland, looks good to me. Please take a look at my question. Once you confirm it's intentional I'll approve this PR.

test/shell/test_coverage_scalatest.sh
mbland: Good catch! But I actually went the other way—the tests missing `--repo_env="SCALA_VERSION=${SCALA_VERSION}"` are only `grep`ping for the expected filename, not `diff`ing against an expected coverage file. So I've added comments in each file to explain the `SCALA_VERSION` setting.

I think the actual solution might be to generate coverage files for each Scala version, and use `$SCALA_VERSION` to select between them instead of setting `--repo_env`. I can take that on as a side quest, perhaps.

mbland:
On another note, I'm going to open a new PR to try to bump `rules_java` to 7.11.1 to see if it resolves this persistent "Bazel green head" CI breakage:

```txt
bazel --nosystem_rc --nohome_rc info
ERROR: Failed to load Starlark extension '@@rules_java//toolchains:toolchain_utils.bzl'.
 - @@rules_java
This could either mean you have to add the '@@rules_java' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
ERROR: cannot load build configuration key because of this cycle
  File "bazelci.py", line 4588, in <module>
    sys.exit(main())
  File "bazelci.py", line 4556, in main
    execute_commands(
  File "bazelci.py", line 1321, in execute_commands
    bazel_version = print_bazel_version_info(bazel_binary, platform)
  File "bazelci.py", line 1623, in print_bazel_version_info
    execute_command(
  File "bazelci.py", line 2750, in execute_command
    return subprocess.run(
  File "/usr/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['bazel', '--nosystem_rc', '--nohome_rc', 'info']' returned non-zero exit status 2.
```

mbland:
Ah, just reminded myself of the problem with `rules_java` and `WORKSPACE`. Filed https://github.com/bazelbuild/rules_scala/issues/1625 for a general discussion, so it doesn't get buried here.

mbland:
Rebased and passing after #1627, including latest Bazel 7.3.2.

simuons (approved):
Thanks, @mbland. Again, it's a pleasure to read descriptions of your PRs.

liucijus (approved):

PR #1623, opened 2024-10-16
Bump to rules_proto 6.0.2, separate protobuf 21.7

mbland:
### Description

Updates the build to depend on rules_proto 6.0.2, and the now separate protobuf v21.7.

Part of #1482.

As explained in the 6.0.0 release notes, rules_proto 6.x no longer depends directly on com_google_protobuf, so we need to import it separately. We're keeping the Protobuf version as 21.7 for now, because Protobuf 22.x requires C++14 at a minimum, which Bazel 6.x doesn't support by default:

- https://protobuf.dev/news/v22/#c11-support
- https://github.com/protocolbuffers/protobuf/releases/tag/v22.0

The downside is that, unline rules_proto-5.3.0-21.7, we now end up building (and occasionally rebuilding) `protoc` as part of our build.

### Motivation

This is in preparation for adding a `MODULE.bazel` file, which will bring in dependencies that use rules_proto 6.x. This change avoids that warning, and changing it now ensures that `WORKSPACE` builds are still compatible.

The flip side of the downside is that rules_proto 6.0.2 also enables clients to try "Protobuf Toolchainization" to download precompiled `protoc` binaries.

- https://github.com/bazelbuild/rules_proto/releases/tag/6.0.0
- https://docs.google.com/document/d/1CE6wJHNfKbUPBr7-mmk_0Yo3a4TaqcTPE0OWNuQkhPs/edit

For now, Protobuf Toolchainization  requires Bazel 6.5.0 or Bazel >= 7 and the `--incompatible_enable_proto_toolchain_resolution` flag:

- https://bazel.build/reference/command-line-reference#flag--incompatible_enable_proto_toolchain_resolution

I've got a [working draft implementation of proto toolchainization in a separate branch](https://github.com/mbland/rules_scala/tree/proto-toolchainization), though that experiment is certainly not urgent.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -45,14 +45,21 @@ rules_scala_setup()
 
 rules_scala_toolchain_deps_repositories(fetch_sources = True)
 
-load("@rules_proto//proto:repositories.bzl", "rules_proto_dependencies", "rules_proto_toolchains")
+load("@rules_proto//proto:repositories.bzl", "rules_proto_dependencies")
 
-# Declares @com_google_protobuf//:protoc pointing to released binary
-# This should stop building protoc during bazel build
-# See https://github.com/bazelbuild/rules_proto/pull/36
 rules_proto_dependencies()
 
+load("@rules_proto//proto:setup.bzl", "rules_proto_setup")
+
+rules_proto_setup()
+
+load("@rules_proto//proto:toolchains.bzl", "rules_proto_toolchains")
+
 rules_proto_toolchains()
+
+load("@com_google_protobuf//:protobuf_deps.bzl", "protobuf_deps")
+
+protobuf_deps()
 
 load("//scala:scala_cross_version.bzl", "default_maven_server_urls")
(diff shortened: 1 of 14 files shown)


simuons (approved):
Thanks, @mbland, especially for reminding about `Proto Toolchainisation`.

Side not for myself: the amount of same changes you had to do in this PR is frightening.

liucijus (approved):

mbland:
> Side not for myself: the amount of same changes you had to do in this PR is frightening.

Heh, on the one hand, this experience has taught my why `WORKSPACE` must be destroyed. 😛 

On the other, I've yet to try it, but I think I can introduce an API shared between `WORKSPACE` and the upcoming module extension that will drop a lot of this boilerplate to a single `load` and macro call. I plan on trying it at some point today in [my Bzlmod working branch](https://github.com/mbland/rules_scala/commits/bzlmod/).

Issue #1625, opened 2024-10-18
Bazel green build CI job now fails persistently due to rules_java versioning


mbland:
All CI jobs except "Bazel green head" pass for my PRs working towards #1482 with the following error:

```txt
bazel --nosystem_rc --nohome_rc info
ERROR: Failed to load Starlark extension '@@rules_java//toolchains:toolchain_utils.bzl'.
Cycle in the workspace file detected. This indicates that a repository is used prior to being defined.
The following chain of repository dependencies lead to the missing definition.
 - @@rules_java
This could either mean you have to add the '@@rules_java' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
ERROR: cannot load build configuration key because of this cycle
Traceback (most recent call last):
  File "bazelci.py", line 4588, in <module>
    sys.exit(main())
  File "bazelci.py", line 4556, in main
    execute_commands(
  File "bazelci.py", line 1321, in execute_commands
    bazel_version = print_bazel_version_info(bazel_binary, platform)
  File "bazelci.py", line 1623, in print_bazel_version_info
    execute_command(
  File "bazelci.py", line 2750, in execute_command
    return subprocess.run(
  File "/usr/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['bazel', '--nosystem_rc', '--nohome_rc', 'info']' returned non-zero exit status 2.
```

After reviewing my notes from https://github.com/bazelbuild/rules_scala/commit/cd22d8896f209c37be8622f1dcc9770c2dc18f48, under `Bump to rules_java 7.9.0 for Bazel 7 compatibility`, it seems like our `WORKSPACE` builds can't be compatible with Bazel 6/7 and Bazel 8 at the same time. The basic issue is that for rules_java > 7.9.0, the `WORKSPACE` prefix in Bazel 6/7 causes Java toolchains to register under a different type than what the rules are expecting later. This issue doesn't affect Bzlmod builds.

Would it be possible to disable the "Bazel green head" builds until the Bzlmod compatibility lands, and then reenable that build with Bzlmod enabled?

simuons:
Hi, @mbland, I've changed `last_green` to `latest` in https://github.com/bazelbuild/rules_scala/pull/1627
[...]

PR #1627, opened 2024-10-21
Build against latest bazel version instead of last_green

simuons:
### Description
Reverts https://github.com/bazelbuild/rules_scala/pull/1349

This task should be optional however it fails all builds because failing `bazel info`.

Error is related to workspace setup. Since we are in the process of moving to bzlmod I think we should not bother with that now. Especially when reproducing locally it gives different error.

### Motivation
Unblock pull requests.


simuons:
Actual reason of failed build can be found here https://github.com/bazelbuild/rules_scala/issues/1625

PR #1608, opened 2024-09-03
Automated creating and updating scala_x_x.bzl repository files

lm1nrt:
### Description
- Updated SHA's, dependencies, artifacts for all of the scala_x_x.bzl repository files with transitive dependencies compliance
- Added a script that simplifies creating and updating scala_x_x.bzl repository files.

It's not the ideal version, but I believe it can be improved by adding support for more Scala artifacts.

### Motivation
Updating and creating new versions for scala_x_x.bzl files is time-consuming, and it's a kind of work that is not performed regularly due to its specifics.

PR #1629, opened 2024-10-23
Add tools/sync-bazelversion.sh, commit results

mbland:
### Description

Ensures Bazelisk uses the same Bazel version in every workspace in the project. Part of #1482.

The `sync-bazelversion.sh` script is currently in `tools/`. #1608 introduced the `scripts/` directory, and I'm happy to move it there, but that directory seems focused on `third_party/repositories/scala_*.bzl` update automation.

### Motivation

I discovered that the workspaces under the main workspace were running the latest mainline Blaze release I had installed, instead of 6.5.0 from `.bazelversion`. Now all the workspaces are synchronized on the same Bazel version.

As noted in the `tools/sync-bazelversion.sh` comments, symlinks or an `import ../.bazelversion` mechanism would be preferable. However, the former can surprise Windows users, and Bazelisk doesn't support the latter.


simuons (changes requested):
Hi, @mbland, thanks for PR!

So `tools`/`scripts` is somewhat confusing and to be honest I'm not 100% sure about difference.

Just to note: `scripts` existed before https://github.com/bazelbuild/rules_scala/pull/1608

After looking at content of each directory, I think `tools` contain something that is used during `bazel build` or can be invoked via `bazel run`. While `scripts` holds anything else (most likely automate repetitive tasks).

I'm not sure such distinction is needed. We gonna deal with it later.

So I you don't mind I'll ask to move `sync-bazelversion.sh` to `scripts/`

Anything else looks awesome.

mbland:
Whoops, you're right—a `git log scripts/` could've told me it was there before #1608. If I'd've thought to check that before, that's where I would've put it to begin with. D'oh!

Just pushed a new commit to move it to `scripts/`.

simuons (approved):

PR #1634, opened 2024-10-23
Add CommandLineParserAdapter for >=2.13.9

mbland:
### Description

Fixes //third_party/utils/src/test:test_util for Scala 2.13.9 and later by providing a thin `scala.tools.cmd.CommandLineParser` adapter. Part of #1482.

### Motivation

Before this change, building under Scala 2.13.14 produced:

```txt
$ bazel build --repo_env=SCALA_VERSION=2.13.14 \
    //third_party/utils/src/test:test_util

ERROR: .../third_party/utils/src/test/BUILD:6:14:
  scala @@//third_party/utils/src/test:test_util failed: (Exit 1):
  scalac failed: error executing Scalac command
  (from target //third_party/utils/src/test:test_util)
  bazel-out/darwin_arm64-opt-exec-ST-d57f47055a04/bin/src/java/io/bazel/rulesscala/scalac/scalac
  @bazel-out/darwin_arm64-fastbuild/bin/third_party/utils/src/test/test_util.jar-0.params

third_party/utils/src/test/io/bazel/rulesscala/utils/TestUtil.scala:10:
  error: object cmd is not a member of package tools
    import scala.tools.cmd.CommandLineParser
                       ^
third_party/utils/src/test/io/bazel/rulesscala/utils/TestUtil.scala:119:
  error: not found: value CommandLineParser
    val options = CommandLineParser.tokenize(compileOptions)
                  ^
```

Turns out `CommandLineParser` disappered in Scala 2.13.9, and its `Parser` replacement is private:

- https://github.com/scala/scala/pull/10057

### DIFF
--- a/third_party/utils/src/test/io/bazel/rulesscala/utils/TestUtil.scala
+++ b/third_party/utils/src/test/io/bazel/rulesscala/utils/TestUtil.scala
@@ -13,7 +13,7 @@ import scala.tools.nsc.Settings
 import scala.tools.nsc.reporters.StoreReporter
 import io.bazel.rulesscala.dependencyanalyzer.DependencyTrackingMethod
 
-object TestUtil extends CompilerAPICompat {
+object TestUtil {
   final val defaultTarget = "//..."
 
   val isWindows: Boolean = System.getProperty("os.name").toLowerCase.contains("windows")
@@ -115,7 +115,7 @@ object TestUtil extends CompilerAPICompat {
     output: AbstractFile
   ): List[StoreReporter#Info] = {
     // TODO: Optimize and cache global.
-    val options = tokenize(compileOptions)
+    val options = CommandLineParserAdapter.tokenize(compileOptions)
     val reporter = new StoreReporter()
     val settings = new Settings(println)
     val _ = new CompilerCommand(options, settings)
--- a/third_party/utils/src/test/io/bazel/rulesscala/utils/TestUtil.scala
+++ b/third_party/utils/src/test/io/bazel/rulesscala/utils/TestUtil.scala
@@ -7,14 +7,13 @@ import scala.reflect.io.AbstractFile
 import scala.reflect.io.Directory
 import scala.reflect.io.PlainDirectory
 import scala.reflect.io.VirtualDirectory
-import scala.tools.cmd.CommandLineParser
 import scala.tools.nsc.CompilerCommand
 import scala.tools.nsc.Global
 import scala.tools.nsc.Settings
 import scala.tools.nsc.reporters.StoreReporter
 import io.bazel.rulesscala.dependencyanalyzer.DependencyTrackingMethod
 
-object TestUtil {
+object TestUtil extends CompilerAPICompat {
   final val defaultTarget = "//..."
 
   val isWindows: Boolean = System.getProperty("os.name").toLowerCase.contains("windows")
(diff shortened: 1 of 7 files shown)


third_party/utils/src/test/scala/tools/cmd/Scala_2_13_9_CommandLineParser.scala
```diff
@@ -0,0 +1,8 @@
+package scala.tools.cmd;
```
WojciechMazur, line 1: It might be unlikely but in theory some other compiler plugin might define the same workaround. Have you maybe considered creating a dedicated proxy explicitly for rulesscala to decrease chances of name clash? 

I had this issue recently when starting work on dependency analyzer plugin port to Scala 3. I've defined it there as following: 
For Scala 2.13.10+
```scala
package scala {
  package rulesscala {
    // proxy to private[scala] compiler API
    object Proxy {
      def tokenize(cmd: String): List[String] = sys.process.Parser.tokenize(cmd)
    }
  }
}

package io.bazel.rulesscala.utils {
  trait CompilerAPICompat {
    def tokenize(cmd: String): List[String] = scala.rulesscala.Proxy.tokenize(cmd)
  }
}
```

For other versions: 
```scala
package io.bazel.rulesscala.utils

import scala.tools.cmd.CommandLineParser

trait CompilerAPICompat {
  def tokenize(cmd: String): List[String] = CommandLineParser.tokenize(cmd)
}

```

Then I've redefined `object TestUtil` to extend this trait and used a proxy method instead of depending on compiler internals

mbland: Oh, this is better. I haven't actually done much Scala in a while, so this approach didn't occur to me, but it's definitely better than monkeypatching `scala.tools.cmd`.

I didn't see your original code in the existing code base, so I literally copied and pasted it into a new commit (crediting you, of course).

If this should live somewhere else, or you're about to check it in somewhere else (a quick `find | xargs grep` didn't turn it up), let me know.

third_party/utils/src/test/io/bazel/rulesscala/utils/TestUtil.scala
```diff
@@ -7,14 +7,13 @@ import scala.reflect.io.AbstractFile
...
 import scala.reflect.io.VirtualDirectory
-import scala.tools.cmd.CommandLineParser
 import scala.tools.nsc.CompilerCommand
 import scala.tools.nsc.Global
 import scala.tools.nsc.Settings
 import scala.tools.nsc.reporters.StoreReporter
 import io.bazel.rulesscala.dependencyanalyzer.DependencyTrackingMethod
 
-object TestUtil {
+object TestUtil extends CompilerAPICompat {
```
simuons, line 16: Would it be possible to make `CompilerAPICompat` an object instead of a trait? It's not a big deal I guess, but on the other hand you are adding `tokenize` to the api of `TestUtil` and I'm not sure it belongs here.

simuons (changes requested):

third_party/utils/src/test/io/bazel/rulesscala/utils/TestUtil.scala
mbland: Done. I was thinking about doing that, too, as I tend to prefer helper objects to mixins for similar reasons.

I also renamed it to `CommandLineParserAdapter` to reflect the very specific reason for its existence.

simuons (approved):
Thanks, @mbland!

PR #1635, opened 2024-10-23
Create the `@scala_compiler_sources` repo

mbland:
### Description

Contains aliases to versioned Scala compiler source repository targets. Part of #1482.

Updates the version specific repo references in the srcs attribute of `//third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/compiler:dep_reporting_compiler`. Now these are references to versioned targets in
`@scala_compiler_sources//`, which are aliases to those versioned compiler source repos.

### Motivation

In a Bzlmod world, this enables `rules_scala` to import only the `scala_compiler_sources` repo in `MODULE.bazel`, instead of importing each individual versioned compiler source repo.

This then allows `rules_scala` clients to set multiple `SCALA_VERSIONS` without requiring them to import this repo or any versioned compiler source repo. The Bzlmodifcation of the test repos under `dt_patches` (coming in a future change) revealed the need for this flexibility.

### DIFF
--- a/third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/compiler/BUILD
+++ b/third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/compiler/BUILD
@@ -1,15 +1,8 @@
 load("//scala:scala.bzl", "scala_library_for_plugin_bootstrapping")
-load("@io_bazel_rules_scala_config//:config.bzl", "SCALA_VERSIONS")
-load("@io_bazel_rules_scala//scala:scala_cross_version.bzl", "version_suffix")
 
 scala_library_for_plugin_bootstrapping(
     name = "dep_reporting_compiler",
-    srcs = select({
-        "@io_bazel_rules_scala_config//:scala_version" + version_suffix(v): [
-            "@scala_compiler_sources//:src%s" % version_suffix(v),
-        ]
-        for v in SCALA_VERSIONS
-    }),
+    srcs = ["@scala_compiler_sources//:src"],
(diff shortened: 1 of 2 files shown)


third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/compiler/BUILD
```diff
@@ -5,7 +5,9 @@ load("@io_bazel_rules_scala//scala:scala_cross_version.bzl", "version_suffix")
 scala_library_for_plugin_bootstrapping(
     name = "dep_reporting_compiler",
     srcs = select({
```
simuons, line 7: Since you are hiding/unifying version specific repositories under `@scala_compiler_sources` maybe this `select` should be moved into that repository. I mean `@scala_compiler_sources` would have a single alias target with select to version specific sources. For example:
```
alias(
  name = "src",
  actual = select({"@io_bazel_rules_scala_config//:scala_version_2_12_19": "@scala_compiler_source_2_12_19//:src"}),
)
```
And here in compilation target you would specify `srcs = ["@scala_compiler_sources//:src"],`

What do you think?

Also I think that compilation target could go into that new repository, but let's not go down that rabbit hole now.

simuons (changes requested):

third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/compiler/BUILD
mbland: I like it. Done.

simuons (approved):
Thanks, @mbland!

PR #1639, opened 2024-10-30
Refactor `scripts/create_repository.py`, fix some issues

mbland:
### Description

Reformats and refactors `scripts/create_repository.py`, prevents artifact downgrades, and improves performance. Part of #1482.

This change is comprised of several commits:

- The first is a pure reformatting and refactoring that produces the same output as before.
- The second is further reformatting and refactoring that still produces the same output.
- The third is further reformatting and refactoring still, but with a performance improvement, and prevents downgrading `protobuf-java`.
- The fourth is a small refactoring, but one I thought worth separating from previous ones.

Commits after that proceed to resolve various issues and improve performance:

- Ensure Scala 2 libs get "_2" repo names in Scala 3
- Bump Scalafmt to 3.8.3 in to match the version set in #1631
- Prevent jar downgrades by comparing new and existing versions
- Write all artifact deps to repo files, not just ones already existing in the original repository file

If this is too much for one pull request, I'm happy to tease it into as many smaller ones as requested.

I tested along the way by first restoring previous repository files at specific `$COMMIT` values via:

```sh
for f in third_party/repositories/scala_*.bzl; do git show $COMMIT:$f > $f; done
```

then I'd:

- run the script the first time to ensure the repository files remained the same as before (for the pure refactorings), or that the output was as expected
- `git add third_party/repositories/scala_*.bzl` if the updates were as expected
- run the script a second time to ensure no new changes appeared in the working directory

### Motivation

Some of the trickiest, most time consuming work I've done while resolving #1482 has been in resolving repository versions by hand. This script is now starting to save me _soooo_ much time. However, it needed some extra love not to undo some of the changes I've needed to make.

The easiest way for me to do that was to start with some pure reformatting and easy refactoring. After a couple iterations, I understood the script better, and could see where to add or update the logic to resolve some issues.

There's more I could do with this script to improve it further, I think, but that's for future pull requests. Many thanks and kudos to @lm1nrt for contributing the original in #1608.

PR #1642, opened 2024-11-07
Refactor `scripts/create_repository.py`, improve efficiency and correctness, and bump Scalatest, protobuf-java, kind-projector

mbland:
### Description

Refactors `scripts/create_repository.py` to make more modular, robust, and efficient, and update Scalatest, protobuf-java, and kind-projector jars. Part of #1482.

There are _a lot_ of changes from individual commits in this one large PR, since @simuons said it wasn't necessary to break up #1639, another large PR to refactor `scripts/create_repository.py`, and all tests continue to pass. That said, I'm always more than happy to break this one apart if it really is too much.

The highlights are:

- Bumps to Scalatest 3.2.19, protobuf-java 4.28.3, and kind-projector 0.13.3
- Each of the output dictionaries are now sorted by key.
  - This makes it much easier to see when and how dependencies change from this point forward, since the ordering of artifacts will remain stable.
- There are now several distinct objects, making the relationships between groups of functions and any shared data much clearer.
- The object based design also has made it easier to tweak details and add features without fear of violating scoping assumptions.
- If the label naming algorithm changes, or any artifacts are added to `EXCLUDED_ARTIFACTS`, the all script will update existing dictionary data appropriately.
  - Even if an artifact isn't upgraded by the script, the script will update its dictionary key and all `deps`  labels referring to it, or remove the artifact and its references altogether.
  - The script does not yet, however, handle updating macros that instantiate artifact repositories from this data. I have ideas for how to do that, but it's out of scope for this change.

Perhaps most significantly, this cuts the number of Coursier commands and artifact downloads significantly, leading to a very large performance improvement. Here are the times for creating a fresh output dir and then updating it before this change (all with a warm Coursier download cache):
    
```txt
$ /usr/bin/time ./scripts/create_repository.py --output_dir before

[ ...snip... ]
       66.01 real        13.25 user         9.99 sys

$ /usr/bin/time ./scripts/create_repository.py --output_dir before

[ ...snip... ]
        1.76 real         1.41 user         0.49 sys
```

Here are the times after:

```txt
$ /usr/bin/time ./scripts/create_repository.py --output_dir after

[ ...snip... ]
        1.16 real         0.84 user         0.39 sys

$ /usr/bin/time ./scripts/create_repository.py --output_dir after

[ ...snip... ]
        0.96 real         0.72 user         0.27 sys
```

### Motivation

Overall, this script is making my further work to update dependencies to ease the Bzlmod migration in #1482 so much faster, easier, and safer. But I found ways to improve it further, after #1639, to make the script itself more maintainable, more correct and robust, and much faster.

I'm now using it to bump ScalaPB, protoc-bridge, gRPC, guava, etc. dependencies, and resolving any breakages, in an effort to make rules_scala able to use newer versions of the Protobuf library. Though independent from (and not strictly required by) the Bzlmod migration work, this may be especially important to future Bzlmod users that include later Protobuf versions in their dependencies.

The script is now in a state where I really trust it to do the right thing as I continue this work. This makes this complex upgrade work far more tractable than before.

### DIFF
--- a/scripts/create_repository.py
+++ b/scripts/create_repository.py
@@ -214,7 +214,8 @@ class ArtifactLabelMaker:
     def _get_label_impl(self, coordinates) -> str:
         group = coordinates.group
         group_label = self._labelize(group)
-        artifact_label = self._labelize(coordinates.artifact.split('_')[0])
+        artifact = self._remove_scala_version_suffix(coordinates.artifact)
+        artifact_label = self._labelize(artifact)
 
         if group in self._SCALA_LANG_GROUPS:
             return self._get_scala_lang_label(artifact_label, coordinates)
@@ -225,8 +226,7 @@ class ArtifactLabelMaker:
         if group in self._SCALA_PROTO_RULES_GROUPS:
             return self._get_scala_proto_label(artifact_label, coordinates)
 
-        # Remove any Scala version suffix from the end, e.g., scopt_2.13.
-        artifact_name = coordinates.artifact_name().rsplit('_', 1)[0]
+        artifact_name = f'{group}:{artifact}'
 
         if artifact_name in self._SPECIAL_CASE_ARTIFACT_LABELS:
             return self._SPECIAL_CASE_ARTIFACT_LABELS[artifact_name]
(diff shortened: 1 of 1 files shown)


simuons (approved):
Thanks!

liucijus (approved):

PR #1643, opened 2024-11-08
Make scala_xxx rules declare that they provide JavaInfo

crt-31:
### Description
Made most scala_xxxx rules declare that they provide JavaInfo (setting the "provides" attribute).
This allows these targets to be visited by aspects that use "required_providers=[JavaInfo]". 

This matches behavior in java_xxx rules, and goal was to address all the equivelant scala_xxx rules.

rules updated are: scala_import, scala_library, scala_binary, scala_test, scala_junit_test, scala_proto, and scala_repl. (scala_proto_library was already correct).

### Motivation
Allows scala_xxx rules to be visited by aspects that use "required_providers=[JavaInfo]"

### DIFF
--- a/scala/private/rules/scala_binary.bzl
+++ b/scala/private/rules/scala_binary.bzl
@@ -92,6 +92,7 @@ def make_scala_binary(*extras):
         ],
         cfg = scala_version_transition,
         incompatible_use_toolchain_transition = True,
+        provides = [JavaInfo],
         implementation = _scala_binary_impl,
     )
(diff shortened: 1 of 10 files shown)


mbland:
Looks like you need to run `bazel run //tools:lint_fix`. Got me on my first few PRs, too.

simuons (approved):
Thanks, @crt-31.

liucijus (approved):

PR #1638, opened 2024-10-29
Add toggle-workspace.sh to turn WORKSPACE on/off

mbland:
### Description

Updates .bazelrc files to enable or disable WORKSPACE builds, or to restore their original state. Part of testing the Bzlmod work in #1482.

### Motivation

Like `scripts/sync-bazelversion.sh` from #1629 does for `.bazelversion` files, this makes switching between `WORKSPACE` and Bzlmod builds fast, easy, and safe. This has made working ahead in [my Bzlmod working branch](https://github.com/mbland/rules_scala/commits/bzlmod) while keeping in sync with pull requests landing on `master` much smoother.

That said, I understand if it _isn't_ accepted. I'm happy to hold onto it in my working branch to suit my needs if folks don't find it generally useful.

PR #1644, opened 2024-11-08
Make every .bazelrc import the top level .bazelrc

mbland:
### Description

Makes the Bazel configuration for every test repository consistent with the main repository. Supersedes #1638.

Removes `test_version/.bazel{rc,version}`, since `test_version.sh` was always copying the top level `.bazel{rc,version}` files into the test repo to begin with. Updates `scripts/sync-bazelversion.sh` such that it won't replace `test_version/.bazelversion` on future runs.

Also fixes a bug in `test/shell/test_scala_proto_library.sh` whereby it never successfully read the top level `.bazelversion` file.

### Motivation

With this change, `scripts/toggle-workspace.sh` from #1638 isn't really necessary, since this allows changes from the top level `.bazelrc` to propagate to all nested repos. I no longer need that script to easily and safely switch between `WORKSPACE` and Bzlmod while working on changes for #1482.

I only realized this while working on [my new branch to use the latest abseil-cpp 20240722.0 and protobuf v28.3](https://github.com/mbland/rules_scala/tree/bzlmod-update-abseil-cpp-and-protobuf), which is based on [my branch to update to ScalaPB 0.11.17, gRPC 1.68.1, and Guava 33.3.1-jre](https://github.com/mbland/rules_scala/tree/bzlmod-update-protocbridge-and-grpc), which is itself based on #1642.

---

As an aside, the upshot of that protobuf upgrade work so far:

- Later versions of Protobuf appear to require ScalaPB 1.0.0-alpha.1, which causes a single test to hang, `test_scala_version 2.11.12` from `test_version.sh`. Appears to be another `ProtoScalaPBRule` aspect worker hanging, this time on `//src/main/scala/scalarules/test/twitter_scrooge:java_jar_not_on_classpath`.

- The latest abseil-cpp and protobuf will build under Bazel 6.5.0 with the compiler flags from https://github.com/bazelbuild/bazel/pull/20785, but those cause protobuf to rebuild even more frequently. These flags aren't required under Bazel 7.4.0, so the problem isn't as pronounced.


mbland:
Hmm, looks like [the `test_reproducibility` job broke due to a failure to download Bazel](https://buildkite.com/bazel/rules-scala-scala/builds/5175#01930e07-2e43-4012-803f-895335ed4853). In other words, the breakage is unrelated to this change; all the other test jobs passed.

```txt
2024/11/08 23:06:26 Downloading https://releases.bazel.build/6.5.0/release/bazel-6.5.0-linux-x86_64...
2024/11/08 23:06:26 could not download Bazel: failed to download bazel: failed to download bazel:
  could not copy from https://releases.bazel.build/6.5.0/release/bazel-6.5.0-linux-x86_64 to
  /var/lib/buildkite-agent/.cache/bazelisk/downloads/_tmp/download1892374034:
  stream error: stream ID 1; INTERNAL_ERROR; received from peer
```

simuons (approved):
Thanks

liucijus (approved):

PR #1598, opened 2024-08-13
[bugfix] normalize paths to enable running external tests

joeljeske:
### Description

Normalize file paths before passing them to runfiles.rlocation() within the test runner. 

Before this change, you could not execute a bazel test target that is not in the primary workspace, it would error with 

For example running `bazel run @my_external_repo//my/path/to:test` 
```
Exception in thread "main" java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at com.google.testing.coverage.JacocoCoverageRunner.main(JacocoCoverageRunner.java:592)
Caused by: java.lang.IllegalArgumentException: path is not normalized: "my_workspace/../my_external_repo/my/path/to/test/test.args"
	at com.google.devtools.build.runfiles.Util.checkArgument(Util.java:45)
	at com.google.devtools.build.runfiles.Runfiles.rlocation(Runfiles.java:337)
	at io.bazel.rulesscala.scala_test.Runner.extendFromFileArgs(Runner.java:51)
	at io.bazel.rulesscala.scala_test.Runner.extendArgs(Runner.java:37)
	at io.bazel.rulesscala.scala_test.Runner.main(Runner.java:33)
	... 5 more
```

### Motivation

We should be able to run tests from external repos

PR #1645, opened 2024-11-09
Pass rlocationpaths to Test Runner

crt-31:
### Description
Changed so that paths passed to the Test Runner are in "rlocationpath" form. This should be more correct way to handle these paths as that is what rlocation() expects.  

This fixes issue where we couldn't test external repo tests on Windows. 

Note: seems to have been chasing this issue in the past (#1598. #1500), hopefully this solves for good.

### Motivation
Handle paths in more robust way. And run scalatests from external repos on windows again.


simuons (approved):
Thanks, @crt-31! Things got cleaner. 

Finding files in bazel is hard :)

liucijus (approved):

PR #1646, opened 2024-11-09
Clean `test_{runner,scalafmt_helper}`, add diagnostic env vars, and improve `test_cross_build` performance

mbland:
### Description

Cleans up some of the formatting of `test/shell/test_{runner,scalafmt_helper}.sh` and adds the `RULES_SCALA_TEST_{ONLY,VERBOSE}` environment variables. Improves `test_cross_build.sh` performance signficantly.

Defining `RULES_SCALA_TEST_ONLY`will cause only one specific test case to run, with full Bazel output. This is helpful for pinpointing and working on a specific test without hunting for it or recreating its commands.

Defining `RULES_SCALA_TEST_VERBOSE` will cause all tests to emit full Bazel output. This is more useful when running all the tests from a specific script from `test/shell/test_*.sh`.

The `test_cross_build.sh` changes speed up its tests substantially, particularly between runs. Here are the times for two runs before this change; the first is after running `bazel clean` and `bazel shutdown` in the `test_cross_version` directory:

```txt
$ /usr/bin/time ./test_cross_build.sh

running test test_cross_build
 Test "test_cross_build" successful (73 sec)
running test test_scalafmt
 Test "test_scalafmt" successful (94 sec)
      167.01 real         0.37 user         0.50 sys

$ /usr/bin/time ./test_cross_build.sh

running test test_cross_build
 Test "test_cross_build" successful (8 sec)
running test test_scalafmt
 Test "test_scalafmt" successful (92 sec)
       99.96 real         0.35 user         0.48 sys
```

Here are the times for two equivalent runs after this change:

```txt
$ /usr/bin/time ./test_cross_build.sh

running test test_cross_build
 Test "test_cross_build" successful (71 sec)
running test test_scalafmt
 Test "test_scalafmt" successful (6 sec)
       77.50 real         0.33 user         0.44 sys

$ /usr/bin/time ./test_cross_build.sh

running test test_cross_build
 Test "test_cross_build" successful (5 sec)
running test test_scalafmt
 Test "test_scalafmt" successful (2 sec)
        7.21 real         0.26 user         0.38 sys
```

### Motivation

The cleanup began as an attempt to get protobuf to stop recompiling so often in [my new branch to use the latest abseil-cpp 20240722.0 and protobuf v28.3](https://github.com/mbland/rules_scala/tree/bzlmod-update-abseil-cpp-and-protobuf). It didn't impact that problem, but they were good cleanups regardless. (Though today I learned that `SECONDS` is a special Bash variable, and marking it `local` is a mistake that makes all tests look like they ran in zero seconds.)

The `test_cross_version.sh` speedups were also a result of working on that branch, on which the speedups described above were even more dramatic.

Previously, every `bazel run` in `test_scalafmt` would rebuild the entire protobuf library. I thought it was hanging, until I used `RULES_SCALA_TEST_ONLY` and could see what was happening. Then I used `RULES_SCALA_TEST_VERBOSE` to see the relationship between `bazel` invocations in `test_cross_build.sh`. That led me to remove the `bazel clean` from `test_cross_build` and move the `bazel shutdown` to the end of the script.

As always, this is spawned from work related to #1482, though it is a bit of a side quest.

### DIFF
--- a/test/shell/test_cross_build.sh
+++ b/test/shell/test_cross_build.sh
@@ -9,11 +9,11 @@ cd test_cross_build
 
 function test_cross_build() {
   bazel test //...
-  bazel clean
-  bazel shutdown;
 }
 
 function test_scalafmt() {
+  bazel build //scalafmt/...
+
   run_formatting scalafmt binary2 binary2
   run_formatting scalafmt binary3 binary3
   run_formatting scalafmt library2 library2
@@ -24,3 +24,10 @@ function test_scalafmt() {
 
 $runner test_cross_build
 $runner test_scalafmt
+
+# `bazel shutdown` used to be in `test_cross_build`, after a `bazel clean`.
+# However, the protobuf library tends to rebuild frequently. Not cleaning and
+# postponing the shutdown to the end of the script helps avoid rebuilding as
+# often, speeding up the tests. It also potentially speeds up debugging by
+# preserving the workspace state when a test fails.
+bazel shutdown
--- a/test/shell/test_runner.sh
+++ b/test/shell/test_runner.sh
@@ -13,9 +13,10 @@ run_test_ci() {
   echo "running test $TEST_ARG"
   eval $TEST_ARG &>$log_file &
   local test_pid=$!
+
   SECONDS=0
-  TIMOUT=${TEST_TIMEOUT-60}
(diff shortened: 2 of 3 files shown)


simuons (approved):
Thanks!

Issue #1647, opened 2024-11-10
Protobuf compatibility findings and suggestions


mbland:
As a side quest to #1482, I've been experimenting with trying to update the [Protobuf library](https://github.com/protocolbuffers/protobuf) beyond [version 21.7](https://github.com/protocolbuffers/protobuf/releases/tag/v21.7), released 2022-09-29. The reason is that I anticipate potential complications after the Bzlmod update lands if users have later Protobuf versions in their `MODULE.bazel` dependency graph.

I've discovered the following details, which we may need to communicate to users through release notes and/or other documentation. Here are the combinations of various dependencies that work with one another. The first row is the current combination.

| Protobuf | abseil-cpp | Bazel 6 | ScalaPB | Scala |
| -------: | ---------: | :-----: | ------: | ----: |
| v21.7 | 20220623.1 | &#x2705; | 0.9.0 | All |
| v21.7 | 20220623.1 | &#x2705; | 0.9.8 | All |
| v21.7 | 20220623.1 | &#x2705; | 0.11.17 | >= 2.12 |
| v25.5 | 20240722.0 | with `--cxxopt` | 0.9.8 | All |
| v25.5 | 20240722.0 | with `--cxxopt` | 0.11.17 | >= 2.12 |
| >= v28.2 | 20240722.0 | with `--cxxopt` | 1.0.0-alpha.1 | >= 2.12 |

## Suggestions

**I'd suggest updating the `.bazelversion` to 7.4.0 and bumping Protobuf to v25.5**. My reasoning is that these changes would bring `rules_scala` closer to recent Bazel and Protobuf updates, without potentially breaking existing users.

If we decide to build with Bazel 7 by default instead of Bazel 6, the Protobuf recompilation sensitivity won't be as much of an issue. In my experience, after merging #1619, #1620, and #1622, I haven't found any other Bazel 6 and 7 incompatibilities outside of building newer Protobuf versions.

It would be nice to bump to ScalaPB 1.0.0-alpha.1 to use the latest Protobuf v28.3. However, it seems like that might be too big a bump for existing users right now.

### ScalaPB and the `scala_proto_aspect` are sensitive to the Protobuf version

[...]

PR #1648, opened 2024-11-12
Bump ScalaPB, protoc-bridge, gRPC, and Guava deps, and add `ProtobufAdapter`s and `ScalaPBCodeGenerator` wrappers

mbland:
### Description

Adds ScalaPB 0.11.17, protoc-bridge 0.9.8, gRPC 1.68.1, and Guava 33.3.1-jre as root artifacts in `scripts/create_repository.py` and `ProtobufAdapter`s for cross Scala version compatibility. Adds `ScalaPBCodeGenerator` wrappers to ensure all ScalaPB workers catch all exceptions to avoid build hangs when updating the Protobuf library. 

This change does _not_ bump the Protobuf library version, which can happen in another PR and is up for discussion in #1647. As always, I'm happy to break this apart, at least into one "version bump and compatibility updates" PR and one "wrap `ScalaPBCodeGenerator`" PR, if so desired.

There are _lots_ of details in the individual commit messages.

### Motivation

These changes are spawned from #1482 and #1647, whereby I investigated what it would take to update the Protobuf version to accommodate future Bzlmod users that depend on newer Protobuf versions.

These changes were essential to the Protobuf experiment, yet aren't necessarily essential to the Bzlmod work, but they're generally beneficial to include for their own sake. All builds and tests are still compatible with all supported Scala versions, on Bazel 6 and Bazel 7.

This is also a good demonstration of how to add new root artifact dependencies to `scripts/create_repository.py`, and how easy it is to run the script and get correct updates across all supported Scala versions.

Regarding the `ScalaPBCodeGenerator` wrappers, I encountered more hanging builds in [my bzlmod-update-abseil-cpp-and-protobuf working branch](https://github.com/mbland/rules_scala/tree/bzlmod-update-abseil-cpp-and-protobuf) that the changes from #1637 didn't catch. I backported [Wrap ScalaPbCodeGenerator, expose Scala 2.11 error](https://github.com/bazelbuild/rules_scala/commit/de8214bc788bfa9b7cdb42793a9c195d0b674fec) from that branch into this one.

See that commit's message and #1647 for the gory details on why the ScalaPB libs are so sensitive to the Protobuf version and why we have to use wrappers to avoid hanging the build.


simuons (approved):
Thanks

liucijus (approved):

PR #1649, opened 2024-11-13
Fix windows Buildkit CI 

crt-31:
Apparently windows CI hasn't been running for a long time (or ever?)

The issue seems to be that for windows, buildkite needs "batch_commands" instead of "shell_commands".

(Update)
Also made minor fixes to tests regarding line endings. _I originally (#1529) had the system work with autocrlf=true, but this isn't handled by BazelCI. And after thinking bout it, its probably best to keep autocrlf=false, since that is what is distributed to users.. even windows users._ 

### Motivation
Make windows CI work so we don't accidently break windows support.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -49,6 +49,7 @@ tasks:
     environment:      
       MSYS2_ARG_CONV_EXCL: "*"
     batch_commands:
+    - "set PATH=/usr/bin;%PATH%" #Make sure bash uses msys commands over windows commands. (i.e. find).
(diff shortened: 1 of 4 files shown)


simuons:
Hi, @crt-31, thanks for PR. You are right. Unfortunately windows support was not a priority. Thanks for taking this.

crt-31:
@simuons, should be proper diff now. RTR.

simuons (approved):
Thanks, @crt-31 !

liucijus (approved):

Issue #1652, opened 2024-11-25
Bazel 8.0.0 compatibility


gergelyfabian:
When testing Bazel 8.0.0 RC versions I discovered rules_scala may be incompatible with the latest rules_java (that you in turn my need for Bazel 8.0.0).

This came out when testing a Bazel bug's solution: https://github.com/bazelbuild/bazel/issues/24235.

After upgrading rules_java to 8.5.0+ rules_scala reported errors:

```
# Using https://github.com/gergelyfabian/bazel-scala-example/tree/bazel-8.0.0rc1
$ USE_BAZEL_VERSION=8.0.0rc3 bazel build //...
INFO: Invocation ID: b8abf68e-49c2-48ee-a5e4-f0d8710f759d
ERROR: Traceback (most recent call last):
	File "/home/user/.cache/bazel/_bazel_user/64527517dacf7170f7d914889f83481c/external/io_bazel_rules_scala/scala/private/rules/scala_doc.bzl", line 52, column 37, in <toplevel>
		_scaladoc_transitive_aspect = aspect(
Error in aspect: Illegal argument: element in 'required_aspect_providers' is of unexpected type. Either all elements should be providers, or all elements should be lists of providers, but got list with an element of type NoneType.
ERROR: Error computing the main repository mapping: at /home/user/.cache/bazel/_bazel_user/64527517dacf7170f7d914889f83481c/external/io_bazel_rules_scala/scala/scala.bzl:20:5: initialization of module 'scala/private/rules/scala_doc.bzl' failed

# Or another one:

$ USE_BAZEL_VERSION=8.0.0rc3 bazel build //...
INFO: Invocation ID: 117a90d7-da17-400c-9f83-908c7ddd7317
WARNING: WORKSPACE support will be removed in Bazel 9 (late 2025), please migrate to Bzlmod, see https://bazel.build/external/migration.
ERROR: Traceback (most recent call last):
	File "/home/user/.cache/bazel/_bazel_user/64527517dacf7170f7d914889f83481c/external/io_bazel_rules_scala/scala/private/common_attributes.bzl", line 18, column 28, in <toplevel>
		"deps": attr.label_list(
Error in label_list: Illegal argument: element in 'providers' is of unexpected type. Either all elements should be providers, or all elements should be lists of providers, but got list with an element of type NoneType.
[...]

PR #1653, opened 2024-11-27
Toolchainize all testing toolchains

mbland:
### Description

Moves the Scalatest, JUnit, and Specs2 toolchains into `@io_bazel_rules_scala_toolchains//testing`. Part of #1482.

Introduces more macros to return a framework's Maven artifact dependencies, rather than inlining them in a `repositories` call. These inlined lists are replaced by macro invocations, and now the `scala_toolchains` macro can invoke these macros to collect artifact IDs to pass to `repositories`. This also allows for future changes to introduce a `scala_version` parameter if necessary, similar to how `scalafmt_artifact_ids` already works.

Updates all `WORKSPACE` files to set the appropriate `scala_toolchains` parameters and to remove the unnecessary repository import and toolchain registration macros.

Adds a `fetch_sources_by_id` parameter to `repositories` from `third_party/repositories/repositories.bzl`.

Updates `scala/scala_maven_import_external.bzl` to generate a `load` line for `//scala:scala_import.bzl` based on the repo's canonical name, not `@io_bazel_rules_scala`.

As usual, includes several other opportunistic removals of the `@io_bazel_rules_scala` repo name prefix to avoid an internal dependency on that name.

### Motivation

The macros returning artifact ID lists help avoid collisions when creating repositories dependencies upon which multiple frameworks depend under Bzlmod. `WORKSPACE` doesn't seem affected by these collisions, but Bzlmod will produce errors like the following, where both `scala_proto` and `twitter_scrooge` depend upon `io_bazel_rules_scala_guava`:

```txt
$ bazel build //src/...

ERROR: .../scala/scala_maven_import_external.bzl:299:24:
    File ".../scala/extensions/deps.bzl",
      line 140, column 21, in _scala_deps_impl
        scala_toolchains(
    File ".../scala/private/macros/toolchains.bzl",
      line 140, column 17, in scala_toolchains
        _scrooge(
    File ".../twitter_scrooge/twitter_scrooge.bzl",
      line 96, column 17, in twitter_scrooge
        repositories(
    File ".../third_party/repositories/repositories.bzl",
      line 113, column 37, in repositories
        _scala_maven_import_external(
    File ".../scala/scala_maven_import_external.bzl",
      line 263, column 30, in scala_maven_import_external
        jvm_maven_import_external(
    File ".../scala/scala_maven_import_external.bzl",
      line 299, column 24, in jvm_maven_import_external
        jvm_import_external(jar_urls = jar_urls, srcjar_urls = srcjar_urls, coordinates = artifact, **kwargs)

Error in repository_rule: A repo named
  io_bazel_rules_scala_guava_2_13_15
  is already generated by this module extension at
  .../scala/scala_maven_import_external.bzl:299:24

ERROR: Analysis of target
  '//src/java/io/bazel/rulesscala/worker:worker_test'
  failed; build aborted:
  error evaluating module extension scala_deps in
  //scala/extensions:deps.bzl
```

Recent updates to `scripts/create_repository.py` (#1639, #1642) make it easy to emit full direct dependency lists for artifacts included in `third_party/repositories/scala_*.bzl`. This increases the likelihood of collisions, since this expanded metadata forces the macros that instantiate artifact repos to instantiate even more repos.

By fetching lists of artifact IDs from these macros, `scala_toolchains` can now consolidate them into dictionary keys. Then it passes these unique keys to `repositories` directly, avoiding the problem of instantiating the same repo multiple times in the same module extension.

This, in turn, also avoids the need to add parameters to the original `WORKSPACE` macros that instantiate dependencies to avoid collisions under Bzlmod. The `scala_toolchains` macro never needs to call these original macros, under either `WORKSPACE` or Bzlmod.

Finally, it also reduces duplication between these artifact ID lists and the `_*_DEPS` symbols originally from `testing/BUILD` (and now in `testing/deps.bzl`). The dependency labels are now generated programatically.

(Aside: As I mentioned, we may eventually need to pass a Scala version argument to these macros. It will be possible to cross that bridge without too much trouble if and when that day comes. Or I can try to future proof it in a follow up pull request.)

Adding the `fetch_sources_by_id` parameter to `repositories()` enables `scala_toolchains` to build the `artifact_ids_to_fetch_sources` mapping from artifact ID lists returned by these new macros. The values assigned to each id match the original `fetch_sources` settings in the corresponding original `WORKSPACE` macros.

Removing `@io_bazel_rules_scala` target prefixes means Bzlmod users won't necessarily have to set the `repo_name` parameter of `bazel_dep` when using `rules_scala`.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -41,10 +41,12 @@ scala_config(enable_compiler_dependency_tracking = True)
 
 load("//scala:scala.bzl", "scala_toolchains")
 
-scala_toolchains(fetch_sources = True)
+scala_toolchains(
+    fetch_sources = True,
+    testing = True,
+)
 
 register_toolchains(
-    "//testing:testing_toolchain",
     "//scala:unused_dependency_checker_error_toolchain",
     "//test/proto:scalapb_toolchain",
     "@io_bazel_rules_scala_toolchains//...:all",
@@ -78,14 +80,6 @@ jmh_repositories()
 load("//scala_proto:scala_proto.bzl", "scala_proto_repositories")
 
 scala_proto_repositories()
-
-load("//scalatest:scalatest.bzl", "scalatest_repositories")
-
-scalatest_repositories()
-
-load("//specs2:specs2_junit.bzl", "specs2_junit_repositories")
-
-specs2_junit_repositories()
 
 load("//scala/scalafmt:scalafmt_repositories.bzl", "scalafmt_default_config", "scalafmt_repositories")
 
--- a/examples/crossbuild/WORKSPACE
+++ b/examples/crossbuild/WORKSPACE
@@ -40,7 +40,7 @@ scala_config(
 
 load("@io_bazel_rules_scala//scala:scala.bzl", "scala_toolchains")
 
-scala_toolchains()
+scala_toolchains(scalatest = True)
 
 register_toolchains("@io_bazel_rules_scala_toolchains//...:all")
 
@@ -59,9 +59,3 @@ rules_proto_toolchains()
 load("@com_google_protobuf//:protobuf_deps.bzl", "protobuf_deps")
 
 protobuf_deps()
-
-load("@io_bazel_rules_scala//testing:scalatest.bzl", "scalatest_repositories", "scalatest_toolchain")
-
-scalatest_repositories()
-
-scalatest_toolchain()
--- a/examples/testing/multi_frameworks_toolchain/BUILD
+++ b/examples/testing/multi_frameworks_toolchain/BUILD
@@ -1,6 +1,4 @@
 load("@io_bazel_rules_scala//scala:scala.bzl", "setup_scala_testing_toolchain")
-load("@io_bazel_rules_scala//scala:scala_cross_version.bzl", "version_suffix")
-load("@io_bazel_rules_scala_config//:config.bzl", "SCALA_VERSION")
 
 setup_scala_testing_toolchain(
     name = "testing_toolchain",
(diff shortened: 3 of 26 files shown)


mbland:
Note that I've updated this to move the new Bzlmodified `scala_toolchains()` macro to `//scala:toolchains.bzl` per #1652, to start the process of partitioning `WORKSPACE`/Bzlmod macros off from `BUILD` macros. I've prepped a bunch of follow up commits to migrate the rest of the toolchains already, which also follow this principle.

For the details on how I think I'll be able to fold in a lot of Bazel 8 + `rules_java` 8 prep at the same time as the Bzlmod changes, see: https://github.com/bazelbuild/rules_scala/issues/1652#issuecomment-2510343584

mbland:
I just added a commit to move `scala/private/macros/toolchains_repo.bzl` to `scala/toolchains_repo.bzl` for reasons explained in the commit message. Then I rebased this on #1671, which should get merged before this pull request, to ensure the build continues to pass.

mbland:
Rebased after #1671 and #1672.

simuons (approved):
Thanks @mbland and sorry again for delay.

As of your comment about breaking changes: i guess we won't be able to avoid them at some point. I think bzlmod justifies some breaking changes and I'm fine with that.

mbland:
> Thanks @mbland and sorry again for delay.

No worries! We all slowed down a bit at the end of the year.

> As of your comment about breaking changes: i guess we won't be able to avoid them at some point. I think bzlmod justifies some breaking changes and I'm fine with that.

OK, then I'll file a documentation PR at some point to advise using `scala_register_toolchains()` from `scala/toolchains.bzl` in `WORKSPACE` to replace all the other toolchain registration macros, just as we're advising the use of `scala_toolchains()`.

liucijus (approved):

PR #1654, opened 2024-11-27
Add dt_patches/compiler_sources repo

mbland:
### Description

Introduces the `dt_patches/compiler_sources` repo. This is part of the Bzlmod prep work from #1482, but also reduces duplication between the `dt_patches/test_dt_patches*/WORKSPACE` files.

### Motivation

The structure of the `dt_patches/compiler_sources/extensions.bzl` declarations accommodates the fact that, unlike `WORKSPACE, `MODULE.bazel` files cannot import constants or contain conditional statements. That is to say, there's no way to port the `if SCALA_VERSION.startswith("2.")` expressions from the previous `WORKSPACE` files to `MODULE.bazel`.

The new `import_compiler_source_repos` macro, however, works for both `WORKSPACE` and Bzlmod builds, as it's trivial to wrap `import_compiler_source_repos` in a module extension.


simuons (approved):
Thanks @mbland

liucijus (approved):

PR #1656, opened 2024-11-29
Extract `dev_deps_repositories()` from WORKSPACE and remove `//private`

mbland:
### Description

`dev_deps_repositories()` encapsulates the instantiation of repositories used only for `rules_scala` development. Also removes the unused `//private` package and its `WORKSPACE` statements. Part of #1482 and #1652.

Replaces existing `new_local_repository` calls with complete nested repositories (with their own `WORKSPACE` files) and `local_repository` calls in the main `WORKSPACE` file. The first commit, however, calls `native.{,new_}local_repository` inside `dev_deps_repositories()`, in case we wish to use that implementation instead.

### Motivation

Both `WORKSPACE` and Bzlmod builds can use this macro. However, the two commits in this pull request reflect different strategies towards handling the existing `new_local_repository` calls.

Either way, we need to update the `new_local_repository` calls because:

- `native.{,new_}local_repository()` aren't available under Bzlmod (#1482).
- `@bazel_tools//tools/build_defs/repo:local.bzl` with the Starlarkified definitions isn't available under Bazel 6.
- Bazel 8 compatibility work is imminent (#1652).

[There are two paths we can go by (but in the long run, there's still time to change the road we're on)](https://youtu.be/QkF3oxziUI4):

1. Replacing the `new_local_repository` instances with complete nested repositories is compatible with Bazel 6, 7, and 8, both under `WORKSPACE` and Bzlmod. (`MODULE.bazel` will use a combination of `bazel_dep()` and `local_path_override()`.) That's the change reflected by the pull request as a whole.

2. If keeping the `new_local_repository` calls is preferable, we'll have to use `native.new_local_repository` and `native.local_repository` in `dev_deps_repositories()`. Then, when we switch to building with Bazel 7 by default, we'll have to load `local.bzl` and strip the `native.` prefixes.


simuons (approved):
Thanks @mbland!

Having dev dependencies in a single place is a nice thing.

[And by the way](https://www.youtube.com/watch?v=8f13FY94BKI)

liucijus (approved):

Issue #1674, opened 2025-01-13
Build a compiler extension and then use it


gergelyfabian:
In rules_scala's scalac BUILD file, we have the following.

```
SCALAC_DEPS = [
    "//scala/private/toolchain_deps:scala_compile_classpath",
    "//src/java/io/bazel/rulesscala/io_utils",
    "@bazel_tools//src/main/protobuf:worker_protocol_java_proto",
    "@io_bazel_rules_scala//src/java/io/bazel/rulesscala/jar",
    "@io_bazel_rules_scala//src/java/io/bazel/rulesscala/worker",
    "@io_bazel_rules_scala//src/protobuf/io/bazel/rules_scala:diagnostics_java_proto",
    "//src/java/io/bazel/rulesscala/scalac/compileoptions",
    "//src/java/io/bazel/rulesscala/scalac/reporter",
]

java_binary(
    name = "scalac",
    srcs = [
        ":scalac_files",
    ],
    javacopts = [
        "-source 1.8",
        "-target 1.8",
    ],
    main_class = "io.bazel.rulesscala.scalac.ScalacWorker",
    visibility = ["//visibility:public"],
    deps = (select_for_scala_version(
        any_2 = [
            "//third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/compiler:dep_reporting_compiler",
        ],
    ) if ENABLE_COMPILER_DEPENDENCY_TRACKING else []) + SCALAC_DEPS,
)
```

"//scala/private/toolchain_deps:scala_compile_classpath" is the critical part in my perspective. It's a:

```
common_toolchain_deps(
    name = "scala_compile_classpath",
    deps_id = "scala_compile_classpath",
    visibility = ["//visibility:public"],
)
```

That is in turn a:

```
_toolchain_type = "@io_bazel_rules_scala//scala:toolchain_type"

def _common_toolchain_deps(ctx):
    return expose_toolchain_deps(ctx, _toolchain_type)

common_toolchain_deps = rule(
    implementation = _common_toolchain_deps,
    attrs = {
        "deps_id": attr.string(mandatory = True),
    },
    toolchains = [_toolchain_type],
    incompatible_use_toolchain_transition = True,
)
```

Where:

```
def _required_deps_id_message(target, toolchain_type_label, deps_id):
    return target + " requires mapping of " + deps_id + " provider id on the toolchain " + toolchain_type_label

def java_info_for_deps(deps):
[...]

PR #1677, opened 2025-01-14
refactor scalac java_binary target creation

gergelyfabian:
I'd like to remove Scala compiler extensions from the Scala toolchain, that I could be potentially building myself (in the same monorepo, build the compiler extension, but also use it).

Till now the only way to add Scala compiler dependencies was a toolchain.
If I left the compiler extension as a toolchain dependency there, I could not build the compiler extension due to a cycle in deps (building the compiler extension would depend on itself through the toolchain).

I'd like to work this around by defining my own scalac java_binary target, where I can define my own dependencies (and then create rule names with macros that override the scalac target).

To do this without duplicating code from rules_scala, refactor SCALAC_DEPS parameter and scalac target definitions to a Starlark file to make them reusable for outside users.

Added:
- `define_scalac()`
- `define_scalac_bootstrap()`

A user could do:
```
EXTENDED_SCALAC_DEPS = DEFAULT_SCALAC_DEPS + [
    "@//my/compiler/dependency:library",
]

define_scalac(
    name = "scalac_with_extensions",
    deps = EXTENDED_SCALAC_DEPS,
)
```

And then build that compiler library in the same monorepo, this way, providing compiler dependencies without  toolchains.

See #1674 for more information.

### DIFF
--- a/src/java/io/bazel/rulesscala/scalac/definitions.bzl
+++ b/src/java/io/bazel/rulesscala/scalac/definitions.bzl
@@ -1,0 +1,48 @@
+load("@io_bazel_rules_scala_config//:config.bzl", "ENABLE_COMPILER_DEPENDENCY_TRACKING")
+load("@rules_java//java:defs.bzl", "java_binary")
+
+DEFAULT_SCALAC_DEPS = [
+    Label(dep)
+    for dep in [
+        "//scala/private/toolchain_deps:scala_compile_classpath",
+        "//src/java/io/bazel/rulesscala/io_utils",
+        "//src/java/io/bazel/rulesscala/jar",
+        "//src/java/io/bazel/rulesscala/worker",
+        "//src/protobuf/io/bazel/rules_scala:diagnostics_java_proto",
+        "//src/java/io/bazel/rulesscala/scalac/compileoptions",
+        "//src/java/io/bazel/rulesscala/scalac/reporter",
+        "@bazel_tools//src/main/protobuf:worker_protocol_java_proto",
+    ]
+]
+
+DEFAULT_SRCS = [
+    Label("//src/java/io/bazel/rulesscala/scalac:scalac_files"),
+]
+
+def define_scalac(name = "scalac", srcs = DEFAULT_SRCS, deps = DEFAULT_SCALAC_DEPS):
+    java_binary(
+        name = name,
+        srcs = srcs,
+        javacopts = [
+            "-source 1.8",
+            "-target 1.8",
+        ],
+        main_class = "io.bazel.rulesscala.scalac.ScalacWorker",
+        visibility = ["//visibility:public"],
+        deps = ([
+            Label("//third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/compiler:dep_reporting_compiler"),
+        ] if ENABLE_COMPILER_DEPENDENCY_TRACKING else []) + deps,
+    )
+
+def define_scalac_bootstrap(name = "scalac_bootstrap", srcs = DEFAULT_SRCS, deps = DEFAULT_SCALAC_DEPS):
+    java_binary(
+        name = name,
+        srcs = srcs,
+        javacopts = [
+            "-source 1.8",
+            "-target 1.8",
+        ],
+        main_class = "io.bazel.rulesscala.scalac.ScalacWorker",
+        visibility = ["//visibility:public"],
+        deps = deps,
+    )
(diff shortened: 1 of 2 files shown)


gergelyfabian:
Rebased with latest master to ensure the build for latest Bazel does not fail.

src/java/io/bazel/rulesscala/scalac/deps.bzl
```diff
@@ -0,0 +1,10 @@
+SCALAC_DEPS = [
+    "//scala/private/toolchain_deps:scala_compile_classpath",
+    "//src/java/io/bazel/rulesscala/io_utils",
+    "@bazel_tools//src/main/protobuf:worker_protocol_java_proto",
+    "@io_bazel_rules_scala//src/java/io/bazel/rulesscala/jar",
+    "@io_bazel_rules_scala//src/java/io/bazel/rulesscala/worker",
+    "@io_bazel_rules_scala//src/protobuf/io/bazel/rules_scala:diagnostics_java_proto",
+    "//src/java/io/bazel/rulesscala/scalac/compileoptions",
+    "//src/java/io/bazel/rulesscala/scalac/reporter",
+]
```
mbland, line 10: I think, to make this more robust and usable outside of `rules_scala` itself, this should be something like the following (which I haven't tried yet):

```py
SCALAC_DEPS = [
    Label(dep) for dep in [
        "//scala/private/toolchain_deps:scala_compile_classpath",
        "//src/java/io/bazel/rulesscala/io_utils",
        "//src/java/io/bazel/rulesscala/jar",
        "//src/java/io/bazel/rulesscala/scalac/compileoptions",
        "//src/java/io/bazel/rulesscala/scalac/reporter",
        "//src/java/io/bazel/rulesscala/worker",
        "//src/protobuf/io/bazel/rules_scala:diagnostics_java_proto",
        "@bazel_tools//src/main/protobuf:worker_protocol_java_proto",
    ]   
]
```

I'm also thinking ahead here to removing all `@io_bazel_rules_scala` prefixes by using `Label()` instead. (And also using `@rules_scala` instead where necessary.) This will avoid Bzlmod users from having to specify `repo_name = "io_bazel_rules_scala"` in their `bazel_dep()` calls. Using `Label()` as illustrated here should work right now, and make the diff for the eventual `@io_bazel_rules_scala` removal just that much smaller.

gergelyfabian: Thank you! I changed to this Label format.

gergelyfabian:
> In fact, I could move the java_binary's definition to the Starlark file too...

Just did that.

src/java/io/bazel/rulesscala/scalac/definitions.bzl
```diff
@@ -0,0 +1,48 @@
...
+        name = name,
+        srcs = srcs,
+        javacopts = [
+            "-source 1.8",
+            "-target 1.8",
+        ],
+        main_class = "io.bazel.rulesscala.scalac.ScalacWorker",
+        visibility = ["//visibility:public"],
+        deps = ([
+            "//third_party/dependency_analyzer/src/main/io/bazel/rulesscala/dependencyanalyzer/compiler:dep_reporting_compiler",
```
mbland, line 33: Don't forget to `Label()` this one, too! 😉

gergelyfabian: Done

gergelyfabian:
Not sure why the build failed, I just rebased with latest master.

mbland:
This looks like the culprit: https://buildkite.com/bazel/rules-scala-scala/builds/5293#0194694d-dea6-481f-85ca-36bcc3da80d3/126-462

```txt
ERROR: /workdir/test/gen_src/BUILD:20:13: scala @//test/gen_src:uses_gen_file failed:
  (Exit 1): scalac failed: error executing command
  (from target //test/gen_src:uses_gen_file)
  bazel-out/k8-opt-exec-2B5CBBC6/bin/src/java/io/bazel/rulesscala/scalac/scalac
  @bazel-out/k8-fastbuild/bin/test/gen_src/uses_gen_file.jar-0.params

bazel-out/k8-fastbuild/bin/test/gen_src/foo.scala:1: error: Invalid literal number
```

BTW, the other `ERROR` further down about a nonexistent `//test/coverage/...` package I've fixed in one of the commits in #1678. That "fix" doesn't actually affect the test result, because _both_ of the builds launched by the script fail in the same way, but it no longer hides the underlying breakage.

The error from this PR seems to be a different issue altogether. I'm not immediately sure what it is.

gergelyfabian:
> The error from this PR seems to be a different issue altogether. I'm not immediately sure what it is.

It seems it's non-deterministic (and probably not related to this PR). Amended my commit, force-pushed, and the rerun of the same test passed.

Issue #1675, opened 2025-01-14
scalafmt compilation breaks after upgrade to "Update to Scalafmt 3.8.3"


gergelyfabian:
If I build my example project with rules_scala (in version 190fbee676632eb67be34b5ee91613391fc8e633), I get the following errors:

```
external/io_bazel_rules_scala/scala/scalafmt/scalafmt/ScalafmtWorker.scala:35: error: Symbol 'type scala.meta.inputs.InputException' is missing from the classpath.
This symbol is required by 'class scala.meta.parsers.ParseException'.
Make sure that type InputException is in your classpath and check for conflicting dependencies with `-Ylog-classpath`.
A full rebuild may help if 'ParseException.class' was compiled against an incompatible version of scala.meta.inputs.
      case e @ (_: org.scalafmt.Error | _: scala.meta.parsers.ParseException) => {
                                           ^
external/io_bazel_rules_scala/scala/scalafmt/scalafmt/ScalafmtWorker.scala:35: warning: fruitless type test: a value of type Throwable cannot also be a scala.meta.parsers.ParseException
      case e @ (_: org.scalafmt.Error | _: scala.meta.parsers.ParseException) => {
                                                              ^
external/io_bazel_rules_scala/scala/scalafmt/scalafmt/ScalafmtAdapter.scala:5: error: object sysops is not a member of package org.scalafmt
import org.scalafmt.sysops.FileOps
                    ^
external/io_bazel_rules_scala/scala/scalafmt/scalafmt/ScalafmtAdapter.scala:9: error: not found: value FileOps
        FileOps.readFile(file.toPath())(codec)
        ^
external/io_bazel_rules_scala/scala/scalafmt/scalafmt/ScalafmtAdapter.scala:5: warning: Unused import
import org.scalafmt.sysops.FileOps

```

How to reproduce:

```
git clone https://github.com/gergelyfabian/bazel-scala-example
git checkout upgrade-rules-scala-to-scalafmt-changes
bazel build //...
```

The version just before 190fbee676632eb67be34b5ee91613391fc8e633 worked ok. Note: in my repo I also upgraded scalafmt to align with the version in rules_scala. Tested the previous version of rules_scala without scalafmt changes.

gergelyfabian:
Same error for 719f353b85129106a745d9825be2c09231d4fcae.

The latest version (6e36591103889e6cce6e2f71266544b313aeeaab) has a bigger blocker:

```
[...]

PR #1678, opened 2025-01-15
Toolchainize //scala/scalafmt:scalafmt_toolchain

mbland:
### Description

Mostly a pretty straightforward and easy update on top of the previous `setup_toolchains` and `scala_toolchains_repo` changes. Part of #1482.

However, it also includes a few opportunistic changes:

- Updates `test_reproducibility.sh` to replace the nonexistent `//test/coverage/...` target with packages that
actually exist, rendering the test more meaningful.
- Adds the `SCALAFMT_TOOLCHAIN_TYPE` constant to reduce string duplication.
- Removes the `dep_providers` default from `scala_toolchain`, since it's the _only_ attribute, so it doesn't make sense to have a default (inspired by #1675).
- Tweaks a few tests to ensure they still pass under Bazel 6 and Bzlmod.
- Updates `//third_party` tests to replace `$(location)` with `$(rootpath)`, ensuring they still pass under Bazel 8 (#1652).

See the commit messages for more details.

### Motivation

The first commit contains the next methodical step in the implementation of the `scala_toolchains()` macro and the `@io_bazel_rules_scala_toolchains` repository. This "toolchainization" is what will enable us to add a `MODULE.bazel` file to register `rules_scala` toolchains automatically, based on optional user configuration values.

Normally I'd split each of the other changes into separate pull requests, but since these are so small, it seemed it might expedite the process to include them all here. All are completely tested, and since they're encapsulated in separate commits, I'm happy to break them down into separate pull requests if preferred.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -50,6 +50,7 @@ load("//scala:toolchains.bzl", "scala_toolchains")
 
 scala_toolchains(
     fetch_sources = True,
+    scalafmt = True,
     testing = True,
 )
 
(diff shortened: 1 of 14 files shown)


simuons (approved):
Thanks

liucijus (approved):

Issue #1657, opened 2024-12-02
Better caching for Scala 3 (via ijar or tasty inspection)


susliko:
Bazel includes the [ijar](https://github.com/bazelbuild/bazel/blob/master/third_party/ijar/README.txt) tool, which strips regular JAR files down to their interfaces. Target hashes are computed based on the resulting ijars, ensuring that only the necessary targets are rebuilt when changes do not affect public interfaces (e.g., modifications to implementations or private method signatures).

However, this approach often falls short in practice. For Scala 2, as noted in [this issue](https://github.com/bazelbuild/rules_scala/issues/125), scalac adds a ScalaSignature annotation to classes, causing cache misses and unnecessary rebuilds.

For Scala 3, ijar generation is explicitly [disabled](https://github.com/bazelbuild/rules_scala/blob/d35448f943e3cefb7e1c7ba2da7982f846bc35f5/scala/private/phases/phase_compile.bzl#L105).

Is the situation with ijars similar for Scala 3?
If ijars are not a viable solution, would analyzing TASTy files for interface-level changes be a feasible alternative?

jjudd:
I agree that it would be great to improve ijar support for Scala or to use some other mechanism to improve detection of interface vs private implementation changes.

We have a discussion started in #1658 that includes a proposal for ijar (as well as many other things). Our current thoughts are to improve the ijar tool to be more Scala aware. If the Bazel folks are not interested in that, then we could explore other alternatives - perhaps there is some way to leverage the Zinc analysis file to accomplish this? Haven't thought much through that.

Part of the proposal covering ijar: https://docs.google.com/document/d/12BmUPpJpworA9ep2P_J00TxXwFRBZDQttUCsoYo0Yao/edit?tab=t.0#heading=h.z65373pwhv7g

jadenPete:
> Is the situation with ijars similar for Scala 3?

We've enabled `ijar` for Scala 3 targets in [lucidsoftware/rules_scala](https://github.com/lucidsoftware/rules_scala) haven't encountered any issues, so I think enabling it for Scala 3 targets should be fine. We have received the following warning:
```
ijar: skipping unknown attribute: "TASTY".
```

[...]

Issue #1658, opened 2024-12-02
Unifying the Scala Rulesets


jadenPete:
Hello! My name is Jaden; I'm a Software Engineer at Lucid Software (you may know us from Lucidchart). Lucid regularly uses Bazel to build our Scala backend that comprises millions of lines of code. We're highly invested in not only the success of the Scala ecosystem, but in Bazel's attractiveness as a build system of choice for large Scala repositories. For those who don't know, we actively maintain an alternative Scala ruleset, [lucidsoftware/rules_scala](https://github.com/lucidsoftware/rules_scala) (previously [higherkindness/rules_scala](https://github.com/higherkindness/rules_scala)). Although the ruleset began as a testbed for incremental compilation with Zinc and new/experimental Bazel features, we've since shifted away from incremental compilation in an effort to maintain correctness and compliance with the Bazel philosophy. Now, we'd like to propose merging various other features from our ruleset into this repository.

We've prepared a comprehensive proposal document that outlines the features our ruleset offers and how we think they'd benefit this one:
https://docs.google.com/document/d/12BmUPpJpworA9ep2P_J00TxXwFRBZDQttUCsoYo0Yao/edit?usp=sharing

As mentioned in the document, we intend to dedicate engineering time to this. From the [discussion](https://bazelbuild.slack.com/archives/CDCKJ2KFZ/p1732130369349749) happening in the #scala channel in the Bazel Slack workspace, I understand this sort of work would be best done after the Bzlmod migration and Bazel 8 upgrade, so know that we're happy to help with that in any way we can.

mbland:
Here are the Bzlmod and Bazel 8 tracking issues:

- #1482
- #1652 

We've got working prototypes (well, almost completely working in the Bazel 8 case), so it shouldn't be a very long wait.

BillyAutrey:
That document outlines _several_ features that you want implemented. Is the goal to port in all of these features, or none at all? Maybe it's worth separating out the discussion of each feature alone. I could see the discussion getting quite twisted if we're talking about seventeen different proposals wrapped into one.

jjudd:
Ideally we would like to port in all these features. We're happy to use this as a meta/tracking issue and open specific issues for each of the individual features. I'll chat with Jaden about that.

mbland:
[...]

PR #1679, opened 2025-01-15
Implement an ijar equivalent for Scala 3

jadenPete:
### Description

From my [comment](https://github.com/bazelbuild/rules_scala/issues/1657#issuecomment-2560292192) in https://github.com/bazelbuild/rules_scala/issues/1657:

> From what I understand, though, TASTy markes a major shift in how type information is stored between compiler invocations from Scala 2.13's "Pickle" format in that it includes a completely typed representation of the program's AST. I'm of the opinion that this defeats the purpose of using `ijar`, since most changes to a program's source code will now induce changes in its TASTy files, thus changing the generated ijar and invalidating reverse dependencies' cached compilation actions.
>
> When you consider that and [the fact](https://docs.scala-lang.org/scala3/guides/tasty-overview.html) that "`scalac` doesn’t have to read their _.class_ files anymore; it can read their _.tasty_ files", it may make more sense to develop an alternative to `ijar` specially for Scala 3 targets, as opposed to modifying `ijar`. That way, we can strip out information from TASTy files not needed for compilation (e.g. positional information, documentation, private members, etc.), similar to what @susliko described.

Given these changes in how symbols are read from the compile classpath in Scala 3 vs Scala 2, I don't think `ijar` is an appropriate solution for producing Scala 3 interface JARs. For that reason, I've developed "dottyijar", a version of `ijar` that produces an interface JAR by removing all `.class` files and performing a few modifications to `.tasty` files:
- Clearing the UUID, "Positions" section, and "Comments" section
- Removing all `val` and `def` definitions that are private and not inline
- Replace the values of all `val` and `def` definitions and class parent arguments with `???`
- Remove unused names from the name table

For context on the TASTy format, see [`dotty.tools.tasty.TastyFormat`](https://github.com/scala/scala3/blob/4d3f7576ccae724e6f83d2f3d68bd4c4e1dd5a14/tasty/src/dotty/tools/tasty/TastyFormat.scala).

Thus, `dottyijar` produces interface JARs that are far less sensitive to implementation changes than those produced by `ijar`. Although it's not as fast as `ijar` (mainly because it's written in Scala and doesn't read and write in a single pass), it's very fast—`dottyijar` produced an interface JAR for the Scala 3 compiler, which contains 1,011 TASTy files, in 4.369 seconds on my machine, whereas `ijar` took 0.292 seconds.

### Motivation

I'm implementing this change to close https://github.com/bazelbuild/rules_scala/issues/1657.

### DIFF
--- a/scala/private/macros/scala_repositories.bzl
+++ b/scala/private/macros/scala_repositories.bzl
@@ -189,6 +189,7 @@ def _artifact_ids(scala_version):
             "io_bazel_rules_scala_scala_library_2",
             "io_bazel_rules_scala_scala_reflect_2",
             "io_bazel_rules_scala_scala_tasty_core",
+            "io_bazel_rules_scala_scala_tasty_inspector",
             "org_jline_jline_native",
             "org_jline_jline_reader",
             "org_jline_jline_terminal",
--- a/scala/private/macros/scala_repositories.bzl
+++ b/scala/private/macros/scala_repositories.bzl
@@ -37,6 +37,12 @@ _COMPILER_SOURCES_ENTRY_TEMPLATE = """
         "@io_bazel_rules_scala_config//:scala_version{scala_version_suffix}":
             "@scala_compiler_source{scala_version_suffix}//:src","""
 
+_JUNIT_DEPS = ["io_bazel_rules_scala_org_hamcrest_hamcrest_core"]
+_SPECS2_DEPS = [
+    "org_portable_scala_portable_scala_reflect",
+    "org_scala_sbt_test_interface",
+]
+
 def _compiler_sources_repo_impl(rctx):
     sources = [
         _COMPILER_SOURCES_ENTRY_TEMPLATE.format(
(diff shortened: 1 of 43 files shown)


scala/private/macros/scala_repositories.bzl
```diff
@@ -167,12 +181,15 @@ def _artifact_ids(scala_version):
 
     if scala_version.startswith("3."):
         result.extend([
+            "com_softwaremill_common_tagging",
+            "dev_zio_izumi_reflect",
```
WojciechMazur, line 185: Are these really needed? Yes, these alllow to define more typesafe TASTy format specification, but at the same time these are introducing additional dependencies. Probably it would be best to keep this list as small as possible.

simuons:
Hi, @jadenPete, thanks for PR. Looks like checks are failing due to strict deps. Could you resolve this? I hadn't had a chance to look at PR yet but I'll try to in upcoming days.

jadenPete:
> Hi, @jadenPete, thanks for PR. Looks like checks are failing due to strict deps. Could you resolve this? I hadn't had a chance to look at PR yet but I'll try to in upcoming days.

Sure thing! I'll look into the test failures.

scala/private/macros/scala_repositories.bzl
jadenPete: `dev_zio_izumi_reflect` is needed during testing so we can get information about which TASTy nodes were read if `dottyijar` fails. Without it, debugging failures becomes near-impossible. `com_softwaremill_common_tagging` can probably be done without, since it's only used to differentiate `SignedInt` and `UnsignedInt`, and `SignedLong` and `UnsignedLong` at the type level. This isn't just for type safety—the compiler needs it so it can summon the appropriate `TastyFormat`. I think I can replicate this with value classes.

jadenPete: Done!

jadenPete:
All the tests are now passing! I've squashed all of the `fixup!` commits I added to make the pull request easier to review.

jadenPete:
Hello! It's no problem. Thanks for the update. This task is among others in this [meta-issue](https://github.com/bazelbuild/rules_scala/issues/1658) we'd like to submit pull requests for, but we're holding off to make the Bzlmod migration as easy as possible. Let me know if we can help with that in any way.

jadenPete:
Hi, @simuons! Now that the Bzlmod and Bazel 8 work with this ruleset is complete, is there any chance of this PR getting merged?

Issue #445, opened 2018-03-19
macros that reference Java libraries don't work


cushon:
The following example shows that a macro that calls a Java library doesn't work. This is related to bazelbuild/bazel#632. The library jar for `:JavaLib` on compile-time classpath only contains the compile-time API of that library, but the macro needs the complete implementation to be available at compile-time.

Demo: https://gist.github.com/cushon/a97470ed69f61d966c94bcde1ed018f0

Setup:

```
$ wget https://gist.github.com/cushon/a97470ed69f61d966c94bcde1ed018f0/archive/31e29b918d9e77cd84ec8bb3cab4cf584e5b2082.zip
$ unzip 31e29b918d9e77cd84ec8bb3cab4cf584e5b2082.zip
$ cd 31e29b918d9e77cd84ec8bb3cab4cf584e5b2082
```

Building `:HelloLib` fails:

```
bazel build :HelloLib
INFO: Analysed target //:HelloLib (23 packages loaded).
INFO: Found 1 target...
INFO: From scala //:MacroTest:
warning: there was one deprecation warning; re-run with -deprecation for details
one warning found
ERROR: /tmp/tmp.UHlgyDbGs0/a97470ed69f61d966c94bcde1ed018f0-31e29b918d9e77cd84ec8bb3cab4cf584e5b2082/BUILD:11:1: scala //:HelloLib failed (Exit 1)
HelloLib.scala:5: error: exception during macro expansion:
java.lang.ClassFormatError: Absent Code attribute in method that is not native or abstract in class file com/test/JavaLib
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at scala.test.MacroTest$.hello_impl(MacroTest.scala:14)

[...]

PR #1681, opened 2025-01-16
Enable scala_macro_library targets to have dependencies

jadenPete:
### Description

This PR should fix https://github.com/bazelbuild/rules_scala/issues/445. Before, macro targets couldn't depend on `ijar`d targets because targets depending on those macros would get the compile (`ijar`d) versions of those `ijar`d targets, which can't be executed at compile time due to necessary information being stripped out. Instead, we need to include macro targets' transitive runtime JARs on the compile classpath. 

This PR makes Scala targets export a `ScalaInfo` provider containing a `contains_macros` attribute. If one of a target's dependencies exports a `ScalaInfo` provider with `contains_macros` set to `True` (i.e. the dependency is a macro), that dependencies' transitive runtime JARs are hoisted to the top of the compile classpath. This is necessary because if they were included anywhere else in the classpath, we'd risk the compile-time versions of those transitive runtime JARs being used instead.

The approach is explained in more detail in `collect_jars` in `scala/private/common.bzl`. Another benefit of this approach is that we can `ijar` macro targets and their transitive dependencies and therefore benefit from reduced incremental build times.

### Motivation

Closing https://github.com/bazelbuild/rules_scala/issues/445.


simuons:
Hi, @jadenPete, thanks for PR. Are failing tests related to your changes?

jadenPete:
No problem! Yes, my changes caused those failures. I just pushed a fix.

simuons (approved):
Thanks, @jadenPete!

jadenPete:
No problem! Let me rebase this branch, after which it can be merged.

jadenPete:
Squashed.

jadenPete:
Done!

liucijus (approved):

PR #1696, opened 2025-02-13
Remove @io_bazel_rules_scala or replace with Label

mbland:
### Description

Removes all remaining internal usages of `@io_bazel_rules_scala` without changing any actual logic. Part of #1482 and #1652, this is the last change required before adding Bzlmod compatibility.

Updates `WORKSPACE` files to use `@rules_scala` and `scala_register_toolchains()` instead of `@io_bazel_rules_scala` and `register_toolchains("@io_bazel_rules_scala_toolchains//...:all")`.

Also adds all the scripts run in CI to `test_all.sh`, plus `dt_patches/dt_patch_test.sh`.

Leaves `@io_bazel_rules_scala_config` as is for now, because it's a documented public interface. It can be easily changed and documented in a future commit if desired.

Also doesn't change the `io_bazel_rules_scala_` prefix for Maven artifact repos. This could also be done and documented in a future commit if desired.

### Motivation

This change enables `WORKSPACE` and Bzlmod users to import `rules_scala` as `rules_scala`, instead of requiring `io_bazel_rules_scala` due to internal dependencies.

i.e., `WORKSPACE` users can still use `http_archive` to import `rules_scala` as `io_bazel_rules_scala`. Bzlmod users will be able to call `bazel_dep(name = "rules_scala", repo_name = "io_bazel_rules_scala")`. However, `io_bazel_rules_scala` is no longer required by `rules_scala` itself.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,4 +1,4 @@
-workspace(name = "io_bazel_rules_scala")
+workspace(name = "rules_scala")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("//scala:deps.bzl", "rules_scala_dependencies")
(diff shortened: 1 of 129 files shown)


simuons (approved):
Thanks, much awaited cleanup!

Issue #1699, opened 2025-02-14
Please add support to Scala 2.13.16 and make new release


bartoszkosiorek:
Please add support for Scala 2.13.16:
https://github.com/scala/scala/releases/tag/v2.13.16

Finally it would be great to release new version as for example scala 2.13.15 is supported only on `master` (not with latest release v6.6.0)

WojciechMazur:
I can help with upgrade (although you've opended the PR already so there wouldn't more to do probably) but I would not be able to make a release. 

@liucijus @simuons 
Is there some plan for releases? Is it blocked by something, eg. ongoing work by @mbland? 
Should users use a fixed commit version? I guess there was a lot of changes since the last release half year ago.

mbland:
The 2.13.16 update could go in before the Bzlmod changes from #1482. There could even be a point release before that; everything I've done so far maintains compatibility with Bazel 6.5.0 and 7.5.0. (Maybe earlier versions of each, too, but if not, I think the burden should be on users to bump their minor versions of Bazel.)

That PR isn't quite ready yet, though. I'll comment on it.

mbland:
BTW, the final Bzlmod changes are imminent. Just to be clear, we don't _have_ to wait for them to land and to cut the next major release before cutting another minor release, but the wait might not be very long, either. It's all up to whatever @simuons and @liucijus prefer.

crt-31:
>  I guess there was a lot of changes since the last release half year ago.

I vote that we should definately have a 6.x release before integrating the bzlmod.  I want some of the changes since July.

mbland:
It just occurred to me that I spoke a little too soon. There are some breaking changes in `master`, described in https://github.com/bazelbuild/rules_scala/issues/1482#issuecomment-2605683509.

They're mostly very easy to resolve, but they are breaking. So the next release should be a major one either way. It would be possible to release 7.0 without Bzlmod, and 7.1 with.

[...]

PR #1700, opened 2025-02-14
Upgrade Scala 2.13 LTS to 2.13.16

bartoszkosiorek:
Fixes: #1699

### Description
Upgrade Scala 2.13 LTS to 2.13.16


third_party/repositories/scala_3_6.bzl
```diff
@@ -205,7 +205,7 @@ artifacts = {
         ],
     },
     "io_bazel_rules_scala_scala_compiler_2": {
-        "artifact": "org.scala-lang:scala-compiler:2.13.15",
+        "artifact": "org.scala-lang:scala-compiler:2.13.16",
         "sha256": "4c200cd193c082bec14a2a2dffe6a1ba5f8130b1b27c79ee54c936dfcafc8ed9",
     },
```
WojciechMazur, line 210: Was it updated using the search and replace? The SHA would probably not match. The files in this directory should be updated using `./scripts/create_repository.py` after updating the Scala 2.13 version specified there.

WojciechMazur (approved):
LGTM

third_party/repositories/scala_3_6.bzl
mbland: Which dependencies weren't updated? If we need to update the script accordingly, we can.

mbland (commented):
I updated `scripts/create_repository.py` locally to use 2.13.16, and got almost the same result...except that the `io_bazel_rules_scala_scala_compiler{,_2}` entries in this pull request are missing the `deps` field:

```json
    "io_bazel_rules_scala_scala_compiler": {
        "artifact": "org.scala-lang:scala-compiler:2.13.16",
        "sha256": "f59982714591e321ba9c087af2c8666e2f5fb92b11a0cef72c2c5e9b342152d3",
        "deps": [ 
            "@io_github_java_diff_utils_java_diff_utils",
            "@org_jline_jline",
        ],
    },
```

The script is set up not to update existing entries, which is why these were never added before (under the "if it ain't broke" principle). But if the scripts adds them on update, and the tests pass, we should be able to trust the new information.

At the same time, by the same principle, the tests _do_ appear to be passing after this update without the new `deps`. So It's not a blocking issue, but I wanted to raise the point that if `scripts/create_repository.py` generates breaking changes, we should update the script; if its changes pass all tests, in general, we should let its output stand.

bartoszkosiorek:
> The script is set up not to update existing entries, which is why these were never added before (under the "if it ain't broke" principle). But if the script adds them on update, and the tests pass, we should be able to trust the new information.

Thanks for Review and quality checks.
How did you force script to update existing entries?

mbland:
> Thanks for Review and quality checks. How did you force script to update existing entries?

To force an update of an artifact, you could remove its existing entry from the `third_party/repositories/scala_*.bzl` file, and `scripts/create_repository.py` will add it back. (Unless the artifact is marked "testonly"; these artifacts _are_ manually updated, and the script leaves them alone.) You could also artificially set the entry to reference an older artifact version.

---

The script will leave alone any entry that already matches the version resolved by `cs` (executed by `ArtifactResolver._fetch_artifact_data()`), has `"testonly": True` set, or has a version _newer_ than the resolved version. In other words, if the script doesn't see a need to update the artifact version, it won't update it at all.

If it _does_ detect the need to update an artifact's entry, it will also set its `deps` field accordingly, even if it didn't have one before. Hence why `org.scala-lang:scala-compiler:2.13.15` didn't have a `deps` field before (that was the existing version when I update and started running the script), and why `org.scala-lang:scala-compiler:2.13.16` generates one.

---

Now that I think about it, we _could_ add a flag to force an update even for matching versions, to add `deps` information. Or better yet, instead of adding that complexity, we could just delete _all_ the generated entries and recreate them. I haven't felt the need to do this yet, since everything seems to be working as is, but we could.

bartoszkosiorek:
> > Thanks for Review and quality checks. How did you force script to update existing entries?
> 
> To force an update of an artifact, you could remove its existing entry from the `third_party/repositories/scala_*.bzl` file, and `scripts/create_repository.py` will add it back. (Unless the artifact is marked "testonly"; these artifacts _are_ manually updated, and the script leaves them alone.) You could also artificially set the entry to reference an older artifact version.

I have updated the dependencies, to be aligned with your results.
Thanks.

In another branch I have also removed all dependencies, and regenerate it from scratch. A lot of dependencies were simply removed. I will create separate Pull Request with such changes.

simuons (approved):
Thanks, @bartoszkosiorek

PR #1703, opened 2025-02-18
Update docs for version 7.0.0, remove old APIs

mbland:
### Description

Describes the new `scala_toolchains()` API, breaking changes, Bazel compatibility matrix, and breaking of the `io_bazel_rules_scala` repo name dependency. Part of #1482, #1647, #1652, and #1699, it also removes obsolete `WORKSPACE` APIs in favor of `scala_toolchains()`, et. al.

Much of the README text regarding `WORKSPACE` configuration, Bazel compatibility, and breaking changes comes from:

- https://github.com/bazelbuild/rules_scala/issues/1482#issuecomment-2605683509

The "Breaking changes in `rules_scala` 7.x" section of `README.md` describes the removed `WORKSPACE` macros, and provides guidance for adopting the `scala_toolchains()` API. Also includes information about the upcoming Bzlmod implementation (#1482) and the Bazel 8 compatibility changes for `rules_scala` 8.x (#1652).

Contains a light editing pass over almost every Markdown file to resolve `markdownlint` warnings when editing in VSCode. Also added `.markdownlint.json` to silence rule `MD033/no-inline-html`, since several docs contain `<br/>` and/or `<table>` elements.

- https://github.com/DavidAnson/vscode-markdownlint?tab=readme-ov-file#markdownlintconfig

Performed other opportunistic grammar edits and added new information, including:

- `docs/coverage.md`: Describes how to determine the default JaCoCo version used by `rules_java`

- `scripts/README.md`: Updates `create_repository.py` docs extensively, and adds a section for `sync_bazelversion.sh`

- Replaces `starlark` code fences with `py`, since the syntax is identical and editors seem to support it better.

### Motivation

Since the next major release is imminent, ensuring the documentation accurately reflects all the changes since v6.6.0 has become an urgent priority.

Rather than leave the old `WORKSPACE` APIs in place, which could lead to surprising failures, this change removes all of them. This also changes some code that still depended on some of these obsolete macros, including `scala_toolchains()`. Since all the toolchainization changes led to the entire project already using `scala_toolchains()` and `scala_register_toolchains()` exclusively, this proved a low risk refactoring.

With some Bzlmod and Bazel 8 information already in place, very little will need to change when these implementations land. The commits that contain those implementations will also include their relevant documentation updates.

PR #1709, opened 2025-02-28
Rename `{io_bazel_,}rules_scala_config`

mbland:
### Description

Shortens `@io_bazel_rules_scala_config` to `@rules_scala_config` in light of the `@io_bazel_rules_scala` removal in #1696. Part of #1482.

Adds a section to `README.md` indicating this as a breaking change, with advice on working around any breakages that aren't immediately fixable.

Includes updates the "Builtin repositories no longer visible by default under Bzlmod" section to improve clarity.

### Motivation

Per @simuons's advice in https://github.com/bazelbuild/rules_scala/pull/1703#discussion_r1971414370:

> @mbland I'm ok with renaming @io_bazel_rules_scala_config since there will be breaking changes as you mentioned and there is no reason for keeping label misaligned.


simuons (approved):
Thanks, especially for documenting these changes.

PR #1687, opened 2025-01-27
Toolchainize //scala_proto:{,deps_}toolchain_type

mbland:
### Description

Adds scala_proto toolchains to `scala_toolchains()`. Part of #1482.

The most significant part of the change is moving all the toolchain rules from `scala_proto/BUILD` to `setup_scala_toolchains()` in `scala_proto/toolchains.bzl`.

Adds the `scala_proto_deps_providers()` macro to replace `//scala_proto:scalapb_{compile,grpc,worker}_deps_provider` targets in the `dep_providers` parameter of `scala_proto_deps_toolchain()`. Examples of this are in `test/proto/custom_generator/BUILD`.

Excludes `@scala_proto_rules_scalapb_protoc_gen` from `DEFAULT_SCALAPB_WORKER_DEPS` in `scala_proto/default/default_deps.bzl` for Scala 2.11. For other Scala versions, this repo name will have the Scala version appended. 

A lot of the other changes are more opportunistic removals of `@io_bazel_rules_scala` label prefixes and application of `Label()` where appropriate.

### Motivation

This is the next step in the toolchainization of the frameworks used to implement various rules, as required to implement Bzlmod support.

Excluding `@scala_proto_rules_scalapb_protoc_gen` from `DEFAULT_SCALAPB_WORKER_DEPS` avoids build failures under Bzlmod, since:

- This repo is required by ScalaPB 0.11.17, but Scala 2.11 is capped at ScalaPB 0.9.8.

- Importing the nonexistent `scala_proto_rules_scalapb_protoc_gen` under Scala 2.11 results in an error under Bzlmod, as does importing it multiple times when configuring multiple Scala versions.

- `MODULE.bazel` can iterate over a list of Scala versions, filtering out Scala 2.11, and call `use_repo()` on each version specific repo.

As in other pull requests, removing `@io_bazel_rules_scala`  label prefixes will allow Bzlmod users to use `rules_scala` without setting `repo_name = "@io_bazel_rules_scala"` in `bazel_dep()`.

PR #1631, opened 2024-10-23
Update to Scalafmt 3.8.3

mbland:
### Description

Ensures that all Scalafmt targets succeed under every supported Scala version. Part of #1482.

Scala 2.11 remains on Scalafmt 2.7.5, since there's not a more recent compatible version.

Between 3.0.0 and 3.8.3, Scalafmt's `FileOps` moved packages, `Config.fromHoconFile` moved to `ScalafmtConfig`, and the methods now take a `java.nio.file.Path` parameter. As a result, this required extracting a thin `ScalafmtAdapter` object, with one implementation for Scala 2.11 and Scalafmt 2.7.5, and one for Scalafmt 3.8.3.

This change also adds the file path to the `Scalafmt.format()` call, allowing error messages to show the actual file path instead of `<input>`.

Removed the `verticalMultiline.newlineBeforeImplicitKW = true` and `unindentTopLevelOperators = false` options from `.scalafmt.conf`. They don't appear to be available in Scalafmt 3.8.3.

Also removes some `@io_bazel_rules_scala` references to make the internal implementation less dependendent on that name. This will allow Bzlmod clients to use `rules_scala` in a `bazel_dep()` without setting `repo_name = "io_bazel_rules_scala"`.

### Motivation

Scalafmt 3.0.0 works with the current default Scala version 2.12.19, but breaks under Scala 2.13.14:

```txt
$ bazel test --repo_env=SCALA_VERSION=2.13.14 //test/scalafmt/...

ERROR: .../test/scalafmt/BUILD:49:20:
  ScalaFmt test/scalafmt/test/scalafmt/unformatted/unformatted-test.scala.fmt.output
  failed: (Exit 1): scalafmt failed: error executing command
  (from target //test/scalafmt:unformatted-test)
  bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/scala/scalafmt/scalafmt
  '--jvm_flag=-Dfile.encoding=UTF-8' ... (remaining 1 argument skipped)

java.util.NoSuchElementException: last of empty IndexedSeq
    at scala.collection.IndexedSeqOps.last(IndexedSeq.scala:110)
    at scala.collection.IndexedSeqOps.last$(IndexedSeq.scala:105)
    at scala.meta.tokens.Tokens.last(Tokens.scala:29)
    at org.scalafmt.internal.Router.$anonfun$getSplitsImpl$90(Router.scala:707)
    at scala.Option.map(Option.scala:242)
[...]

PR #1710, opened 2025-02-28
Bazel 8 + `rules_java` 8 compatibility updates

### LINKED ISSUE #1652: Bazel 8.0.0 compatibility

mbland:
### Description

Bumps dependencies to versions that are compatible with both Bazel 7.5.0 and 8.0.0. Part of #1482. Closes #1652.

- ScalaPB jars: 0.11.17 => 1.0.0-alpha.1
- `rules_python`: 0.38.0 => 1.2.0
- `rules_cc`: 0.0.9 => 0.1.1
- `abseil-cpp`: 20220623.1 => 20250127.0
- `rules_java`: 7.12.4 => 8.9.0
- `protobuf`: 21.7 => 29.3
- `rules_proto`: 6.0.2 => 7.1.0
- `google-common-protos`: 2.52.0 => 2.53.0

This precipitated the following updates also included in this commit:

- Loads `java_proto_library` from `com_google_protobuf`.
- Bumps `.bazelversion` to 7.5.0.
- Sets `.bazelci/presubmit.yml` to use Bazel 7.5.0 instead of 6.5.0, and `last_rc` in place of `7.x`.
- Sets `common --enable_workspace --noenable_bzlmod` in `.bazelrc` and `tools/bazel\.rc\.buildkite`.
- Adds `allow_empty = True` to a `glob` expression in `//test/semanticdb:lib_with_tempsrc`.
- Removes Scala 2.11 test cases.

Bazel 6 is officially unsupported as of this change and the upcoming `rules_scala` 7.0.0 release. `WORKSPACE` builds succeed under Bazel 6.5.0 (with C++ compiler flags) for the time being, but are not guaranteed to continue working. (Bazel 6 Bzlmod builds would break anyway, because Bazel 6 doesn't provide `use_repo_rule`, required by `rules_jvm_external` 6.3, which is required by `protobuf` v29.)

The `README` now documents that `scala_proto` or any rules otherwise depending on `protobuf` are no longer supported out of the box for Scala 2.11. Such users will have to ensure they register their own downgraded versions, at their own risk of future `rules_scala` 7.x incompatibility.

Version bump related updates to `.bazelversion`,`WORKSPACE`, and `third_party/repositories` comprise the bulk of this change. Other small, yet important changes (other than those due to the version bumps noted above) include:

- Adding a new `examples/overridden_artifacts` repository and the `overridden_artifacts_example` test case in `test/shell/test_examples.sh`.

- Making `_validate_scalac_srcjar()` and `dt_patched_compiler_setup()` in `scala/private/macros/scala_repositories.bzl` more tolerant of dictionaries containing keys mapped to `None`. The new `overridden_artifacts_example` test covers this.

### Motivation

We're no longer planning to support Bazel 6 in the next major release per @simuons's decision in:

- https://github.com/bazelbuild/rules_scala/issues/1482#issuecomment-2690786093

The plan is now to land the Bazel 7 and 8 compatibility updates first, then land the Bzlmod change. This enables us to make only one new major version release, instead of two (the first supporting with Bazel 6, the second dropping support).

Bazel 8 and `rules_java` 8 require `protobuf` >= v29. After the `protobuf` v29 bump, and before the ScalaPB 1.0.0-alpha.1 bump, `scala_proto` targets would fail with the following error:

```txt
ERROR: .../external/com_google_protobuf/src/google/protobuf/BUILD.bazel:23:14:
  ProtoScalaPBRule
  external/com_google_protobuf/src/google/protobuf/any_proto_jvm_extra_protobuf_generator_scalapb.srcjar
  failed: (Exit 1): scalapb_worker failed:
  error executing ProtoScalaPBRule command
  (from target @@com_google_protobuf//src/google/protobuf:any_proto)
  bazel-out/.../bin/src/scala/scripts/scalapb_worker
    ... (remaining 2 arguments skipped)

--jvm_extra_protobuf_generator_out:
  java.lang.NoSuchMethodError:
  'java.lang.Object com.google.protobuf.DescriptorProtos$FieldOptions.getExtension(com.google.protobuf.GeneratedMessage$GeneratedExtension)'
    at scalapb.compiler.DescriptorImplicits$ExtendedFieldDescriptor.fieldOptions(DescriptorImplicits.scala:329)
  [ ...snip... ]

java.lang.RuntimeException: Exit with code 1
    at scala.sys.package$.error(package.scala:30)
    at scripts.ScalaPBWorker$.work(ScalaPBWorker.scala:44)
    at io.bazel.rulesscala.worker.Worker.persistentWorkerMain(Worker.java:96)
    at io.bazel.rulesscala.worker.Worker.workerMain(Worker.java:49)
    at scripts.ScalaPBWorker$.main(ScalaPBWorker.scala:39)
    at scripts.ScalaPBWorker.main(ScalaPBWorker.scala)

ERROR: .../external/com_google_protobuf/src/google/protobuf/BUILD.bazel:23:14
  Building source jar external/com_google_protobuf/src/google/protobuf/any_proto_scalapb-src.jar
  failed: (Exit 1): scalapb_worker failed:
  error executing ProtoScalaPBRule command
  (from target @@com_google_protobuf//src/google/protobuf:any_proto)
  bazel-out/darwin_arm64-opt-exec-ST-a828a81199fe/bin/src/scala/scripts/scalapb_worker
    ... (remaining 2 arguments skipped)
```

Here's why the other changes were necessary in light of the version bumps:

- `java_proto_library` from `rules_java` is now officially deprecated, hence loading it from `com_google_protobuf`.

- Setting `common --enable_workspace --noenable_bzlmod` (instead of `build`) fixes `test_semanticdb_handles_removed_sourcefiles`. This test relies on `bazel query`, which is also affected by these flags, hence `common` instead of `build`. Bazel 8 defaults to `--enable_bzlmod --noenable_workspace`, causing the `WORKSPACE` run of this test to fail.

- `glob` requires an explicit `allow_empty = True` parameter now that `--incompatible_disallow_empty_glob` defaults to `True` in Bazel 8.

- ScalaPB 0.9.8, the last version compatible with Scala 2.11, does not support `protobuf` >= 26.0. For this reason, we must remove the Scala 2.11 test cases.

We should consider dropping Scala 2.11 support at this point, since there's no ScalaPB release for it that supports later versions of `protobuf`. That, and we could remove some of the special case code added in the following changes, amongst other 2.11 support details:

- #1631
- #1648
- #1687
- #1688

Finally, the motivation for the non-version bump changes:

- The design bug in the upcoming Bzlmod API for `overridden_artifacts` reported by @dmivankov precipitated the `examples/overriden_artifacts` repo and test. See: https://github.com/bazelbuild/rules_scala/issues/1482#issuecomment-2676123661 https://github.com/bazelbuild/rules_scala/issues/1482#issuecomment-2676338407

- The `_validate_scalac_srcjar()` update maintains the strict checks for mutually exclusive values, while preventing client code from having to explicitly filter out `None` entries. This pairs with the change in `dt_patched_compiler_setup()` that uses the `compiler_srcjar` dictionary. These changes helps keep the upcoming module extension a bit cleaner.


mbland:
Hmm, [an odd failure on the Windows build](https://github.com/bazelbuild/rules_scala/pull/1710):

```txt
ERROR: C:/tools/msys64/home/b/_bazel_b/xknd5zlq/external/com_google_protobuf/src/google/protobuf/compiler/java/BUILD.bazel:87:11: Compiling src/google/protobuf/compiler/java/java_features.pb.cc [for tool] failed: (Exit 2): cl.exe failed: error executing CppCompile command (from target @@com_google_protobuf//src/google/protobuf/compiler/java:java_features_bootstrap) C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.39.33519\bin\HostX64\x64\cl.exe ... (remaining 1 argument skipped)
--
  | external/com_google_protobuf/src/google/protobuf/compiler/java/java_features.pb.cc(6): fatal error C1083: Cannot open include file: 'google/protobuf/compiler/java/java_features.pb.h': No such file or directory
```

[src/google/protobuf/compiler/java/java_features.pb.h most definitely exists in protobuf v29.3](https://github.com/protocolbuffers/protobuf/blob/v29.3/src/google/protobuf/compiler/java/java_features.pb.h).

All the other builds seem to pass. If they do, I'll kick this with an empty commit to see if we can get Windows to play along, in case it was some weird transient error.

mbland:
Yep, just windows. And my empty commit bump just failed in the same way. Hmm... why can't it find the `#include` file right next to the source?

crt-31:
I looked at the windows issue... the issue is that the full include path for that file is over the char limit (260 chars).
I think its a cc_library issue, but still investigating when it got broken... I think it has to do with the param file limitations, but I'm still looking into it. 

Also, I don't remember having to compile those protobuf tools locally before...  I feel like they came precompiled before?

crt-31:
OK, I found the related issue here: https://github.com/protocolbuffers/protobuf/issues/12947.  (I'm going to look into it more cause I use MSVC all the time and haven't run into these issues.)

I wonder if we can go back to using the prebuilt protobuf libraries, then we don't need to require the c++ compiler in order to use rules_scala.

mbland:
@crt-31 Heh, you beat me to it...I snuck a look on my phone while out this evening and found the same issue. Got online to post it, but was too late. 😛 What's more, my next [Bzlmod blog post](https://blog.engflow.com/category/bzlmod/) is going to talk about how I had to patch `rules_rust` to work around Windows path length breakages.

And I saw the paragraph that you just responded to:

> Since we don't have any good way to mitigate this issue, and because bazel+msvc usage is low compared to other ways to build on Windows, we are planning to drop support for bazel+msvc in Protobuf 34 (the Q1 2026 release). Bazel 30 already disables support for this combination, with a flag to reenable support for it which gives time to gather feedback on this as well as give folks a longer horizon to migrate their builds to either make+msvc or bazel+clang-cl which don't suffer from this artificial limitation without being stuck on Protobuf 29 or lower:
>
> https://protobuf.dev/news/v30/#poison-msvc--bazel

The linked announcement makes reference to protocolbuffers/protobuf#20085: "Breaking Change: Dropping support for Bazel+MSVC".

But to your point above:

> I wonder if we can go back to using the prebuilt protobuf libraries, then we don't need to require the c++ compiler in order to use rules_scala.

I _did_ experiment with using prebuilt Protobuf toolchains in [my proto-toolchainization branch in mbland/rules_scala](https://github.com/mbland/rules_scala/tree/proto-toolchainization) back in October, when I was first experimenting with Bzlmodification. I think I may dust it off and play with it again in light of this development; it may also provide a performance win, since `protobuf` wants to recompile constantly. (I'm also not aware of what "going back to using prebuilt protobuf libraries" would involve. We're already using the prebuilt `protobuf-java` artifact; I think `protobuf` keeps wanting to recompile `protoc`.)

mbland:
The good news is, I went ahead and implemented `protoc` toolchainization in the latest commit, and it works _really well_. Please check out what I did in https://github.com/bazelbuild/rules_scala/pull/1710/commits/7b7ab2a7a63d2f1a297982f3508fe9ab5421db5f and let me know what you think of the `README` update.

The bad news is, [the Windows build is still broken in the same way](https://buildkite.com/bazel/rules-scala-scala/builds/5398#01955417-cbac-4212-9337-5fd4f6a7ffda/75-182). I'm pretty sure it's because `protobuf` v29.3 still uses `native.java_proto_library`, which still depends on `@com_google_protobuf//:protoc`:

- https://github.com/protocolbuffers/protobuf/blob/v29.3/bazel/java_proto_library.bzl#L11-L16

```py
def java_proto_library(**kwattrs):
    # Only use Starlark rules when they are removed from Bazel
    if not hasattr(native, "java_proto_library"):
        _java_proto_library(**kwattrs)
    else:
        native.java_proto_library(**kwattrs)
```

I believe this is what @fmeum is trying to fix in protocolbuffers/protobuf#19679 (which I borrowed from in `scala_proto/scala_proto_toolchain.bzl`).

@simuons Until this gets resolved one way or another, what do you think of making Windows builds optional for now? Then when a fix comes, we can make it required again. Or maybe I can hack something to downgrade the Windows build down to `protobuf` v21.7 again, as it was before this pull request (i.e., currently remains on `master`).

@crt-31 I know you really don't want that to happen, but it's effectively blocking the next `rules_scala` release at this point.

@fmeum If you've any better ideas, or my understanding is incorrect in any way, I'm all ears.

At any rate, this gives me time to think about maybe adding more tests for this latest `protoc` toolchainization, if we need them.

mbland:
Hold the presses... I just pushed an experiment that fixed the Windows build. I created a patch from https://github.com/protocolbuffers/protobuf/pull/19679, then added a line to `WORKSPACE` to register the prebuilt toolchain early. Success!

Everything else is failing right now (all the tests for all the other `WORKSPACE` repos I haven't modified), but I'll fix that up later. The main question is whether to actually commit this patch.

mbland:
OK, please ignore [my update from yesterday](https://github.com/bazelbuild/rules_scala/pull/1710#issuecomment-2692476188). I've learned a lot, including that the `protoc` failure was due to the new toolchain not getting registered early enough, not that `protobuf` was selecting `native` rules. The latest commit polishes the `protoc` toolchainization implementation, and [the Windows build is passing again (along with all the others)](https://buildkite.com/bazel/rules-scala-scala/builds/5402#0195593f-4a07-497a-83cd-3615346bc298).

The majority of the changes in the PR are still the boilerplate updates to `WORKSPACE` and `.bazelversion` files, and automated updates to `third_party/repositories/scala_*.bzl` files. The most substantial changes are in:

- Updating Bazel versions in `.bazelci/presubmit.yml`
- Adding `common --incompatible_enable_proto_toolchain_resolution` to `.bazelrc` and `tools/bazel.rc.buildkite`
- `README.md`
- The new `examples/overridden_artifacts` repo and new test in `test/shell/test_examples.sh`
- The new `//protoc` package with the `protoc` toolchainization logic
- The `protobuf` patch in `scala/deps.bzl` from protocolbuffers/protobuf#19679
- The slight update to `scala/private/macros/scala_repositories.bzl`
- Updates to `scala/toolchains.bzl` and `scala/toolchains_repo.bzl` for toolchainization
- Minor tweaks to `scala_config.bzl`
- `scala_proto/scala_proto_toolchain.bzl` updates based on protocolbuffers/protobuf#19679
- Version bumps in `scripts/create_repository.py`
- The new `scripts/update_protoc_integrity.py`
- Adding `allow_empty = True` to a `glob()` in `test/semanticdb/BUILD`
- Dropping Scala 2.11 tests from `test_thirdparty_version.sh` and `test_version.sh`

Still, this is a _lot_ more than I'd originally planned to include in this pull request. Some of these updates can probably be teased out into separate PRs.

_And/or_, we _could_ go back to the original 7.x and 8.x release plan. This would mean backing out the `protobuf` upgrade from v21.7 to v29.3, and other updates along with it, most notably `protoc` toolchainization.

In either case, Bzlmod could land in 7.0 or 7.1. I'll prepare my Bzlmod branch, but it would remain a relatively light update on top of whatever changes we decide to include, and when.

Anywho, as usual, options... @simuons Let me know what you'd prefer to do here.

mbland:
P.S. I can confirm that this branch is still compatible with Bazel 6.5.0, both with the C++ compiler flags and with protocol compiler toolchainization enabled, as well as Bazel 8.1.1.

I have protoc toolchainized Bzlmod builds working on [my unstable 'bzlmod-enable' branch](https://github.com/mbland/rules_scala/tree/bzlmod-enable) and both of my stable [bzlmod](https://github.com/mbland/rules_scala/tree/bzlmod) and [bzlmod-bazel-8](https://github.com/mbland/rules_scala/tree/bzlmod-bazel-8) branches. Can confirm they work with Bazels 7.5.0, 8.1.1, `rolling`, and `last_green`.

```txt
# rolling
Starting local Bazel server (9.0.0-pre.20250210.1) and connecting to it...

# last_green
2025/03/02 22:24:18 Using unreleased version at commit 15a3526621559170c2a28e9dae5e9d7c72a82f02                                                                  
2025/03/02 22:24:18 Downloading https://storage.googleapis.com/bazel-builds/artifacts/macos_arm64/15a3526621559170c2a28e9dae5e9d7c72a82f02/bazel...
```

The thing that's at once a bug and a feature of the Bzlmod branches is the fact that I had to manually copy the `protobuf` patch and add a `single_version_override` to each nested module to enable protoc toolchainization. This is because you can only apply patches in the main module, and they must be regular files in the same module. In a pathological case like this, with having to apply a patch to all these nested modules, it's a drag. In a typical case, when you're working with a single module, it's nice that Bzlmod forces things to be so explicit and correct, as opposed to `WORKSPACE` macros doing [mysterious and important work](https://lumonindustries.net/).

And of course, in the best possible worlds, protocolbuffers/protobuf#19679 will get merged, and we can upgrade to at least v29.4 and forget about this patching and `single_version_override` business.

At any rate, for now, the patch is helping everything work, and I've documented the hell out of it in the `README`.

bartoszkosiorek (approved):

mbland:
I've just gone ahead and added these bumps as well, and propagated them to [bzlmod-enable](https://github.com/mbland/rules_scala/tree/bzlmod-enable), [bzlmod](https://github.com/mbland/rules_scala/tree/bzlmod), and [bzlmod-bazel-8](https://github.com/mbland/rules_scala/tree/bzlmod-bazel-8):

- Go: 1.24.0 => 1.24.1
- Scalafmt: 3.9.1 => 3.9.2
- `org.scala-sbt:compiler-interface`: 1.10.7 => 1.10.8
- `org.scala-sbt:util-interface`: 1.10.7 => 1.10.10
- `rules_java`: 8.9.0 => 8.10.0

mbland:
@simuons and @liucijus Before you review, approve, and merge this pull request, consider reviewing and merging these others first, which I just extracted from this one:

- #1711 
- #1713 
- #1714 

Each of those is orthogonal to one another, and is much easier to understand. Merging them will render this pull request much smaller and more focused itself.

Also, what do you think of bumping the default Scala version from the current 2.12.20 to 2.13.16?

Finally, #1712 (which is _not_ for review at this point) shows that even `protobuf` v25.5 breaks MSVC builds by default. I may play with that branch/pull request some more, but I think it's proving that upgrading from `protobuf` v21.7 very rapidly leads to the toolchainization solution in this PR.

mbland:
I've found out that protocol compiler toolchainization requires `protobuf` v29 and later. Details are in https://github.com/bazelbuild/rules_scala/pull/1710/commits/116dd192d1429fa4b30c4e471588d6255c26f567, both in its `README.md` updates and its commit message.

As a result, it seems Windows users will have to make the leap to `protobuf` v29 and protocol compiler toolchainization. I don't think it's worth the time trying to keep bumping #1712 to see if there's a version between 21.7 and 25.5 that will work without toolchainization. So I'm going to close that PR.

I also found out `protobuf` v30 is right out for Bazel 6.5.0, so I've updated that section of `README.md` as well.

Still thinking it's worth landing #1711, #1713, and #1714 before this one. #1715 should definitely go in soon as well. Then I can rebase this pull request to something much more reasonable. And then, [bzlmod-enable](https://github.com/mbland/rules_scala/tree/bzlmod-enable) is waiting in the wings.

simuons:
Thanks, @mbland. I'll go over other PRs first and then come back to this one.

> Also, what do you think of bumping the default Scala version from the current 2.12.20 to 2.13.16?

I'm not against it, I'm for it I would say, but would like to know why. Are there technical reasons?

mbland:
> I'm not against it, I'm for it I would say, but would like to know why. Are there technical reasons?

Not really. I'll leave it alone if you don't see a compelling reason to change it.

Rebasing and squashing and updating the PR description as we speak...

mbland:
@simuons I just created #1716 after discovering a problem with some `test_version.sh` test cases. It can land either before or after this one, but it's both tiny and important. I did confirm locally that this PR passes with the changes from #1716.

mbland:
Just rebased after #1716 and added a new commit to update `README.md` slightly.

crt-31:
Hi @mbland, Is this in a good state to test on windows (with Bazel 7.5). I was a little confused... does this pr have everything you had broken out in the other prs?

mbland:
> Hi @mbland, Is this in a good state to test on windows (with Bazel 7.5). I was a little confused... does this pr have everything you had broken out in the other prs?

@crt-31 Yes, it is. `master` includes all the changes originally broken out of this pull request, and I've subsequently rebased this pull request on `master`. The Windows build has been passing using Bazel 7.5.0 for a while now.

mbland:
@crt-31 P.S. It looks like `protobuf` v30.1 _just_ came out. About to push a new commit to bump that.

crt-31:
@mbland, cool I'm gonna try it out on our codebase.

simuons (approved):
Thanks, @mbland, especially for documenting all of this. It's a pity that protoc requires a lot of massaging and dancing around.
All looks good to me. Just want to clarify are we still sticking to the paln of releasing rules for bazel 7 and then for bazel 8 or we will jump stright to bazel 8 (since PR messge mentions it)

simuons:
Hi, @crt-31, should we wait your confirmation regarding builds on windows?

mbland:
> Thanks, @mbland, especially for documenting all of this. It's a pity that protoc requires a lot of massaging and dancing around.

My pleasure. Hopefully it won't be required forever, but also, at least it makes the build faster.

> All looks good to me. Just want to clarify are we still sticking to the paln of releasing rules for bazel 7 and then for bazel 8 or we will jump stright to bazel 8 (since PR messge mentions it)

We'll jump straight to Bazel 8, while fully supporting Bazel 7. The PR updates the `.bazelversion` files and CI builds for Bazel 7.5.0 (though [7.6.0 appears imminent](https://github.com/bazelbuild/bazel/releases/tag/7.6.0rc1)). I've been testing this branch locally with Bazel 8, and the pending [bzlmod-enable][] branch with Bazel 7, 8, `rolling`, and `last_green`.

[bzlmod-enable]: https://github.com/mbland/rules_scala/tree/bzlmod-enable

The only things that would block releasing `rules_scala` 7.0.0 after this are:

- Whether to wait for #1718 now. Note that it needs a section under "Breaking changes in rules_scala 7.x" in `README.md`. Since it would be a breaking change on top of a breaking change, it would be best to get it in now, if ever.
- Whether to wait to merge the next PR from [bzlmod-enable][]. If not, that PR could precipitate `rules_scala` 7.1.0.

Also, if you decide to include #1718, I'd have to tweak [bzlmod-enable][] slightly to accommodate the change from the `scala_proto_enable_all_options` boolean argument to `scala_toolchains()` to the `scala_proto_options` list argument. This wouldn't be difficult, though.

cc: @kczulko

crt-31:
@simuons, 
Yeah, I'm in the process of testing on windows on our codebase. (Taking a little longer cause we are on airgap, and the changing dependencies are quite a challenge to deal with, but I'm close).

crt-31:
@mbland,  I'm having trouble figuring out what the mechanism is now to setup the custom scala toolchains that I have.  Before, I setup toolchains with `setup_scala_toolchain()` and then register them with `register_toolchains()`.  I tried to still do that , but I need to also setup the scala_proto toolchain (precompiled), but when I call scala_toolchains() it's failing saying the scala_version I'm using is unsupported. (we are on 2.13.14).  I think its cause its trying to use a list of jars that it knows about for 2.13.16, but it doesn't know what to do when I'm using 2.13.14, maybe?  Didn't have this issue in rules_scala 6.  Any ideas?

mbland:
> @mbland, I'm having trouble figuring out what the mechanism is now to setup the custom scala toolchains that I have. Before, I setup toolchains with `setup_scala_toolchain()` and then register them with `register_toolchains()`. I tried to still do that , but I need to also setup the scala_proto toolchain (precompiled), but when I call scala_toolchains() it's failing saying the scala_version I'm using is unsupported. (we are on 2.13.14). I think its cause its trying to use a list of jars that it knows about for 2.13.16, but it doesn't know what to do when I'm using 2.13.14, maybe? Didn't have this issue in rules_scala 6. Any ideas?

Set `scala_toolchains(validate_scala_version = False)`:

https://github.com/bazelbuild/rules_scala/blob/11d3b32db8060f59bea536fe4ce5bf1b2d9a3106/scala/toolchains.bzl#L74-L75

The default, `scala_toolchains(validate_scala_version = True)`, enables the check that the version configured via `scala_config(scala_version = "...")` matches the version at the top of the corresponding `third_party/repositories/scala_*.bzl` file:

https://github.com/bazelbuild/rules_scala/blob/11d3b32db8060f59bea536fe4ce5bf1b2d9a3106/third_party/repositories/repositories.bzl#L102-L108

For example, the currently supported version from `third_party/repositories/scala_2_13.bzl` is 2.13.16:

https://github.com/bazelbuild/rules_scala/blob/11d3b32db8060f59bea536fe4ce5bf1b2d9a3106/third_party/repositories/scala_2_13.bzl#L6

mbland:
@crt-31 I've pushed a new commit to include this explanation across `README.md` and `docs/scala_toolchain.md`.

@simuons I know we decided to have `scala_toolchains()` always instantiate a default Scala toolchain (https://github.com/bazelbuild/rules_scala/pull/1633#discussion_r1834378901), but this does appear to be a case where we might not want that.

Or, perhaps the precompiled protocol compiler toolchain should have its own repo, with its own separate repository rule. Or both. Or we keep things as they already are in this pull request, with everything consolidated behind `scala_toolchains()` and no special `scala = True` parameter for the default toolchain.

As eager as I am to land this PR and get on with the Bzlmod PR immediately after (implying I'm partial to the "keep things as they are" option), I'm happy to implement (and document) whichever approach you ultimately prefer.

mbland:
I just pushed an update to enable `scala_toolchains(scala = False)`. This will skip instantiating the default Scala toolchain and its dependencies, as well as the Scala version check.

@crt-31 Please give this version a try, instead of `scala_toolchains(validate_scala_version = False)`. That argument should still work, but this new version won't even pull in the default compiler JARs.

@simuons Please take a look at this latest commit. Implementing `scala = False` was a lot less of a big deal than it seemed in https://github.com/bazelbuild/rules_scala/pull/1633#discussion_r1834378901. I also improved the `//protoc` toolchains implementation by moving the `toolchain_type` and `proto_lang_toolchain` targets into the `//protoc` package (originally from the generated repo).

The underlying `protoc` executables package _could_ live in a separate repo, but I think it's a better API to continue using `scala_toolchains()` to instantiate it. Otherwise we'd have to add extra `load()` and `scala_protoc_repositories()` (for example) statements to all the `WORKSPACE` files.

Then again, requiring a separate `load()` wouldn't require users to patch `rules_scala` if they _really_ need to keep using a `protobuf` version earlier than v28. I'll prepare a commit to extract the `protoc` binaries into a separate repo, just to have it handy, in case you decide that's best.

mbland:
To be clear, with the latest commit, the protocol compiler toolchain API no longer requires loading and calling `scala_toolchains()`, but now requires:

```py
# Must come after loading `platforms` and `com_google_protobuf`.
load("@rules_scala//protoc:toolchains.bzl", "scala_protoc_toolchains")

scala_protoc_toolchains()
```

The `register_toolchains("@rules_scala//protoc:all")` call much earlier in `WORKSPACE` remains the same.

mbland:
@crt-31 @simuons OK, last API update, I swear... 🙂 

I've now tried updating the [bzlmod-enable][] branch, and landed on making this a pure repo rule. So now the API is this (as documented in `README.md` and at the top of `//protoc:toolchains.bzl`):

[bzlmod-enable]: https://github.com/mbland/rules_scala/tree/bzlmod-enable

```py
# WORKSPACE

# Register this toolchain before any others.
register_toolchains("@rules_scala_protoc_toolchains//...:all")

load("@platforms//host:extension.bzl", "host_platform_repo")

# Instantiates the `@host_platform` repo to work around:
# - https://github.com/bazelbuild/bazel/issues/22558
host_platform_repo(name = "host_platform")

# ...load `com_google_protobuf`, `rules_proto`, etc...

load("@rules_scala//protoc:toolchains.bzl", "scala_protoc_toolchains")

# The name can be anything, but we recommend `rules_scala_protoc_toolchains`.
# Only include `platforms` if you need additional platforms other than the
# automatically detected host platform.
scala_protoc_toolchains(
    name = "rules_scala_protoc_toolchains",
    platforms = ["linux-x86_64"],
)
```

No need to use `scala_toolchains()` if you don't need it.

I've already pushed the corresponding module extension to [bzlmod-enable][], as `//scala/extensions:protoc.bzl`. This really should do it, I think.

crt-31:
@mbland , Tried latest version from today on my codebase and it worked.  (Note I'm not using scalapb, not calling scala_toolchains()).

Also, thanks so much for setting up mechanism for prebuilt protoc. I can't beleive they don't have prebuilt support in the rules_proto or protobuf repo anymore, but awesome that you did that.

* So is the idea that we would now just only use prebuilt protoc in the rules_scala repo? That would make it simpler. My understanding was that rules_scala 6.x used prebuilt protoc, so this might be just fine. If someone wants to build protoc themselves, they could do it themeselves outside of rules_scala and bind it in.
* If rules_scala uses prebuilt protoc, then wouldn't this reduce a lot of the ScalaPB/abseil/protobuf versioning complexity described in #1647 ?
* Do we still need absil-cpp dependency if we are using prebuilts? Also maybe remove rules_cc as well. (built fine for me when I commented them out).
* Maybe for future: rules_proto is deprecated, so I think we might be able to remove dependency on that repo.
* In the scenario that I make my own custom scala_toolchain() like I do, but then still wanna use scala_proto. If I call scala_toolchains() with everything false except scalapb=true, I get an error that it can't find the scala_reflect jar. How would I do this? Shouldn't it use the already configured toolchain to get the compiler jars? Or should I be able to make a custom scala_proto_toolchain?

mbland:
> @mbland , Tried latest version from today on my codebase and it worked. (Note I'm not using scalapb, not calling scala_toolchains()).

Awesome! 😄 

> Also, thanks so much for setting up mechanism for prebuilt protoc. I can't beleive they don't have prebuilt support in the rules_proto or protobuf repo anymore, but awesome that you did that.

Heh, thanks. But I view it less as awesomeness on my part than necessity, since the Windows build was dead in the water without it. I'm just lucky I dabbled in exploring precompiled `protoc` toolchains back in October, when I was just starting to figure out how to update `rules_scala` for Bzlmod. Dusting off my old experimental branch greatly expedited the process.

That said, there are hints that precompiled `protoc` support in `protobuf` itself is forthcoming:

- [Protobuf News: News Announcements for Version 29.x](https://protobuf.dev/news/v29/)
- https://github.com/protocolbuffers/protobuf/issues/19558#issuecomment-2528810726
- https://github.com/protocolbuffers/protobuf/issues/20182#issuecomment-2643462353

> * So is the idea that we would now just only use prebuilt protoc in the rules_scala repo? That would make it simpler. My understanding was that rules_scala 6.x used prebuilt protoc, so this might be just fine. If someone wants to build protoc themselves, they could do it themeselves outside of rules_scala and bind it in.

I wouldn't go that far yet, because _consumers_ might not be prepared for or comfortable with using the precompiled `protoc` toolchain. This here pull request does enable the precompiled `protoc` toolchain by default for `rules_scala` development and CI, and makes it available to consumers, but doesn't force consumers to adopt it.

That said, I would imagine many, if not most, consumers might take advantage of the option once they realize it exists.

Also, FWIW, I'm not aware that `rules_scala` 6.x used a prebuilt `protoc` toolchain at all. `protobuf` didn't really begin to support it until v27.0 (protocolbuffers/protobuf#16323), which isn't compatible with any version of `rules_scala`, per #1647. I couldn't even get it to work with `protobuf` < v29, as I mention in the new "Why this requires `protobuf` v29 or later" section in `README.md` (from this PR).

> * If rules_scala uses prebuilt protoc, then wouldn't this reduce a lot of the ScalaPB/abseil/protobuf versioning complexity described in [Protobuf compatibility findings and suggestions #1647](https://github.com/bazelbuild/rules_scala/issues/1647) ?

Sadly, no, not entirely, not as I understand it. The `abseil-cpp` "incompatibiliy" is whether or not the Bazel version uses a new enough C++ flavor by default while compiling `protoc`. Once you've got a `protoc` binary, prebuilt or compiled as a build dependency (possibly with explicit C++ compiler flags), the more significant incompatibility lies within the `scala_proto` implementation.

The `scala_proto` rules use ScalaPB libraries and the related `protoc-bridge` library to implement an aspect that compiles proto files into Scala. That aspect creates a Unix pipe (which I suppose MSYS2 emulates on Windows?), launches `protoc`, and uses the pipe to communicate between the `protoc` process and the aspect. Both ends of that pipe have to agree on the version of the [protobuf-java](https://github.com/protocolbuffers/protobuf/tree/main/java) runtime and how to interpret its features.

The good news is that the `protobuf-java` Maven releases are backwards compatible with earlier versions of `protoc`. A lot of the build hangs described in #1647 were fixed by bumping `protobuf-java` alone. The bad news is that not all of the problems were fixed by an easy `protobuf-java` bump. Newer versions of ScalaPB and `protoc-bridge` only support `protobuf` v28 and later. I don't know if future versions of `ScalaPB` and `protoc-bridge` will be as backward compatible as `protobuf-java` is, or if there'll be another compatibility gap like there is between ScalaPB 0.11.17 and 1.0.0-alpha1 (i.e., no version supports `protobuf` v25.6 through v27).

I would _hope_ these `protobuf` related dependencies would become less version sensitive over time, but I wouldn't assume that we're there yet.

> * Do we still need absil-cpp dependency if we are using prebuilts? Also maybe remove rules_cc as well. (built fine for me when I commented them out).

Technically, no, we don't still need them when using the prebuilt `protoc` toolchain. But we still need to keep them so that consumers that _don't_ opt for the prebuilt `protoc` can still build with `rules_scala` successfully (i.e., can still recompile `protoc` as part of the build).

> * Maybe for future: rules_proto is deprecated, so I think we might be able to remove dependency on that repo.

It is, but we can't remove it yet for two reasons: one hard requirement, and one nice to have. And in writing this, I realize I need to add this information to `README.md`, which I'll do in a new commit shortly.

The hard requirement is that the precompiled `protoc` toolchain implementation depends on the `toolchains` helper struct from `@rules_proto//proto:proto_common.bzl`. The equivalent `toolchains` helpers from `rules_proto` are currently _private_ in `protobuf`, and first appeared in v29 (protocolbuffers/protobuf#17397).

The nice to have is that it enables both older and newer `protobuf` versions to still work with `rules_scala` by smoothing API changes between them. Even though we _officially_ don't support `protobuf` versions earlier than v28 with this change, by keeping `rules_proto` for now, it's still _possible_ to use v25.5 and earlier with `rules_proto` 6.0.2.

Specifically, this pull request ended up using doing the following in `protoc/private/protoc_toolchain.bzl`:

First, it loads the following directly from `protobuf`, since the `protobuf_version.bzl` file has contained `PROTOC_VERSION` since v21.0 (added in protocolbuffers/protobuf#9900).

```py
load("@com_google_protobuf//:protobuf_version.bzl", "PROTOC_VERSION")
```

Then it loads the `toolchains` helper struct from `rules_proto`, because it's the same API from `rules_proto` 6.0.2 to 7.1.0. But check out the implementations:

- [6.0.2 wraps the Bazel native 'proto_common' constants itself](https://github.com/bazelbuild/rules_proto/blob/6.0.2/proto/proto_common.bzl)
- [7.1.0 wraps @com_google_protobuf//bazel/common:proto_common.bzl](https://github.com/bazelbuild/rules_proto/blob/7.1.0/proto/proto_common.bzl), which doesn't exist until `protobuf` v27.0 (protocolbuffers/protobuf#16323).

```py
load("@rules_proto//proto:proto_common.bzl", "toolchains")
```

Finally, it pushes the following `load()` statements back into the generated `rules_scala_protoc_toolchains` repo, since these files also don't exist until `protobuf` v27 (same protocolbuffers/protobuf#16323). Since we require v29 to even generate this package, it avoids breaking builds using earlier versions.

```py
load(
    "@com_google_protobuf//bazel/toolchains:proto_lang_toolchain.bzl",
    "proto_lang_toolchain",
)
load(
    "@com_google_protobuf//bazel/toolchains:proto_toolchain.bzl",
    "proto_toolchain",
)
```

> * In the scenario that I make my own custom scala_toolchain() like I do, but then still wanna use scala_proto. If I call scala_toolchains() with everything false except scalapb=true, I get an error that it can't find the scala_reflect jar. How would I do this? Shouldn't it use the already configured toolchain to get the compiler jars? Or should I be able to make a custom scala_proto_toolchain?

`scala_proto_toolchain()` should contain an argument like (from `docs/scala_toolchain.md`):

```py
scala_library_classpath = [
    "@maven//:org_scala_lang_scala_library",
    "@maven//:org_scala_lang_scala_reflect",
],
```

If it doesn't contain a `scala_library_classpath` argument at all, it will default to creating one from the builtin toolchain repos. In that case, you'll need to set `scala_toolchains(validate_scala_version = False, scala_proto = True)`, with `scala = True` left as the default. (I just updated `docs/scala_toolchain.md` earlier today to mention this.)

mbland:
@crt-31 I assume this change is good to go now from your point of view?

@simuons I just extracted #1720 to trim this PR down again. Up to you if you want to merge #1720 first. If you merge this pull request first, I'll repurpose #1720 to only include the Guava updates (which aren't in this PR).

mbland:
@simuons Just rebased on `master` after #1719 and #1720 and everything's passing.

mbland:
Rebased on `master` after #1721.

mbland:
Rebased on `master` after #1718.

simuons (approved):
Thanks @mbland and sorry it took longer than necessary.

mbland:
@simuons Thanks, and no worries. We definitely straightened out a few things in the process.

I've rebased and updated my local `bzlmod-enable` branch and am running `./test_all.sh` on it now. Once it passes, I'll post what could possibly be the final pull request for #1482. 🙂 

That said, it will include changes to update the release workflow to publish to the Bazel Central Registry, replete with provenance attestations. (There'll be comments and a `.bcr/README.md` with references to explain this.) If you wanted me to have that in a separate pull request, I could break it out, since it's technically orthogonal to the rest of the change.

PR #1711, opened 2025-03-06
Bazel 7 and MSVC compatible version updates

mbland:
### Description

Dependency version updates that still work with Bazel 6.5.0 and 7.5.0. Broken out from #1710, and part of #1482 and #1652.

Updates `.bazelversion` files to 7.5.0 and the CI builds in `.bazelci/presubmit.yml` to use Bazel 7.5.0. Sets `common --enable_workspace --noenable_bzlmod` in `.bazelrc` and `tools/bazel.rc.buildkite`.

Bumps the following dependencies, which should not cause build breakages on Windows + MSVC:

- Go: 1.24.0 => 1.24.1
- Scalafmt: 3.9.1 => 3.9.2
- `abseil-cpp`: 20220623.1 => 20250127.0
- `grpc`: 1.70.0 => 1.71.1
- `protobuf-java`: 4.29.3 => 4.30.0
- `sbt-compiler-interface`: 1.10.7 => 1.10.8
- `sbt-compiler-util`: 1.10.7 => 1.10.10
- `google-common-protos`: 2.52.0 => 2.53.0

Defers the following updates, which are already present in #1710:

- `protobuf`: v21.7 => v28.3 (or v30.0)
- `rules_cc`: 0.0.9 => 0.1.1
- `rules_proto`: 6.0.2 => 7.1.0
- `rules_python`: 0.38.0 => 1.2.0
- `scalapb`: 0.11.17 => 1.0.0-alpha1

These deferred updates all need to happen together, as updating only a subset of them will break the build.

### Motivation

This change is smaller and more focused than #1710, and should ultimately make that pull request smaller and/or easier to review.

`--[no]enable_workspace` and `--[no]enable_bzlmod` affect `bazel query`, and Bazel 8 defaults to `--noenable_workspace --enable_bzlmod`. Using `common` ensures `bazel query` sees the same settings as `bazel build`. This prevents the `WORKSPACE` run of `test_semanticdb_handles_removed_sourcefiles` from failing under Bazel 8.

Bazel 6.5.0 doesn't define `--[no]enable_workspace`, so it makes sense to include it in the change that sets all `.bazelversion` files to 7.5.0.


simuons (approved):
Thanks, @mbland!  So this is a farewell to bazel 6

PR #1712, opened 2025-03-06
Upgrade to `protobuf` v25.6

mbland:
### Description

Updates the `protobuf` version to one that's still compatible with Bazel 6.5.0 and 7.5.0 without updating `scalapb`. Broken out from #1710, and part of #1482 and #1652.

Like #1711, updates `.bazelversion` files to 7.5.0 and the CI builds in `.bazelci/presubmit.yml` to use Bazel 7.5.0.

Unlike #1711, contains only these updates:

- `abseil-cpp`: 20220623.1 => 20250127.0
- `protobuf`: v21.7 => v25.6

### Motivation

This change aims make #1710 smaller and more focused, and should ultimately make that pull request smaller and/or easier to review.

Specifically, this is an attempt to see whether MSVC will build successfully with `protobuf` v25.6. If it doesn't, I will update this change to include the protocol compiler toolchainization changes from #1710.

PR #1714, opened 2025-03-07
7.0.0 README updates and minor code changes

mbland:
### Description

Contains many editorial improvements to and some extra information in the README, along with a few small improvements to the code. Broken out from #1710, and part of #1482 and #1652.

Specifically:

- Adds more info on translating `@rules_scala` to `@io_bazel_rules_scala` for dependencies via repo mapping.

- Merges information about the previously planned `rules_scala` 8.0.0 release into the information for 7.0.0, since it seems we may make only one major release.

- Improves information about `protobuf` support for versions before v28, Scala 2.11, and the upcoming Bzlmod `compatibility_level` setting.

- In `scala_config.bzl`, changes the private `_default_scala_version()` to the public `DEFAULT_SCALA_VERSION`.

- Adds `allow_empty = True` to a `glob` expression in `//test/semanticdb:lib_with_tempsrc`.

- Removes Scala 2.11 test cases from `test_thirdparty_version.sh` and `test_version.sh`.

### Motivation

This change is smaller and more focused than #1710, and should ultimately make that pull request smaller and/or easier to review.

The motivations for the individual changes are:

- The public `DEFAULT_SCALA_VERSION` constant makes this value accessible to the upcoming module extension.

- `glob` requires an explicit `allow_empty = True` parameter in Bazel 8, in which `--incompatible_disallow_empty_glob` defaults to `True`.

- ScalaPB 0.9.8, the last version compatible with Scala 2.11, does not support `protobuf` v25.6 or later. For this reason, we must remove the Scala 2.11 test cases, as documented in the `README.md` updates. See also #1712.

We should consider dropping Scala 2.11 support at this point, since there's no ScalaPB release for it that supports later versions of `protobuf`. That, and we could remove some of the special case code added in the following changes, amongst other 2.11 support details:

- #1631
- #1648
- #1687
- #1688


bartoszkosiorek (approved):
Looks great to me.

mbland:
BTW, we _could_ bump `DEFAULT_SCALA_VERSION` to 2.13.16. It's been working fine in my [bzlmod](https://github.com/mbland/rules_scala/tree/bzlmod) and [bzlmod-bazel-8](https://github.com/mbland/rules_scala/tree/bzlmod-bazel-8) branches for a while.

simuons (approved):
Thanks, @mbland!

PR #1718, opened 2025-03-17
feat: bit more configurable proto toolchain

kczulko:
### Description

Hello,

Over a year ago I had the experience of porting a project from sbt to rules_scala. One of the quirks of such a migration was protobuf. That time I've used `rules_proto_grpc` to overcome this part but I thought it should be possible with built-in proto toolchain from `rules_scala`. Now I finally had some time to take this journey and check it. Here are the results:
1. `proto_enable_all_options` removed - I think that user should define the options. This `all` seems to be a bit out of date, e.g. `scala3_sources` was not included there.
2. `grpc` dep id removed - Similar to above. The code that was merging compile deps was checking `with_grpc` flag that also got removed. I took an approach that here, as well, user knows what he wants so if `grpc` flag was added to the toolchain configuration, that means you need to provide proper dependencies as well. IMHO this grpc dep id was a bit artificial but I'm keen to hear some other opinion.
3. `main_generator` removed - for this approach the main_generator becomes the default one and got moved to `named_generators` with its default name set to `scala`.

I assume there might be a places which are not ready yet (like `docs`). Let's see what CI will tell us.

### Motivation

- User can define proto flags per plugin (either scalapb, fs2grpc, pekko-grpc, etc)
- Introducing support for more than a 3 proto flags (e.g. `scala3_sources`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -819,6 +819,35 @@ under WORKSPACE](#6.5.0), with the maximum dependency versions specified in
 that section. While this may continue to work for some time, it is not
 officially supported.
 
+### `scala_proto_toolchain` changes
+
+Since #1718 `scala_proto_toolchain`'s `main_generator` was removed in flavor of more
+flexible protobuf plugins configuration. Now each generator (plugin) will get a corresponding name
+that can be used for further plugin options setup:
+
+```
+scala_proto_toolchain(
+    name = "example",
+    generators_opts = {
+        "scala": [
+            "grpc",
+            "single_line_to_proto_string",
+        ],
+        "jvm_extra_protobuf_generator": [
+            "grpc",
+            "single_line_to_proto_string",
+        ],
+    },
+    generators = {
+        "scala": "scripts.ScalaPbCodeGenerator",
+        "jvm_extra_protobuf_generator": "scalarules.test.extra_protobuf_generator.ExtraProtobufGenerator",
+    },
+)
+```
+
+Also, `scalapb_grpc_deps` was removed since this changes moves the responsibility
+on the user side to configure dependencies based on the provided generators and their options.
+
 ### Removal of `bind()` aliases for `twitter_scrooge` dependencies
 
 `rules_scala` 7.x removes all of the obsolete [`bind()`][] aliases under
--- a/README.md
+++ b/README.md
@@ -828,6 +828,10 @@ that can be used for further plugin options setup:
 ```
 scala_proto_toolchain(
     name = "example",
+    generators = {
+        "scala": "scripts.ScalaPbCodeGenerator",
+        "jvm_extra_protobuf_generator": "scalarules.test.extra_protobuf_generator.ExtraProtobufGenerator",
+    },
     generators_opts = {
         "scala": [
             "grpc",
@@ -837,10 +841,6 @@ scala_proto_toolchain(
             "grpc",
             "single_line_to_proto_string",
         ],
-    },
-    generators = {
-        "scala": "scripts.ScalaPbCodeGenerator",
-        "jvm_extra_protobuf_generator": "scalarules.test.extra_protobuf_generator.ExtraProtobufGenerator",
     },
 )
 ```
--- a/README.md
+++ b/README.md
@@ -494,7 +494,7 @@ with the following `scala_toolchains()` parameters:
 ```py
 scala_toolchains(
     scala_proto = True,
-    scala_proto_enable_all_options = True,
+    scala_proto_options = [],
 )
 ```
 
--- a/README.md
+++ b/README.md
@@ -860,7 +860,7 @@ to configure dependencies based on the provided generators and their options.
 The new `scalapb_toolchain` convenience macro wraps `scala_proto_toolchain`
 to provide the default [ScalaPB](https://scalapb.github.io/) implementation:
 
-.``py
+```py
 load("//scala_proto:scala_proto_toolchain.bzl", "scalapb_toolchain")
 
 scalapb_toolchain(
@@ -871,8 +871,7 @@ scalapb_toolchain(
     ],
     visibility = ["//visibility:public"],
 )
-.``
-
+```
 
 ### Removal of `bind()` aliases for `twitter_scrooge` dependencies
 
--- a/README.md
+++ b/README.md
@@ -819,13 +819,22 @@ under WORKSPACE](#6.5.0), with the maximum dependency versions specified in
 that section. While this may continue to work for some time, it is not
 officially supported.
 
-### `scala_proto_toolchain` changes
+### `scala_proto_toolchain` changes and new `scalapb_toolchain` macro
 
-Since #1718 `scala_proto_toolchain`'s `main_generator` was removed in flavor of more
-flexible protobuf plugins configuration. Now each generator (plugin) will get a corresponding name
+`scala_proto_toolchain` has a more flexible plugin configuration schema. The
+new `generators` and `generators_opts` attributes replace the following
+attributes:
+
+- `with_grpc`
+- `with_flat_package`
+- `with_single_line_to_string`
+- `main_generator`
+- `named_generators`
+
+Now each generator (plugin) will get a corresponding name
 that can be used for further plugin options setup:
 
-```
+```py
 scala_proto_toolchain(
     name = "example",
     generators = {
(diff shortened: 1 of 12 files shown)


scala/toolchains.bzl
```diff
@@ -42,7 +42,7 @@ def scala_toolchains(
         scalafmt = False,
         scalafmt_default_config_path = ".scalafmt.conf",
         scala_proto = False,
-        scala_proto_enable_all_options = False,
+        scala_proto_options = [],
```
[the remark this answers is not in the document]
mbland: Ah, you did this while I was writing. I see it's already done.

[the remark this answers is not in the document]
kczulko: @mbland Is there any automated tooling for docs generation like bzl_library/gazelle?

simuons:
Hi, @kczulko, thanks for PR. Your approach looks good to me. I remember there was a discussion in the past "list of opts vs individual attribute in toolchain for each opt" but I don't remember remember where it was. I'll try to look around and get back to this.

test/proto/custom_generator/BUILD.bazel
```diff
@@ -100,7 +102,9 @@ toolchain(
 
 scala_proto_toolchain(
     name = "failing_scala_proto_toolchain_def",
-    main_generator = "test.proto.custom_generator.FailingGenerator",
+    named_generators = {
```
simuons, line 105: If only `name_generators` are left then probably we could get rid of `name_` prefix

simuons (changes requested):
As I said overall I like the change. But I think we should handle "default" generator differently. Waiting for your comment.

scala_proto/toolchains.bzl
```diff
@@ -32,22 +35,13 @@ def setup_scala_proto_toolchains(name, enable_all_options = False):
...
-        toolchain_impl_name = "%s_enable_all_options_toolchain_impl" % name
-        toolchain_options["with_flat_package"] = True
-        toolchain_options["with_single_line_to_string"] = True
 
     scala_proto_toolchain(
         name = toolchain_impl_name,
         visibility = ["//visibility:public"],
-        **toolchain_options
+        generators_opts = {
+            "scala": default_gen_opts,
```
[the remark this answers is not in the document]
kczulko: Starting from the beginning, you'd need this knowledge now. I mean default is named "scala" and it's done here https://github.com/bazelbuild/rules_scala/blob/11d3b32db8060f59bea536fe4ce5bf1b2d9a3106/scala_proto/scala_proto_toolchain.bzl#L10 . I just decided to keep this convention.

IMHO differentiating `main_generator` and other `generators` is a bit artificial. However one cannot live without the other (at least for fs2Grpc case) but for me having them in one place / one attr is better than merging on the private side of the code. I don't see the other cases/options now (e.g. using non-scalapb generator) but it gives more flexible configuration possibilities for the user (IMHO).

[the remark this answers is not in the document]
simuons: Can this be solved like this?
- remove defaults from `scala_proto_toolchain` 
- add new macro `scalapb_toolchain` which knows about `scala` prefix in generator and opts

Something like this:
```
scala_proto_toolchain = rule(
    implementation = _scala_proto_toolchain_impl,
    attrs = {
        "generators": attr.string_dict(allow_empty = False, mandatory = True),
        "generators_opts": attr.string_list_dict(),
        }
)

def scalapb_toolchain(name, opts = [], **kwargs):
    scala_proto_toolchain(
        name = name,
        generators = {
            "scala": "scripts.ScalaPbCodeGenerator",
        },
        generator_opts = {
            "scala": opts,
        },
        **kwargs
    )
```

This way you would have flexibility to configure multiple generators and anyone interested only in `scalapb` wouldn't have to know that it must use "scala" key in opts dict.

> IMHO differentiating main_generator and other generators is a bit artificial. 
Agree on that. However IMHO attribute on toolchain/macro is more discoverable than a key in a dict.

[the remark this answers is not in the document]
kczulko: Thank you Simonas! I'll apply your suggestions. Hopefully today :)

Best regards

test/proto/custom_generator/BUILD.bazel
kczulko: Done

README.md
```diff
@@ -819,6 +819,35 @@ under WORKSPACE](#6.5.0), with the maximum dependency versions specified in
 that section. While this may continue to work for some time, it is not
 officially supported.
 
+### `scala_proto_toolchain` changes
+
+Since #1718 `scala_proto_toolchain`'s `main_generator` was removed in flavor of more
+flexible protobuf plugins configuration. Now each generator (plugin) will get a corresponding name
+that can be used for further plugin options setup:
```
mbland, line 826:
```md
`scala_proto_toolchain` has a more flexible plugin configuration schema. The 
new `generators` and `generators_opts` attributes replace the following
attributes:

- `with_grpc`
- `with_flat_package`
- `with_single_line_to_string`
- `main_generator`
- `named_generators`
```

README.md
```diff
@@ -819,6 +819,35 @@ under WORKSPACE](#6.5.0), with the maximum dependency versions specified in
...
+
+```
```
mbland, line 828: <code>```py</code>

README.md
```diff
@@ -819,6 +819,35 @@ under WORKSPACE](#6.5.0), with the maximum dependency versions specified in
...
+            "grpc",
+            "single_line_to_proto_string",
+        ],
+    },
+)
+```
+
+Also, `scalapb_grpc_deps` was removed since this changes moves the responsibility
+on the user side to configure dependencies based on the provided generators and their options.
+
```
mbland, line 850:
```md
`scalapb_grpc_deps` no longer exists since it's now the user's responsibility
to configure dependencies based on the provided generators and their options.
```

scala_proto/scala_proto_toolchain.bzl
```diff
@@ -132,6 +103,24 @@ scala_proto_toolchain = rule(
     },
 )
 
+def scalapb_toolchain(name, opts = [], **kwargs):
+    """Setups default scala protobuf (scalapb) toolchain
```
mbland, line 107:
```py
"""Sets up a scala_proto_toolchain using ScalaPB.
```

scala_proto/scala_proto_toolchain.bzl
```diff
@@ -132,6 +103,24 @@ scala_proto_toolchain = rule(
...
+
+    Args:
+    name: A unique name for this target
+    opts: scalapb generator options like 'grpc' or 'flat_package'
```
mbland, line 111: Indent `name:` and `opts:` four more spaces. This will prevent any future breakages if we ever turn on Stardoc.

It also needs:

```py
        kwargs: remaining arguments to `scala_proto_toolchain`
```

README.md
```diff
@@ -819,6 +819,35 @@ under WORKSPACE](#6.5.0), with the maximum dependency versions specified in
 that section. While this may continue to work for some time, it is not
 officially supported.
 
+### `scala_proto_toolchain` changes
```
mbland, line 822:
```md
### `scala_proto_toolchain` changes and new `scalapb_toolchain` macro
```

README.md
mbland, line 850: Note I had to set the <code>```py</code> code fence to <code>.``py</code>  to nest it in the markdown example here.

```md
The new `scalapb_toolchain` convenience macro wraps `scala_proto_toolchain` 
to provide the default [ScalaPB](https://scalapb.github.io/) implementation:

.``py
load("//scala_proto:scala_proto_toolchain.bzl", "scalapb_toolchain")

scalapb_toolchain(
    name = "my_toolchain",
    opts = [
        "grpc",
        "single_line_to_proto_string",
    ],  
)                               
.``
```

mbland (commented):
Thanks for adding the docs. I've some editorial suggestions that hopefully make things a little clearer.

README.md
```diff
@@ -819,6 +819,61 @@ under WORKSPACE](#6.5.0), with the maximum dependency versions specified in
...
+)
+```
+
+`scalapb_grpc_deps` no longer exists since it's now the user's responsibility
+to configure dependencies based on the provided generators and their options.
+
+The new `scalapb_toolchain` convenience macro wraps `scala_proto_toolchain`
+to provide the default [ScalaPB](https://scalapb.github.io/) implementation:
+
+.``py
```
mbland, line 863: The <code>.``</code> patterns here and below should become <code>```</code>. (Sorry if you get this twice; I commented directly on the commit moments ago, instead of within the pull request.)

simuons:
Hi, @kczulko, looks good to me now. Please address comments by @mbland and I'll merge it.

README.md
kczulko: Done

PR #1719, opened 2025-03-18
Add Java 24 compatibility

srdo-humio:
The SecurityManager is a hollow shell that just throws exceptions in Java 24. We should avoid calling it. This will make tests that call System.exit crash the worker.

Bazel is doing the same thing for now, see
https://github.com/bazelbuild/bazel/issues/24354

It is often possible to edit code to avoid calling System.exit, and guarding against it requires more effort in Java 24+, likely involving attaching an agent. Until someone actually has that need, it doesn't seem worth doing.

### DIFF
--- a/scala/private/phases/phase_write_executable.bzl
+++ b/scala/private/phases/phase_write_executable.bzl
@@ -1,3 +1,5 @@
+load("//scala/private:common.bzl", "rlocationpath_from_file")
+
 #
 # PHASE: write executable
 #
@@ -5,16 +7,13 @@
 #
 load(
     "//scala/private:rule_impls.bzl",
-    "allow_security_manager",
     "expand_location",
     "first_non_empty",
     "is_windows",
     "java_bin",
     "java_bin_windows",
     "runfiles_root",
-    "specified_java_runtime",
 )
-load("//scala/private:common.bzl", "rlocationpath_from_file")
 
 def phase_write_executable_scalatest(ctx, p):
     # jvm_flags passed in on the target override scala_test_jvm_flags passed in on the
@@ -29,7 +28,7 @@ def phase_write_executable_scalatest(ctx, p):
         jvm_flags = [
             "-DRULES_SCALA_MAIN_WS_NAME=%s" % ctx.workspace_name,
             "-DRULES_SCALA_ARGS_FILE=%s" % rlocationpath_from_file(ctx, p.runfiles.args_file),
-        ] + expand_location(ctx, final_jvm_flags) + _allow_security_manager_for_specified_java_runtime(ctx),
+        ] + expand_location(ctx, final_jvm_flags),
         use_jacoco = ctx.configuration.coverage_enabled,
     )
     return _phase_write_executable_default(ctx, p, args)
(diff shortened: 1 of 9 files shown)


srdo-humio:
In a future Java release, the SM classes will go away entirely. At that point, we'll have to either access the SM via reflection, or split the worker into an old-JDK version and a new-JDK version if we want to maintain support for blocking System.exit. 

I think it might be fine to just drop support for the SM entirely across all versions instead of trying to maintain partial support.

Do you have opinions on this?

simuons (approved):
Thanks, @srdo-humio, if I remember correctly JEP promised a replacement for SecurityManager. If it's not the case anymore then I think we will get rid of SM entirely, because all other options you listed doesn't seem to be worth the effort/maintainance.

srdo-humio:
@simuons The tracking issue for the JDK at https://bugs.openjdk.org/browse/JDK-8199704 seems to have no movement, and the JEP that degrades the SM is not suggesting a replacement API, but instead contains an appendix describing a bytecode rewriting agent that people can use as a starting point if they really need this functionality https://openjdk.org/jeps/486. 

I don't think there's a built-in replacement for this in Java 24, and I don't get the impression a replacement is planned. 

Assuming we agree that we should drop the System.exit-blocking code entirely, would you prefer that I updated this PR to remove the SM code, or should we delay doing that until the JDK that removes the SM classes releases?

simuons:
@srdo-humio please update PR and remove SM.

Since replacement is not planned and everyone will eventually move to jdk 24 we will be left with obsolate code.

scala/private/phases/phase_write_executable.bzl
```diff
@@ -1,20 +1,19 @@
+load("//scala/private:common.bzl", "rlocationpath_from_file")
```
srdo-humio, line 1: Buildifier moves this line up here. Should I leave these changes in?

srdo-humio:
@simuons Sounds good. I've updated the PR to remove the SM.

mbland:
@srdo-humio @simuons I noticed that `src/java/io/bazel/rulesscala/worker/WorkerTest.java` is now broken:

```txt
$ bazel build //src/...

INFO: Analyzed 33 targets (0 packages loaded, 0 targets configured).
ERROR: /Users/mbland/src/bazelbuild/rules_scala/src/java/io/bazel/rulesscala/worker/BUILD:12:10: Building src/java/io/bazel/rulesscala/worker/worker_test.jar (1 source file) failed: (Exit 1): java failed: error executing Javac command (from target //src/java/io/bazel/rulesscala/worker:worker_test) external/remotejdk21_macos_aarch64/bin/java '--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED' '--add-exports=jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED' ... (remaining 19 arguments skipped)
src/java/io/bazel/rulesscala/worker/WorkerTest.java:40: error: cannot find symbol
        assertThrows(Worker.ExitTrapped.class, () -> Worker.workerMain(new String[] {}, worker))
                           ^
  symbol:   class ExitTrapped
  location: class Worker
src/java/io/bazel/rulesscala/worker/WorkerTest.java:40: error: unreported exception Exception; must be caught or declared to be thrown
        assertThrows(Worker.ExitTrapped.class, () -> Worker.workerMain(new String[] {}, worker))
                                                                      ^
INFO: Elapsed time: 0.211s, Critical Path: 0.07s
INFO: 2 processes: 2 internal.
ERROR: Build did NOT complete successfully
```

This test isn't exercised by our CI, or by `./test_all.sh`, but I often build this package as a smoke test before `./test_all.sh`. Should this test be updated or removed?

PR #1721, opened 2025-03-26
Cleanup WorkerTest after SecurityManager was removed

simuons:
### Description
WorkerTest was failing after https://github.com/bazelbuild/rules_scala/pull/1719


mbland (approved):
LGTM

This works for me locally.

By the way, as I was referring to in https://github.com/bazelbuild/rules_scala/pull/1719#issuecomment-2754231968, I run `bazel build` and `bazel test` on these targets:

```txt
//src/... //test/... //scala/... //third_party/... //scala_proto/...
```

I don't know if we need all of them, but we could at least add `//src/...` somewhere in this block of lines from `test_rules_scala.sh` (in a future PR as you prefer, of course):

```sh
$runner bazel build test/...
#$runner bazel build "test/... --all_incompatible_changes"
$runner bazel test test/...
$runner bazel test third_party/...
```

PR #1726, opened 2025-04-27
Establish minimum compatible Bazel, dep versions

mbland:
### Description

Separates the latest dependency versions that we test against from the minimum required versions and bumps `protobuf` to v30.2. Part of #1482.

Adds `scala/latest_deps.bzl` for `WORKSPACE` and `deps/latest` for Bzlmod, used by our internal test repos.

Tests used to validate these dependency versions will land in a future change.

Also:

- Fixes a bug in the failure message of `_default_platform()` in `protoc_toolchains.bzl` by calling `string.join()` on the `HOST_CONSTRAINTS` list.

- Downloads the `win64` build for Windows ARM64.

### Motivation

This avoids forcing users to upgrade to the latest versions that `rules_scala` tests against.

Inspired by a thread in the #bzlmod channel of the Bazel Slack workspace on 2025-01-01 indicating that rules should require the minumum versions possible:

- https://bazelbuild.slack.com/archives/C014RARENH0/p1743597941149639

I didn't notice the `_default_platform()` error until building on Windows ARM64, since there's not yet a binary `protobuf` release for that platform. The `win64` workaround fixes the actual error, since Windows ARM64 can execute x86 and x64 binaries:

- https://learn.microsoft.com/en-us/windows/arm/apps-on-arm-x86-emulation


mbland:
Hmm, at this point I can't tell if the Windows breakage is a fluke or not. I can't reproduce it on my local Windows ARM64 VM because of `rules_java`'s dependency on `rules_python` that I mentioned in https://github.com/bazelbuild/rules_scala/issues/1482#issuecomment-2821936058. (i.e., The build break before the failing test target even buiilds.)

Perhaps #1724 should land first (to add `--test_output=errors`), then I can rebase this pull request and try again. If it's not a fluke, we'll see it then.

simuons:
Hi, @mbland, I've just merged https://github.com/bazelbuild/rules_scala/pull/1724 could you please try rebasing.

mbland:
> Hi, @mbland, I've just merged #1724 could you please try rebasing.

Done. The build is passing now. It seems that previous test failure on Windows was some kind of fluke, or perhaps this change depended on #1724, #1727, or #1728 in some way after all.

simuons (approved):
Thanks, @mbland for the PR and for the pointers (was a good read as usual)

PR #1728, opened 2025-04-27
Remove `Label` wrappers for target string literals

mbland:
### Description

Removes all the `Label` wrappers added in #1696 that aren't strictly necessary.

Removed instances:

```sh
$ git show -p | grep '^- .*Label(' | wc -l
    88
```

Remaining instances:

```sh
$ git ls-files | xargs grep '[^A-Za-z0-9]Label(' | wc -l
    31
```

### Motivation

Inspired by research I did while writing a blog post reflecting upon the motivations behind #1696.

Bazel evaluates target string literals in the context of the repository that appears to contain the assignment of the literal to a `Label` attribute. `Label` wrappers are necessary for target string literals within macros, passed to external functions, and included in files generated by a repository rule. This ensures that the targets refer to the originating repository by preventing Bazel from interpreting them within the context of the repository using them.

Keeping the `Label` wrappers doesn't hurt, but leaving only the ones that are required helps those instances stand out.

### DIFF
--- a/jmh/jmh.bzl
+++ b/jmh/jmh.bzl
@@ -39,12 +39,12 @@ scala_generate_benchmark = rule(
         "_generator": attr.label(
             executable = True,
             cfg = "exec",
-            default = Label(
-                "//src/scala/io/bazel/rules_scala/jmh_support:benchmark_generator",
+            default = (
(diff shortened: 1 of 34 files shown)


simuons (approved):
Thanks, @mbland

PR #1729, opened 2025-04-29
Add test_dependency_versions, update test_runner

mbland:
### Description

Adds `test_dependency_versions.sh` to validate the minimum Bazel and dependency versions required by `rules_scala`, with and without the precompiled protocol compiler toolchain. Extracts setup, run, and teardown helpers from `test/shell/test_bzlmod_helpers.sh` into `test_runner.sh` for reuse by `test_dependency_versions.sh`. Part of #1482.

Also:

- Adds test files supporting `test_dependency_versions.sh` in `deps/test`.

- Adds a mechanism to skip tests by prefixing their name with `_` to `run_test_local` and `run_test_ci`.

- Adds the `RULES_SCALA_TEST_REGEX` environment variable to `test_runner.sh`.

- Adds documentation for `RULES_SCALA_TEST_{ONLY,REGEX,VERBOSE}` to the header comment of `test_runner.sh`.

- Adds `./test_dependency_versions` jobs to `.bazelci/presubmit.yml`.

### Motivation

Continuation of the previous change to ensure we don't force users to upgrade their dependencies beyond the minimum versions supported by `rules_scala`. Only builds using Bzlmod, as `WORKSPACE` is considered legacy.

Inspired by a thread in the #bzlmod channel of the Bazel Slack workspace on 2025-01-01 indicating that rules should require the minumum versions possible:

- https://bazelbuild.slack.com/archives/C014RARENH0/p1743597941149639


simuons (commented):
Thanks, @mbland 

> Note that this adds three new CI jobs. Of the three, the macOS version is dog slow for some reason. So if you want to only run a Linux job, Linux and Windows, or leave test_dependency_versions.sh out of the CI build altogether (like dt_patches/dt_patch_test.sh) let me know.

simuons:
For some reason github doesn't allow me to merge and suggest to try later. I will :) 
So after this lands I'll make new release.

PR #1731, opened 2025-04-29
Use the Publish to BCR reusable GitHub workflow

mbland:
### Description

Updates `.github/workflows/release.yml` and adds `publish-to-bcr.yml` for publishing to the Bazel Central Registry. Part of #1482 (originally broken out from #1722).

`release.yml` now uses the `release_ruleset` workflow from `bazel-contrib/.github`, which does everything `release.yml` did previously and adds SLSA provenance attestations. `release.yml` then invokes the new `publish-to-bcr.yml` workflow after publishing a successful release to GitHub.

Requires that the `BCR_PUBLISH_TOKEN` GitHub secret and the `registry_fork` specified in `.github/workflows/publish-to-bcr.yml` are in place.

See `.bcr/README.md` for all the details and references.

### Motivation

This will enable automated publishing to https://registry.bazel.build/.

### DIFF
--- a/.github/workflows/publish-to-bcr.yml
+++ b/.github/workflows/publish-to-bcr.yml
@@ -24,7 +24,7 @@ on:
 
 jobs:
   publish-to-bcr:
-    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v0.0.3
+    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v0.1.0
     with:
       tag_name: ${{ inputs.tag_name }}
       # bazelbuild/bazel-central-registry fork used to open a pull request.
--- a/.github/workflows/publish-to-bcr.yml
+++ b/.github/workflows/publish-to-bcr.yml
@@ -28,7 +28,7 @@ jobs:
     with:
       tag_name: ${{ inputs.tag_name }}
       # bazelbuild/bazel-central-registry fork used to open a pull request.
-      registry_fork: simuons/bazel-central-registry
+      registry_fork: bazel-contrib/bazel-central-registry
     permissions:
       attestations: write
       contents: write
--- a/.github/workflows/publish-to-bcr.yml
+++ b/.github/workflows/publish-to-bcr.yml
@@ -1,0 +1,38 @@
+# Publishes to the Bazel Central Registry.
+#
+# Based on .github/workflows/publish.yaml from aspect-build/rules_lint v1.3.5.
+# See .bcr/README.md.
+name: Publish to the Bazel Central Registry
+
+on:
+  # Run from release.yml.
+  workflow_call:
+    inputs:
+      tag_name:
+        required: true
+        type: string
+    secrets:
+      bcr_publish_token:
+        required: true
+
+  # In case of problems, enable manual dispatch from the GitHub UI.
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        required: true
+        type: string
+
+jobs:
+  publish-to-bcr:
+    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v0.0.3
+    with:
+      tag_name: ${{ inputs.tag_name }}
+      # bazelbuild/bazel-central-registry fork used to open a pull request.
+      registry_fork: simuons/bazel-central-registry
+    permissions:
+      attestations: write
+      contents: write
+      id-token: write
+    secrets:
+      # Necessary to push to the BCR fork and open a pull request.
+      publish_token: ${{ secrets.bcr_publish_token }}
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-to-bcr.yml
```diff
@@ -0,0 +1,38 @@
...
+  # In case of problems, enable manual dispatch from the GitHub UI.
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        required: true
+        type: string
+
+jobs:
+  publish-to-bcr:
+    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v0.0.3
```
kormide, line 27: I suggest using the latest `v0.1.0`.

mbland: @kormide Thanks, done in https://github.com/bazelbuild/rules_scala/pull/1731/commits/7f2594d9baff5f27abe3d3323b9dc282f25c13f7.

BTW, I see [release_ruleset.yaml is at v7.2.2](https://github.com/bazel-contrib/.github/releases/tag/v7.2.2). Is it worth updating to that, too? At a first quick glance, it doesn't look like the API has changed.

kormide: Probably not load bearing for your case, but it does have some fixes so worth upgrading.

mbland:
@simuons @liucijus I set the `registry_fork` to [bazel-contrib/bazel-central-registry](https://github.com/bazel-contrib/bazel-central-registry) in https://github.com/bazelbuild/rules_scala/pull/1731/commits/4c4fcaf6d08a0be9aefaa6112d309fbfbc5f3eef given @meteorcloudy's initiation of the transfer per https://github.com/bazelbuild/rules_scala/issues/1616#issuecomment-2839699762.

That just leaves creating the `BCR_PUBLISH_TOKEN` Personal Access Token in this repo. Then once this and #1730 (and possibly #1729) land, tagging `v7.0.0` and pushing should work.

mbland:
BTW, I just noticed another a [bunch of similar errors in the CI build](https://buildkite.com/bazel/rules-scala-scala/builds/5569) to the one I noted in https://github.com/bazelbuild/rules_scala/pull/1729/commits/900847b7d0dc28979c458ae973c695093a31ce2f earlier today:

```txt
WARNING: Download from
https://github.com/bufbuild/buf/releases/download/v1.27.0/protoc-gen-buf-lint-Windows-x86_64.exe
failed: class java.io.IOException GET returned 618 jwt:jwt-not-provided
```

Interesting that it looks like some invented a HTTP 618 error code somewhere.

.github/workflows/publish-to-bcr.yml
mbland: Thanks, also done in https://github.com/bazelbuild/rules_scala/pull/1731/commits/b27b6b582727c0a4b87ec04fabc970786ae75b30.

meteorcloudy:
I think the github infra is a bit unstable recently, I'm also seeing similar issue in bazel builds.

alexeagle (approved):

alexeagle:
Note, I've populated a PAT for @aspect-marvin in the BCR_PUBLISH_TOKEN secret for this repo.
If you'd like to use a different identity, you can make a different PAT.

mbland:
Oh, I was expecting @simuons or @liucijus to approve and merge this...I hope they don't mind in this case.

So to overstate the obvious to make absolutely sure I'm understanding correctly:

- Since this repo now belongs to bazel-contrib (#1616), that gives anyone in the org rights to update its secrets.
- This repo now has a `BCR_PUBLISH_TOKEN` secret, so neither @simuons nor @liucijus need to update it if they're OK with the workflow publishing as @aspect-marvin.
- If they'd prefer not to use that identity, they can create a different Personal Access Token and replace `BCR_PUBLISH_TOKEN` with that one.

So @alexeagle: If they choose to replace `BCR_PUBLISH_TOKEN`, does that complicate the BCR publishing workflow in any way? My guess is no, but I don't want to assume.

And to be clear, I'd recommend against pushing a new tag to publish a new release until #1730 has landed.

.github/workflows/publish-to-bcr.yml
```diff
@@ -0,0 +1,38 @@
...
+        required: true
+        type: string
+
+jobs:
+  publish-to-bcr:
+    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v0.1.0
+    with:
+      tag_name: ${{ inputs.tag_name }}
+      # bazelbuild/bazel-central-registry fork used to open a pull request.
+      registry_fork: bazel-contrib/bazel-central-registry
```
kormide, line 31: If you are using @aspect-marvin as the PR author, I suggest setting `draft: false` (it defaults to true). Removes one step you have to do to merge the PR.

The draft option was added because authors can't approve their own PRs on the BCR. The BCR team wants to use clicking the "Ready for review" button as a mechanism to approve the PR in those cases.

PR #1752, opened 2025-07-08
Make java library earlier on classpath

fzakaria:
### Description
I have a `scala_library` with some java files where I would like to override the class for a particular dependency (i.e. something from `@maven//`).

> This is a common idiom for monkey patching some functionality.

In order for this to work, the Java file needs to be first on the _classpath_.
@rules_scala seems to be placing the generated `java_library` at the end of the CLASSPATH.

This change makes it so that it's first.


fzakaria:
This patch properly lets me override the Java files of my Maven dependencies from my `scala_library`

PR #1753, opened 2025-07-16
expose build_ijar for scala_library

gergelyfabian:
### Description

Expose build_ijar to enable compiling scala code, that is not for macro
definitions, but would be used in inlining (so ijar support should be
disabled for it).

You may need strict dependency checks (transitive dependencies not on the
classpath) for code that is to be used in inlining.

Till now one would need to use scala_macro_library for that, but it was
recently switched to include all transitive dependencies on the compiler's
classpath, hence strict dependencies are not enforced there.

The new attribute enables having both:
a) code compiled for inlining
b) strict dependency checks

### DIFF
--- a/docs/scala_library.md
+++ b/docs/scala_library.md
@@ -172,5 +172,21 @@ In order to have a Java rule use this jar file, use the `java_import` rule.
         </p>
       </td>
     </tr>
+    <tr>
+      <td><code>build_ijar</code></td>
+      <td>
+        <p><code>Boolean; optional (default True)</code></p>
+        <p>
+          Enables building an interface jar.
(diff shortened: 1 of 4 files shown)


gergelyfabian:
Added a basic test and documentation.

simuons:
Hi, @gergelyfabian, thanks for looking into it. I did quick look and was wondering could it be solved with an attribute on the existing rule instead of new one? Looks like it's very similar to `scala_library`  or I missed the point why new rule is needed?

gergelyfabian:
@simuons, updated PR to expose a `build_ijar` attribute on `scala_library` instead of creating a new rule. This solves the problem as well.

simuons (approved):
Thanks, @gergelyfabian. We will release it as part of 7.1.1

PR #1764, opened 2025-08-27
fix: string + label concatenation error

kczulko:
### Description

Removal of string + label concatenation error when someone do not register rules_scala's PROTOC_TOOLCHAIN_TYPE

### Motivation

-


simuons (approved):
Thanks @kczulko

PR #1767, opened 2025-09-03
Add BCR presubmit checks to CI, bump versions

mbland:
### Description

Adds a matrixed `bcr_presubmit` job to `.bazelci/presubmit.yml` to parallel `.bcr/presubmit.yml`. Bumps the `MODULE.bazel` version to 7.1.2 and updates a number of Bazel module and Maven artifact dependencies.

- `com.thesamet.scalapb`: 1.0.0-alpha.1 => 1.0.0-alpha.2
- `io.grpc`: 1.74.0 => 1.75.0
- `org.scala-sbt:util-interface`: 1.11.4 => 1.11.5
- `proto-google-common-protos`: 2.60.0 => 2.61.0
- `rules_cc`: 0.1.4 => 0.2.3
- `rules_go`: 0.56.1 => 0.57.0
- `rules_java`: 8.15.1 => 8.15.2
- `rules_python`: 1.5.3 => 1.6.0-rc2
- `rules_shell`: 0.5.1 => 0.6.0

Also contains minor updates to `test_dependency_versions.sh`.

### Motivation

The matrixed `bcr_presubmit` job should prevent future failed Bazel Central Registry releases like bazelbuild/bazel-central-registry#5490 for 7.1.0. If the CI job passes, particularly the daily build job, then the BCR release pull request should pass.

The `MODULE.bazel` version bump to 7.1.2 prepares us for the next release. All the version updates other than `rules_python` are routine updates.

The `rules_python` version bump to 1.6.0-rc2 contains aarch64-pc-windows-msvc build support. This enables `rules_scala` building, testing, and debugging on the Windows Arm64 platform without using `python.single_version_platform_override`. See:

- https://github.com/bazel-contrib/rules_python/issues/2276#issuecomment-3221026183

The `test_dependency_versions.sh` updates include:

- resolvng a silent `cp` failure on `deps/test/*.{scala,bzl}`
- invoking `set -e` earlier to catch such failures
- reorganizing logic for readability
- improving the `do_build_and_test()` function comment


mbland:
Since GitHub didn't automatically assign any reviewers, I suppose I should ping @simuons and @WojciechMazur explicitly here.

MODULE.bazel
```diff
@@ -304,5 +304,5 @@ use_repo(
     "org_golang_x_tools",
 )
 
-bazel_dep(name = "rules_python", version = "1.5.3", dev_dependency = True)
-bazel_dep(name = "rules_shell", version = "0.5.1", dev_dependency = True)
+bazel_dep(name = "rules_python", version = "1.6.0-rc0", dev_dependency = True)
```
WojciechMazur, line 307: The upgrade to `-rc0` cought my attention. It looks like tags [1.6.0](https://github.com/bazel-contrib/rules_python/releases/tag/1.6.0) and [1.6.0-rc0](https://github.com/bazel-contrib/rules_python/releases/tag/1.6.0-rc0) point to the same commits, but that's only their house-keeping method -in case of rc1 the tag of 1.6.0 would point to rc1. 
It seems however like they're going have a final release shortly https://github.com/bazel-contrib/rules_python/issues/3188 
Since that's only a dev_dependency it should be fine, or we can simply stick to last stable release 1.5.4

WojciechMazur (approved):

MODULE.bazel
mbland: Huh, I didn't realize they manage their tags like that. I'd definitely prefer a more official release, but [only 1.6.0-rc0 is available on the BCR right now](https://registry.bazel.build/modules/rules_python) and [supports aarch64-pc-windows-msvc builds](https://github.com/bazel-contrib/rules_python/blob/1.6.0-rc0/python/versions.bzl#L482).

In contrast, [1.5.4 doesn't support aarch64-pc-windows-msvc builds](https://github.com/bazel-contrib/rules_python/blob/1.5.4/python/versions.bzl). Like you said, since this is a `dev_dependency`, I felt like using a release candidate is OK, since it'll enable testing and debugging on Windows Arm64 sooner than later.

PR #1771, opened 2025-09-14
Expand `$(location)` and "Make variables" in `env`

### LINKED ISSUE #1770: Conform to `*_test` `$(location)` and "Make variable" substitution behavior from native `*_test` rules
Right now `scala_test` (and I assume `scala_binary`, although I did not check), does not emulate the behavior of native `*_test` rules in terms of having the `env` attribute expand `$(location)` or substitute "Make variable"s.

To add this functionality, we would add the `env` attr and populate `RunEnvironmentInfo`.

I am not sure if this has some tradeoffs or design decisions outside of the typical cost of implementing a new feature, but this behavior is very useful for instrumenting environment variables in tests, and brings `rules_scala` behavior in line with native and other rulesets.

Bazel Slack context: https://bazelbuild.slack.com/archives/CDCKJ2KFZ/p1757624627518829

mbland:
### Description

Implements the "values are subject to `$(location)` and "Make variable" substitution" contract for the common `env` attribute for binary and test rules. Fixes #1770.

### Motivation

Makes `env` attribute semantics for Scala binary and test targets conformant with those documented in the Bazel docs:

- https://bazel.build/reference/be/common-definitions#common-attributes-binaries
- https://bazel.build/reference/be/common-definitions#common-attributes-tests
- https://bazel.build/reference/be/make-variables#use
- https://bazel.build/rules/lib/providers/RunEnvironmentInfo

Most of the new functions could live in a separate repository at some point.

As for why we're using the "deprecated" `ctx.expand_make_variables`:

- https://github.com/bazelbuild/bazel/issues/5859
- https://github.com/bazelbuild/bazel-skylib/pull/486

The "deprecated" comment in the `ctx.expand_make_variables` docstring has existed from the beginning of the file's existence (2018-05-11):

- https://github.com/bazelbuild/bazel/commit/abbb9002c41bbd53588e7249756aab236f6fcb4b


WojciechMazur (approved):
LGTM

PR #1790, opened 2025-11-05
scala_export macro

vinnybod:
### Description

This adds a drop in replacement for `java_export` that works for scala.

We've been using this internally at Confluent for some time now in our rules_jvm_external fork.
https://github.com/confluentinc/rules_jvm_external/blob/master/private/rules/scala_export.bzl

### Motivation

We needed this at Confluent and figure other folks might find it useful. Originally I tried to upstream this to rules_jvm_external, but it was decided that it would be more appropriate here and would help to avoid circular dependencies between rules_scala and rules_jvm_external if rules_scala ever chose to adopt rules_jvm_external in the future.


MODULE.bazel
```diff
@@ -45,6 +45,12 @@ single_version_override(
     version = "8.16.1",
 )
 
+bazel_dep(name = "rules_jvm_external", version = "6.9")
+single_version_override(
+    module_name = "rules_jvm_external",
+    version = "6.9",
+)
```
mbland: Please also add a `bazel_dep(name = "rules_jvm_external", version = "6.9")` to `deps/latest/MODULE.bazel`.

test/scala_export/Dependency.scala
```diff
@@ -0,0 +1,7 @@
+package scalarules.test.scala_export
+
+class Dependency {
+    def getName(): String = {
+        "scala_export test"
+    }
+}
```
mbland: This is such a small thing, but please add newlines to the end of these test `.scala` files.

MODULE.bazel
```diff
@@ -45,6 +45,12 @@ single_version_override(
     version = "8.16.1",
 )
 
+bazel_dep(name = "rules_jvm_external", version = "6.9")
```
mbland: We'll also need to set this up for legacy `WORKSPACE` builds. That means updating:

- `scala/deps.bzl`
- `scala/latest_deps.bzl`
- Probably every legacy `WORKSPACE` file in the project

Another reason the legacy `WORKSPACE` system must be destroyed, but I'm not ready to drop support just yet.

mbland (changes requested):
Overall, I think it looks mostly pretty good, modulo the few requests I've made so far, particularly around legacy `WORKSPACE` support. I also need to study it a bit to make sure I understand the whole thing, and I'll eventually pull the branch and run the tests locally.

That said, I'm going to play the BazelCon card and decline to commit to looking any more closely until about a week from now. (If changes come in, and I'm inclined, I may respond sooner. But no promises until next week.)

MODULE.bazel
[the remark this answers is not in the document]
vinnybod: No, because we aren't actually using any external maven deps.

scala/private/rules/scala_export.bzl
```diff
@@ -0,0 +1,132 @@
...
+            deps = [":" + lib_name],
+        )
+
+        scaladocs_jar_name = name + "-scaladocs"
+        create_jar(
+            name = scaladocs_jar_name,
+            inputs = [":" + scaladocs_name] + doc_resources,
+            out = name + "-scaladocs.jar",
+        )
+        classifier_artifacts["scaladoc"] = scaladocs_jar_name
```
WojciechMazur: Shouldn't that be a `javadoc`? `scaladoc` is definitely a thing, but `javadoc` is standard classifier for Maven artifacts containing documentation.

As an example all artifacts published to Maven Sonatype need to have a required `javadoc` artifact, nevertheless if that's a Scala or Java library. 

It also seems like maven_export would try to build `javadoc`, I guess it might fail for Scal sources, so maybe we should prevent it from doing so. 
https://github.com/bazel-contrib/rules_jvm_external/blob/a23c85cf0af3a2b10d3d84da7a0160261b38e2bb/private/rules/java_export.bzl#L307-L328

I expect that as result we should have the same outputs, as can be found by sbt (or other Scala dedicated build tools) releases: `–javadoc`, `-sources`, actual jar and soruces

scala/private/rules/scala_export.bzl
```diff
@@ -0,0 +1,132 @@
+load("@bazel_skylib//rules:run_binary.bzl", "run_binary")
```
WojciechMazur: Unused load

test_reproducibility.sh
```diff
@@ -21,7 +21,9 @@ md5_util() {
 }
 
 non_deploy_jar_md5_sum() {
-    find bazel-bin/test -name "*.jar" ! -name "*_deploy.jar" ! -path 'bazel-bin/test/jmh/*' | xargs -n 1 -P 5 $(md5_util) | sort
+    # Exclude *-docs.jar (javadoc jars from maven_export in rules_jvm_external)
+    # because javadoc generates non-deterministic output.
```
WojciechMazur: Do you maybe remember what was the issue here? Maybe we should fix it upstream in the scaladoc tool or add some options to make the outputs more deterministic

deps/latest/MODULE.bazel
```diff
@@ -7,10 +7,12 @@ module(
 )
 
 bazel_dep(name = "bazel_skylib", version = "1.8.2")
+bazel_dep(name = "aspect_bazel_lib", version = "2.22.0")
```
WojciechMazur: This one is used only in tests, so can be marked as dev dependency

PR #1792, opened 2025-11-11
Upgrade Scala 3.7.3 to 3.7.4

WojciechMazur:
### Description
Upgrades Scala 3.7.3 to 3.7.4

### Motivation

https://github.com/scala/scala3/issues/24344


mbland:
> Yeah, I'll be dealing with these shortly. Seems likes for some reason the GH automatically requested reviews - have you enabled that recently?

I don't _think_ I changed that...I think that's the way was set up by the time we became maintainers. @simuons Do you know if that's the case?

I wouldn't mind turning it off if y'all prefer. I don't have a strong opinion one way or another.

mbland (approved):
LGTM

PR #1794, opened 2025-11-23
Add support for Scala 3.8

WojciechMazur:
[Scala 3.8.1](https://github.com/scala/scala3/releases/tag/3.8.1) was released

- Add Scala 3.8.0-RC1 for repositories 
- Build now uses JDK 17 instead of default JDK 11 (required by Scala 3.8) 
- To test with 3.8.0-RC1 version parsing of Scalla version now strips the `-RC` suffix 
- Upgrade to 3.8.1 (3.8.0 was retired due runtime regression)

Piggybacked fixed: 
- Fix issues with builds on Windows - unlocks hard requirements for the CI. The latest_green builds are still failing

### DIFF
--- a/test/shell/test_bzlmod_macros.sh
+++ b/test/shell/test_bzlmod_macros.sh
@@ -28,7 +28,8 @@ setup_suite() {
   if is_windows; then
     # On Windows, use absolute Windows path for local_path_override because
     # Bazel cannot create junctions with Unix-style paths from MSYS2.
-    rules_scala_dir="$(cygpath -w "$original_dir")"
+    # Use -m (mixed format with forward slashes) to avoid sed escaping issues.
+    rules_scala_dir="$(cygpath -m "$original_dir")"
   else
     rules_scala_dir="$(relative_path_to_parent "$original_dir" "$test_tmpdir")"
   fi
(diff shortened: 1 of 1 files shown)


third_party/repositories/scala_3_8.bzl
```diff
@@ -0,0 +1,958 @@
...
+        "artifact": "org.scala-lang:scala3-library_3:3.8.0",
+        "sha256": "3607f6666d33f8d4d6870d31053e7c74ca65603adf61b622ba1cb24f656f9e4b",
+        "deps": [
+            "@io_bazel_rules_scala_scala_library_2",
+        ],
+    },
+    "io_bazel_rules_scala_scala_library_2": {
+        "artifact": "org.scala-lang:scala-library:3.8.0",
+        "sha256": "c9242f95e9e2cf2e0fc07e6334885c258204f6c83ba7dc5565dae237be9f0618",
+    },
```
WojciechMazur: This might be confusing, but it's correct. At the point when `io_bazel_rules_scala_scala_library_2` was choosen nobody expected  `scala_library_2` whould have version `3.8.0` 
We should not change that for source compatibility

PR #1795, opened 2025-11-24
[Compiler Toolchains] Add support for plugins

FrankPortman:
### Description

Adds a `plugins` attr to `scala_toolchain`, merged with each target's own `plugins` in `phase_compile`, so compiler plugins can be enabled for every target using the toolchain.

### Motivation

Sometimes you want to apply your `plugin`s to your entire monorepo, or apply the plugins broadly based on flags. The current ways to do so would involve "wrapping" `scala_library` and injecting these specifics into it via `macro`, which feels regrettable because of how foundational of a concept it is. On the other hand, if you "wrap" `scala_library` and create `new_scala_library`, you must then know when/where to use this other version.

### Testing

Analysis tests (`//test/toolchain_plugins`) assert the toolchain's plugin jar appears in the `Scalac` command line under an `--extra_toolchains` transition, and doesn't otherwise. The plugin jar is never executed, so the tests run on every Scala version — including Scala 3, where a real plugin repo like `@org_typelevel_kind_projector` doesn't even analyze (its generated BUILD references `_2`-suffixed compiler repos that only exist under Scala 2; this is what broke CI on the previous revision). Plugin *execution* via `--Plugins` is unchanged and already covered by the target-level plugins tests; toolchain plugins join the same list.

Verified locally under 2.13.18 and 3.7.4. `plugins` documented in `docs/scala_toolchain.md`.


FrankPortman:
@mbland 

Reworked per your feedback:

- Replaced the manual `--extra_toolchains` builds with analysis tests (`//test/toolchain_plugins`) asserting the plugin jar's presence/absence in the `Scalac` command line. The CI failures came from the old test targets compiling kind-projector syntax under the default (pluginless) toolchain.
- The analysis tests never execute the plugin jar, so they pass on every Scala version. This is a design choice by me - LMK if the analysis test alone is convincing enough re functionality. I would posit that it is, because other tests exercise that plugins work at all, and this is just verifying they're passed in where expected.
  - `@org_typelevel_kind_projector//jar` has hardcoding to Scala 2
  - Plugin execution via `--Plugins` is already covered by the target-level plugins tests; toolchain plugins join the same list in `phase_compile`.
- Trailing newlines fixed

Verified locally under 2.13.18 and 3.7.4.

test/toolchain_plugins/toolchain_plugins_test.bzl
```diff
@@ -0,0 +1,54 @@
...
+toolchain_plugin_present_test = analysistest.make(
+    _toolchain_plugin_present_test,
+    config_settings = {
+        "//command_line_option:extra_toolchains": [
+            "//test/toolchain_plugins:with_plugin",
+        ],
+    },
+)
+
+toolchain_plugin_absent_test = analysistest.make(_toolchain_plugin_absent_test)
```
bor-p-s: This test doesn't pin `--extra_toolchains`, so it only proves "no plugin jar" under whatever extra toolchain the invoking build happens to pass. `test_rules_scala.sh` reruns all of `test/...` under three different `--extra_toolchains` values (`high_level_transitive_deps_strict_deps_error`, `minimal_direct_source_deps`, `plus_one_deps`), and it passes today only because none of those three set `plugins`. If any of them (or a future one added to that list) ever does, this test would silently stop proving what its name says.

Can you pin it the same way `toolchain_plugin_present_test` does, just with an empty list:

```python
toolchain_plugin_absent_test = analysistest.make(
    _toolchain_plugin_absent_test,
    config_settings = {
        "//command_line_option:extra_toolchains": [],
    },
)
```

That way the test is immune to whatever toolchain override the outer build happens to use.

FrankPortman:
@bor-p-s - good catch. updated with your suggestion and I also tested it both ways to make sure it works (more to better understand it myself).

bor-p-s (approved):
LGTM. The toolchain-level plugins attr follows the same merge pattern already used for scalacopts, and the analysis test now pins `--extra_toolchains` on both sides so it stays correct regardless of what the outer build passes.

PawelLipski (approved):

PR #1797, opened 2025-12-02
Make java library earlier on classpath

vinnybod:
### Description

I have a scala_library with some java files where I would like to override the class for a particular dependency (i.e. something from @maven//).

> This is a common idiom for monkey patching some functionality.

In order for this to work, the Java file needs to be first on the classpath.
@rules_scala seems to be placing the generated java_library at the end of the CLASSPATH.

This change makes it so that it's first.

This is a copy of https://github.com/bazel-contrib/rules_scala/pull/1752 but with a test added


test_rules_scala.sh
```diff
@@ -19,10 +19,10 @@ $runner bazel build src/... test/...
 $runner bazel test "${test_output_flag}" src/... test/...
 $runner bazel test "${test_output_flag}" third_party/...
 $runner bazel build --extra_toolchains=//test/toolchains:high_level_transitive_deps_strict_deps_error -- test/...
-$runner bazel build --extra_toolchains=//scala:minimal_direct_source_deps -- test/...
+$runner bazel build --extra_toolchains=//scala:minimal_direct_source_deps --build_tag_filters=-no-strict-deps -- test/...
 #$runner bazel build --extra_toolchains=//test/toolchains:high_level_transitive_deps_strict_deps_error --all_incompatible_changes -- test/...
 $runner bazel test "${test_output_flag}" --extra_toolchains=//test/toolchains:high_level_transitive_deps_strict_deps_error -- test/...
-$runner bazel test "${test_output_flag}" --extra_toolchains=//scala:minimal_direct_source_deps -- test/...
+$runner bazel test "${test_output_flag}" --extra_toolchains=//scala:minimal_direct_source_deps --build_tag_filters=-no-strict-deps --test_tag_filters=-no-strict-deps -- test/...
```
vinnybod: I don't love that I had to do this, but afaik there isn't a way to disable "strict deps" on a target level like you can with "unused deps"

PR #1802, opened 2026-01-25
chore(docs): add starter repo instructions to README

alexeagle:
Added instructions for using the starter repository.

Note, this is the same link reachable from https://bazel.build/start


mbland:
I'm not sold on the start project link as an appropriate inclusion here. Closing without merge.

Issue #1816, opened 2026-02-27
Will removing`compatibility_level` in v7.2.3 break users?


mbland:
Today I released v7.2.3, which eagerly removes `compatibility_level` via #1815, since [compatibility_level is a no-op in Bazel 8.6.0 and will be in 9.1.0](https://github.com/bazelbuild/bazel/pull/28600):

- https://github.com/bazel-contrib/rules_scala/releases/tag/v7.2.3
- https://registry.bazel.build/modules/rules_scala/7.2.3

I had to override the `compatibility_level` check in https://github.com/bazelbuild/bazel-central-registry/pull/7729 to unblock the release flow. Only after that did I ask myself, "Will this break any of our users?"

Pinging @meteorcloudy and @Wyverald for their opinions. I could [yank 7.2.3](https://bazel.build/external/registry#metadata-json) and restore `compatibility_level = "7"` if this might become a problem.

mbland:
#1817 effectively closed this.

PR #1817, opened 2026-03-03
Restore compatibility_level, yank 7.2.3

mbland:
### Description

Restores `compatibility_level = 7` to the `module` directive and yanks version 7.2.3 from the Bazel Central Registry.

### Motivation

Per #1816, #1815 removed the `compatibility_level` attribute of the `module` directive too soon. Builds using Bazel <8.6.0 or <9.1.0 with other `rules_scala` 7.x dependencies in the build graph would break once any module introduces `rules_scala` 7.2.3.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -3,7 +3,7 @@ validate_config: 1
 
 matrix:
   bcr_platform: ["debian10", "macos", "ubuntu2004", "windows"]
-  bcr_bazel: ["7.x", "8.x", "9.x", "rolling", "last_green"]
+  bcr_bazel: [7.x, 8.x, 9.x, rolling, last_green]
 
 tasks:
   ubuntu2004:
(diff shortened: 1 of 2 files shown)


mbland:
Uh, wow...something funny's going on with Bazel's CI system. Filed: bazelbuild/continuous-integration#2486

mbland:
> Re-removal of the compatibility level, needs to wait for the new minor version being cutoff, right?

I don't _think_ so. By yanking 7.2.3 from the BCR, it should be almost like it never happened. (And it has no effect at all on legacy `WORKSPACE` users.)

BTW, I pushed https://github.com/bazel-contrib/rules_scala/pull/1817/changes/b6cbb95b4db36b103b7f7023e819f618a17d5346 to see if it would help with the breakage from https://github.com/bazelbuild/continuous-integration/issues/2486, but it didn't.

PR #1825, opened 2026-04-11
Fix cross-version default dep providers

gmethvin:
### Description

Change the default `declare_deps_provider` targets in `@rules_scala_toolchains` to use `select()` over `SCALA_VERSIONS` instead of hardcoding `SCALA_VERSION` (the workspace default).

Custom `scala_toolchain()` callers that don't pass explicit `dep_providers` (using `_default_dep_providers()`) were always getting the default version's compiler JARs, regardless of the transitioned `scala_version`. This made cross-compilation between Scala 2.x and 3.x impossible with per-target `scala_version`.

### Motivation

When building a `scala_library` with `scala_version = "3.7.4"` while the workspace default is `"2.13.18"`, the scalac worker's reporter fails to compile. The toolchain resolves correctly (the Scala 3 toolchain is selected), but its `dep_providers` default to unprefixed provider targets that were hardcoded to `SCALA_VERSION`:

    declare_deps_provider(
        deps = default_deps(deps_id, SCALA_VERSION),  # always "2.13.18"
    )

The reporter's `select_for_scala_version` picks Scala 3 Java sources (importing `dotty.tools.dotc.*`), but `scala_compile_classpath` provides Scala 2.13 compiler JARs, causing:

    error: symbol not found dotty.tools.dotc.reporting.AbstractReporter

`setup_scala_toolchain` correctly creates per-version providers, but `_default_dep_providers()` in `scala_toolchain.bzl` returns the unprefixed (default-version) labels. Using `--repo_env=SCALA_VERSION` works around it by changing the default globally.

### DIFF
--- a/scala/toolchains_repo.bzl
+++ b/scala/toolchains_repo.bzl
@@ -126,7 +126,7 @@ load(
 )
 load("@rules_scala//scala:providers.bzl", "declare_deps_provider")
 load("@rules_scala//scala:scala_cross_version.bzl", "version_suffix")
-load("@rules_scala_config//:config.bzl", "SCALA_VERSION", "SCALA_VERSIONS")
+load("@rules_scala_config//:config.bzl", "SCALA_VERSIONS")
 
 [
     setup_scala_toolchain(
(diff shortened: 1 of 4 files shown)


mbland (approved):
LGTM

Thanks for this! Will try to get a release out with this fix in the next couple of days.

PR #1826, opened 2026-04-14
Upgrade build_jacocorunner.sh script for Bazel 9

gergelyfabian:
Added new patch file for Bazel 9.
Treat Bazel 8 as Bazel 7 for earlier usage.

### DIFF
--- a/scripts/build_jacocorunner/0001-Build-Jacoco-for-Bazel-9.0.patch
+++ b/scripts/build_jacocorunner/0001-Build-Jacoco-for-Bazel-9.0.patch
@@ -1,0 +1,29 @@
+From 712d62a8238f3a7fe51e1cf4cc2520b5f249e1d3 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Gergely=20F=C3=A1bi=C3=A1n?= <gergo.fb@gmail.com>
+Date: Tue, 14 Apr 2026 12:28:00 +0200
+Subject: [PATCH] Build Jacoco for Bazel
+
+---
+ org.jacoco.build/pom.xml | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/org.jacoco.build/pom.xml b/org.jacoco.build/pom.xml
+index 8aae1543..067cc6a7 100644
+--- a/org.jacoco.build/pom.xml
++++ b/org.jacoco.build/pom.xml
+@@ -691,11 +691,11 @@
+
+                 buildDate = qualifier.substring(0, 4) + "/" + qualifier.substring(4, 6) + "/" + qualifier.substring(6, 8);
+                 project.getProperties().setProperty("build.date", buildDate);
+
+                 commitId = project.getProperties().get("build.commitId");
+-                pkgName = commitId.substring(commitId.length() - 7, commitId.length());
++                pkgName = "29a6edd";
+                 project.getProperties().setProperty("jacoco.runtime.package.name", "org.jacoco.agent.rt.internal_" + pkgName);
+
+                 void loadLicense(String libraryId) {
+                     version = project.getProperties().get(libraryId + ".version");
+                     path = project.getBasedir().toPath().resolve("../org.jacoco.build/licenses/" + libraryId + "-" + version + ".html");
+-- 
+2.25.1
+
--- a/scripts/build_jacocorunner/build_jacocorunner.sh
+++ b/scripts/build_jacocorunner/build_jacocorunner.sh
@@ -2,25 +2,22 @@
 #
 # Script to build custom version of `JacocoCoverage_jarjar_deploy.jar` from Jacoco and Bazel repositories.
 #
-# The script has three flavours: Bazel 5, 6 and 7. Documenting the details for Bazel 7 here.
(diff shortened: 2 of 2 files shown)


scripts/build_jacocorunner/0001-Build-Jacoco-for-Bazel-9.0.patch
```diff
@@ -0,0 +1,29 @@
...
+--- a/org.jacoco.build/pom.xml
++++ b/org.jacoco.build/pom.xml
+@@ -691,11 +691,11 @@
+
+                 buildDate = qualifier.substring(0, 4) + "/" + qualifier.substring(4, 6) + "/" + qualifier.substring(6, 8);
+                 project.getProperties().setProperty("build.date", buildDate);
+
+                 commitId = project.getProperties().get("build.commitId");
+-                pkgName = commitId.substring(commitId.length() - 7, commitId.length());
++                pkgName = "29a6edd";
```
mbland, line 21: I understand hardcoding a `pkgName` follows the pattern from the preexisting patch files in this directory. But why is it necessary? It seems like each of these values are magic numbers.

scripts/build_jacocorunner/build_jacocorunner.sh
```diff
@@ -72,12 +72,16 @@ bazel_major_version=$1
 if [ -z "$bazel_major_version" ]; then
   echo "Please provide Bazel major version"
   exit 1
-elif [ "$bazel_major_version" != "5" ] && [ "$bazel_major_version" != "6" ] && [ "$bazel_major_version" != "7" ]; then
+elif [ "$bazel_major_version" != "5" ] && [ "$bazel_major_version" != "6" ] && [ "$bazel_major_version" != "7" ] && [ "$bazel_major_version" != "8" ] && [ "$bazel_major_version" != "9" ]; then
```
mbland: The shebang is `#!/usr/bin/env bash`, so there's no need to remain tied to pure `sh` syntax. Let's tidy this up to:

```bash
elif [[ "$bazel_major_version" -lt 5 || "$bazel_major_version" -gt 9 ]]; then
```

mbland (changes requested):

scripts/build_jacocorunner/0001-Build-Jacoco-for-Bazel-9.0.patch
gergelyfabian: If I don't do this change, then I'll get an error (when using the previously built JacocoRunner for Bazel 7/8):

`Cause: java.lang.NoClassDefFoundError: org/jacoco/agent/rt/internal_29a6edd/Offline`

This was the direct motivation for adding a new patch file.

gergelyfabian: I get the same exception if I build JacocoRunner without this patch file.

gergelyfabian: The buildnumber-maven-plugin in JaCoCo's pom.xml captures the current git HEAD commit hash and takes its last 7 characters as the package suffix. This suffix gets baked into the jar via Maven shading - all internal classes end up under `org.jacoco.agent.rt.internal_<suffix>`.

The problem is that the fork's 0.8.11-scala branch HEAD is commit `993411620...d88945e`, so an unpatched build would produce the package `org.jacoco.agent.rt.internal_d88945e`. But Bazel's own bundled JaCoCo 0.8.14 jars (in `third_party/java/jacoco/`) were built from upstream JaCoCo commit `2eb24836...29a6edd` (the v0.8.14 release tag), so they contain `org.jacoco.agent.rt.internal_29a6edd`.

At build time, Bazel's `OfflineInstrumentationAccessGenerator` reads `RUNTIMEPACKAGE` from jacoco.properties inside the core jar and injects bytecode into every instrumented class that calls `org.jacoco.agent.rt.internal_29a6edd.Offline.getProbes()`. At runtime, if the agent jar has the Offline class under a different suffix (`_d88945e`), we get `NoClassDefFoundError`.

So the hardcoding is necessary because the fork is built from a different commit than the upstream release that Bazel bundles. Even without any patches, the fork's commit hash differs from upstream's. The value must match what Bazel's own JaCoCo jars were built with - i.e., the last 7 characters of the upstream JaCoCo release commit that the specific Bazel version ships. You can verify: `unzip -p org.jacoco.core-0.8.14.jar org/jacoco/core/jacoco.properties` shows `COMMITID=2eb248366f0eb63fd964fc7a81804b27229a6edd` and `RUNTIMEPACKAGE=org.jacoco.agent.rt.internal_29a6edd`.

(Used AI a bit to confirm this)

gergelyfabian: This also shows, that I need to update the fork to diverge from Jacoco 0.8.14.

gergelyfabian: Upgraded to 0.8.14. This caused a lot more changes.

scripts/build_jacocorunner/build_jacocorunner.sh
gergelyfabian: Fixed.

WojciechMazur (approved):

PR #1830, opened 2026-04-25
Enable semanticdb phase during plugin bootstrapping

thirtyseven:
### Description
Add `phase_semanticdb` to the `scala_library_for_plugin_bootstrapping` phase list so the semanticdb plugin jar is loaded whenever the toolchain enables semanticdb.

Previously, `scala_library_for_plugin_bootstrapping` (used to compile the dep analyzer plugin) skipped `phase_semanticdb`. When a toolchain sets `enable_semanticdb = True` and includes a `-P:semanticdb:...` scalacopt (e.g. `-P:semanticdb:synthetics:on`), the bootstrap compilation received the flag but had no semanticdb plugin loaded — scalac failed with `error: bad option: -P:semanticdb:synthetics:on`.

A regression test that builds `//test/semanticdb:all_lib` with a synthetics-enabled semanticdb toolchain is also added to verify the bootstrap path handles these scalacopts.

### Motivation
When a toolchain sets `enable_semanticdb = True` and passes `-P:semanticdb:...` scalacopts (e.g. `-P:semanticdb:synthetics:on`), the dep analyzer plugin bootstrap compilation failed with `error: bad option: -P:semanticdb:synthetics:on` because `phase_semanticdb` was not included in the bootstrap phase list and the plugin jar was never loaded.

### DIFF
--- a/scala/private/rules/scala_library.bzl
+++ b/scala/private/rules/scala_library.bzl
@@ -163,7 +163,7 @@ def _scala_library_for_plugin_bootstrapping_impl(ctx):
             ("dependency", phase_dependency_library_for_plugin_bootstrapping),
             ("collect_jars", phase_collect_jars_common),
             ("scalacopts", phase_scalacopts),
-            #("semanticdb", phase_semanticdb), noneed for semanticdb in bootstrap
+            ("semanticdb", phase_semanticdb),  # enable semanticdb during bootstrap since user scalacopts may include flags for it
             ("compile", phase_compile_library_for_plugin_bootstrapping),
             ("merge_jars", phase_merge_jars),
             ("runfiles", phase_runfiles_library),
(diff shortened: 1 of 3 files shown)


scala/private/rules/scala_library.bzl
```diff
@@ -163,7 +163,7 @@ def _scala_library_for_plugin_bootstrapping_impl(ctx):
             ("dependency", phase_dependency_library_for_plugin_bootstrapping),
             ("collect_jars", phase_collect_jars_common),
             ("scalacopts", phase_scalacopts),
```
thirtyseven, line 165: An alternate approach could be to disable this phase during bootstrap instead, but I think this has a higher (but still small) risk of breaking existing users

WojciechMazur (approved):

PR #1837, opened 2026-06-06
fix: Expose SemanticdbInfo.target_root when semanticdb is bundled in jar

### LINKED ISSUE #1831: SemanticdbInfo.target_root is None when bundle_in_jar=True even though files are on disk
Metals Doctor doesn't show a green checkmark for SemanticDb.

`scala/private/phases/phase_semanticdb.bzl` writes `.semanticdb` files to `<bin>/<pkg>/_scalac/<name>/classes/META-INF/semanticdb/` regardless of `semanticdb_bundle_in_jar`. With `bundle_in_jar=True` they get packed into the jar but also remain on disk. `SemanticdbInfo.target_root` is set to `None` in that mode, so external consumers can't locate the on-disk files without reconstructing the path.

```python
semanticdb_provider = SemanticdbInfo(
    semanticdb_enabled = True,
    target_root = semanticdb_target_root,   # was: None if bundled
    is_bundled_in_jar = toolchain.semanticdb_bundle_in_jar,
    plugin_jar = plugin_jar,
)
```

`is_bundled_in_jar` still signals "also in jar."

WojciechMazur:
When semanticdb_bundle_in_jar is enabled, semanticdb files are still written under _scalac/<target>/classes/META-INF/semanticdb/ on disk and also packed into the output jar. 

Always populate target_root with the on-disk path and keep is_bundled_in_jar to indicate jar bundling. 
Update semanticdb tests to assert target_root and on-disk files in bundled mode.

Fixes #1831

### DIFF
--- a/scala/private/phases/phase_semanticdb.bzl
+++ b/scala/private/phases/phase_semanticdb.bzl
@@ -60,7 +60,7 @@ def phase_semanticdb(ctx, p):
 
         semanticdb_provider = SemanticdbInfo(
             semanticdb_enabled = True,
-            target_root = None if toolchain.semanticdb_bundle_in_jar else semanticdb_target_root,
+            target_root = semanticdb_target_root,
             is_bundled_in_jar = toolchain.semanticdb_bundle_in_jar,
             plugin_jar = plugin_jar,
         )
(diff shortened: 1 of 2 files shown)


oharboe:
😌

PR #1839, opened 2026-06-17
worker: dispatch persistent workers via Bazel's WorkRequestHandler

gergelyfabian:
The custom persistentWorkerMain loop in Worker.java hijacked System.out / System.err into a per-request ByteArrayOutputStream shipped via WorkResponse.output, but had no fallback if the JVM exited before the WorkResponse was written: an uncaught Error (e.g. OutOfMemoryError, which escapes the existing `catch (Exception)`) tore down the JVM with the captured buffer still in memory. The worker server saw an EOF on the codec, no stderr on the pipe, and exit status 1 with no diagnostic. The non-persistent ephemeralWorkerMain path did not have this problem because it never hijacked the streams.

Replace persistentWorkerMain with a WorkRequestHandler-based dispatch, mirroring rules_kotlin's PersistentWorker.kt and Bazel's own BazelJavaBuilder. WorkRequestHandler snapshots the original stderr (`realStdErr`), installs an uncaught-exception handler on each request thread that writes Error stack traces to that snapshotted stderr and then calls System.exit(1) — bytes land on the real FD 2, the worker server captures them, and the diagnostic reaches operators on the kill line. WorkerIO inside WorkRequestHandler still captures per-request System.out / System.err for the normal-completion path, so successful actions continue to ship their output in WorkResponse.output as before.

Side benefits beyond the regression fix: multiplex worker support is now available without further plumbing, cancellation semantics route through the handler's mechanism, and the GC cadence is bounded by WorkRequestHandler's cpuUsageBeforeGc / idleTimeBeforeGc knobs instead of an unconditional System.gc() per request.

SmartByteArrayOutputStream and its corresponding test go away with the custom loop they served. WorkerTest is restructured to start the worker on a background thread, write the request, read the response, then close stdin — the new code dispatches each request to its own thread and exits the read loop on stdin EOF, so the previous "pre-close stdin then run worker synchronously" pattern races with that worker thread. The argsfile test now expects the trimmed output that
WorkRequestHandler.respondToRequest produces.

### Motivation

Found this issue as testing a Remote Persistent Worker feature I'm developing for BuildBarn, with rules_scala error-handling issue became evident.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -72,6 +72,10 @@ single_version_override(
 # See //src/java/io/bazel/rulesscala/worker:worker_protocol_java_proto.
 bazel_dep(name = "bazel_worker_api", version = "0.0.10")
 
+# WorkRequestHandler-based persistent worker dispatch. See
+# //src/java/io/bazel/rulesscala/worker:Worker.java.
+bazel_dep(name = "bazel_worker_java", version = "0.0.10")
+
 scala_config = use_extension(
     "//scala/extensions:config.bzl",
     "scala_config",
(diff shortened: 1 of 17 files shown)


gergelyfabian:
Will return to this later.

gergelyfabian:
Superseded by #1841 in debugging.

PR #1841, opened 2026-06-18
worker: flush captured stderr on abnormal exit, gate trace markers behind env var

gergelyfabian:
The persistent-worker loop redirects System.out / System.err into an in-memory buffer (outStream) so the bytes ride back in WorkResponse.output. When the JVM exits before the next WorkResponse is written — direct System.exit(N) from inside work() (scalac's reporter, a compiler plugin, any wrapper that bails via Runtime.exit), uncaught Throwable unwinding the main thread, signal-induced shutdown that still runs hooks — the captured buffer dies in memory and the worker server receives nothing but an EOF on the codec. SIGKILL, native JVM crashes, and Runtime.halt are unavoidably uncovered.

Snapshot System.err into a `realStdErr` local before any setErr() call and install a JVM shutdown hook that, on every termination path that runs hooks, flushes the in-flight outStream onto realStdErr (still wired to the worker server's pipe). The normal-completion path resets outStream after every request, so a clean stdin-EOF shutdown leaves the buffer empty; the hook is removed from the runtime in the outer finally on normal exit. Track in-flight argv and a completed-request counter so the abnormal-exit diagnostic can name the action that was running. Captured tail is bounded to 64 KB (last bytes wins; the failure-causing output is typically at the end), preventing a runaway scalac from flooding the worker server's stderr ring buffer.

Per-request start/end markers and lifecycle banners are gated behind the RULES_SCALA_WORKER_TRACE env var — off by default, since the shutdown-hook diagnostic already names the in-flight argv. Set to a truthy value (e.g. RULES_SCALA_WORKER_TRACE=1) when debugging worker behavior.

Tests cover the summarizeArgv helper and exercise the shutdown-hook / trace-mode paths via a forked JVM subprocess (testing them in-process would require taking the test JVM down).

The ephemeral (non-persistent) path is unchanged; it never hijacks the streams and so never needed the recovery path.

### Motivation

Used to debug #1840.

### DIFF
--- a/src/java/io/bazel/rulesscala/worker/Worker.java
+++ b/src/java/io/bazel/rulesscala/worker/Worker.java
@@ -10,6 +10,8 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * A base for JVM workers.
@@ -19,6 +21,15 @@ import java.util.List;
  *
  * <p>Worker implementations should implement the `Worker.Interface` interface and provide a main
  * method that calls `Worker.workerMain`.
+ *
+ * <p>The persistent-worker loop hijacks {@link System#out} / {@link System#err} into an
+ * in-memory buffer so the bytes can ride back in {@code WorkResponse.output}. The original
+ * {@code System.err} is snapshotted into {@code realStdErr} so a JVM shutdown hook can flush
+ * the captured tail of the in-flight request onto it on an abnormal exit that still runs
+ * shutdown hooks: {@link System#exit} from inside {@code work()}, uncaught {@link Throwable}
+ * unwinding the main thread, or signal-induced shutdown (e.g. SIGTERM). It does NOT cover
+ * SIGKILL, native JVM crashes, or {@link Runtime#halt} — those terminate without running hooks
+ * and the captured tail is unavoidably lost.
  */
 public final class Worker {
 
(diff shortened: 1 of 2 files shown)


gergelyfabian:
I'm not sure this should land in master, just sending for the record. Very helpful for debugging if anything goes south in error handling.

PawelLipski:
Hey, thanks for the change - let me add some comments on possible test improvements/additions, otherwise good to go

src/java/io/bazel/rulesscala/worker/Worker.java
```diff
@@ -94,22 +136,124 @@ private static void persistentWorkerMain(Interface workerInterface) {
...
+   * partial-buffer read — still better than no diagnostic at all.
+   */
+  private static Thread installAbnormalExitHook(
+      PrintStream realStdErr,
+      ByteArrayOutputStream outStream,
+      PrintStream out,
+      AtomicReference<String> inFlightArgv,
+      AtomicLong requestsCompleted,
+      String prefix) {
+    final int maxTailBytes = 64 * 1024;
```
PawelLipski, line 204: Consider adding a test for this truncation logic as well (AFAICS not covered yet)

src/java/io/bazel/rulesscala/worker/Worker.java
```diff
@@ -94,22 +136,124 @@ private static void persistentWorkerMain(Interface workerInterface) {
...
+              try {
+                out.flush();
+                byte[] buf = outStream.toByteArray();
+                int total = buf.length;
+                int start = total > maxTailBytes ? total - maxTailBytes : 0;
+                int len = total - start;
+                String inFlight = inFlightArgv.get();
+
+                realStdErr.println();
+                realStdErr.println(prefix + "abnormal-exit diagnostic");
```
PawelLipski, line 217: Is there a test for happy path - hook not run, no `abnormal-exit diagnostic` printed out if the worker shuts down successfully? in case the logic around `removeShutdownHook` is ever removed

src/java/io/bazel/rulesscala/worker/Worker.java
```diff
@@ -94,22 +136,124 @@ private static void persistentWorkerMain(Interface workerInterface) {
...
   }
 
+  private static boolean isTraceEnabled() {
+    String v = System.getenv("RULES_SCALA_WORKER_TRACE");
+    if (v == null || v.isEmpty()) return false;
+    return !v.equals("0") && !v.equalsIgnoreCase("false");
+  }
+
+  /**
+   * Installs a JVM shutdown hook that, on abnormal exit, flushes the persistent worker's captured
```
PawelLipski, line 187: Maybe explain what's the mechanism for making sure it only fires on an abnormal exit (`removeShutdownHook`, I suppose)

PawelLipski (approved):

PR #1842, opened 2026-07-01
Declare a "scalac" exec_group so per-action cpu sizing can be set

gergelyfabian:
Adds an empty `exec_group("scalac")` to every scala rule that compiles Scala sources (scala_library, scala_library_for_plugin_bootstrapping, scala_macro_library, scala_binary, scala_test, scala_junit_test, scala_repl) and wires the Scalac action to it via
`exec_group = "scalac"` in `compile_scala`. Also declares the same exec_group on the two aspects that call `compile_scala` (scrooge_scala_aspect, scala_proto_aspect) — aspects do not inherit exec_groups from the rule they attach to, so they need their own declaration.

The group is intentionally empty (no toolchains / no exec_compatible_with) — its sole purpose is to be a target for the `<group>.<key>` syntax on `exec_properties`. This lets a user set `exec_properties = {"scalac.<key>": "<value>"}` on a target to configure the compile action independently of the target's other actions (test execution, jar packaging, etc.). Without a declared exec_group, Bazel rejects the prefixed key at analysis time.

The action's toolchain resolution and exec platform selection are unchanged: no toolchains or constraints are attached to the group, so the rule's default toolchains still apply to the Scalac action.

`scala_proto_library` and `scrooge_scala_library` intentionally do not declare the exec_group on the top-level rule: Bazel does not propagate target-level `exec_properties` to actions created by attached aspects, so declaring it there would silently drop the user's config. The analysis-time error surfaces the limitation; users needing per-action config for these rules should set exec_properties at the platform level.

Regression coverage:

- //test/scalac_exec_group/... holds a target per rule, each setting a
  scalac.* exec_property, covering the seven rules that support the
  feature. Removing the rule's exec_group declaration fails analysis on
  these targets under `bazel build test/...`.
- :scalac_action_exec_properties_test runs a single nested `bazel
  aquery` over all of them to verify the property reaches the Scalac
  action's ExecutionInfo, requiring one match per target so a partial
  regression cannot hide behind a passing pattern match. This catches
  removal of `exec_group = "scalac"` from `ctx.actions.run` in
  compile_scala (which would silently reroute the property to the
  default exec group without failing analysis). Starlark's Action API
  exposes no execution info, so this cannot be an analysistest.
- :scala_proto_library_rejects_scalac_exec_property and
  :scrooge_scala_library_rejects_scalac_exec_property exercise the
  intentional gap on those two rules, asserting each fails with the
  expected "non-existent exec groups: scalac" message, so the docs'
  claim stays honest if Bazel changes behavior or someone adds the
  exec_group to them. Bazel raises that error while configuring the
  target, before the rule is analyzed, so `--allow_analysis_failures`
  never sees it and expect_analysis_failure_test cannot be used.

Documents the feature and the aspect-propagation limitation under docs/scalac_exec_group.md.


PawelLipski:
@bor-p-s  if you could take look

test/shell/test_scalac_exec_group.sh
```diff
@@ -0,0 +1,56 @@
...
+  //test:ScalacExecGroupJunitTest
+  //test:ScalacExecGroupRepl
+)
+
+test_scalac_action_receives_exec_property() {
+  local target output
+  for target in "${SCALAC_EXEC_GROUP_TARGETS[@]}"; do
+    output="$(bazel aquery "mnemonic(\"Scalac\", ${target})" 2>&1)"
+    assert_matches 'ExecutionInfo:.*cpu: 1' "$output" \
+      "Scalac action on ${target} did not receive scalac.cpu=1 via exec_properties"
```
bor-p-s: `test_scalac_action_receives_exec_property` runs a separate `bazel aquery` per target, so 7 Bazel calls on every CI push. `test/shell/test_scala_proto_library.sh` already does this with one combined query: `mnemonic("Scalac", //test:ScalacExecGroupLib + ... + //test:ScalacExecGroupRepl)`.

One thing to be careful with: if you just run `assert_matches` on the combined output, it only checks the pattern shows up once, so it would still pass even if 6 of the 7 targets lost the property. To keep the check strict, count the matches (e.g. `grep -c 'ExecutionInfo:.*cpu: 1'`) and compare against `${#SCALAC_EXEC_GROUP_TARGETS[@]}`.

gergelyfabian: Done

bor-p-s (approved):
Looked at the whole diff and the aquery fix — LGTM.

test/BUILD
```diff
@@ -775,6 +776,76 @@ scala_library(
     deps = [":InlinableExported"],
 )
 
+# Regression targets for the "scalac" exec_group. Each sets a scalac.*
```
PawelLipski: Could you extract these targets into separate subdirectories under test/ ? rather than keeping them under top-level test/BUILD

gergelyfabian: Done.

test/BUILD
```diff
@@ -6,6 +6,7 @@ load(
     "scala_doc",
     "scala_junit_test",
     "scala_library",
+    "scala_library_for_plugin_bootstrapping",
```
PawelLipski: Nit: import redundant now

test/shell/test_scalac_exec_group.sh
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env bash
```
PawelLipski: Whoops could you rewrite this script to the new format - BUILD files with regular targets? we're migrating away from these scripts (see most of PRs in the recent months)

test/BUILD
gergelyfabian: Done

test/shell/test_scalac_exec_group.sh
gergelyfabian: Yes, did that. Hopefully I managed to match the new conventions.

test/scalac_exec_group/BUILD
```diff
@@ -0,0 +1,172 @@
...
+        "'cpu: 1'",
+        "//test/scalac_exec_group:Lib",
+        "//test/scalac_exec_group:Macro",
+        "//test/scalac_exec_group:LibForPluginBootstrapping",
+        "//test/scalac_exec_group:Binary",
+        "//test/scalac_exec_group:Test",
+        "//test/scalac_exec_group:JunitTest",
+        "//test/scalac_exec_group:Repl",
+    ],
+    data = [
```
bor-p-s, line 115: These nested-Bazel tests don't have the rule-implementation `.bzl` files as `data`, so a regression in `compile_scala` or the rule definitions can hit a stale cached PASS instead of rerunning (confirmed by removing `exec_group = "scalac"` and getting `(cached) PASSED`). Add them to `data`, same as `code_under_test` in `expect_build_failure.bzl`.

gergelyfabian: Tried satisfying your request, but ongoing changes on master were a bit hard to navigate, so two tests had to become external.

bor-p-s: Thanks, this will work.

PawelLipski (approved):

PR #1844, opened 2026-07-02
ci: Exercise `bazel build //...` to catch wildcard-only breakages

PawelLipski:
# Based on PR #1843

## Chain of upstream PRs as of 2026-07-02

* PR #1843:
  `master` ← `fix/bazel-analysis`

  * **PR #1844 (THIS ONE)**:
    `fix/bazel-analysis` ← `chore/ci-analysis-check`

### Description

Extend CI to actually exercise the whole repo so wildcard-only breakages are caught going forward. Changes to `.bazelci/presubmit.yml`:

- Add an `analyze_all` task running `bazel build //... --nobuild` — loading + analysis of the whole repo with no execution, a cheap tripwire for regressions like the protoc alias visibility error fixed in #1843.
- Expand the `ubuntu2004` and `macos` `build_targets` from `//test/...` to `//...`, so the full repo is actually built (not just the test tree).
- Broaden the `ubuntu2004` and `macos` `test_targets` the same way, from `//test/...` to `//...`, so tests outside `//test/...` actually run: the `//third_party/dependency_analyzer/...` suite, the `//src/java` `worker`/`compileoptions` unit tests, and `//tools:lint_check`.

All three wildcards exclude `//test_expect_failure/...`, whose targets are negative tests meant to fail (some already at analysis) and are exercised explicitly by the shell scripts. Other conditional targets (`manual_test/*`, the empty-stats-file genrule) are tagged `manual` in #1843 and are therefore already skipped by wildcard expansion.

### Motivation

The whole class of bugs fixed in #1843 slipped through because CI only ever built and tested `//test/...` — a plain `bazel build //...` / `bazel test //...` was never run anywhere, so a wildcard-only regression (e.g. a visibility/alias change) produced green CI while the wildcard build was broken, and real tests outside `//test/...` were never executed. This change closes that gap by making the full repo wildcard-clean under CI. Depends on #1843, which makes `bazel build //...` pass in the first place.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -6,6 +6,23 @@ matrix:
   bcr_bazel: [7.x, 8.x, 9.x, rolling, last_green]
 
 tasks:
+  analyze_all:
+    # Loading + analysis of the whole repo (no execution). This is the cheap
+    # tripwire for wildcard-only breakages (e.g. a visibility/alias regression)
+    # that never surface when CI only builds //test/..., since a plain
+    # `bazel build //...` is not otherwise exercised anywhere. test_expect_failure/
+    # is excluded: its targets are negative tests meant to fail (some already at
+    # analysis) and are exercised explicitly by the shell scripts instead.
+    name: "bazel build //... --nobuild"
+    platform: ubuntu2004
+    shell_commands:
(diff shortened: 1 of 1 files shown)


WojciechMazur (approved):

PawelLipski:
Merged.

Issue #1819, opened 2026-03-05
scala_macro_library incorrectly builds ijar for Scala 2.x, causing ClassFormatError during macro expansion


tylerprete:
## Summary

`scala_macro_library` targets fail to compile for Scala 2.x with:

```
java.lang.ClassFormatError: Absent Code attribute in method that is not native or abstract
    in class file com/example/MyMacro$
...
Build failure during macro expansion. You may have declared a target containing a macro
as a `scala_library` target instead of a `scala_macro_library` target
```

The target *is* declared as `scala_macro_library`. Scala 3 targets are unaffected.

## Root cause

`scala_macro_library` uses `phase_compile_library`, which only sets `buildijar` when the rule defines a `build_ijar` attr:

```python
# phase_compile.bzl
def phase_compile_library(ctx, p):
    args = { ... }
    if hasattr(ctx.attr, "build_ijar"):
        args["buildijar"] = ctx.attr.build_ijar
    return _phase_compile_default(ctx, p, struct(**args))
```

`_scala_macro_library_attrs` doesn't define `build_ijar`, so the check fails. `_phase_compile_default` then falls back to:

```python
buildijar_default_value = toolchain.scala_version.startswith("2.")
```

For Scala 2.x this is `True`, so macro libraries get ijar'd. Ijar strips method bodies, and when the Scala compiler loads the macro class to execute at compile time, it fails with `ClassFormatError: Absent Code attribute`.

Note: https://github.com/bazel-contrib/rules_scala/pull/1753 added `build_ijar` as a configurable attr for `scala_library` but did not extend it to `scala_macro_library`.

## Suggested fix

Add `build_ijar` to `_scala_macro_library_attrs` with `default = False`:

```python
_scala_macro_library_attrs = {
    "main_class": attr.string(),
    "exports": attr.label_list(allow_files = False),
    "build_ijar": attr.bool(default = False),  # macros must not be ijar'd
}
```

## Workaround

Use `make_scala_macro_library` to inject the missing attr (note: uses a private API):

```python
load(
    "@io_bazel_rules_scala//scala/private:rules/scala_library.bzl",
    "make_scala_macro_library",
)

scala_macro_library = make_scala_macro_library({
    "phase_providers": [],
    "attrs": {"build_ijar": attr.bool(default = False)},
})
```

mbland:
Sorry for the long delay. Would you be willing to create a pull request with the fix and a test?

PawelLipski:
[...]

PR #1846, opened 2026-07-06
test: Migrate `test_macros.sh` to `bazel test` targets

PawelLipski:
## Tree of downstream PRs as of 2026-07-07

* **PR #1846 (THIS ONE)**:
  `master` ← `refactor/test-macros-as-bazel-test`

    * PR #1848:
      `refactor/test-macros-as-bazel-test` ← `refactor/test-scalacopts-invalid-as-bazel-test`

      * PR #1849:
        `refactor/test-scalacopts-invalid-as-bazel-test` ← `refactor/test-expect-failure-simple-as-bazel-test`

## Summary

Migrates the bespoke `test/shell/test_macros.sh` integration script to Bazel-native test targets under `//test/macros`, so the macro checks run as part of `bazel test //...` rather than only via the standalone `test_rules_scala.sh` harness. This is the first step toward moving as many shell-based tests as possible into `bazel test`.

Since Bazel has no native "expect this target to fail to build" rule, this PR also introduces the reusable machinery for that (described below), with the macro tests as its first consumer.

## Reusable "expect build failure" machinery

> [!NOTE]
> This machinery arguably deserves its own PR, but it's bundled here with its first consumer so it can be reviewed in context — a helper is much easier to judge next to a real use than in isolation. The downstream PRs (#1848, #1849, ...) reuse it as-is.

Introduced under a neutral `//test/expect_build_failure` package:

- **`nested_bazel.sh`** — shared library holding the nested-`bazel` plumbing: locating the real source workspace, reusing the parent's repository cache, restoring the real `$HOME`, and running the inner build under its own output base. Sourced (not re-implemented) by the assertion scripts.
- **`expect_build_failure.sh`** — asserts that a target fails to build. Flag-based CLI: `--target`, repeatable `--build-arg` (forwarded to the nested build, e.g. `--repo_env=SCALA_VERSION=...`), and repeatable `--expect-file` / `--reject-file` assertions on the build output. Messages are passed as files (not inline args) because Bazel Bourne-tokenizes `sh_test` `args`, which would split messages containing spaces.
- **`expect_build_failure_test`** (`expect_build_failure.bzl`) — a macro wrapping `sh_test` so callers declare intent (`target`, `expect`, `reject`, `build_args`) instead of hand-assembling raw script args, `$(rootpath ...)` expansions, runfiles, and boilerplate tags.

## Macro tests (`//test/macros`)

- Positive cases (`correct-macro-user`, `macro-with-dependencies-user`, `macro-export-user`) become `@bazel_skylib//rules:build_test` targets.
- The negative case (`incorrect-macro-user` must fail during macro expansion) becomes an `expect_build_failure_test` asserting the expected error from `incorrect_macro_user_expected_error.txt`.
- `scala_macro_library` must not emit an ijar (#1819): `expect_no_ijar.sh` — a second consumer of `nested_bazel.sh` — builds the macro library per Scala 2.x version and asserts the ijar is absent. (This checks build *success* plus a filesystem condition, so it fits neither `expect_build_failure_test` nor `build_test`.)
- Removes `test/shell/test_macros.sh` and its sourcing from `test_rules_scala.sh`.

### Why a nested `bazel build`?

Verifying a **compile-time** macro-expansion failure can't be expressed with `analysistest` (analysis-phase only) or `build_test` (success only), so the helper shells out to a nested `bazel build`. That nested invocation:

- uses its own output base, since the parent `bazel test` server holds the build lock;
- shares the parent's repository cache so most external repos resolve locally;
- runs with the real `$HOME` so it reads the user's `~/.bazelrc` (e.g. a corporate Maven download-proxy config), which `bazel test` would otherwise scrub.

Such tests are tagged `local` + `requires-network` accordingly.

## Test plan

- [x] `bazel test //test/macros:all` passes locally
- [x] CI green on `bazel test //...`
- [x] CI green on the remaining `./test_rules_scala.sh` shell suite


test/expect_build_failure/expect_build_failure.sh
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/env bash
+#
+# Verifies that `bazel build` of a target fails, and that the combined build
+# output contains every expected message and none of the rejected messages.
+#
+# Bazel has no native "expect this target to fail to build" rule, so this runs a
+# *nested* `bazel build` from inside an `sh_test`: nested in the sense of a
+# `bazel` invocation running inside the action of a `bazel test`, not nested on
+# disk. The nested invocation uses its own output base (the parent server holds
+# an exclusive lock on its own) while sharing the parent's repository and disk
```
PawelLipski: TBD if true about disk cache

test/macros/BUILD
```diff
@@ -60,3 +65,51 @@ scala_library(
...
+)
+
+# Regression test for https://github.com/bazel-contrib/rules_scala/issues/1819:
+# scala_macro_library must not emit an ijar, since ijar strips the macro
+# implementations that consumers need at compile time. Checked across the
+# Scala 2.x versions where ijar interpolation applies.
+sh_test(
+    name = "scala_macro_library_no_ijar_test",
+    size = "large",
+    srcs = ["expect_no_ijar.sh"],
```
PawelLipski, line 103: Can't be implemented in terms of the generic `expect_build_failure_test` (nor `build_test`) as we're checking for build success && missing ijar, not a failure

test/expect_build_failure/expect_build_failure.bzl
```diff
@@ -0,0 +1,65 @@
...
+    for reject_file in reject:
+        args += ["--reject-file", "$(rootpath %s)" % reject_file]
+
+    # The helper resolves the workspace root via `MODULE.bazel` in runfiles, so
+    # it must always be present alongside any expect/reject message files.
+    data = ["//:MODULE.bazel"] + expect + reject + kwargs.pop("data", [])
+    deduped_data = []
+    for entry in data:
+        if entry not in deduped_data:
+            deduped_data.append(entry)
```
PawelLipski, line 68: Paranoia check, Bazel fails on duplicated entries in `data` (and probably all in `label_list`s)

WojciechMazur (approved):

PR #1864, opened 2026-07-15
refactor: Migrate scala_library_suite build-failure test to Bazel

PawelLipski:
# Based on PR #1865

## Chain of upstream PRs as of 2026-07-17

* PR #1865:
  `master` ← `fix/scala-library-suite-aggregate-tags`

  * **PR #1864 (THIS ONE)**:
    `fix/scala-library-suite-aggregate-tags` ← `refactor/analysis-build-failures-as-bazel-test`

### Description

Migrates the `scala_library_suite` build-failure case out of the shell-test scripts and the skipped `test_expect_failure/` tree into `test/scala_library_suite`, driven by `expect_build_failure_test` (replacing the `test_scala_library_suite` case in `test/shell/test_scala_library.sh`). Also reworks the misnamed `library_suite_dep_on_children` fixture so it actually tests inter-child dependency rejection (`DependsOnSibling.scala` → `Sibling.scala`, expecting `not found: value Sibling`), renaming the old invalid-source case to `library_suite_with_invalid_source`.

The `tags`-forwarding fix this relies on now lives in the base PR #1865. The resource-strip-prefix migration that used to be part of this PR was dropped: it landed independently on `master` via #1874 (into `test/mismatching_resource_strip_prefix`).

### Motivation

Part of migrating bespoke shell tests to native `bazel test`, so `bazel test //...` covers these negative cases and `test_expect_failure/` can eventually be dropped.

PR #1865, opened 2026-07-15
fix: Propagate `tags` from `scala_library_suite` to its aggregate target

PawelLipski:
## Tree of downstream PRs as of 2026-07-17

* **PR #1865 (THIS ONE)**:
  `master` ← `fix/scala-library-suite-aggregate-tags`

    * PR #1864:
      `fix/scala-library-suite-aggregate-tags` ← `refactor/analysis-build-failures-as-bazel-test`

### Description

`scala_library_suite` compiles each source into its own child `scala_library` and defines an aggregate target (of the same `name`) that `deps`/`exports` all the children. The caller's `tags` reached the children (via `**kwargs`) but not the aggregate. This forwards them to the aggregate too.

It's a no-op for callers that pass no `tags` (the aggregate keeps its previous empty tag set); the only observable change is for callers that do pass `tags`.

### Motivation

`manual` only removes a target from wildcard *expansion*, not from being built as a dependency. So an untagged aggregate stayed in `//...` and pulled in every child, which made `tags = ["manual"]` on a suite effectively a no-op — the suite and all its children built under a wildcard anyway. The same applied to `--build_tag_filters` and other tags, which never reached the aggregate. Propagating `tags` makes them behave as a user would expect (and lets a suite be excluded from wildcard builds).

### DIFF
--- a/scala/private/rules/scala_library.bzl
+++ b/scala/private/rules/scala_library.bzl
@@ -145,6 +145,10 @@ def scala_library_suite(
         visibility = visibility,
         exports = exports + ts,
         deps = ts,
+        # Forward tags to the aggregate target too (children already get them via
+        # **kwargs), so e.g. `tags = ["manual"]` can keep the whole suite -- which
+        # depends on every child -- out of wildcard builds.
+        tags = kwargs.get("tags", []),
     )
 
 ##


WojciechMazur (approved):
LGTM, makes sense to me, shouldn't negatively affect the downstream users

PawelLipski:
> LGTM, makes sense to me, shouldn't negatively affect the downstream users

For the record, we'll still check this change on a few downstream projects - a new group of tests (probably manually executed in CI, TBD) is WIP by @bor-p-s

PawelLipski:
Tested against 3rd party projects (downstream tests recently included in the CI: https://github.com/bazel-contrib/rules_scala/pull/1868)

PR #1871, opened 2026-07-21
refactor: Migrate test_unused_dependency.sh to Bazel

bor-p-s:
## Summary

Continues retiring test_expect_failure/ in favor of native Bazel tests.

Moves test_expect_failure/unused_dependency_checker (including filtering/)
and test_expect_failure/plus_one_deps/with_unused_deps into test/, driven by
expect_build_failure_test and a new expect_build_success_test instead of
test/shell/test_unused_dependency.sh's ten cases. That file is now empty, so
this deletes it and its test_rules_scala.sh entry.

expect_build_success_test fills the one gap the existing macros left:
asserting a bazel build succeeds while checking its output for a message.
Its warmup_build_args guards against a real flake: a successful nested build
can hit the shared output base's on-disk action cache on a repeat run and
silently skip recompilation, printing no warning even though nothing broke.


PawelLipski (commented):
Consider splitting this PR into smaller ones (e.g. one for plus_one, one for unused deps)

bor-p-s:
Splitting into #1872 (unused_dependency_checker) and #1873 (plus_one_deps, stacked on #1872) per review feedback.

PR #1872, opened 2026-07-21
refactor: Migrate unused_dependency_checker tests to Bazel

bor-p-s:
## Summary

Continues retiring test_expect_failure/ in favor of native Bazel tests.

Moves test_expect_failure/unused_dependency_checker (including filtering/)
into test/unused_dependency_checker, driven by expect_build_failure_test and
a new expect_build_success_test instead of six cases in
test/shell/test_unused_dependency.sh (the file's remaining plus_one_deps
cases are covered by a follow-up PR).

expect_build_success_test fills the one gap the existing macros left:
asserting a bazel build succeeds while checking its output for a message.
Its warmup_build_args guards against a real flake: a successful nested build
can hit the shared output base's on-disk action cache on a repeat run and
silently skip recompilation, printing no warning even though nothing broke.


test/expect_build_failure/nested_bazel.sh
```diff
@@ -144,5 +144,13 @@ nested_bazel_run() {
...
-  "${cmd[@]}"
+  # On Windows this script runs under MSYS2 bash, which auto-converts
+  # POSIX-path-looking argv entries before exec'ing a native Windows binary
+  # like bazel.exe. A bare `//pkg:target` label gets corrupted by this (observed:
+  # arrived at Bazel as `/pkg:target`), but `--output_base=/tmp/...` above
+  # *needs* the same conversion to become a real Windows path -- so disabling
+  # it outright (tried first) broke that instead. MSYS2_ARG_CONV_EXCL only
+  # excludes args matching a given prefix, leaving everything else (like
+  # /tmp/...) converted as before.
+  MSYS2_ARG_CONV_EXCL='//' "${cmd[@]}"
```
PawelLipski, line 155: Hmmm is this needed specifically on this PR? or is it an unrelated fix?

test/expect_build_failure/expect_build_failure.sh
```diff
@@ -116,6 +131,10 @@ _resolve_message_file() {
 
 nested_bazel_setup "rules_scala_expect_build_failure_output_base"
 
+if [[ "${#warmup_bazel_args[@]}" -gt 0 ]]; then
+  nested_bazel_run build ${warmup_bazel_args[@]+"${warmup_bazel_args[@]}"} "${target}" >/dev/null 2>&1 || true
```
PawelLipski: This can't be simulated with `bazel shutdown` to kill the server right?
Or some `--no-build-cache`-like flag?

test/expect_build_failure/expect_build_failure.sh
```diff
@@ -116,6 +116,19 @@ _resolve_message_file() {
 
 nested_bazel_setup "rules_scala_expect_build_failure_output_base"
 
+# The nested output base's on-disk action cache persists across separate
+# invocations of this same test: a prior run that already built `target` with
+# these exact args can leave a cache hit that silently skips recompilation --
+# fine when we only check the exit code, but a real problem when asserting on
+# output content (below), since a cache hit reprints nothing. `bazel clean`
```
PawelLipski: Nice, much better (the original approach looked pretty brittle TBH)

test/expect_build_failure/nested_bazel.sh
bor-p-s: Yep, this is specific. `mode_warn_test` here uses `target = "//test:UnusedDependencyCheckerWarn"`, a bare //pkg:target label with no further path segments. That's exactly the shape MSYS2 was mangling into `/test:...` on the Windows job (Bazel rejected it as an invalid package name in CI run). None of the other already-merged expect_build_failure_test targets hit this because they resolve via package-relative labels with multiple path segments (e.g. //test/transitive/scala_to_scala:d), which MSYS apparently converts differently/harmlessly AFAICS.

test/expect_build_failure/expect_build_failure.sh
bor-p-s: Good catch, actually `bazel clean` works fine, let me remove all that.

test/expect_build_failure/expect_build_failure.sh
bor-p-s: 💯

test/expect_build_failure/expect_build_failure.sh
```diff
@@ -116,6 +116,19 @@ _resolve_message_file() {
...
+# (no --expunge) clears the action cache and build outputs without touching
+# the fetched external repos, so it's cheap and doesn't defeat the shared
+# output base's purpose (see nested_bazel_setup).
+if [[ "${expect_success}" == "true" ]] &&
+  { [[ "${#expect_files[@]}" -gt 0 ]] || [[ "${#reject_files[@]}" -gt 0 ]]; }; then
```
PawelLipski: Why check for `expect_files` and `reject_files` specifically here? 🤔 `bazel clean` not needed when neither is present?

PawelLipski: Maybe control this `clean` behavior with an explicit flag instead?

test/BUILD
```diff
@@ -187,6 +187,12 @@ scala_library(
     ],
 )
 
+# The warn-mode test for this fixture lives in //test/unused_dependency_checker
+# (not here) so that this package does not load @rules_shell (a dev-only
+# dependency, invisible when rules_scala is consumed as a module): //test is
+# built by test_dependency_versions.sh as an external-module consumer (e.g.
```
PawelLipski: Hmm test_dependency_versions.sh will probably be gone soon (scripts will be migrated in BUILD files)

In fact the entire comment feels over-detailed/redundant

test/expect_build_failure/expect_build_failure.sh
bor-p-s: done

test/BUILD
bor-p-s: updated

PawelLipski:
In general: avoid merge commits when dealing with stacked branches (this can get messy very quickly), prefer [linear history](https://slides.com/plipski/git-linear-history#/1)

PawelLipski (approved):

PR #1873, opened 2026-07-21
refactor: Migrate plus_one_deps unused-dep tests to Bazel

bor-p-s:
Stacked on #1872 -- merge that first.

## Summary

Moves test_expect_failure/plus_one_deps/with_unused_deps into
test/plus_one_deps, driven by expect_build_failure_test/expect_build_success_test
instead of test/shell/test_unused_dependency.sh's five ast-analyzer cases.

The scala_test-doesn't-need-scalatest-dep case moves into
test/unused_dependency_checker (referencing its fixture in test/scala_test)
rather than test/scala_test itself, so that package doesn't load @rules_shell
-- matching //test/jmh's precedent.

This was the last content in test/shell/test_unused_dependency.sh, so this
deletes that file and its test_rules_scala.sh entry.


test/plus_one_deps/with_unused_deps/BUILD
```diff
@@ -0,0 +1,53 @@
...
+    srcs = ["B.scala"],
+    deps = [":c"],
+)
+
+scala_library(
+    name = "c",
+    srcs = ["C.scala"],
+)
+
+expect_build_failure_test(
```
PawelLipski, line 30: Do we want to check for error messages here?

PawelLipski: (and in the tests below)

bor-p-s: done!

PawelLipski (approved):

PR #1875, opened 2026-07-23
refactor: Migrate scala_library_suite build-failure test to Bazel

bor-p-s:
Replaces the one-line `action_should_fail build` case in test_scala_library.sh with
an expect_build_failure_test, and asserts the scalac parse error is surfaced (the
shell test only checked the build failed at all).

### DIFF
--- a/test/scala_library_suite/BUILD
+++ b/test/scala_library_suite/BUILD
@@ -1,0 +1,11 @@
+# Migrated from test/shell/test_scala_library.sh.
+
+load("//test/expect_build_failure:expect_build_failure.bzl", "expect_build_failure_test")
+
+package(default_testonly = 1)
+
+expect_build_failure_test(
+    name = "scala_library_suite",
+    expect = ["expected_invalid_scala.txt"],
+    target = "//test_expect_failure/scala_library_suite:library_suite_dep_on_children",
+)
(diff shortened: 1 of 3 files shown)


test/scala_library_suite/expected_invalid_scala.txt
```diff
@@ -0,0 +1 @@
+expected class or object definition
```
PawelLipski: Hmm the error is `expected class or object definition` even though the test is named `library_suite_dep_on_children`? 🤔 this doesn't sound like an expected error for such a case (?)

PawelLipski (approved):

test/scala_library_suite/BUILD
```diff
@@ -0,0 +1,8 @@
+load("//test/expect_build_failure:expect_build_failure.bzl", "expect_build_failure_test")
+
+package(default_testonly = 1)
+
+expect_build_failure_test(
+    name = "scala_library_suite",
+    target = "//test_expect_failure/scala_library_suite:library_suite_dep_on_children",
```
PawelLipski, line 7: Pls move the target here to the package + mark it manual

PawelLipski (commented):
Actually, I've got a PR for this change already 😅 https://github.com/bazel-contrib/rules_scala/pull/1864

So you can close this one. Mine has a bit more test cases added, waits for #1865 to merge tho - which is better merged after downstream tests land, in turn

bor-p-s:
Closed in favor of #1864

PR #1877, opened 2026-07-23
refactor: Migrate proto_source_root build test to Bazel

bor-p-s:
Replaces the `bazel build //test_expect_failure/proto_source_root/... --strict_proto_deps=off`
line in test_rules_scala.sh with an expect_build_success_test running the same build.


test/proto_source_root/BUILD
```diff
@@ -0,0 +1,9 @@
+load("//test/expect_build_failure:expect_build_failure.bzl", "expect_build_success_test")
+
+package(default_testonly = 1)
+
+expect_build_success_test(
+    name = "proto_source_root",
+    build_args = ["--strict_proto_deps=off"],
```
PawelLipski: Maybe add a test ensuring that this would fail without the flag, like:

```starlark

expect_build_failure_test(
    name = "proto_source_root_...",
    expect = "<some message file>"
    target = "//test_expect_failure/proto_source_root/...",
)
```

(at least I'd expect that the build only passes with `--strict_proto_deps=off`; also the expected error message will make it clearer why the flag is needed)

test/proto_source_root/BUILD
```diff
@@ -0,0 +1,9 @@
...
+    target = "//test_expect_failure/proto_source_root/...",
```
PawelLipski: Also, I'd move the target out of `test_expect_failure` to this package, and mark it `manual` (this is convention in other similar PRs; we'd like to get rid of `test_expect_failure` eventually, and keep targets close to their tests instead)

test/proto_source_root/BUILD
bor-p-s: AFAICS `--strict_proto_deps=off` isn't needed here anymore.

It was added in 2018 (#506), together with rules_scala's old custom "flatten transitive proto paths" code, to work around strict proto-dep checking failing on `strip_import_prefix`. That old code was removed in 2021 (#1213, the FileDescriptorSets rewrite), so the flag has done nothing since. The target now builds fine at the default `--strict_proto_deps=error`. 
`bazel build //test/proto_source_root/user:user_scala` passes both with and without the flag, so a "fails without the flag" test can't work, nothing fails.
- The native `--strict_proto_deps` is also being removed in favor of the Starlark `--@protobuf//bazel/flags:strict_proto_deps` / proto toolchain resolution.

So I'd drop `--strict_proto_deps=off` and keep this as a plain build test that checks
the `strip_import_prefix=""` proto still builds. WDYT?

test/proto_source_root/BUILD
bor-p-s: done

test/proto_source_root/BUILD
PawelLipski: Okay! LGTM

PawelLipski (approved):

PR #1878, opened 2026-07-23
refactor: Migrate persistent-worker tests to Bazel

bor-p-s:
Replaces test/shell/test_persistent_worker.sh with expect_build_success_test /
expect_build_failure_test running the same `--strategy=Scalac=worker` builds of the
same targets.

Note: `worker_sandboxing = True` reproduces the shell file's non-Windows-only
`--worker_sandboxing` (Bazel has no worker sandboxing on Windows) — the existing
`worker_sandboxing` param in expect_build_failure.bzl expands it into a Windows-excluding
`select`.


test/persistent_worker/BUILD
```diff
@@ -0,0 +1,21 @@
...
+    name = "persistent_worker_success",
+    build_args = ["--strategy=Scalac=worker"],
+    target = "//test:ScalaBinary",
+    worker_sandboxing = True,
+)
+
+expect_build_failure_test(
+    name = "persistent_worker_failure",
+    build_args = ["--strategy=Scalac=worker"],
+    target = "//test_expect_failure/diagnostics_reporter:error_file",
```
PawelLipski: Pls move the target here to the package + mark it manual

test/persistent_worker/BUILD
```diff
@@ -0,0 +1,21 @@
+load(
+    "//test/expect_build_failure:expect_build_failure.bzl",
+    "expect_build_failure_test",
+    "expect_build_success_test",
+)
+
+package(default_testonly = 1)
+
+expect_build_success_test(
+    name = "persistent_worker_success",
```
PawelLipski, line 19: For clarity, these tests would pass/fail the same without `build_args = ["--strategy=Scalac=worker"],` right?

bor-p-s: Good question. Yes, pass/fail is the same without it, Scalac already runs as a worker by default (a plain build shows worker processes). I kept `--strategy=Scalac=worker` to make the intent explicit and to make sure `--worker_sandboxing` actually applies (it only affects worker-strategy actions), so the test won't quietly stop exercising the worker if the default ever changes.

test/persistent_worker/BUILD
bor-p-s: Moved it: the failure test now builds a small local `:broken` scala_library (manual) instead of the shared `diagnostics_reporter:error_file`, so this package no longer depends on an unmigrated fixture. The success target stays `//test:ScalaBinary`.

PawelLipski (approved):

Issue #1867, opened 2026-07-16
Migrate the tests out of `test_*.sh` scripts to regular bazel test rules


PawelLipski:
Will involve running `bazel` from inside the tests at times (in the tests where we're asserting that certain constructs will **fail**).
See recent PRs: https://github.com/bazel-contrib/rules_scala/pull/1863 + its entire stack (#1864, #1856 etc.)

PawelLipski:
Let's also make sure that `test_expect_failure` failure top-level directory is gone, let's keep the failing targets as `manual` next to their tests)

PR #1880, opened 2026-07-24
refactor: Migrate compiler-dependency-tracker dep-checking tests to Bazel

bor-p-s:
### Description

Migrates the dependency-checking failure cases from test/shell/test_compiler_dependency_tracking.sh to Bazel-native expect_build_failure_test: unused dep, missing compile dep, strict dep, the Scala 3 unused-dep variant, and the "no unused-dep warning when the source is broken" case. Each asserts the exact buildozer fix suggestion (or, for the broken case, that no "remove deps" suggestion is printed).

The A-F fixture and the ast_plus_error / ast_plus_warn toolchains move into the test package, tagged manual so a wildcard build skips them. This package does not use package(default_testonly): a toolchain registered via --extra_toolchains is resolved by non-test infra (scala_compile_classpath), which cannot depend on a testonly target.

First of a 3-PR stack for this file. The sdeps and given-import cases still run from the shell script (pointed at the moved toolchains) until the follow-up PRs.

### Motivation

Part of #1867 (move tests out of test_*.sh into regular bazel test rules).


bor-p-s:
scala_3_fails_for_unused_dep pins Scala 3.3.8 (LTS), not a newer 3.x, on purpose.

As a nested `bazel test //...` target, its inner `bazel build` runs under the CI agent's local JDK, which is Java 11 on the macOS runner (the `bazel test //...` tasks overwrite .bazelrc, so the committed Java 17 setting does not reach the inner build). Scala 3.8.4's compiler is Java 17 bytecode and cannot start under Java 11, so the inner build fails there.

Two ways to fix it:
- Pin the inner build to Java 17 (all four java_*_version flags). This works, but it keeps the heavy 3.8.4 compiler in the inner build: it measured ~840s on the Windows agent, close to the 900s test timeout, so on a loaded agent it risks a flaky timeout. The repo already sees this on other slow tests (#1884, #1885).
- Use the Scala 3 LTS (3.3.8), whose compiler runs on Java 8+. This is a one-line change, needs no JDK pinning, stays light, and the bug under test is not version-specific.

Went with the LTS.

PawelLipski (approved):

PR #1881, opened 2026-07-24
refactor: Migrate compiler-dependency-tracker sdeps test to Bazel

bor-p-s:
### Description

Migrates the sdeps case to a Bazel-native expect_test_success_test. The sdeps fixture moves into the test package; sdeps_test is tagged manual (its .sdeps resources are only emitted by an ast-plus toolchain) and driven by the new :sdeps target under ast_plus_warn.

SdepsTest.scala loads the .sdeps resources by workspace path, so its prefix constant changes from /test_expect_failure/... to /test/... to match the new location.

### Motivation

Part of #1867. Stacked on the dep-checking PR, which moved the shared ast_plus_warn toolchain into the test package.


test/compiler_dependency_tracker/sdeps/BUILD
```diff
@@ -1,10 +1,14 @@
 load("//scala:scala.bzl", "scala_library", "scala_specs2_junit_test")
+load("//test/expect_build_failure:expect_build_failure.bzl", "expect_test_success_test")
 
+# Both carry an unused dep, which only warns under ast_plus_warn and errors under
+# the default toolchain, so a wildcard build must skip them.
```
PawelLipski, line 5: Maybe also add a test to check that the failure under the default toolchain actually happens (?)

bor-p-s: Added it on `only_scala_sources`, asserting the buildozer line so it cannot pass on an unrelated failure like a compile error.
One target rather than both: `mixed_sources` carries the same unused dep, and what differs (the Java source) is about `.sdeps` emission, which `sdeps_test` covers.
Checked both ways: dropping `scala_unused_dep` from the fixture turns it red, and it still passes in the runs that repeat `test/...` under another toolchain, since the nested build uses its own flags and always sees the default from `MODULE.bazel`.

PawelLipski (approved):

Issue #1885, opened 2026-07-24
`//test/jmh/reports_failure:jmh_reports_failure_build_test` sometimes times out on Ubuntu


PawelLipski:
Similar to #1884, but it's a different target.
See [this build](https://buildkite.com/bazel/rules-scala-scala/builds/6422/list?sid=019f9425-3e35-49f6-9e9a-f8ef06d4e5a0&tab=output):

<img width="1725" height="796" alt="Image" src="https://github.com/user-attachments/assets/2003616c-ae18-4358-a333-6a6ef8338696" />

bor-p-s:
The recent timeout here (build 6422, 2026-07-24) matches the lock-contention pattern described in #1894: `jmh_reports_failure_build_test` sat behind the shared nested-bazel output base lock along with the other `expect_build_failure_test` targets, and the wait counted toward its own 900s timeout.

That was fixed for the whole `expect_*` test family by #1939 (3 lanes instead of one shared lock), merged 2026-09-04. No repeat of this timeout on this target since then in the builds I checked. I'll leave this open for a bit longer to confirm across a few more CI runs before closing it as fixed by #1939.

bor-p-s:
Checked every CI run from the last two days (builds #6929-#6940 on bazel/rules-scala-scala, covering master plus every active PR branch): no TIMEOUT or hard_failed on any nested-bazel/lane-lock test. `jmh_reports_failure_build_test` itself passed in build #6937 (137.2s, cached), well under the 900s limit.

No repeat of this timeout since #1939 merged. Closing as fixed by #1939.

Issue #1884, opened 2026-07-24
`//test/scala_junit_test:specs2_exception_terminates_without_timeout` sometimes times out on Windows (15 minutes)


PawelLipski:
Pls check [this build](https://buildkite.com/bazel/rules-scala-scala/builds/6420/list?jid=019f93ec-9529-4475-942e-a7075c7f1152&tab=output), which has been retried and succeeded:

<img width="1719" height="796" alt="Image" src="https://github.com/user-attachments/assets/5a0a0856-0c3c-480d-a6ed-c4daea2c322d" />

It's took 15 minutes, the timeout should probably be decreased to a mere minute (or even less), to avoid blocking everything.

And crucially, either fix on Windows or exclude on Windows if unfixable (hopefully not).

PawelLipski:
Even worse, in [this build](https://buildkite.com/bazel/rules-scala-scala/builds/6421/list?jid=019f93fc-225e-48c1-96b4-04498f67fe7b&tab=output) it failed twice :/

PR #1890, opened 2026-07-27
TEMP: probe K=4 parallel nested output bases (throwaway, do not merge)

bor-p-s:
Throwaway, do not merge. Probing the fix for the nested-test timeouts (#1884, #1885): the ~45 `expect_build_*_test` targets all share one nested output base, so under `bazel test //...` they serialize on a single lock and the slow ones hit the 900s timeout. This spreads them across 4 output bases (keyed by a hash of the test target) with a shared `--disk_cache`, so they run in parallel. Locally a 13-test cross-package set went 730s serial to 80s. Watching whether CI clears the timeouts and how much /tmp it uses (~1.4G per lane).


bor-p-s:
Throwaway probe, closing. The chosen fix (exclusive) merged in #1894, and the lanes approach is written up there under Alternatives considered.

PR #1891, opened 2026-07-27
fix: Run nested expect-test builds in parallel output bases to avoid timeouts

bor-p-s:
### Description

Every `expect_build_*_test` / `expect_test_*_test` runs a nested `bazel build/test`, and they all shared one nested output base under /tmp. Under `bazel test //...` those nested invocations serialize on that base's lock (you can see "Another command holds the client lock ... Waiting" in the test logs), so the slow ones queue behind each other and can hit the 900s test timeout. That looks like the cause of the flaky timeouts in #1884 and #1885.

This spreads the nested invocations across a small number of output bases, keyed by a stable hash of the test target, so they run in parallel, plus a shared `--disk_cache` so the lanes reuse each other's compiled actions instead of each rebuilding the toolchain.

I went with two lanes. In the CI runs I checked, the tests that had been timing out passed (`scala_junit_test:specs2_exception_terminates_without_timeout`, `jmh/reports_failure:jmh_reports_failure_build_test`, `scala_test_testfilter:testfilter_class_selection`). But these timeouts are flaky, so I can't claim two lanes always clears them; if they come back, raising the lane count is a one-line change. Each lane is a full nested output base (roughly 1.4G in /tmp on the agents I saw, mostly JDK toolchains), so I kept the count low to bound disk.

### Motivation

Meant to address #1884 and #1885. Given how flaky these timeouts are, I'd rather confirm over a few runs before we call them closed.

### DIFF
--- a/test/expect_build_failure/nested_bazel.sh
+++ b/test/expect_build_failure/nested_bazel.sh
@@ -98,7 +98,19 @@ nested_bazel_setup() {
   local parent_output_base
   parent_output_base="$(_nested_bazel_find_parent_output_base)"
 
-  _nested_bazel_output_base="/tmp/${output_base_name}"
+  # Under `bazel test //...` these nested `bazel` invocations would all contend
+  # for a single output base's lock and serialize. Spread them across a few bases,
+  # keyed by a stable hash of the test target, so they run in parallel.
+  local lanes=2
+  local lane=0
+  if [[ -n "${TEST_TARGET:-}" ]]; then
+    # Pick the lane by a cheap stable hash of the target: cksum prints
+    # "<CRC32> <byte count>", so take the CRC and reduce it modulo the lane count.
+    local target_crc
+    target_crc="$(printf '%s' "${TEST_TARGET}" | cksum | cut -d' ' -f1)"
+    lane=$((target_crc % lanes))
+  fi
+  _nested_bazel_output_base="/tmp/${output_base_name}_lane${lane}"
   mkdir -p "${_nested_bazel_output_base}"
   cd "${NESTED_BAZEL_WORKSPACE}"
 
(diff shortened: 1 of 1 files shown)


PawelLipski (commented):
Actually... are you sure this solution is worse than the #1894?

The execution times are clearly better:

Image

vs

Image

The risk of timeouts still lingers... but we can add further lanes and/or fall back to exclusive tag if this becomes unmanagable otherwise.

test/expect_build_failure/nested_bazel.sh
```diff
@@ -98,7 +98,19 @@ nested_bazel_setup() {
   local parent_output_base
   parent_output_base="$(_nested_bazel_find_parent_output_base)"
 
-  _nested_bazel_output_base="/tmp/${output_base_name}"
+  # Under `bazel test //...` these nested `bazel` invocations would all contend
+  # for a single output base's lock and serialize. Spread them across a few bases,
+  # keyed by a stable hash of the test target, so they run in parallel.
```
WojciechMazur, line 103: I wonder should we also add some custom resources and add them to the .bzl macros we're using.  I think currently we would run all tests at once anyway, so with increased number of tests the issue might be reintroduced again.

We should be able to limit concurrency of these tests executions with custom `--local_extra_resources=bazel_instance=2` to .bazelrc and adding   `tags = ["resources:bazel_instance:1"]`  to https://github.com/bazel-contrib/rules_scala/blob/2f175c60436b4b20738ff8132872ced4d5bb6132/test/expect_build_failure/expect_build_failure.bzl#L135 so bazel can control invocations based on the available resources. 

I think that in such case the timeout would be counted since the moment the resource (bazel instance) was acquired, so even if we increase the number of tests we should not have a new timeout failures

PawelLipski: Nice, I wasn't aware of this mechanism, +1

bor-p-s:
Closing in favor of #1894 (exclusive), which merged as the fix. This lanes approach and its tradeoffs are captured in that PR under Alternatives considered. If we pick up the performance follow-up later, resource-gating (single output base) came out ahead of lanes in testing.

PR #1893, opened 2026-07-27
fix: return struct from write_executable phase so jacoco_metadata.txt lands in runfiles

### LINKED ISSUE #1892: `bazel coverage` on scala tests produces empty coverage output at times
### `bazel coverage` on scala tests produces empty coverage output at times

Running `bazel coverage` on any scala test or scala test suite target gives me empty `_coverage_report.dat` and no coverage data is recorded even if the test says passing.

##### Possible root cause

`phase_write_executable.bzl` returns `[jacoco_metadata_file]` which is a plain Python list from the coverage branch of `_write_executable_non_windows`. `phase_default_info` only picks up runfiles from phases that return a struct with a `runfiles` attribute; a bare list is ignored, as a result `jacoco_metadata.txt` is never added to the test's runfiles tree.

At test execution time `JACOCO_METADATA_JAR` is set to a path that does not exist in the sandbox, `JacocoCoverageRunner` finds no metadata amd skips all instrumeneted classes and wirtes empty output.

Vanessamae23:
Fixes #1892

## Root cause

`_write_executable_non_windows` returned `[jacoco_metadata_file]` (a plain Python list) from the Jacoco coverage branch. In `api.bzl`, `run_phases` stores each phase result by name into `global_provider`, so `p.write_executable` ended up being a bare list.

`phase_default_info` iterates over all phase results and only collects runfiles from results that have a `runfiles` attribute (`hasattr(phase, "runfiles")`). A plain list has no such attribute, so `jacoco_metadata.txt` was silently dropped from the test's runfiles tree.

At execution time the stub script sets:
```
JACOCO_METADATA_JAR="$JAVA_RUNFILES/<workspace>/<path>/test.jacoco_metadata.txt"
```
but that path does not exist in the sandbox. `JacocoCoverageRunner` finds no metadata, skips all instrumented classes, and writes empty coverage output.

## Fix

Return `struct(runfiles = depset([jacoco_metadata_file]))` instead of `[jacoco_metadata_file]`, matching the pattern used by every other phase that contributes files to runfiles (e.g. `phase_runfiles`, `phase_scalafmt`).

## Testing

Existing coverage tests in `test/shell/test_coverage_scalatest.sh` and `test/shell/test_scala_jacocorunner.sh` cover this path. The fix makes `jacoco_metadata.txt` appear in the runfiles tree so `JACOCO_METADATA_JAR` resolves correctly and coverage data is recorded.

### DIFF
--- a/scala/private/phases/phase_write_executable.bzl
+++ b/scala/private/phases/phase_write_executable.bzl
@@ -151,7 +151,7 @@ def _write_executable_non_windows(ctx, executable, rjars, main_class, jvm_flags,
             },
             is_executable = True,
         )
-        return [jacoco_metadata_file]
+        return struct(runfiles = depset([jacoco_metadata_file]))
     else:
         # RUNPATH is defined here:
         # https://github.com/bazelbuild/bazel/blob/0.4.5/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt#L227


PawelLipski:
Thanks for the contribution! could you provide a (regression) test for the fix?

bor-p-s:
Thanks @Vanessamae23 for the fix! Merged as #1983, with the regression test that was requested here. Closing this in favor of that one.

PR #1894, opened 2026-07-28
fix: Run nested expect_* tests one at a time (exclusive)

bor-p-s:
### Description

The `expect_*` macros each run a nested `bazel` against one shared output base. Only one `bazel` can hold that base's lock at a time. So when `bazel test //...` starts several of these tests together, all but one appear to sit and wait for the lock, and that wait counts toward each test's own timeout. On a busy agent that is enough for one of them to go over the limit, which is what the timeouts in #1884 (Windows) and #1885 (Ubuntu) look like.

This tags them `exclusive` in `_nested_bazel_test`, so Bazel runs them one at a time. The running test takes the lock immediately. The tag is a default in the shared macro, so it covers the whole class of nested tests (51 today), the two in the issues included.

It keeps the single shared output base, so disk stays flat. The trade-off is on the critical path: `exclusive` makes these tests run one after another instead of alongside the rest of the suite, so a step that runs them takes longer. On Buildkite, this branch vs master was up to about +8 minutes on the steps that run the nested tests (`bazel test //...` and `test_rules_scala`), but only about +3 minutes on the whole build. The whole-build number is smaller because those steps run in parallel with other long steps, so most of the growth hides behind them. These are single runs, so treat the numbers as rough.

### Alternatives considered

All four options clear the timeouts. I picked `exclusive` as the cleanest, the one that removes the lock contention with the least extra machinery.

- **Parallel output bases / lanes (#1891, probed at K=4 in #1890):** each test hashes to one of N output bases, so up to N run at once. Fastest option. The cost: two tests that hash to the same lane still share a lock, parallel lanes write one shared `--disk_cache` together, and each lane is a full ~1.7G output base.
- **Custom-resource gating (#1895, `--local_extra_resources`):** caps how many nested tests run at once via a Bazel resource, keeping one output base. It held the timeouts on two runs and matched master on Linux. On macOS it showed a repeatable slowdown (`test_rules_scala` ~2300s vs ~2040s on master), which lines up with bazel#18153 (resource-tagged tests waiting for their turn can hold back the rest of the suite). A good second choice for getting the Linux speed later.
- **Raise the test timeout (`size=enormous`, already used for one test in 2938914c):** leaves the contention in place and gives each test more time. The needed timeout then grows as we migrate more nested tests, `enormous` (3600s) is the ceiling, and a genuinely stuck test burns the whole timeout before it fails.
- **Lower the timeout and skip the two targets on their failing OS (suggested in #1884):** covers the two named targets, while `exclusive` covers the whole class.

### Motivation

These tests take under a minute locally. In the #1885 run they each reported 250-790s and `jmh_reports_failure` tipped over the 900s timeout, with the whole class of nested tests running at once. That inflation looks like time spent waiting for the shared lock rather than real work, so running them one at a time should remove it. I am inferring this from the per-test wall times alone, so treat it as the most likely cause rather than a proven one. A single shared output base does give a single lock by construction, which is why the lock is the prime suspect.


test/expect_build_failure/expect_build_failure.bzl
```diff
@@ -132,7 +132,10 @@ def _nested_bazel_test(
         srcs = [_HELPER],
         args = args,
         data = deduped_data,
-        tags = tags,
+        # Each of these runs a nested `bazel` that takes the shared nested output
+        # base's lock. Run them one at a time (`exclusive`) so they do not queue
```
PawelLipski, line 136: Note that AFAIK these tests are exclusive not only with one another, but also with all other tests/build actions. Although they probably dominate test/build time so much that it doesn't make much difference anyway - i.e. we wouldn't save much even if we somehow made them exclusive with one another but NOT with all other tests/builds.

test/expect_build_failure/expect_build_failure.bzl
```diff
@@ -132,7 +132,10 @@ def _nested_bazel_test(
...
+        # behind each other and blow the test timeout.
```
PawelLipski, line 137: One optimization to consider if we see total test time go through the roof - define a constant number of "shards" for output bases.

I.e. all tests from shard 1 use output base 1, all tests from shard 2 use output base 2 etc., with a low (2? 4? 8?) number of output bases.

Let's see how the total test time grows over time as we get to the end of the migration (can you estimate how many bazel-in-bazel tests are still left to migrate?)

PawelLipski: > adds hashing to pick a lane

Aww okay, the other PR does that already 😅

PawelLipski: Yeah, TBH I like the other PR more (unless it turns out it doesn't help with timeouts, but this can only be verified experimentally)

test/expect_build_failure/expect_build_failure.bzl
bor-p-s: Agreed. And the numbers support your point that it barely costs anything: on Buildkite the nested-test steps grew by up to ~8 min, but the whole build only grew ~3 min, because those steps run in parallel with other long steps and most of the growth hides behind them. So even a hypothetical "serialize only among themselves" tag would save at most those ~3 min, which matches your read that it wouldn't make much difference.

PawelLipski: > because those steps run in parallel with other long steps

Hmmm wait, `exclusive` targets **don't** run in parallel with anything else? 🤔  or do you mean the state from master (not this PR) where these targets aren't exclusive?

bor-p-s: You're right, the exclusive tests don't run in parallel. I meant the Buildkite steps (separate agents), not the tests. The test_rules_scala step grew ~8 min, but in master it wasn't the slowest step (~1730s vs ~2040s), so most of that growth stayed under the critical path. Only ~3 min showed up on the whole build.

test/expect_build_failure/expect_build_failure.bzl
bor-p-s: What I don't like is that PR1891 approach eats much more space. I'm not sure what is our CI configuration is and would it cause issues potentially. Also, AFAICS PR1891 carries more potential conflicts that could cause more flakiness. I'd say it's your call.

PawelLipski: Okay... could you restart (by amending&pushing) the CI on that other PR for a couple times to see if the errors (either due to disk pressure and/or flakiness) show up?

Also, I think you should be able to determine what machine spec we're using in the CI + estimate how much disk is consumed for each output base

Still, I'd like @WojciechMazur to take a look in case we're missing sth obvious

WojciechMazur: If the increased disk usage per lane (~1.4G) is the main concern then I think we can go for parallel builds - I'm not sure about workers used in our CI, but the smallest BuildKite-hosted runner provide 47GB, I'm not sure about the agents pool we're using (can't get any info on that, so some runtime check might be required) 
Both solutions look fine, if we need to perform additional checks we can go with exclusiveness for now to fix the build stability and resolve performance issue in the follow up

PawelLipski (approved):

PR #1895, opened 2026-07-30
TEMP: probe custom-resource gating for nested tests (do not merge)

bor-p-s:
Throwaway probe, do not merge. Tests the resource-gating idea from #1891 (cap nested tests to one at a time via `--local_extra_resources`, single output base) and prints the agent disk/spec.


bor-p-s:
Throwaway probe, closing. exclusive merged in #1894. The resource-gating results are summarized there as the runner-up: it holds the timeouts and matches master on Linux, with a repeatable macOS slowdown (bazel#18153).

PR #1896, opened 2026-07-30
refactor: Replace test_deps.sh with an analysis test

bor-p-s:
### Description

Replaces `test/shell/test_deps.sh` with an `analysistest` on the same two binaries, renamed from `test_binary_to_ensure_no_host_deps` to `test_binary_to_ensure_no_exec_deps`.

The shell test ran `readlink` over every `*.jar` in the binaries' runfiles and grepped the output for `bazel-out/host`. Bazel 7 names no output directory that way, so the grep never matched and both cases passed whatever ended up in runfiles. In this checkout the runfiles of both binaries contain files from two roots only: the source root and `bazel-out/<cpu>-fastbuild/bin`.

The replacement fails on any runfile whose root is neither of those two, instead of looking for a directory name. One plain build already has two exec configurations with different output directories (`darwin_arm64-opt-exec-ST-154afa76daca` and `...-ST-d57f47055a04`, differing in `java_runtime_version`), so a check pinned to a single exec directory would stay quiet about a leak from the other one, which is how the old check rotted. I checked that the new one does fail, with throwaway rules that put an exec-configuration file and a transitioned file into their runfiles.

This is stricter than the original: a runfile from any configuration other than the target one fails the test, not only from an exec configuration. Neither binary has such a file today, and if a dependency later goes through a legitimate transition, the test fails and prints the path.

### Motivation

Part of #1867. What this test checks is visible at analysis time, so the replacement needs no nested `bazel` invocation.


test/no_exec_config_deps_test.bzl
```diff
@@ -0,0 +1,17 @@
+load("@bazel_skylib//lib:unittest.bzl", "analysistest", "asserts")
+
+def _no_exec_config_deps_test(ctx):
```
PawelLipski, line 12: Somewhat a pre-existing problem, but it's unclear what this test is actually testing for, pls add some """-doc

test/no_exec_config_deps_test.bzl
```diff
@@ -0,0 +1,17 @@
...
+    env = analysistest.begin(ctx)
+
+    runfiles = analysistest.target_under_test(env)[DefaultInfo].default_runfiles.files.to_list()
+
+    # A file's root is empty for sources and otherwise names the configuration that
```
PawelLipski: Also, the comment looks it's missing some context 🤔 _file's root_ - unclear what that refers to

test/no_exec_config_deps_test.bzl
```diff
@@ -0,0 +1,17 @@
...
+    # built it. Accept only the target configuration: a build has more than one exec
+    # configuration, so matching exec output dirs would miss leaks from the others.
+    foreign = [f.path for f in runfiles if f.root.path not in ["", ctx.bin_dir.path]]
+
+    asserts.equals(env, [], foreign)
```
PawelLipski: Add some assertion message if the API allows, otherwise the failure's gonna be very confusing

test/no_exec_config_deps_test.bzl
bor-p-s: done

test/no_exec_config_deps_test.bzl
bor-p-s: done

test/no_exec_config_deps_test.bzl
bor-p-s: done

test/no_exec_config_deps_test.bzl
```diff
@@ -0,0 +1,29 @@
...
+    """
+    env = analysistest.begin(ctx)
+
+    # Every File has a root: an empty path for a source, otherwise the output directory
+    # of the configuration that built it. ctx.bin_dir is that directory for the
+    # configuration the test and its target under test are analyzed in. A build has more
+    # than one exec output directory, so the check accepts that single root instead of
+    # matching exec ones.
+    runfiles = analysistest.target_under_test(env)[DefaultInfo].default_runfiles.files.to_list()
+    foreign = [f.path for f in runfiles if f.root.path not in ["", ctx.bin_dir.path]]
```
PawelLipski: This is pretty esoteric... could you, as a verification that it really checks what it should (esp. when the supported Bazel version is bumped one day), add a test (with a required failure) where a build-time dependency is present on the runtime classpath? 

You can use some ad-hoc faulty toolchain to achieve that setup

Not insisting tho, if it's going to be very contrived then probably not worth it

This entire suite looks like a very specific regression test

bor-p-s: Added it. `//test:exec_config_dep_leak` is a rule that pulls scalac in with `cfg = "exec"` and puts it in its own runfiles, which is the mistake we are looking for, and `//test:detects_exec_config_deps_test` asserts the check reports it. I used a second analysistest with the inverted assertion rather than `expect_test_failure_test`, so it stays free of a nested bazel invocation.

To make sure the new test cannot pass for free, I neutered the check (made the predicate return nothing) and re-ran it: `detects_exec_config_deps_test` failed with `runfiles of @@//test:exec_config_dep_leak hold scalac, built in the exec configuration`, while `//test/proto:no_exec_deps_test` kept passing. That gap is what the new test now covers.

A faulty toolchain would give a more lifelike leak, but it needs much more setup for the same signal, so I stayed with the small fixture.

test/no_exec_config_deps_test.bzl
```diff
@@ -0,0 +1,62 @@
...
+    )
+
+    return analysistest.end(env)
+
+no_exec_config_deps_test = analysistest.make(_no_exec_config_deps_test)
+
+detects_exec_config_deps_test = analysistest.make(_detects_exec_config_deps_test)
+
+def _exec_config_dep_leak(ctx):
+    return [DefaultInfo(runfiles = ctx.runfiles(files = ctx.files._build_time_tool))]
```
PawelLipski, line 51: Okay! ended up pretty easy & clear, nice

PawelLipski (approved):

PR #1901, opened 2026-08-04
refactor: Migrate jvm-flags toolchain-success tests to Bazel

bor-p-s:
### Description

Migrates the five toolchain-success cases from `test/shell/test_scalac_jvm_flags.sh` and `test/shell/test_scala_jvm_flags.sh`.

The fixtures land in the existing `test/toolchains/scalac_jvm_opts` and `test/toolchains/scala_test_jvm_flags` rather than new packages: those already hold the failing halves of the same fixtures, migrated in #1852 and #1856 out of `test_expect_failure/`, under the same target names and with their own copies of `Empty.scala` / `EmptyTest.scala`. `empty_build` and `empty_test` are now each driven by one failure test and one success test differing only in the toolchain.

That merge turned up a dead fixture. `test/toolchains/scala_test_jvm_flags/EmptyTest.scala` and `test/toolchains/scala_test_jacocorunner/EmptyTest.scala` import `org.scalatest.FunSuite`, which scalatest 3.2.19 does not have, so neither compiles:

```
test/toolchains/scala_test_jvm_flags/EmptyTest.scala:3: error: object FunSuite is not a member of package org.scalatest
```

Both are tagged `manual`, so no wildcard build touched them, and both negative tests were green on that compile error rather than on the failure they name: neither had an `expect`, so any failure counted as a pass.

Fixed both fixtures and added `expect` to both tests, pinning them to `Too small maximum heap` and the jacocorunner `ClassNotFoundException`. Verified by putting `FunSuite` back locally: both tests failed again, now on the `expect` mismatch.

Unlike the failure fixtures next to them, these ones also succeed under the default toolchain, so their `manual` tag saves a redundant second run rather than avoiding a failure.

### Motivation

Part of #1867 (move tests out of `test_*.sh` into regular bazel test rules).

The five new tests are `exclusive`, so they extend the serialized path. This PR's two CI runs both land above master's range on `./test_rules_scala` ubuntu (39-43 min vs. master's 28-38 min, median 28): the added tests cost several real minutes.


PawelLipski (commented):
Whoops looks that the total CI time spirals out of control under `exclusive`, consider the alternatives then 😅

PawelLipski (approved):

PR #1903, opened 2026-08-04
fix: Fail string_in_filename_in_jar.sh when grep itself fails

bor-p-s:
### Description

`string_in_filename_in_jar.sh` now fails when `grep` itself fails. `grep` exits above 1 on error, and the "expect not to find" branch treated anything other than 0 as a pass, so a broken input passed for free. The pattern is also quoted and matched literally with `-F`, so a pattern with a space stays one string instead of becoming a second grep argument, and `.` in a name like `resource.txt` matches only a dot.

`string_in_filename_in_jar_required_failures` covers the cases: an unreadable jar has to report `Could not list`, a failing `grep` has to report `Could not search`, and a pattern with a space has to come back as a plain "not found". The first two run for both the "true" and the "false" expectation. The test makes `grep` fail by putting a stub that exits 2 first on `PATH`, so it behaves the same on BSD and GNU grep instead of leaning on a pattern odd enough to break the real one.

### Motivation

This is the same problem as #1897, one line further down. That PR made an unreadable jar fail instead of returning an empty listing, and left `grep` able to fail just as quietly, which the "false" expectation would report as a pass. The guard #1897 added had no test of its own, and the new test covers it too.

I checked each case is live by taking out one piece at a time: without the grep guard, without the quoting and `-F`, and without the `jar tf` guard from #1897 the test goes red each time, and green with all three in place.

Stacked on #1900, which moves the script to `test/sh_tests/`. The test itself stays in the scrooge package, where the jar it needs lives, and loads the script the way the neighbouring tests do.

Release impact: tests and test helpers only.


test/sh_tests/BUILD
```diff
@@ -106,6 +106,21 @@ sh_test(
     ],
 )
 
+sh_test(
+    name = "string_in_filename_in_jar_required_failures",
+    size = "small",
+    srcs = ["string_in_filename_in_jar_required_failures.sh"],
+    args = [
+        "$(rootpath string_in_filename_in_jar.sh)",
```
PawelLipski, line 117: For clarity, as we're testing `string_in_filename_in_jar.sh` here... what is this script itself used for? 🤔

bor-p-s: Added a comment above the target explaining that.

PawelLipski (approved):

PR #1904, opened 2026-08-06
refactor: Migrate classpath namespace-conflict test to Bazel

PawelLipski:
### Description

Migrates the last case in `test/shell/test_scala_classpath.sh` to a native `bazel test`. That case built `//test/src/main/scala/scalarules/test/classpath_resources:classpath_resource_duplicates` (a `scala_binary` with two `classpath_resources` of the same name) and grepped the build output for the namespace-conflict warning `ScalacWorker` prints.

Replaced with an `expect_build_success_test` in `test/classpath_resource_conflict/`: it drives a nested `bazel build` of the same fixture, asserts the build succeeds, and asserts the warning is surfaced. Because `expect` files are supplied, the helper runs `bazel clean` first, so a cached action can't silently skip recompilation and hide the warning.

The fixture `scala_binary` stays in `test/src/main/...` — it's a stable shared target that builds cleanly under wildcards — so the test references it cross-package (with the usual glob-limitation comment). `test/shell/test_scala_classpath.sh` and its `test_rules_scala.sh` entry are removed.

### Motivation

Part of migrating bespoke shell tests to native `bazel test`, so `bazel test //...` covers these cases directly.


test/classpath_resource_conflict/BUILD
```diff
@@ -0,0 +1,19 @@
+load("//test/expect_build_failure:expect_build_failure.bzl", "expect_build_success_test")
+
+package(default_testonly = 1)
+
+# `//test/src/main/scala/scalarules/test/classpath_resources:classpath_resource_duplicates`
+# is a scala_binary with two `classpath_resources` of the same name, so it builds
+# successfully but ScalacWorker prints a namespace-conflict warning. Bazel has no
+# native "build and assert on output" rule, so drive a nested `bazel build` via
+# expect_build_success_test and assert the warning is surfaced (with expect files
+# the helper runs `bazel clean` first, so a cached action can't hide the warning).
```
PawelLipski: ❓

bor-p-s: Lines 5-10 restate the `expect_build_success_test` docstring in `expect_build_failure.bzl` (the bazel clean caveat about `expect`/`reject`). Maybe just point there instead of repeating it, so the two stay in sync if the macro's behavior changes. Is this what the ❓ was about?

bor-p-s (approved):

test/classpath_resource_conflict/BUILD
PawelLipski: Oh, it was actually about _with expect files the helper runs `bazel clean` first_ which I wasn't sure is true (it is)

But still, good point, lemme tidy up

PR #1907, opened 2026-08-07
test: Declare the inputs of the nested expect_* tests

bor-p-s:
### Description

The nested `expect_*` tests run a `bazel` of their own against the real source tree, invisible to the outer Bazel. Two problems follow: a cached pass can survive the change it was meant to catch, and `local` keeps every one of the 61 tests out of the shared cache, so every build re-runs all of them.

Each test now declares what decides its outcome: a fingerprint of the actions its fixture and its dependencies run (an aspect in `collect_actions.bzl`), the toolchains a `--extra_toolchains` flag registers (a transition), and the repo `.bazelrc`, `MODULE.bazel`(`.lock`), `.bazelversion`, and the compiler/runner jars. `local` becomes `no-sandbox` + `no-remote-exec`.

Analysis fails if the fingerprint carries no rules_scala action -- a fingerprint that can't notice a rules change must not serve a stale pass. 11 tests hit this (`scala_proto`/`semanticdb` compiling through their own aspect, two plain `proto_library`/`java_library` targets, and two `test/strict_dependency/scala_proto_deps` tests whose shared `--extra_toolchains` puts them under the same config as `:some_proto`'s own custom aspect, which Bazel reports as an `ActionConflictException`). Each says why with `no_fingerprint_reason` and is tagged `external`.

A fixture in another package used to be `external` unconditionally, since the macro can't verify a caller's `data` covers every source. Now that's only true when `target` is a pattern with no single label to fingerprint (`test/twitter_scrooge_jdk11`'s `.../...`); a concrete label gets the same real fingerprint a same-package fixture gets. It still can't safely add the fixture's own build output to `data` for full source coverage -- these fixtures are routinely `tags = ["manual"]`, built to fail or to only succeed under a toolchain the nested `bazel` supplies, and this macro can't read that tag from another package's BUILD file. Same tradeoff a near-empty fingerprint already accepts.

Two fixtures that fail during analysis moved to `analysistest` (#1922, merged, this PR's base).

`test/expect_build_failure/README.md` has the rest.

### Motivation

Removes a risk (a regression a nested test can no longer catch) and a cost (every nested test re-running on every build). The cross-package fixtures now cache: running them twice, the second run shows `(cached) PASSED`.

CI, same two steps against pre-branch master (#1918's build #6615) and this branch (#6637):

| | master | this branch |
|---|---|---|
| whole build | 1h19m05s | 1h16m48s |
| `bazel test //...` ubuntu | 2847s | 1484s |
| `bazel test //...` macOS | 3477s | 823s |
| `./test_rules_scala` (5 platform variants) | 3137-4712s | 3661-4600s |

Release impact: tests only.

### DIFF
--- a/test/classpath_resource_conflict/BUILD
+++ b/test/classpath_resource_conflict/BUILD
@@ -9,9 +9,8 @@ package(default_testonly = 1)
 # expect_build_success_test and assert the warning is surfaced. (See that macro's
 # docstring for how it keeps a cached action from hiding the warning.)
 #
-# `target` lives in another package, so its sources aren't tracked as this test's
-# runfiles (see expect_build_failure.bzl) and an edit there won't force a re-run;
-# the fixture is long-standing and rarely touched.
+# `target` lives in another package, so the macro tags this test `external`
+# (see expect_build_failure.bzl) and it re-runs on every invocation.
 expect_build_success_test(
     name = "classpath_resource_namespace_conflict_test",
     expect = ["expected_namespace_conflict.txt"],
--- a/test/classpath_resource_conflict/BUILD
+++ b/test/classpath_resource_conflict/BUILD
@@ -9,9 +9,8 @@ package(default_testonly = 1)
 # expect_build_success_test and assert the warning is surfaced. (See that macro's
 # docstring for how it keeps a cached action from hiding the warning.)
 #
-# `target` lives in another package, but it's one concrete label, so the macro
-# still fingerprints it via `fixture_actions` (see expect_build_failure.bzl)
-# instead of tagging this test `external`.
+# `target` lives in another package; see expect_build_failure.bzl's docstring
+# for why that still gets a real fingerprint here.
 expect_build_success_test(
     name = "classpath_resource_namespace_conflict_test",
     expect = ["expected_namespace_conflict.txt"],
--- a/test/classpath_resource_conflict/BUILD
+++ b/test/classpath_resource_conflict/BUILD
@@ -9,8 +9,9 @@ package(default_testonly = 1)
 # expect_build_success_test and assert the warning is surfaced. (See that macro's
 # docstring for how it keeps a cached action from hiding the warning.)
 #
-# `target` lives in another package, so the macro tags this test `external`
-# (see expect_build_failure.bzl) and it re-runs on every invocation.
+# `target` lives in another package, but it's one concrete label, so the macro
+# still fingerprints it via `fixture_actions` (see expect_build_failure.bzl)
+# instead of tagging this test `external`.
 expect_build_success_test(
     name = "classpath_resource_namespace_conflict_test",
     expect = ["expected_namespace_conflict.txt"],
--- a/test/expect_build_failure/README.md
+++ b/test/expect_build_failure/README.md
@@ -1,0 +1,77 @@
+# Asserting that a build or a test fails
+
+Two mechanisms live here. Pick by *when* the failure happens.
+
+## Analysis-time failure: use `expect_analysis_failure_test`
+
+A target that Bazel rejects while analysing it -- a `fail()` in a rule, a bad
+attribute combination -- needs no build at all. `expect_analysis_failure_test`
+(see `expect_analysis_failure.bzl`) asks Bazel in-process through
+`bazel_skylib`'s `analysistest`. It is hermetic, cacheable and runs in under a
+second.
+
+## Execution-time failure: use the `expect_*` macros
+
+A compile error, a failing test binary, a warning printed by an action: these
+only appear once actions run, and Bazel has no native rule for "run a build and
+assert it fails". So `expect_build_failure.bzl` wraps an `sh_test` around a
+**nested `bazel`** invocation.
+
+That nested build reads the real source tree, which the outer Bazel cannot see,
+so the test has to declare by hand everything that decides its outcome. The
+macro does that for you; the docstring in `expect_build_failure.bzl` lists what
+goes into the cache key and what stays outside it.
+
+Nothing re-runs these tests without the result cache on a schedule. That would be
+the way to catch an input nobody thought to declare.
+
+## What to delete later
+
+- **On Bazel 8 or newer**, an aspect can reach toolchain dependencies through
+  `aspect(toolchains_aspects = ...)`. That replaces `_SCALAC_JAR` and
+  `_RUNNER_JARS`, which exist only because the fingerprint carries the tools'
+  paths and not their content. This repo pins 7.7.1 in `.bazelversion`, where
+  the parameter does not exist.
+- **The success cases do not need a nested build at all.** A Starlark transition
+  can set `//command_line_option:extra_toolchains`, verified on Bazel 7.7.1: the
+  same fixture fails with the `-Xmx1M` toolchain and builds with the `-Xmx1G` one
+  when the flag is set that way. Of the 61 tests here, 41 assert a failure and do
+  need the nested build, because a target that fails cannot be a dependency. The
+  other 20 assert success: 17 of those only set flags and check nothing about the
+  output, so a transition plus `build_test` covers them, and 1 sets no flags at
+  all and needs only `build_test`. The last 2 assert on build output, which only
+  a real invocation produces. Those 18 took 180s of the 653s a full local run
+  spends here, all of it on the critical path because of `exclusive`.
+- **The whole nested driver** would go away by moving these tests to
+  [`rules_bazel_integration_test`](https://github.com/bazel-contrib/rules_bazel_integration_test),
+  which gives each test its own child workspace and an explicit `workspace_files`
+  attribute. Then the inputs are declared by construction rather than by hand,
+  the shared output base disappears, and with it the `exclusive` tag that
+  serialises these tests today. The cost is moving 27 fixture packages into
+  child workspaces, so it is a project rather than a patch.
+
+## When you add a test here
+
+Nothing in a caller has to know about caching. Two things are worth knowing:
+
+- If the fixture lives in another package, make it visible to yours; the
+  fingerprint is a real dependency edge, unlike the label string the nested
+  `bazel` receives.
+- A nested build runs with a scrubbed `HOME`, so your `~/.bazelrc` is ignored.
+  If you need it (a download proxy, say), set
+  `RULES_SCALA_NESTED_BAZEL_USE_REAL_HOME=1`. Any failing nested build says so.
+
+## Registering a toolchain
+
+A toolchain the nested build registers decides whether that build fails, and it
+is not part of the fixture's own graph. The macro handles this on its own: it
+reads the `--extra_toolchains` flags out of the arguments it forwards and
+analyses the fixture under exactly those toolchains, through a Starlark
+transition on `//command_line_option:extra_toolchains`. Change a toolchain and
+the command lines in the fingerprint change with it, so the test re-runs.
+
+Nothing to declare in the toolchain's package, in other words. Verified by
+setting `unused_dependency_checker_mode`, `strict_deps_mode` and
+`dependency_mode` in `//scala:minimal_direct_source_deps` to values that would
+make the assertion vacuous: each moves the fingerprint of the test that
+registers it.
--- a/test/expect_build_failure/README.md
+++ b/test/expect_build_failure/README.md
@@ -52,7 +52,7 @@ the way to catch an input nobody thought to declare.
 
 ## When you add a test here
 
-Nothing in a caller has to know about caching. Two things are worth knowing:
+Nothing in a caller has to know about caching. Three things are worth knowing:
 
 - If the fixture lives in another package, make it visible to yours; the
   fingerprint is a real dependency edge, unlike the label string the nested
@@ -60,6 +60,23 @@ Nothing in a caller has to know about caching. Two things are worth knowing:
 - A nested build runs with a scrubbed `HOME`, so your `~/.bazelrc` is ignored.
   If you need it (a download proxy, say), set
   `RULES_SCALA_NESTED_BAZEL_USE_REAL_HOME=1`. Any failing nested build says so.
+- Analysis fails if the fingerprint carries no rules_scala action, because such
+  a fingerprint cannot notice a change in the rules and the test it keys would
+  stay green through a regression. Either point `fingerprint_target` at a label
+  the nested build really compiles, or say why you cannot with
+  `no_fingerprint_reason`, which tags the test `external` so it re-runs every
+  time rather than being served a pass it cannot vouch for. Seven tests do that
+  today: `scala_proto` generates through an aspect of its own, whose actions
+  this aspect cannot read, and the `semanticdb` fixtures only build toolchain
+  deps, so they compile nothing at all.
+
+What the fingerprint does not see: an action that writes a file instead of
+running a command line (`ctx.actions.write`, template expansion) contributes its
+mnemonic and nothing else, which is 446 of the 5135 lines it collects today. A
+change to what such an action writes leaves the fingerprint identical. Declaring
+those files instead would put output paths -- and with them the configuration --
+back into the key, which is what makes the tests re-run under every flag the
+outer build sets.
 
 ## Registering a toolchain
 
--- a/test/expect_build_failure/README.md
+++ b/test/expect_build_failure/README.md
@@ -54,12 +54,13 @@ the way to catch an input nobody thought to declare.
 
 Nothing in a caller has to know about caching. Three things are worth knowing:
 
-- If the fixture lives in another package, make it visible to yours; the
-  fingerprint is a real dependency edge, unlike the label string the nested
-  `bazel` receives.
+- If the fixture lives in another package, the macro tags the test `external`,
+  because it cannot verify that caller data covers every source the nested target
+  reads.
 - A nested build runs with a scrubbed `HOME`, so your `~/.bazelrc` is ignored.
-  If you need it (a download proxy, say), set
-  `RULES_SCALA_NESTED_BAZEL_USE_REAL_HOME=1`. Any failing nested build says so.
+  If you need it (a download proxy, say), pass `--nocache_test_results` and
+  `--test_env=RULES_SCALA_NESTED_BAZEL_USE_REAL_HOME=1` to `bazel test`. The
+  home file is outside this test's key, so keep this opt-in uncached.
 - Analysis fails if the fingerprint carries no rules_scala action, because such
   a fingerprint cannot notice a change in the rules and the test it keys would
   stay green through a regression. Either point `fingerprint_target` at a label
--- a/test/expect_build_failure/README.md
+++ b/test/expect_build_failure/README.md
@@ -79,11 +79,16 @@ Nothing in a caller has to know about caching. Three things are worth knowing:
 
 What the fingerprint does not see: an action that writes a file instead of
 running a command line (`ctx.actions.write`, template expansion) contributes its
-mnemonic and nothing else, which is 446 of the 5135 lines it collects today. A
-change to what such an action writes leaves the fingerprint identical. Declaring
-those files instead would put output paths -- and with them the configuration --
-back into the key, which is what makes the tests re-run under every flag the
-outer build sets.
+mnemonic and nothing else, so a change to what such an action writes leaves the
+fingerprint identical. The same blind spot applies more broadly: an action's
+command line carries a file's path, not its content, so a content-only change
+to a file rules_scala itself reads by path (rather than compiles from a
+fixture's own sources) also leaves the fingerprint identical -- `_SCALAC_JAR`
+and `_RUNNER_JARS` below are declared as explicit `data` for exactly this
+reason. Declaring the file such a write action produces as a test input
+instead would put its output path -- and with it the configuration -- back
+into the key, which is what makes the tests re-run under every flag the outer
+build sets.
 
 ## Registering a toolchain
 
--- a/test/expect_build_failure/README.md
+++ b/test/expect_build_failure/README.md
@@ -32,16 +32,16 @@ the way to catch an input nobody thought to declare.
   `_RUNNER_JARS`, which exist only because the fingerprint carries the tools'
   paths and not their content. This repo pins 7.7.1 in `.bazelversion`, where
   the parameter does not exist.
-- **The success cases do not need a nested build at all.** A Starlark transition
-  can set `//command_line_option:extra_toolchains`, verified on Bazel 7.7.1: the
-  same fixture fails with the `-Xmx1M` toolchain and builds with the `-Xmx1G` one
-  when the flag is set that way. Of the 61 tests here, 41 assert a failure and do
-  need the nested build, because a target that fails cannot be a dependency. The
-  other 20 assert success: 17 of those only set flags and check nothing about the
-  output, so a transition plus `build_test` covers them, and 1 sets no flags at
-  all and needs only `build_test`. The last 2 assert on build output, which only
-  a real invocation produces. Those 18 took 180s of the 653s a full local run
-  spends here, all of it on the critical path because of `exclusive`.
+- **Most success cases do not need a nested build at all.** A Starlark
+  transition can set `//command_line_option:extra_toolchains`, verified on
+  Bazel 7.7.1: the same fixture fails with the `-Xmx1M` toolchain and builds
+  with the `-Xmx1G` one when the flag is set that way. A test asserting a
+  build *failure* still needs the nested build, because a target that fails
+  cannot be a dependency. A test asserting *success* only needs it if it also
+  checks the build output (`expect`/`reject`); one that just checks the build
+  passed under a flag or toolchain override can be replaced with a transition
+  plus `build_test`, which runs in-process and isn't `exclusive`, so it drops
+  off the critical path these nested tests share today.
 - **The whole nested driver** would go away by moving these tests to
   [`rules_bazel_integration_test`](https://github.com/bazel-contrib/rules_bazel_integration_test),
   which gives each test its own child workspace and an explicit `workspace_files`
@@ -68,10 +68,14 @@ Nothing in a caller has to know about caching. Three things are worth knowing:
   stay green through a regression. Either point `fingerprint_target` at a label
   the nested build really compiles, or say why you cannot with
   `no_fingerprint_reason`, which tags the test `external` so it re-runs every
-  time rather than being served a pass it cannot vouch for. Seven tests do that
-  today: `scala_proto` generates through an aspect of its own, whose actions
-  this aspect cannot read, and the `semanticdb` fixtures only build toolchain
-  deps, so they compile nothing at all.
+  time rather than being served a pass it cannot vouch for. A few tests do
+  that today, for one of four reasons: `scala_proto` generates through an
+  aspect of its own, whose actions this aspect cannot read; the `semanticdb`
+  fixtures only build toolchain deps, so they compile nothing at all; a
+  fixture targets a plain `proto_library`/`java_library`, which runs no
+  rules_scala action either; or fingerprinting two fixtures under the same
+  `--extra_toolchains` would analyse a shared dependency under two different
+  aspect stacks, which Bazel reports as an `ActionConflictException`.
 
 What the fingerprint does not see: an action that writes a file instead of
 running a command line (`ctx.actions.write`, template expansion) contributes its
--- a/test/expect_build_failure/README.md
+++ b/test/expect_build_failure/README.md
@@ -54,9 +54,11 @@ the way to catch an input nobody thought to declare.
 
 Nothing in a caller has to know about caching. Three things are worth knowing:
 
-- If the fixture lives in another package, the macro tags the test `external`,
-  because it cannot verify that caller data covers every source the nested target
-  reads.
+- If the fixture lives in another package, it still gets a real fingerprint
+  through `fixture_actions` when `target` is one concrete label. Only a
+  cross-package *pattern* (a wildcard like `//pkg/...`) gets tagged `external`,
+  because there's no single label a macro could fingerprint or a caller could
+  declare as `data`.
 - A nested build runs with a scrubbed `HOME`, so your `~/.bazelrc` is ignored.
   If you need it (a download proxy, say), pass `--nocache_test_results` and
   `--test_env=RULES_SCALA_NESTED_BAZEL_USE_REAL_HOME=1` to `bazel test`. The
--- a/test/expect_build_failure/collect_actions.bzl
+++ b/test/expect_build_failure/collect_actions.bzl
@@ -35,7 +35,7 @@ def _without_configuration(line):
         out += "bazel-out/" + ("<cfg>" + part[slash:] if slash != -1 else part)
     return out
 
-def _collect_impl(target, ctx):
+def _collect_actions_aspect_impl(target, ctx):
     # One line per action, joining mnemonic and argv so each action keeps its own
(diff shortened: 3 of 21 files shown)


test/expect_build_failure/collect_actions.bzl
```diff
@@ -0,0 +1,100 @@
+"""Serializes the actions a target would run, so they can key a nested test.
```
PawelLipski, line 1: Wow, I start to get it

test/expect_build_failure/collect_actions.bzl
```diff
@@ -0,0 +1,100 @@
...
+)
+
+# Every action this ruleset runs invokes a tool built from here, so a fingerprint
+# without it describes a target the rules never touch.
+_RULESET_TOOLS = "src/java/io/bazel/rulesscala"
+
+def _exposed_impl(ctx):
+    lines = sorted(ctx.attr.target[0][_ActionsInfo].lines.to_list())
+    if not [line for line in lines if _RULESET_TOOLS in line]:
+        fail(("%s runs no rules_scala action, so it cannot tell whether the rules " +
```
PawelLipski, line 86: 🧠

test/strict_dependency/scala_proto_deps/BUILD
```diff
@@ -85,12 +85,14 @@ expect_build_failure_test(
...
     expect = ["expected_stamp_by_convention.txt"],
+    no_fingerprint_reason = "under this toolchain combination collides with custom_phase_stamping_test below on :some_proto's aspect application (see that test's own no_fingerprint_reason)",
     target = ":uses_transitive_scala_proto",
 )
 
 expect_build_failure_test(
     name = "custom_phase_stamping_test",
     build_args = ["--extra_toolchains=//test/toolchains:ast_plus_one_deps_strict_deps_error,//test/strict_dependency/scala_proto_deps:stamp_by_convention_toolchain"],
     expect = ["expected_custom_phase_stamping.txt"],
+    no_fingerprint_reason = "fingerprinting this target under the same --extra_toolchains as stamp_by_convention_test above analyses :some_proto under two different aspect stacks (scala_proto_aspect alone vs. scala_proto_aspect + this package's custom_stamping_apsect) in the same configuration, which Bazel reports as an ActionConflictException on :some_proto's generated jar",
```
PawelLipski: Typo: `apsect`

PawelLipski: (oh, pre-existing, pls rename then)

test/strict_dependency/BUILD
```diff
@@ -7,6 +7,7 @@ load("//test/expect_build_failure:expect_build_failure.bzl", "expect_build_failu
 expect_build_success_test(
     name = "plus_one_deps_only_works_for_java_info_targets_test",
     build_args = ["--extra_toolchains=//test_expect_failure/plus_one_deps:plus_one_deps"],
+    no_fingerprint_reason = "test_proto is a plain proto_library, so it runs no rules_scala action to fingerprint",
```
PawelLipski, line 10: Nit: would it make sense to add an attr to allow for no rules_scala actions in the given test? although AFAICS it's just these 2 tests, so maybe not worth it

test/twitter_scrooge_jdk11/BUILD
```diff
@@ -16,12 +16,6 @@ expect_build_success_test(
         "--java_runtime_version=remotejdk_11",
         "--tool_java_runtime_version=remotejdk_11",
     ],
-    # The built target's sources live in another package, so the macro's
-    # automatic glob of this package doesn't cover them for staleness. Depend
-    # on a public target that transitively covers most of that package's
-    # scrooge sources (:twitter_scrooge_tests itself is package-private) so a
-    # source change still invalidates this test's cache key.
-    data = ["//test/src/main/scala/scalarules/test/twitter_scrooge:justscrooges_scala_and_java"],
```
PawelLipski, line 24: This is now redundant right? as `fixture_actions` cover it anyway?

Or actually, `fixture_actions` don't "include" targets as such in any sense, right? only text dumps of actions from the transitive closure of the `target`?

test/expect_build_failure/nested_bazel.sh
```diff
@@ -91,6 +91,16 @@ _nested_bazel_common_opts=()
 # then serialize on Bazel's output-base lock (each inner `bazel --batch` waits for
 # the lock rather than failing), which keeps the extracted external repos warm and
 # avoids multiplying ~1GB of Scala jars across a separate output base per test.
+_nested_bazel_home_hint() {
+  if [[ -n "${RULES_SCALA_NESTED_BAZEL_USE_REAL_HOME:-}" ]]; then
+    return
+  fi
+  echo "note: the nested \`bazel\` ran with a scrubbed HOME, so your ~/.bazelrc was" >&2
```
PawelLipski, line 98: Actually, is this fix specifically needed for this PR? or would it already be useful before (just the error not surfaced)?

PawelLipski: Coz scrubbed HOME was already there before (?)

test/expect_build_failure/expect_build_failure.bzl
```diff
@@ -318,7 +471,9 @@ def expect_test_success_test(
         expect = [],
         reject = [],
         size = "large",
-        tags = ["local", "requires-network"],
+        tags = ["no-sandbox", "requires-network"],
```
PawelLipski, line 479: `local` is uncached but `no-sandbox` is, right?

test/classpath_resource_conflict/BUILD
```diff
@@ -9,9 +9,8 @@ package(default_testonly = 1)
 # expect_build_success_test and assert the warning is surfaced. (See that macro's
 # docstring for how it keeps a cached action from hiding the warning.)
 #
-# `target` lives in another package, so its sources aren't tracked as this test's
-# runfiles (see expect_build_failure.bzl) and an edit there won't force a re-run;
-# the fixture is long-standing and rarely touched.
+# `target` lives in another package, so the macro tags this test `external`
+# (see expect_build_failure.bzl) and it re-runs on every invocation.
```
PawelLipski: Hmm okay - so this test is **silently** un-cached right (b/c the reason isn't lacking rules_scala in fingerprint but cross-package tested target)?

Maybe there should be an explicit opt-out boolean attr for these cases? like, mostly to increase visibility; not insisting tho

test/expect_build_failure/collect_actions.bzl
```diff
@@ -0,0 +1,100 @@
...
+The fixture is analysed under the toolchains the nested `bazel` registers, so a
+change in one of those lands in the command lines too.
+
+Command lines carry output paths, and those name the configuration they were
+built in, so the fingerprint would otherwise differ under every flag the outer
+build happens to set and re-run the test for nothing. The configuration segment
+is dropped, in the spirit of Bazel's own path mapping. What is left out: actions
+that write a file rather than run a command line (`ctx.actions.write`, template
+expansion) contribute their mnemonic and nothing else, so a change to what such
+an action writes does not reach the fingerprint.
```
PawelLipski, line 18: Hmmm and what if the **contents** of input files that the fingerprinted actions consume change?

Like, if `ScalaCompile` read a file (defined here in rules_scala, not in a project that consumes rules_scala) with the content `foobar`, and then it became `barfoo`? does it make sense?  I think the action dump won't change in such case. Although this is probably a niche corner case (as most file inputs are coming from the user's project, not from rules_scala)

test/expect_build_failure/collect_actions.bzl
```diff
@@ -0,0 +1,100 @@
...
+def _without_configuration(line):
+    # Output paths carry the configuration (`bazel-out/k8-fastbuild-ST-1234/`),
+    # so the same fixture yields a different fingerprint under every flag the
+    # outer build happens to set, and every such build re-runs the test. The
+    # nested `bazel` does not inherit those flags, so drop the segment.
+    parts = line.split("bazel-out/")
+    out = parts[0]
+    for part in parts[1:]:
+        slash = part.find("/")
+        out += "bazel-out/" + ("<cfg>" + part[slash:] if slash != -1 else part)
```
PawelLipski, line 35: Why two cases with/without slash?

test/expect_build_failure/collect_actions.bzl
```diff
@@ -0,0 +1,100 @@
...
+    return out
+
+def _collect_impl(target, ctx):
+    # One line per action, keeping the arguments in order. Collecting the tokens
+    # separately would lose which value belongs to which flag: a depset of tokens
```
PawelLipski: This is unclear 🤔  is it just about keeping the command line as an (ordered) list rather than an (unordered) set?

test/expect_build_failure/collect_actions.bzl
```diff
@@ -0,0 +1,100 @@
...
+    # does not change when a flag's value is swapped for one that some other flag
+    # already carries, and `unused_dependency_checker_mode = "error"` -> `"off"`
```
PawelLipski: Also, the example is unclear 🤔 (why this one specifically)

test/expect_build_failure/collect_actions.bzl
```diff
@@ -0,0 +1,100 @@
...
+    implementation = _collect_impl,
+    attr_aspects = ["*"],
+)
+
+def _toolchains_transition_impl(_settings, attr):
+    # Analyse the fixture under the toolchains the nested `bazel` will register,
+    # so that changing one of them changes the command lines collected above.
+    # Whatever the outer build registered is replaced, not extended: the nested
+    # `bazel` gets these and nothing else.
+    return {"//command_line_option:extra_toolchains": attr.extra_toolchains}
```
PawelLipski, line 71: Wow, is it possible to swap ALL the available toolchains to a new list so easily? 😯

test/strict_dependency/scala_proto_deps/BUILD
bor-p-s: done, missed that.

test/strict_dependency/BUILD
bor-p-s: Yeah, doesn't worth it.

test/expect_build_failure/collect_actions.bzl
```diff
@@ -0,0 +1,100 @@
...
+    implementation = _toolchains_transition_impl,
+    inputs = [],
+    outputs = ["//command_line_option:extra_toolchains"],
+)
+
+# Every action this ruleset runs invokes a tool built from here, so a fingerprint
+# without it describes a target the rules never touch.
+_RULESET_TOOLS = "src/java/io/bazel/rulesscala"
+
+def _exposed_impl(ctx):
```
PawelLipski: Nit: why "exposed" impl?

PawelLipski: (and not just e.g. `_fixture_actions_impl`)

test/expect_build_failure/README.md
```diff
@@ -0,0 +1,95 @@
+# Asserting that a build or a test fails
```
PawelLipski, line 1: Nit: before reading in detail, is both the markdown and bzl doc needed? as in, do they overlap in the content? do both have sth that the other doesn't?

PawelLipski: Or maybe a better question: how is the division b/w what's here and what's in bzl defined?

test/expect_build_failure/expect_build_failure.bzl
```diff
@@ -115,10 +169,76 @@ def _nested_bazel_test(
     #     directory the nested `bazel <command>` must run in. See
     #     _nested_bazel_find_workspace in nested_bazel.sh.
     #   - nested_bazel.sh: sourced at runtime by the helper script.
+    # Everything the nested `bazel` reads and this test's own key would
+    # otherwise miss: the repo `.bazelrc` (CI steps append to it, so two steps
```
PawelLipski, line 178: Do CI steps still append to .bazelrc? wasn't CI bazelrc aligned with local recently in your PR? 🤔

test/expect_build_failure/expect_build_failure.bzl
```diff
@@ -21,15 +21,67 @@ A label inside a `build_args`/`bazel_args` flag value (e.g.
...
+_SCALAC_JAR = "//src/java/io/bazel/rulesscala/scalac:scalac_deploy.jar"
+
+# The compiler jar above does not pull in the test runners, and a nested
+# `bazel test` runs through them, so their content decides its outcome too.
+_RUNNER_JARS = [
+    "//src/java/io/bazel/rulesscala/test_discovery:test_discovery.jar",
+    "//src/java/io/bazel/rulesscala/specs2:specs2_test_discovery.jar",
+    "//src/java/io/bazel/rulesscala/scala_test:librunner.jar",
+    "//src/java/io/bazel/rulesscala/coverage/instrumenter:instrumenter.jar",
+]
```
PawelLipski, line 87: Actually, how was the list of these jars determined? (as in, how do know we that these ones and not more/less)

test/classpath_resource_conflict/BUILD
bor-p-s: Right catch — the comment was stale. This target is a concrete label (not a pattern) in another package, so it actually goes through `fixture_actions` and gets a real fingerprint now, it's not tagged `external`. Fixed the comment to say that.

PawelLipski: Okay... so only **patterns**  don't go through fingerprinting right? could you re-iterate what's the problem with them?

test/twitter_scrooge_jdk11/BUILD
bor-p-s: `fixture_actions` doesn't declare `target` as `data`, it dumps the action command lines (mnemonic + argv) from the transitive closure, so a content-only source edit that doesn't change the command line wouldn't be caught. Moot here anyway: this fixture is a `...` pattern in another package, so it's tagged `external`, not fingerprinted.

test/expect_build_failure/nested_bazel.sh
bor-p-s: Yes, needed for this PR. Before this change, the nested `bazel` always read your real `~/.bazelrc`, so its content could silently affect the test's outcome without being part of the cache key. That was already true before this PR, but it didn't matter yet, because nothing checked whether these tests were cached correctly. This PR is exactly what starts checking that, so the undeclared `~/.bazelrc` had to go.

test/expect_build_failure/expect_build_failure.bzl
bor-p-s: Correct. `local` opts the test out of remote/shared caching entirely. `no-sandbox` only skips the sandbox and still lets the result be cached, which is what we want here.

test/expect_build_failure/collect_actions.bzl
bor-p-s: Good catch, and yes, real gap. An action's command line only carries a file's path, not its content, so a content-only change to a file rules_scala itself bakes in wouldn't move the fingerprint. This is exactly why `_SCALAC_JAR`/`_RUNNER_JARS` in `expect_build_failure.bzl` are declared as explicit `data` instead of relying on the fingerprint. If another such file ever shows up, the fix is the same: add it to `data` by hand. Not generalizing this (e.g. hashing every input) since it's rare and the per-case fix is cheap.

test/expect_build_failure/collect_actions.bzl
bor-p-s: Added a comment: the no-slash branch is for a token that ends right at the configuration directory, nothing follows it, so there's no anchor to replace up to; it's left unscrubbed in that rare case.

test/expect_build_failure/collect_actions.bzl
bor-p-s: Reworded. It's not about order, it's about pairing: several rules_scala flags share the same off/warn/error values, so a flat depset of individual tokens can't tell which flag holds which value, and can miss a value moving from one flag to another.

test/expect_build_failure/collect_actions.bzl
bor-p-s: Picked that pair because it's real: `unused_dependency_checker_mode`, `strict_deps_mode`, and `compiler_deps_mode` all draw from the same off/warn/error set, so a value can genuinely migrate between them. Reworded the comment to name that instead of just one flag.

test/expect_build_failure/collect_actions.bzl
bor-p-s: Yep. `//command_line_option:extra_toolchains` written by a transition replaces the whole flag value for that configuration, it doesn't merge with whatever the outer build registered.

test/expect_build_failure/collect_actions.bzl
bor-p-s: Fair, renamed to `_fixture_actions_impl` to match the rule it implements.

test/expect_build_failure/README.md
bor-p-s: Some overlap, yes. README is the entry point: pick between the two mechanisms, the caching rationale, what to delete later. The `expect_build_failure.bzl` docstring is the source of truth for the actual cache-key list, kept next to the code so it can't drift from it. The overlap you're pointing at (`ctx.actions.write` actions) is one fact stated in both places; happy to trim it down to one if you'd rather.

test/expect_build_failure/expect_build_failure.bzl
bor-p-s: Yes, still true: `.bazelci/presubmit.yml` has several steps doing `echo ... >>.bazelrc` (`lockfile_mode=update`, `test_env=PATH`, the JDK 21 flags on one step). Checked just now.

test/expect_build_failure/expect_build_failure.bzl
bor-p-s: Each jar is a test rule's default dependency for the path a nested `bazel test` actually runs through: `scala_junit_test.bzl:74` (JUnit runner), `scala.bzl:47` (Specs2 discovery), `scala_test.bzl:78` (Scala test runner), `common_attributes.bzl:78` (coverage instrumenter). Found by searching the repo for every default label under `src/java/io/bazel/rulesscala` that a test rule wires in.

bor-p-s: Added a comment on the list: no automatic check keeps it complete, it's kept in sync by hand, same tradeoff as _SCALAC_JAR.

test/twitter_scrooge_jdk11/BUILD
PawelLipski: Yes, the point was more general about `fixture_actions`, not just about `test/twitter_scrooge_jdk11` - still, such minor omission feels acceptable

test/expect_build_failure/collect_actions.bzl
PawelLipski: Oh! okay
So not really such a corner case after all

test/expect_build_failure/README.md
PawelLipski: Okay, I'll read both in more detail now and comment if needed

test/expect_build_failure/expect_build_failure.bzl
PawelLipski: Oh okay I've missed that

test/classpath_resource_conflict/BUILD
PawelLipski: (@bor-p-s in case you've missed - mostly out of curiosity, not critical)

test/twitter_scrooge_jdk11/BUILD
PawelLipski: > content-only source edit that doesn't change the command line wouldn't be caught

Oh okay, addressed elsewhere, I see now

test/expect_build_failure/collect_actions.bzl
```diff
@@ -0,0 +1,104 @@
...
+        slash = part.find("/")
+        # No slash means the token ends right at the configuration directory (no
+        # path below it), so there's nothing to anchor the replacement on; leave
+        # it as-is rather than guess where the configuration name ends.
+        out += "bazel-out/" + ("<cfg>" + part[slash:] if slash != -1 else part)
+    return out
+
+def _collect_impl(target, ctx):
+    # One line per action, joining mnemonic and argv so each action keeps its own
+    # flag-to-value pairing. Collecting individual tokens into one depset would
```
PawelLipski, line 40: Wait, why are actually "tokens" (as in, shell tokens?) stored in a **depset**? isn't a depset meant to keep, well, deps? is it some off-the-label use?

test/classpath_resource_conflict/BUILD
```diff
@@ -9,9 +9,9 @@ package(default_testonly = 1)
 # expect_build_success_test and assert the warning is surfaced. (See that macro's
 # docstring for how it keeps a cached action from hiding the warning.)
 #
-# `target` lives in another package, so its sources aren't tracked as this test's
-# runfiles (see expect_build_failure.bzl) and an edit there won't force a re-run;
-# the fixture is long-standing and rarely touched.
+# `target` lives in another package, but it's one concrete label, so the macro
```
PawelLipski: Comment slightly redundant (up to you if it's worth keeping)

test/expect_build_failure/expect_build_failure.bzl
```diff
@@ -21,15 +21,71 @@ A label inside a `build_args`/`bazel_args` flag value (e.g.
...
+Editing `.bazelversion` re-runs the tests. What that file cannot catch is a CI
+step which leaves it alone and points bazelisk at another binary. One step does
+that (`last_green`), and it keeps its own cache key regardless, because it
+appends three flags to `.bazelrc`, which is on the list above.
+
+Outside the list is whatever the network serves while external repositories are
+fetched. `MODULE.bazel.lock` pins the versions, so what is exposed is a flaky
+download rather than a silent change in what is being tested. Catching an input
+nobody thought of would take a run with `--nocache_test_results` on the branch
+we merge into, which means a step in `.bazelci/presubmit.yml`; this change does
```
PawelLipski, line 57: Wait, is there an existing step like that? or is such a step only proposed here?

test/expect_build_failure/expect_build_failure.bzl
```diff
@@ -115,10 +174,76 @@ def _nested_bazel_test(
...
+    # `target` may be a pattern (a wildcard like `//pkg/...`), which has no
+    # single configured target to read actions from and no single label a
+    # cross-package caller could declare as `data`. A concrete label doesn't
+    # have that problem either way, so only a cross-package *pattern* is
+    # unfixable without the caller's help.
+    fingerprint_target = fingerprint_target or target
+    fingerprint_target_is_pattern = "..." in fingerprint_target or fingerprint_target.endswith((":all", ":*"))
+
+    if no_fingerprint_reason or (cross_package_fixture and fingerprint_target_is_pattern):
+        # The package glob above cannot cross a Bazel package boundary, and a
```
PawelLipski: > package glob above

Does it refer to `code_under_test`? maybe state explicitly

test/expect_build_failure/expect_build_failure.bzl
```diff
@@ -115,10 +174,76 @@ def _nested_bazel_test(
...
+        # elsewhere in this file.
+        fixture_actions(
+            name = "%s_fixture_actions" % name,
+            target = fingerprint_target,
+            extra_toolchains = extra_toolchains,
+            testonly = True,
+        )
+        data.append(":%s_fixture_actions" % name)
+
+    # `no-sandbox` rather than `local`: both run the nested `bazel` outside the
```
PawelLipski: Somewhat misplaced comment, `no-sandbox` isn't declared here

test/expect_build_failure/expect_build_failure.bzl
```diff
@@ -143,10 +268,10 @@ def _nested_bazel_test(
         srcs = [_HELPER],
         args = args,
         data = deduped_data,
-        # Each of these runs a nested `bazel` that takes the shared nested output
-        # base's lock. Run them one at a time (`exclusive`) so they do not queue
-        # behind each other and blow the test timeout.
-        tags = tags + ["exclusive"],
+        # `exclusive`: each runs a nested `bazel` that takes the shared output
```
PawelLipski: Same here, misplaced comment

test/strict_dependency/scala_proto_deps/BUILD
```diff
@@ -85,12 +85,14 @@ expect_build_failure_test(
     name = "stamp_by_convention_test",
     build_args = ["--extra_toolchains=//test/toolchains:ast_plus_one_deps_strict_deps_error,//test/strict_dependency/scala_proto_deps:stamp_by_convention_toolchain"],
     expect = ["expected_stamp_by_convention.txt"],
+    no_fingerprint_reason = "under this toolchain combination collides with custom_phase_stamping_test below on :some_proto's aspect application (see that test's own no_fingerprint_reason)",
```
PawelLipski, line 88: Actually, how did you detect that tests like this can't have a fingerprint? does analysis fail?

test/expect_build_failure/collect_actions.bzl
```diff
@@ -0,0 +1,104 @@
...
+    out = parts[0]
+    for part in parts[1:]:
+        slash = part.find("/")
+        # No slash means the token ends right at the configuration directory (no
+        # path below it), so there's nothing to anchor the replacement on; leave
+        # it as-is rather than guess where the configuration name ends.
+        out += "bazel-out/" + ("<cfg>" + part[slash:] if slash != -1 else part)
+    return out
+
+def _collect_impl(target, ctx):
```
PawelLipski: Same as before, rename to `_collect_actions_aspect_impl` per convention

test/expect_build_failure/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+That nested build reads the real source tree, which the outer Bazel cannot see,
+so the test has to declare by hand everything that decides its outcome. The
+macro does that for you; the docstring in `expect_build_failure.bzl` lists what
+goes into the cache key and what stays outside it.
+
+Nothing re-runs these tests without the result cache on a schedule. That would be
+the way to catch an input nobody thought to declare.
+
+## What to delete later
```
PawelLipski, line 28: Pls create relevant issues once this PR lands

PawelLipski: Or in fact, just move this section to GitHub issues (and only leave links with short summaries)

test/expect_build_failure/README.md
```diff
@@ -0,0 +1,101 @@
...
+  time rather than being served a pass it cannot vouch for. A few tests do
+  that today, for one of four reasons: `scala_proto` generates through an
+  aspect of its own, whose actions this aspect cannot read; the `semanticdb`
+  fixtures only build toolchain deps, so they compile nothing at all; a
+  fixture targets a plain `proto_library`/`java_library`, which runs no
+  rules_scala action either; or fingerprinting two fixtures under the same
+  `--extra_toolchains` would analyse a shared dependency under two different
+  aspect stacks, which Bazel reports as an `ActionConflictException`.
+
+What the fingerprint does not see: an action that writes a file instead of
```
PawelLipski, line 80: Also mention the change in rules_scala-declared file inputs

test/expect_build_failure/README.md
```diff
@@ -0,0 +1,101 @@
...
+running a command line (`ctx.actions.write`, template expansion) contributes its
+mnemonic and nothing else, which is 446 of the 5135 lines it collects today. A
```
PawelLipski: Hmm it's unclear what the specific numbers refer to

Is it from the fingerprint of some specific target?

test/expect_build_failure/README.md
```diff
@@ -0,0 +1,101 @@
...
+change to what such an action writes leaves the fingerprint identical. Declaring
+those files instead would put output paths -- and with them the configuration --
```
PawelLipski: Also unclear 🤔  what _those_ refer to?

test/classpath_resource_conflict/BUILD
bor-p-s: Right, only patterns. The problem: `fixture_actions`'s aspect needs one concrete target to read actions from, a pattern like `//pkg/...` has no single configured target. And there's no single label the caller could declare as `data` either, so there's no fallback. A concrete label doesn't have either problem.

test/expect_build_failure/collect_actions.bzl
bor-p-s: Not shell tokens. Each action's mnemonic and args are joined into one string per action (line 48). The depset holds these strings, not individual words.

Depset here merges each action's line with the lines already collected from its dependencies, across the whole graph, without flattening the list at every level.

test/classpath_resource_conflict/BUILD
bor-p-s: Agreed, trimmed to a cross-reference to the macro's docstring instead of restating the mechanism.

test/expect_build_failure/expect_build_failure.bzl
bor-p-s: Only proposed, doesn't exist. Checked .bazelci/presubmit.yml: no --nocache_test_results step there.

test/expect_build_failure/collect_actions.bzl
PawelLipski: Okay! so off-label use IIUC, but smart/justified

test/expect_build_failure/expect_build_failure.bzl
bor-p-s: Yes. Renamed the reference to `code_under_test` explicitly in the comment.

test/expect_build_failure/expect_build_failure.bzl
bor-p-s: Right on both. Consolidated into one comment at the actual `execution_tags = tags + [...]` line -- that's where `exclusive`/`no-remote-exec` get added and where the `no-sandbox` vs `local` tradeoff belongs, since `no-sandbox` is a caller default, not decided here.

test/expect_build_failure/expect_build_failure.bzl
bor-p-s: Fixed the same way -- both comments got folded into the one at `execution_tags = tags + [...]` (see the reply above).

test/strict_dependency/scala_proto_deps/BUILD
bor-p-s: Yes, analysis fails -- that's exactly what happened when I tried fingerprinting this one: Bazel's analysis phase itself failed with a real `ActionConflictException` on `:some_proto`'s generated jar, not something the macro's own `fail()` guard caught. The two tests analyse `:some_proto` under different aspect stacks in the same configuration, and Bazel won't allow two different action graphs to produce the same output path.

test/expect_build_failure/collect_actions.bzl
bor-p-s: Done.

test/expect_build_failure/README.md
bor-p-s: Added: the same blind spot applies to any file rules_scala reads by path rather than content -- `_SCALAC_JAR`/`_RUNNER_JARS` are declared as explicit `data` for exactly that reason.

test/expect_build_failure/README.md
bor-p-s: Removed.

test/expect_build_failure/README.md
bor-p-s: Fixed.

test/expect_build_failure/README.md
bor-p-s: Agreed, will make issues after this lands and link them here.

PawelLipski (approved):
Good job! it  was very non-trivial 😅

PR #1908, opened 2026-08-07
refactor: Migrate test_toolchain_deps.sh and test_twitter_scrooge.sh to Bazel

bor-p-s:
### Description

Migrates test_toolchain_deps.sh to 4 expect_build_success_test targets in
test/semanticdb (one per Scala version, building
//scala/private/toolchain_deps:semanticdb), and test_twitter_scrooge.sh to one
in test/src/main/scala/scalarules/test/twitter_scrooge, building that
package's wildcard under JDK 11. Continues #1900/#1882/#1881/#1880/#1864.

toolchain_deps is production code outside test/semanticdb, so the macro's
package-scoped glob doesn't track it for staleness; each test also depends on
the built target directly (see comment in test/semanticdb/BUILD).

The scrooge target already builds under the repo's default
java_runtime_version=17, so this test confirms it also builds under 11 rather
than only under 11 - dropping the JDK 11 flags does not turn it red.

### Motivation

Deletes both scripts and their test_rules_scala.sh lines.
test/shell/scala_versions.sh stays; test_version.sh still uses it.

Release impact: tests only. Adds 4 more serialized (exclusive) nested-bazel
tests; worth checking the real CI step-time delta (locally, warm cache, they
run ~5-6s each).


test/semanticdb/BUILD
```diff
@@ -75,3 +79,26 @@ scala_library(
         allow_empty = True,
     ),
 )
+
+# Scala versions the toolchain_deps semanticdb classpath is resolved against,
+# mirroring test/shell/scala_versions.sh.
```
PawelLipski: Will `test/shell/scala_versions.sh` be gone eventually?
If so, don't reference it

test/twitter_scrooge_jdk11/BUILD
```diff
@@ -0,0 +1,29 @@
+load(
+    "//test/expect_build_failure:expect_build_failure.bzl",
+    "expect_build_success_test",
+)
+
+# Not declared in test/src/main/scala/scalarules/test/twitter_scrooge itself:
+# that package's :twitter_scrooge_tests is built directly (as a non-root bzlmod
+# dependency) by test_dependency_versions.sh, where rules_shell -- a
+# dev_dependency in the root MODULE.bazel -- is invisible. Loading this
+# macro's `@rules_shell` reference from that package broke that job. This
```
PawelLipski: _that job_ - the reference is unclear 🤔

test/twitter_scrooge_jdk11/BUILD
```diff
@@ -0,0 +1,29 @@
+load(
+    "//test/expect_build_failure:expect_build_failure.bzl",
+    "expect_build_success_test",
+)
+
+# Not declared in test/src/main/scala/scalarules/test/twitter_scrooge itself:
+# that package's :twitter_scrooge_tests is built directly (as a non-root bzlmod
+# dependency) by test_dependency_versions.sh, where rules_shell -- a
```
PawelLipski: Also, once `test_dependency_versions.sh` is gone, will the comment still stand?

test/semanticdb/BUILD
bor-p-s: fixed!

test/twitter_scrooge_jdk11/BUILD
bor-p-s: reworded, sorry.

test/twitter_scrooge_jdk11/BUILD
bor-p-s: fixed!

test/twitter_scrooge_jdk11/BUILD
```diff
@@ -0,0 +1,30 @@
...
+    "//test/expect_build_failure:expect_build_failure.bzl",
+    "expect_build_success_test",
+)
+
+# Not declared in test/src/main/scala/scalarules/test/twitter_scrooge itself:
+# rules_shell is a dev_dependency in the root MODULE.bazel, so it is invisible
+# whenever something else consumes rules_scala as a non-root bzlmod
+# dependency. test_dependency_versions.sh does exactly that, building
+# :twitter_scrooge_tests directly, so loading this macro's `@rules_shell`
+# reference from that same package broke test_dependency_versions.sh's build.
```
PawelLipski: Hmmm `test_dependency_versions.sh` will also be gone right? 😅  this comment will get stale quickly, rewrite in a time-proof way

PawelLipski (approved):

PawelLipski (approved):

test/twitter_scrooge_jdk11/BUILD
bor-p-s: fixed

PR #1910, opened 2026-08-07
Build the scrooge worker input with Args so it is path-mapping safe

FrankPortman:
`ScroogeRule` fails under `--experimental_output_paths=strip`:

```
scrooge_worker @bazel-out/k8-fastbuild/bin/foo/test_thrift_scala_worker_input
   ^ executable mapped (bazel-out/cfg/bin/...)    ^ argfile not mapped
NoSuchFileException: bazel-out/k8-fastbuild/bin/foo/test_thrift_scala_worker_input
```

The worker input is built with `ctx.actions.write`, so it is materialized at analysis time and its content embeds the output path and every input path. Path mapping rewrites command lines, not file contents, so those paths are never rewritten. Passing the argfile itself as a `File` would not help — the stale paths are inside the file.

Building it with `Args` and a multiline param file lets Bazel write it at execution time, where paths are mapped. The `Scalac` action in this repo already does exactly this.

The format is unchanged — output, compile thrifts, include thrifts, two padded blanks, flags. `add_joined` preserves the sort and `":"` separator `_colon_paths` provided, and `omit_if_empty = False` keeps the blank lines that the worker-argument padding relies on. The declared argfile goes away since Bazel owns the param file.

### Verification

Patched into a repo with 89 `thrift_library` targets and `--modify_execution_info=ScroogeRule=+supports-path-mapping`:

- the target that previously failed with the error above builds
- 25 thrift targets built, and the scrooge jars are **byte-identical** to an unpatched build with mapping off

Separate from #1909, which fixes three unrelated one-line cases of the same underlying problem.

### DIFF
--- a/twitter_scrooge/twitter_scrooge.bzl
+++ b/twitter_scrooge/twitter_scrooge.bzl
@@ -17,9 +17,6 @@ load(
 load("//thrift:thrift.bzl", "merge_thrift_infos")
 load("//thrift:thrift_info.bzl", "ThriftInfo")
 
-def _colon_paths(data):
-    return ":".join([f.path for f in sorted(data)])
-
 ScroogeAspectInfo = provider(fields = [
     "thrift_info",
     "src_jars",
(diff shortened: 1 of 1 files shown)


PawelLipski (approved):

bor-p-s (approved):
Looked at this and the fix looks right. Args with `add_joined(omit_if_empty=False)` writes the same lines as the old `ctx.actions.write` version (output, compile thrifts, include thrifts, two pad lines, flags), so the worker input format stays the same. `use_param_file("@%s", use_always=True)` keeps the old behavior of always passing `@paramfile`.

On the test question: a build test with `--experimental_output_paths=strip` would hit unrelated toolchain failures before it even reaches Scrooge (per Frank's notes above), and a structural test only sees analysis time, while path mapping happens at execution time. The unchanged file format plus the byte-identical check against 25 real targets covers this well enough.

LGTM, worth merging.

PR #1911, opened 2026-08-11
refactor: Migrate test_diagnostics_reporter.sh to sh_test targets

bor-p-s:
### Description

Moves `test_diagnostics_reporter.sh` off the legacy shell runner. Its two checks (diagnostics reporter produces the right proto diagnostics per fixture, plain and with semanticdb) become two `sh_test` targets in `test/diagnostics_reporter`, next to the fixtures and the existing verifier binary. The 5 fixture `scala_library` targets move there too (from `test_expect_failure/diagnostics_reporter`); 3 of them (`error_file`, `two_errors_file`, `error_and_warning_file`) fail to compile by design, so they're tagged `manual` to stay out of the default `test/...` wildcard build. The other two (`warning_file`, `info_file`) compile fine and carry no such tag.

This check needs two nested `bazel` invocations (build the fixtures under a toolchain, then run the verifier against the resulting `.diagnosticsproto` files), which no existing `expect_*_test` macro supports (they assert on one). New custom `diagnostic_proto_files_test.sh`, following the same pattern as the existing `test/macros/expect_no_ijar.sh`.

`test_version.sh`'s `test_diagnostic_proto_files` (a separate cross-Scala-version CI job) also hardcoded the old `test_expect_failure/diagnostics_reporter` labels directly against this repo; fixed those too, and switched its `:all` wildcard build to an explicit target list, since a package's `:all` wildcard excludes manual targets the same way `...` does.

Dropped `--build_event_publish_all_actions` from both the new test and `test_version.sh`'s nested build. Traced it to the diagnostics reporter's original PR (#1068, 2020): it's already there in the initial diff, carried forward unchanged since, with nothing nearby explaining it. Verified empirically instead: dropping it, the fixtures still produce the same set of `.diagnosticsproto` files locally.

### Motivation

Deletes `test_diagnostics_reporter.sh` and its `test_rules_scala.sh` line.

Release impact: tests only. Adds 2 more serialized (`exclusive`) nested-bazel tests to the suite. Measured locally across several runs, ~12-16s each; CI numbers will come from this PR's own run.


test_version.sh
```diff
@@ -133,8 +133,18 @@ test_diagnostic_proto_files() {
   local SCALA_VERSION="$1"
   local SCALA_TOOLCHAIN="$2"
 
-  compilation_should_fail build --build_event_publish_all_actions -k --repo_env=SCALA_VERSION=${SCALA_VERSION} --extra_toolchains=${SCALA_TOOLCHAIN} //test_expect_failure/diagnostics_reporter:all
-  diagnostics_output="$(bazel info bazel-bin)/test_expect_failure/diagnostics_reporter"
+  # Listed explicitly, not as //test/diagnostics_reporter:all: 3 of these 5
+  # fixtures are tagged "manual" so a plain `test/...` build skips them, and
+  # :all excludes manual targets the same way `...` does. Using :all here
```
PawelLipski, line 138: Whoops was it the problem already previously? that the two manual targets were silently skipped?

test/diagnostics_reporter/BUILD
```diff
@@ -16,3 +19,127 @@ java_binary(
...
+    "//test/diagnostics_reporter:error_file",
+    "//test/diagnostics_reporter:two_errors_file",
+    "//test/diagnostics_reporter:warning_file",
+    "//test/diagnostics_reporter:error_and_warning_file",
+    "//test/diagnostics_reporter:info_file",
+]
+
+# Bazel has no native "run bazel, then run a verifier against its output"
+# rule, and unlike expect_build_failure.bzl's macros this test needs two
+# nested-bazel steps (build the fixtures, then run the checker against the
```
PawelLipski, line 119: Wow that's a new case 😅

test/diagnostics_reporter/diagnostic_proto_files_test.sh
```diff
@@ -0,0 +1,38 @@
...
+  exit 2
+fi
+
+nested_bazel_setup "rules_scala_diagnostics_reporter_output_base"
+
+# Swallowing this build's exit code doesn't make the check blind: the
+# verifier below asserts each fixture's specific diagnostic message, so if a
+# fixture stopped failing (or reporting) the way it's supposed to, the
+# expected message just wouldn't be found and the verifier would fail.
+nested_bazel_run build -k --extra_toolchains="${toolchain}" "${fixtures[@]}" || true
```
PawelLipski: Does `-k` has a long option equivalent? short flags look unclear

test_expect_failure/diagnostics_reporter/BUILD
```diff
@@ -1,78 +0,0 @@
-load("//scala:scala.bzl", "scala_library")
```
PawelLipski, line 1: Anything changed from this BUILD file? looks like moved 1:1 to `test/diagnostics_reporter/BUILD` right?

PawelLipski (approved):

test_version.sh
bor-p-s: 3 of the 5, actually: error_file, two_errors_file, error_and_warning_file. And no, wasn't already broken: in test_expect_failure they were plain rules with no tags, package lived outside test/... entirely, so :all just worked there. Became an issue only once they moved into test/..., which is why the build here lists targets explicitly instead.

test/diagnostics_reporter/BUILD
bor-p-s: First time we needed this for expect_build_failure.bzl. But it's not new to the repo: test/macros/expect_no_ijar.sh already builds, then checks the output, the same way, for a case the macros don't cover.

test/diagnostics_reporter/diagnostic_proto_files_test.sh
bor-p-s: Yes, --keep_going. Switched to it here and in test_version.sh's build too.

test_expect_failure/diagnostics_reporter/BUILD
bor-p-s: Mostly yes. The scala_toolchain/toolchain/scala_library rules moved as-is, I just added tags = ["manual"] to 3 of the 5 scala_library rules, didn't need it before since this package was outside test/... entirely. The two sh_test targets and the fixture list are new, not part of the move.

PawelLipski (approved):

PR #1912, opened 2026-08-11
refactor: Migrate plus-one dependency-mode compilation tests to Bazel

PawelLipski:
### Description

Migrates `test/shell/test_compilation.sh` (plus-one dependency *mode*) to native `bazel test`. That script ran nested `bazel build`s of the `test_expect_failure/plus_one_deps/{internal,external,exports_of,deps_of}_*` fixtures — one success case per aspect of plus-one propagation, plus one failure case under the default toolchain — and asserted the exit code.

- Converts the 5 shell cases into 6 native tests (the `deps_of_exports` package wildcard is split into its two top targets):
  - `internal_deps`: `expect_build_success_test` (with the plus-one toolchain) + `expect_build_failure_test` (default toolchain → strict-deps error).
  - `external_deps`, `exports_of_deps`, `deps_of_exports` (facade + real-dep): `expect_build_success_test` with the plus-one toolchain.
- Colocates the fixtures under `test/plus_one_deps/` (next to the existing `with_unused_deps` package migrated in #1873). Every fixture `scala_library` is tagged `manual` so a default-toolchain wildcard build skips the ones that only compile under plus-one; the nested build in each test names the target explicitly (which bypasses `manual`) and supplies the plus-one toolchain.
- The shared `plus_one_deps` toolchain stays in `test_expect_failure/plus_one_deps` because `test_strict_dependency.sh` and the global `test_rules_scala.sh` pass still reference it.
- Removes `test/shell/test_compilation.sh` and its `test_rules_scala.sh` entry.

### Motivation

Part of migrating bespoke shell tests to native `bazel test` and shrinking `test_expect_failure/` (four more subpackages moved out here).


test/plus_one_deps/exports_of_deps/BUILD.bazel
```diff
@@ -0,0 +1,40 @@
...
+
+scala_library(
+    name = "exported_dep",
+    srcs = ["D.scala"],
+    tags = ["manual"],
+)
+
+expect_build_success_test(
+    name = "plus_one_deps_exports_of_deps_test",
+    build_args = ["--extra_toolchains=//test_expect_failure/plus_one_deps:plus_one_deps"],
```
PawelLipski, line 38: The toolchain is kept in `//test_expect_failure` because two still-unmigrated shell consumers reference it, so it can't move without editing those.

test/plus_one_deps/deps_of_exports/A.scala
```diff
@@ -0,0 +1,5 @@
+package scalarules.test_expect_failure.plus_one_deps.deps_of_exports
```
PawelLipski, line 1: Also fixed the packages while I was here (was `...internal_deps` everywhere, copy-paste mistake)

test/plus_one_deps/deps_of_exports/BUILD.bazel
```diff
@@ -0,0 +1,78 @@
...
+    target = ":test_target_using_facade",
+)
+
+expect_build_success_test(
+    name = "plus_one_deps_deps_of_exports_test",
+    build_args = ["--extra_toolchains=//test_expect_failure/plus_one_deps:plus_one_deps"],
+    target = ":test_target",
+)
+
+# Negative control: as above, but reaching the exported dep through `deps` +
```
PawelLipski: Added missing negative controls (missing from original tests)

test/plus_one_deps/deps_of_exports/BUILD.bazel
```diff
@@ -0,0 +1,78 @@
...
+expect_build_success_test(
+    name = "plus_one_deps_deps_of_exports_facade_test",
+    build_args = ["--extra_toolchains=//test_expect_failure/plus_one_deps:plus_one_deps"],
+    target = ":test_target_using_facade",
+)
+
+# Negative control: without the plus-one toolchain, the plus-one dep hanging off
+# the exported (hence direct) dep is invisible, so the build must fail.
+expect_build_failure_test(
+    name = "plus_one_deps_deps_of_exports_facade_undefined_fails_test",
```
bor-p-s: `plus_one_deps_deps_of_exports_facade_undefined_fails_test` and `plus_one_deps_deps_of_exports_undefined_fails_test` in this file fail for the same reason. Both `test_target_using_facade` and `test_target` depend on `:direct_dep`, and it's `direct_dep` that fails to compile:

```
test/plus_one_deps/deps_of_exports/B.scala:3: error: Symbol 'type ...D' is missing from the classpath.
This symbol is required by 'class ...C'.
```

So the facade case fails on the same shared-target error before it ever reaches the facade-specific export resolution it's meant to test. Could we route it through a fixture where the facade's own export path is what's missing, or drop the duplicate?

PawelLipski: Done, dropped `plus_one_deps_deps_of_exports_facade_undefined_fails_test`

bor-p-s (approved):

PR #1913, opened 2026-08-12
refactor: Migrate missing_direct_deps checks in test_scala_library.sh

bor-p-s:
### Description

Moves the 10 missing-direct-deps checks from `test_scala_library.sh` into `test/missing_direct_deps/BUILD` as `expect_build_failure_test`/`expect_build_success_test` targets, covering each strict-deps mode (error, warn, off, default) plus the two custom-Java-provider label cases.

The fixture stays at `test_expect_failure/missing_direct_deps/` (not colocated): two other in-flight branches still migrate shell files that use it. Colocation is a follow-up once all three land.

The "off mode" case now asserts `not found: value C` instead of the original path+line message (which had a separate Windows variant): the new macro only does substring matching, not regex, so it can't branch on OS. Still confirms the intended behavior, a raw compiler error instead of a buildozer hint.

### Motivation

Deletes the 10 functions and their `$runner` lines; the file's 4 recompilation tests are untouched.

Release impact: tests only.

### DIFF
--- a/test/missing_direct_deps/BUILD
+++ b/test/missing_direct_deps/BUILD
@@ -1,0 +1,99 @@
+load(
+    "//test/expect_build_failure:expect_build_failure.bzl",
+    "expect_build_failure_test",
+    "expect_build_success_test",
+)
+
+# Fixture lives in test_expect_failure/missing_direct_deps -- shared by this
+# package and by test/shell/test_scala_binary.sh and
+# test/shell/test_strict_dependency.sh (migrating separately), so it stays put
+# instead of being colocated here.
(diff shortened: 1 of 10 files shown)


test/missing_direct_deps/BUILD
```diff
@@ -0,0 +1,99 @@
+load(
+    "//test/expect_build_failure:expect_build_failure.bzl",
+    "expect_build_failure_test",
+    "expect_build_success_test",
+)
+
+# Fixture lives in test_expect_failure/missing_direct_deps -- shared by this
```
PawelLipski, line 7: Ok, to be cleaned up soon

PawelLipski (approved):
Looks like a straightforward translation, approving

PR #1915, opened 2026-08-12
refactor: Migrate test_strict_dependency.sh to sh_test targets

bor-p-s:
### Description

Moves the 10 checks from `test_strict_dependency.sh` into 12 `sh_test` targets: `test/strict_dependency` plus three subpackages, `scala_import`, `filtering`, `scala_proto_deps`. One check splits into 3 targets, one per toolchain it asserts against.

`scala_import`, `filtering`, and `scala_proto_deps` fixtures move with their tests; nothing else in the repo references them. `internal_deps`, `external_deps`, and the `plus_one_deps` toolchain stay in `test_expect_failure/`: `test_scala_library.sh` and `test_scala_binary.sh` still use them (#1913, #1914, in flight). Colocate those once both land.

4 of the 10 checks assert a buildozer message naming a bzlmod repo or a self-repo label, neither a fixed string across the Bazel-version matrix this suite runs under (on 7.7.1: `@@+scala_deps+io_bazel_rules_scala_guava_2_12_21`). Added `expect_regex`/`reject_regex` to `expect_build_failure_test`/`expect_build_success_test` (`grep -E` instead of `--fixed-strings`) for these.

Folded the resulting four near-identical match loops in `expect_build_failure.sh` into one helper.

### Motivation

Deletes `test_strict_dependency.sh` and its `test_rules_scala.sh` line.

Release impact: tests only. Adds 12 `exclusive` nested-bazel tests; 7-17s each locally (warm disk cache), CI numbers from this PR's run.


test/strict_dependency/BUILD
```diff
@@ -0,0 +1,47 @@
...
+# The fixtures these tests build (test_expect_failure/missing_direct_deps/
+# {internal_deps,external_deps}, test_expect_failure/plus_one_deps) are left in
+# place rather than colocated: they're also referenced by the not-yet-migrated
+# test_scala_library.sh and test_scala_binary.sh, so moving them here would
+# break those. Once both migrate, re-check whether these are still referenced
+# elsewhere before colocating them as a follow-up.
+
+expect_build_success_test(
+    name = "plus_one_deps_only_works_for_java_info_targets_test",
+    build_args = ["--extra_toolchains=//test_expect_failure/plus_one_deps:plus_one_deps"],
```
PawelLipski, line 9: > plus_one_deps toolchain stay in test_expect_failure/: test_scala_library.sh and test_scala_binary.sh still use them (https://github.com/bazel-contrib/rules_scala/pull/1913, https://github.com/bazel-contrib/rules_scala/pull/1914, in flight)

Looks that both these PRs are merged, can these be moved out of `test_expect_failure`?

test/strict_dependency/scala_proto_deps/BUILD
```diff
@@ -0,0 +1,95 @@
...
+    # Missing a direct dep under strict deps even by default; the test below
+    # builds it under a strict-deps-error toolchain either way.
+    tags = ["manual"],
+    deps = [":custom_transitive"],
+)
+
+# The stamped labels below name a self-repo canonical target
+# (@@//test/strict_dependency/scala_proto_deps:...); whether Bazel prints the
+# leading "@@" varies across the Bazel-version matrix this suite runs under,
+# hence expect_regex (with an optional "@@?"), not expect.
```
PawelLipski: Aww okay this is a regex 🤦🏻

test/expect_build_failure/expect_build_failure.bzl
```diff
@@ -15,6 +15,12 @@ runfiles, and boilerplate tags on every call.
   only passes under a specific `--test_filter` or inherited env var).
 
 All four share the same script and nested-Bazel plumbing.
+
+A label inside a `build_args`/`bazel_args` flag value (e.g.
```
PawelLipski, line 19: Pls extract these regex changes to a separate PR

Alternatively, would it be possible to craft the test cases so that the `@@`/`@` aren't printed? but not insisting if regex is cleaner

test/strict_dependency/BUILD
bor-p-s: Fixed — moved into `test/missing_direct_deps/`. Also fixed 7 expect files there and in `test/scala_binary_missing_direct_deps/` still pointing at the old path.

test/strict_dependency/scala_proto_deps/BUILD
bor-p-s: No longer one — see below.

test/expect_build_failure/expect_build_failure.bzl
bor-p-s: Found a way to avoid regex: `grep --fixed-strings` matches a substring, so the expect file just starts right after where `@@`/the separator would be. Reverted `expect_regex`/`reject_regex`, nothing left to extract.

test/expect_build_failure/expect_build_failure.bzl
```diff
@@ -15,6 +15,12 @@ runfiles, and boilerplate tags on every call.
...
+`--extra_toolchains=//pkg:toolchain`) is passed through as-is: unlike
+`target`, it is not absolutized against the caller's package, because the
+nested `bazel` runs from the workspace root. Always write such a label out in
+full, even for a toolchain defined in the same package as the caller.
```
PawelLipski, line 23: Does this comment reflect a behavior change? or was this statement always true, just not written explictly?

test/expect_build_failure/expect_build_failure.sh
```diff
@@ -169,26 +169,38 @@ if [[ "${expect_success}" != "true" && "${status}" -eq 0 ]]; then
...
-  expected_message="$(tr -d '\n' <"${resolved}")"
-  if ! grep --quiet --fixed-strings -- "${expected_message}" <<<"${output}"; then
-    echo "Nested \`bazel ${command}\` finished as expected, but output did not contain the expected message." >&2
-    echo "Expected (from ${expect_file}): ${expected_message}" >&2
-    echo "Output:" >&2
-    echo "${output}" >&2
-    exit 1
-  fi
-done
+# Checks each file in $2... against $output, failing if its (fixed-string)
```
PawelLipski, line 172: Is it a logic change? or just a refactor?

test/strict_dependency/scala_proto_deps/BUILD
```diff
@@ -0,0 +1,96 @@
...
+    # Missing a direct dep under strict deps even by default; the test below
+    # builds it under a strict-deps-error toolchain either way.
+    tags = ["manual"],
+    deps = [":custom_transitive"],
+)
+
+# The stamped labels below name a self-repo canonical target
+# (@@//test/strict_dependency/scala_proto_deps:...); whether Bazel prints the
+# leading "@@" varies across the Bazel-version matrix this suite runs under.
+# The expected message starts right after where that "@@" would be, at the
```
PawelLipski, line 74: 👍🏻

test/expect_build_failure/expect_build_failure.bzl
bor-p-s: Always true, not a behavior change — `_absolutize()` only ever applied to `target`, never `build_args`/`bazel_args`. Just consolidated the note that was duplicated across two docstrings.

test/expect_build_failure/expect_build_failure.sh
bor-p-s: Just a refactor — same trigger condition and same messages as the original two loops, verified by diffing against the pre-refactor code and by breaking each check locally.

PawelLipski (approved):

PR #1917, opened 2026-08-12
refactor: Migrate `test_junit.sh` to Bazel

PawelLipski:
# Based on PR #1923

## Chain of upstream PRs as of 2026-08-14

* PR #1923:
  `master` ← `refactor/expect-test-bazel-arg-file`

  * **PR #1917 (THIS ONE)**:
    `refactor/expect-test-bazel-arg-file` ← `refactor/junit-test-selection-as-bazel-test`

## Summary

Migrates all of `test/shell/test_junit.sh` to native Bazel targets in `//test/scala_junit_test` and deletes the script (and its `test_rules_scala.sh` line), continuing the move of bespoke shell tests off the legacy runner. The targets land next to the existing `scala_junit_test` failure-case tests, so all `scala_junit_test` behaviour lives in one package.

- **Class discovery + `--test_filter`** (`multiple_junit_suffixes_test`, `multiple_junit_prefixes_test`, `multiple_junit_patterns_test`, `junit_test_filter_test`, `junit_test_filter_custom_runner_test`) become `expect_test_success_test` wrappers. Each case's signal is the JUnit runner's own stdout -- the `print_discovered_classes` listing and each method's `println` -- which the nested `bazel test --test_output=all` surfaces, so `expect`/`reject` files match on it directly. `junit_test_filter_test` keeps the upstream cross-class alternation regex, passed through a file via `bazel_arg_files` (see #1923) so its metacharacters survive the Windows `sh_test` launcher.

- **`junit_generates_xml_logs`** becomes a plain `sh_test`. It asserts on the generated JUnit `test.xml` report, an on-disk artifact `bazel test` never prints to stdout, so it can't reuse the stdout-matching `expect_build_failure.bzl` macros. Instead it runs a nested `bazel test` and greps the report directly -- mirroring `test/diagnostics_reporter` (#1911), which likewise runs bazel then inspects its output.

`//test` grants `//test/scala_junit_test` visibility to its fixtures; each target under test is added to the wrapping test's `data` so editing its sources re-keys the test (the nested `bazel` reads the real source tree and is otherwise invisible to the cache key).

## Test plan

- [x] `bazel test //test/scala_junit_test:multiple_junit_suffixes_test //test/scala_junit_test:multiple_junit_prefixes_test //test/scala_junit_test:multiple_junit_patterns_test //test/scala_junit_test:junit_test_filter_test //test/scala_junit_test:junit_test_filter_custom_runner_test //test/scala_junit_test:junit_generates_xml_logs_test` -- all 6 pass.


test/sh_tests/BUILD
```diff
@@ -171,3 +172,119 @@ scala_test(
...
+
+# `--test_filter` narrows a run to a single method: only SecondFilterTest#method2
+# runs (its `println` appears), and no other method of any of the three classes
+# does (the eight others are negative controls).
+expect_test_success_test(
+    name = "junit_test_filter_test",
+    bazel_args = [
+        "--test_output=all",
+        "--nocache_test_results",
+        "--test_filter=scalarules.test.junit.SecondFilterTest#method2",
```
PawelLipski: The upstream shell test filtered a cross-class subset via a regex, but a regex's metacharacters (`(`, `|`) can't ride the `sh_test` args list -- the Windows launcher's `bash -c` would interpret them -- so this uses a single, metacharacter-free class#method selector instead, with the eight other methods as negative controls.

test/scala_junit_test/BUILD
```diff
@@ -86,3 +91,119 @@ expect_test_failure_test(
...
+
+# `--test_filter` narrows a run to a single method: only SecondFilterTest#method2
+# runs (its `println` appears), and no other method of any of the three classes
+# does (the eight others are negative controls).
+expect_test_success_test(
+    name = "junit_test_filter_test",
+    bazel_args = [
+        "--test_output=all",
+        "--nocache_test_results",
+        "--test_filter=scalarules.test.junit.SecondFilterTest#method2",
```
PawelLipski: The upstream shell test filtered a cross-class subset via a regex, but a regex's metacharacters (`(`, `|`) can't ride the `sh_test` args list -- the Windows launcher's `bash -c` would interpret them -- so this uses a single, metacharacter-free class#method selector instead, with the eight other methods as negative controls.

test/scala_junit_test/BUILD
```diff
@@ -86,3 +91,119 @@ expect_test_failure_test(
...
+        "junit_suffix_e2e.txt",
+    ],
+    target = "//test:JunitPrefixesAndSuffixes",
+)
+
+# `--test_filter` narrows a run to a single method: only SecondFilterTest#method2
+# runs (its `println` appears), and no other method of any of the three classes
+# does (the eight others are negative controls).
+expect_test_success_test(
+    name = "junit_test_filter_test",
```
bor-p-s, line 156: This drops the alternation case entirely (only `SecondFilterTest#method2` now, vs. the original 4-method group filter). Re your Windows note above: can we pass the filter through something that skips shell re-tokenization (data file / env var) instead of dropping alternation coverage?

PawelLipski: Done, extracted to `test/scala_junit_test/junit_test_filter_flags.txt`; added the support for args files in a new base PR https://github.com/bazel-contrib/rules_scala/pull/1923

test/scala_junit_test/junit_generates_xml_logs_test.sh
```diff
@@ -0,0 +1,35 @@
...
+
+testlogs="$(nested_bazel_run info bazel-testlogs)"
+report="${testlogs}/test/JunitTestWithDeps/test.xml"
+
+expected=(
+  "testcase name='hasCompileTimeDependencies'"
+  "testcase name='hasRuntimeDependencies'"
+)
+for testcase in "${expected[@]}"; do
+  if ! grep -qF "${testcase}" "${report}"; then
```
bor-p-s: Optional: original asserted exact match counts (3 / 2), so a duplicate entry would fail. Current `expect`/`reject` checks only presence. Worth restoring the count if cheap.

PawelLipski: Done, restored

bor-p-s (approved):

PR #1879, opened 2026-07-23
refactor: Colocate migrated-test fixtures next to their tests

PawelLipski:
### Description

The recently migrated Bazel tests (`expect_build_failure_test`) still referenced their fixture targets in the skipped `test_expect_failure/` tree. This moves each fixture that is only used by a migrated test next to that test, tagging the deliberately-broken ones `manual` so wildcard builds skip them:

- **`mismatching_resource_strip_prefix`** (migrated in #1874): the `noSrcsJarWithWrongStripPrefix` `scala_library` and its `resource.txt` move into `test/mismatching_resource_strip_prefix`; `test_expect_failure/mismatching_resource_strip_prefix` is removed entirely.
- **`scalacopts_invalid`** (migrated in #1848): the invalid-scalacopt `empty` `scala_library` and `Empty.scala` move into `test/scalacopts_invalid`; `test_expect_failure/scalacopts_invalid` is removed entirely.
- **`plus_one_deps`** (migrated in #1873): the `plus_one_deps_with_unused_error` toolchain moves into `test/plus_one_deps/with_unused_deps`. The plain `plus_one_deps` toolchain stays in `test_expect_failure/plus_one_deps` because unmigrated shell tests (`test_compilation.sh`, `test_strict_dependency.sh`) still use it.

### Motivation

Part of migrating bespoke shell tests to native `bazel test` and eventually dropping `test_expect_failure/`. Keeping each fixture next to its test also lets `expect_build_failure_test`'s per-package source glob track the fixture, so editing it now invalidates the test's cache (previously an edit to a cross-package fixture wouldn't force a re-run).

PR #1918, opened 2026-08-12
refactor: Migrate test_scaladoc.sh to Bazel-native sh_test targets

bor-p-s:
### Description

Migrates the 3 `plus_one_deps` scaladoc checks out of `test_scaladoc.sh`. The two file-existence checks (`test_scaladoc_transitive`/`test_scaladoc_intransitive`) don't fit the 4 existing macros in `expect_build_failure.bzl`, since they need `bazel cquery --output=files` plus a check for a specific generated file on disk. They become two new `sh_test` targets following `expect_no_ijar.sh`'s pattern. The third check is a plain build-succeeds case, so it uses the existing `expect_build_success_test` macro.

Fixtures move from `test_expect_failure/scaladoc/` to a new colocated `test/scaladoc/` package (see #1879). `C`/`scaladoc_C` stay tagged `manual`: they only build under `--extra_toolchains=...ast_plus_one_deps_unused_deps_warn`.

### Motivation

Deletes `test_scaladoc.sh` and its `test_rules_scala.sh` line.

Release impact: tests only.


test/scaladoc/BUILD
```diff
@@ -0,0 +1,105 @@
...
+)
+
+# Test that scaladoc for transitive deps generates docs for both B and A.
+sh_test(
+    name = "transitive_scaladoc_test",
+    size = "large",
+    srcs = ["scaladoc.sh"],
+    args = [
+         "--target=//test/scaladoc:scaladoc_transitive",
+         "--expect-file=B$$.html",
```
PawelLipski, line 73: At first I WTF'ed at this `$`, then I reminded myself it's Scala 😅  (companion object)

bor-p-s: yes, that looks scary at the first glance 🙂

test/scaladoc/scaladoc_test.sh
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env bash
```
PawelLipski, line 1: Nit: `test/diagnostics_reporter/diagnostic_proto_files_test.sh` has `_test` suffix, this looks like a better convention (the script is _testing_ scaladoc, not just running/wrapping scaladoc)

test/scaladoc/scaladoc.sh
```diff
@@ -0,0 +1,76 @@
...
+           --target=*)
+            TARGET="${1#*=}"
+            shift
+              ;;
+          --expect-file=*)
+            EXPECT_FILE="${1#*=}"
+            EXPECT_FILES+=("${EXPECT_FILE}")
+            shift
+              ;;
+         --do-not-expect-file=*)
```
PawelLipski: Nit: isn't it called `reject` in the macros? or the logic is different here?

test/scaladoc/scaladoc.sh
```diff
@@ -0,0 +1,76 @@
...
+              ;;
+           *)
+            echo "Unexpected argument: $1" >&2
+            exit 1
+              ;;
+    esac
+done
+
+if [[ -z "${TARGET:-}" ]]; then
+    echo "Usage: scaladoc.sh --target=label [--expect-file=file]... [nested_bazel.sh]" >&2
```
PawelLipski: Nit: flag `--do-not-expect-file`/`--reject-file` not mentioned here

test/scaladoc/scaladoc.sh
PawelLipski: Okay slightly different coz you're checking for file existence, not a string

test/scaladoc/BUILD
```diff
@@ -0,0 +1,105 @@
+load("@rules_shell//shell:sh_test.bzl", "sh_test")
+load("//scala:scala.bzl", "scala_doc", "scala_library")
+load("//scala:scala_import.bzl", "scala_import")
+load("//test/expect_build_failure:expect_build_failure.bzl", "expect_build_success_test")
+load("scaladoc.bzl", "scala_doc_intransitive")
```
PawelLipski, line 5: I know this is pre-existing... but why is `scala_doc_intransitive` only spawned for the sake of tests and not exposed publicly?

Although `scaladoc_intransitive_aspect` seems to be public OTOH 🤔

test/scaladoc/BUILD
```diff
@@ -0,0 +1,105 @@
...
+)
+
+# B depends on A
+scala_library(
+    name = "B",
+    srcs = ["B.scala"],
+    deps = ["A"],
+)
+
+# "Simulate" A & B as imported libraries (no srcs attribute)
```
PawelLipski, line 23: Why is this `scala_import` needed? can't we rely just on `deps`?

test/scaladoc/BUILD
bor-p-s: Emulates a srcs-less import target (like `java_import`) so the test actually hits the transitive external-dep path (`C` -> `ImportedB` -> `ImportedA`). Plain `deps` would just be a scala_library dep.

PawelLipski: (I assume this is a reply to https://github.com/bazel-contrib/rules_scala/pull/1918#discussion_r3782866767 below)

test/scaladoc/BUILD
bor-p-s: Emulates a srcs-less import target (like java_import) so the test actually hits the transitive external-dep path (C -> ImportedB -> ImportedA). Plain deps would just be a scala_library dep.

PawelLipski: Okay, and these cases are supposed to be handled differently by scaladoc logic right? I'm not familiar

bor-p-s: Yes -- the aspect only collects src_files/compile_jars for a dep when it has `srcs` (`scala_doc.bzl`'s `if hasattr(ctx.rule.attr, "srcs")`). `scala_import` has none, so it contributes no doc content itself, but the aspect still walks through it transitively via `deps` -- that's how it reaches `ImportedA` through `ImportedB`. Exactly the case this fixture tests.

test/scaladoc/scaladoc.sh
bor-p-s: Done, renamed to --reject-file.

test/scaladoc/scaladoc.sh
bor-p-s: Done, added to the usage line.

test/scaladoc/BUILD
bor-p-s: Confirmed via blame -- added in #1584 as a one-off for this test, no alias since. Not a design call, just never wired up.

PawelLipski (approved):

twitter_scrooge/twitter_scrooge.bzl
```diff
@@ -17,9 +17,6 @@ load(
 load("//thrift:thrift.bzl", "merge_thrift_infos")
```
PawelLipski, line 17: Looks like some extra changes sneaked in

bor-p-s: Fixed by a proper rebase

PR #1920, opened 2026-08-13
refactor: Migrate test_scala_jacocorunner.sh to Bazel

bor-p-s:
### Description

Moves `test_scala_jacocorunner.sh`'s single check off the legacy shell runner. It asserted that a nested `bazel coverage` under a passing Jacoco toolchain succeeds; that becomes an `expect_test_success_test` target next to the existing fixture in `manual_test/scala_test_jacocorunner`.

`expect_test_success_test` only supported a nested `bazel test`; added a `command` param (defaulting to `"test"`, mirroring the existing param on `expect_test_failure_test`) so it can also drive `bazel coverage`.

### Motivation

Deletes `test_scala_jacocorunner.sh` and its `test_rules_scala.sh` line.

Release impact: tests only.


manual_test/scala_test_jacocorunner/BUILD
```diff
@@ -18,3 +19,16 @@ scala_test(
     name = "empty_test",
     srcs = ["EmptyTest.scala"],
 )
+
+# Jacoco coverage support is not implemented on Windows (mirrors the
+# is_windows guard in the former test/shell/test_scala_jacocorunner.sh).
```
PawelLipski: As usual, don't mention historical code 😅  maybe add that to AGENTS.md or a similar file

PawelLipski (approved):

manual_test/scala_test_jacocorunner/BUILD
bor-p-s: fixed. the rule was there already 😅 but...

PawelLipski (approved):

PawelLipski (commented):
Whoops 63 files 😅

PawelLipski (approved):

PR #1921, opened 2026-08-14
refactor: Migrate test_misc.sh shell tests to Bazel-native sh_test targets

bor-p-s:
### Description

Migrates 5 of the 7 cases in `test/shell/test_misc.sh` to native `bazel test` targets:

- `test_disappearing_class`: colocates the fixture into `test/disappearing_class/` and adds a bespoke `sh_test` (`disappearing_class_test.sh`, following the `test/macros/expect_no_ijar.sh` pattern), since the case needs two chained nested `bazel build`s with a source swap in between, and none of the 4 `expect_*_test` macros support that. Builds `uses_class` once, overwrites its dependency's source with a version missing the class it references, and asserts the rebuild fails rather than reusing a stale cached result.
- `test_repl`: plain `sh_test` (`test/sh_tests/repl_test.sh`) that runs the 5 already-built `scala_repl` binaries directly and greps their output. Also builds `PhaseRepl` (granted visibility to `test/sh_tests`) as a 6th target, matching the original's `bazel query 'kind(scala_repl, //test/...)'` coverage of it. The 6 current targets are hardcoded in `data`; a 7th `scala_repl` target added later would need adding here by hand.
- `test_benchmark_jmh`: plain `sh_test` (`test/jmh/benchmark_test/benchmark_test.sh`) that runs the built JMH binary directly with `-i1 -f1 -wi 1` and checks its output. Needed a one-line fix in `jmh.bzl`: `scala_benchmark_jmh` built its underlying `scala_binary` without forwarding a caller's `visibility`, so the new test package couldn't see `test_benchmark`.
- `scala_test_test_filters`: two `expect_test_success_test`s (`test/sh_tests/BUILD`) against `//test:TestFilterTests`, one with a wildcard `--test_filter` and one with a single-class filter, both asserting on the same "tests a"/"tests b" output lines via `expect`/`reject` files.
- `test_multi_service_manifest`: plain `sh_test` (`test/sh_tests/multi_service_manifest_test.sh`) that unzips the already-built deploy jar's service manifest and diffs it against the checked-in expected file.

`test_override_javabin` and `xmllint_test` stay in `test_misc.sh` for now:
- `test_override_javabin` needs `bazel run` support and an exposed `env` parameter in `expect_build_failure.bzl`, neither of which the macro has today. That's a separate mechanism PR, with the usage stacked on top per this repo's convention of keeping a new macro capability and its first caller apart.
- `xmllint_test` checks every XML file under `bazel-testlogs` after the whole suite has run, which is a property of the whole run rather than one target's fixture (a different shape of check than a per-target `sh_test`). Also a separate PR.

Every new/changed check was verified locally to actually go red when broken: swapped the fixture's source back to keep its class, corrupted an `expect`/`reject` file's content, and widened a `--test_filter` to the value the `reject` check was guarding against.

### Motivation

Removes 5 of `test_misc.sh`'s 7 functions and their `$runner` lines.

Release impact: tests only.


test/disappearing_class/ClassProviderWithoutClass.scala
```diff
@@ -0,0 +1,3 @@
+package scalarules.test
+
+object BackgroundNoise{}
```
PawelLipski, line 3: Hmmm this is a new file (not moved one)... has the test logic been expanded?

test/disappearing_class/disappearing_class_test.sh
```diff
@@ -0,0 +1,73 @@
...
+      printf '%s' "${candidate}"
+      return
+    fi
+  done
+  printf '%s' "${file}"
+}
+replacement_path="$(_resolve_runfile "${replacement_relpath}")"
+
+nested_bazel_setup "rules_scala_disappearing_class_output_base"
+
```
PawelLipski, line 53: Actually... where does the disappearance actually happen? 🤔

test/jmh/benchmark_test/BUILD
```diff
@@ -0,0 +1,12 @@
+load("@rules_shell//shell:sh_test.bzl", "sh_test")
+
+# This lives in its own package rather than //test/jmh so that //test/jmh does
+# not load @rules_shell (a dev-only dependency, invisible when rules_scala is
```
PawelLipski, line 4: Actually... what would happen if `@rules_shell` got loaded? it's already appeared in multiple places but I forgot the reasoning

PawelLipski: Would it somehow expose `@rules_shell` to rules_scala end-users?

test/sh_tests/BUILD
```diff
@@ -171,3 +172,68 @@ scala_test(
     srcs = ["ScalaTestResourcesFromLocalTargetTest.scala"],
     resources = [":py_resource_binary"],
 )
+
+expect_test_success_test(
+    name = "test_filter_wildcard_test",
+    target = "//test:TestFilterTests",
+    bazel_args = [
+        "--test_output=streamed",
```
PawelLipski: Why `streamed`? 🤔  this mode doesn't seem conducive to testing (?)

PawelLipski: (as in, I think of it as something for interactive runs, but maybe I'm wrong)

test/sh_tests/BUILD
```diff
@@ -171,3 +172,68 @@ scala_test(
     srcs = ["ScalaTestResourcesFromLocalTargetTest.scala"],
     resources = [":py_resource_binary"],
 )
+
```
PawelLipski, line 175: Also, is `sh_tests` the right place for these tests? the new ones, and the existing ones as well

PawelLipski: It seems to suggest the rule kind of the test target, which is of secondary importance (and also, these `sh_test`s are now wrapped into macros anyway); maybe better move to "thematic" folders?

test/shell/test_misc.sh
```diff
@@ -4,86 +4,6 @@ dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
...
-  tmp=$(mktemp -d -t test_disappearing_class-XXXXXXXXXX)
-  out=$tmp/out
-  cache=$tmp/cache
-  git checkout test_expect_failure/disappearing_class/ClassProvider.scala
-  bazel --output_base=$out build --disk_cache=$cache test_expect_failure/disappearing_class:uses_class
-  echo -e "package scalarules.test\n\nobject BackgroundNoise{}" > test_expect_failure/disappearing_class/ClassProvider.scala
-  set +e
-  bazel --output_base=$out build --disk_cache=$cache test_expect_failure/disappearing_class:uses_class
-  RET=$?
-  git checkout test_expect_failure/disappearing_class/ClassProvider.scala
```
PawelLipski, line 17: Nit: hmmm now it's just `cp` right? would it make sense to use `git checkout`? for the sake of better idempotency/reproducibility (given that we're outside of sandbox & actively modifying the source code anyway)

test/disappearing_class/BUILD
```diff
@@ -0,0 +1,32 @@
...
+sh_test(
+    name = "disappearing_class_test",
+    srcs = ["disappearing_class_test.sh"],
+    args = [
+        "//test/disappearing_class:uses_class",
+        "test/disappearing_class/ClassProvider.scala",
+        "test/disappearing_class/ClassProviderWithoutClass.scala",
+    ],
+    data = [
+        "ClassProviderWithoutClass.scala",
```
PawelLipski, line 25: Should UsesClass.scala or ClassProvider.scala be included here as well?

test/disappearing_class/disappearing_class_test.sh
```diff
@@ -0,0 +1,73 @@
...
+class_provider_path="${NESTED_BAZEL_WORKSPACE}/${class_provider_relpath}"
+backup_path="$(mktemp)"
+cp "${class_provider_path}" "${backup_path}"
+restore() {
+  cp "${backup_path}" "${class_provider_path}"
+  rm -f "${backup_path}"
+}
+trap restore EXIT
+
+if ! nested_bazel_run build "${target}" >/dev/null 2>&1; then
```
PawelLipski: Hmmm discard the errors? I'd rather keep them for visibility

test/sh_tests/BUILD
PawelLipski: Maybe `all` instead?

test/disappearing_class/ClassProviderWithoutClass.scala
bor-p-s: Not new logic, same content as the old test's `echo -e "package scalarules.test\n\nobject BackgroundNoise{}"`, just moved into a checked-in file instead of generated inline. Easier to diff and it fails loudly if it doesn't parse.

test/disappearing_class/disappearing_class_test.sh
bor-p-s: Line 68 copies the replacement file over the real `ClassProvider.scala` in the source tree, then line 70 rebuilds the same target against the same nested output base. The class "disappears" because the replacement file no longer defines it. The test fails if that second build still succeeds, since that would mean Bazel reused a stale cached result instead of noticing the dependency broke.

test/jmh/benchmark_test/BUILD
bor-p-s: When a bzlmod repo is loaded with `dev_dependency = True` in MODULE.bazel, it stays out of the dependency graph anyone consuming rules_scala as a module sees, so `@rules_shell` never reaches them. That's why this test sits in its own package instead of `//test/jmh`: so `//test/jmh` (and anything depending on it) never has to load `@rules_shell` at all.

test/disappearing_class/disappearing_class_test.sh
PawelLipski: Aww okay

PawelLipski: `test/disappearing_class/ClassProvider.scala` is hidden by GitHub diff, hence I didn't notice the missing class

test/sh_tests/BUILD
bor-p-s: `--test_output=streamed` isn't for interactive use, it gets the test's own stdout into `bazel test`'s output so `expect=` can grep for it (default `--test_output` only shows output on failure). Switched to `--test_output=all`, same content since we only read the output after the run finishes, without `streamed`'s side effect of forcing tests to run one at a time locally.

test/disappearing_class/BUILD
bor-p-s: No, they don't need to be. The nested `bazel build` runs directly against the real source tree (`nested_bazel_setup` cd's into it), not through this test's runfiles, so those two files are read straight from disk. Only `ClassProviderWithoutClass.scala` is in `data`, because that one is read through the test's own runfiles before the cd happens.

test/jmh/benchmark_test/BUILD
PawelLipski: Hmmm okay but I still don't get it - is it just an optimization to reduce the scope of dependencies (the dep on `@rules_shell` is narrowed down just the a small package rather than a larger one)? or would anything actually break if you merged test/jmh/benchmark_test back to its parent package?

test/disappearing_class/BUILD
PawelLipski: Okay but for the sake of cache invalidation? even if not read from runfiles?

test/jmh/benchmark_test/BUILD
bor-p-s: It would actually break, not just narrow scope. `rules_shell` is `dev_dependency = True` in rules_scala's MODULE.bazel, so bzlmod doesn't propagate it to consumers. `test_dependency_versions.sh` builds `@rules_scala//test/jmh:jmh_command_line_parsing_test` from a consumer whose MODULE.bazel never declares `rules_shell` -- if `test/jmh/BUILD` loaded it, that build would fail on an unresolved repository.

test/disappearing_class/BUILD
bor-p-s: Good catch, I was wrong. It's cache invalidation, not runfiles: without these two files in `data`, editing them didn't change the test's inputs, so Bazel served a stale `(cached) PASSED`. Confirmed and fixed by adding both to `data`.

test/sh_tests/BUILD
bor-p-s: Agree it's worth doing, but it's a real reshuffle on its own (touches ~20 targets and label references in 6 other files), not something to fold into this PR. OK if I open a follow-up issue for it?

PawelLipski: Sure! pls open a separate issue

PawelLipski (approved):

test/disappearing_class/disappearing_class_test.sh
bor-p-s: Fixed, both builds now print their captured output on unexpected result instead of discarding it.

test/shell/test_misc.sh
bor-p-s: Good point, changed `restore()` to `git checkout -- ...` instead of the backup copy. Simpler, matches the old test.

PR #1925, opened 2026-08-14
refactor: Migrate `test_custom_reporter_class.sh` to Bazel

PawelLipski:
### Description

Migrate `test/shell/test_custom_reporter_class.sh` to a native Bazel `sh_test` in `//test/custom_reporter` (and drop its `test_rules_scala.sh` line). The wrapper runs `//test/scala_test:custom_reporter` -- a `scala_test` configured with a `reporter_class` -- via a nested `bazel test`, then asserts the reporter wrote its `custom_reporter_check` marker into the test's undeclared outputs.

That marker is an on-disk artifact `bazel test` never prints to stdout, so it can't reuse the stdout-matching `expect_build_failure.bzl` macros; it mirrors `test/diagnostics_reporter` (a plain `sh_test` that runs bazel then inspects its output). `--zip_undeclared_test_outputs=false` leaves the marker as a plain file so the check is a simple file-existence test with no unzip step (the original had to `unzip` + `chmod`). The `custom_reporter` fixture rides the test's `data` (with a matching visibility grant) so editing its sources re-keys the wrapper.

### Motivation

### Test plan

- [x] `bazel test //test/custom_reporter:custom_reporter_writes_check_file_test` -- passes.


test/custom_reporter/custom_reporter_writes_check_file_test.sh
```diff
@@ -0,0 +1,37 @@
...
+nested_bazel_setup "rules_scala_custom_reporter_output_base"
+
+nested_bazel_run test //test/scala_test:custom_reporter \
+  --nocache_test_results \
+  --zip_undeclared_test_outputs=false
+
+testlogs="$(nested_bazel_run info bazel-testlogs)"
+marker="${testlogs}/test/scala_test/custom_reporter/test.outputs/custom_reporter_check"
+
+if [[ ! -f "${marker}" ]]; then
```
PawelLipski, line 32: Maybe worth extracting it at some point as another abstraction in the macros (but let's land the PRs that use this pattern first, like #1917, #1918, this one)

bor-p-s (approved):

PR #1928, opened 2026-08-18
refactor: Migrate `test_scala_config.sh` to Bazel

PawelLipski:
### Description

Migrate `test/shell/test_scala_config.sh` to native Bazel `sh_test` targets in `//test/scala_config` (and drop its `test_rules_scala.sh` line), keeping the original's three cases as three separate targets:

- `scalac_javac_uses_scala_library_{2_12,2_13}_test`: `aquery` the scalac `Javac` action under `SCALA_VERSION=2.12.21` / `2.13.18` and assert the matching `scala-library` rides its inputs (one parameterised script, one target per major version), and
- `config_bzl_reflects_scala_version_test`: build `@rules_scala_config` under a fake `SCALA_VERSION=0.0.0` and assert that version lands in the generated `config.bzl`.

These inspect `aquery` stdout and a generated external-repo file -- neither is the build/test *outcome* the stdout-matching `expect_build_failure.bzl` macros model -- so they stay bespoke `sh_test`s, mirroring `test/diagnostics_reporter` (which likewise runs bazel then inspects its output). All three share one nested output base; being `exclusive`, they serialize rather than multiply the ~1GB of extracted Scala jars a separate base would cost.

### Motivation

### Test plan

- [x] `bazel test //test/scala_config:all` -- all 3 targets pass.


bor-p-s:
[test/scala_config/config_bzl_reflects_scala_version_test.sh#L24-L35](https://github.com/bazel-contrib/rules_scala/blob/76ad218e9512f1252842b24ff670254596c1896a/test/scala_config/config_bzl_reflects_scala_version_test.sh#L24-L35)

The shared output base can keep a stale `*rules_scala_config` dir from an earlier bzlmod hash, so this glob can match more than one file. The check only looks at `config_bzl[0]`, but the original test grepped every match. Use `"${config_bzl[@]}"` in both the existence check and the grep.

test/scala_config/config_bzl_reflects_scala_version_test.sh
```diff
@@ -0,0 +1,36 @@
...
+  echo "Could not find a generated rules_scala_config config.bzl under" \
+    "${output_base}/external." >&2
+  exit 1
+fi
+
+if ! grep --quiet --fixed-strings "SCALA_MAJOR_VERSION='0.0'" "${config_bzl[0]}"; then
+  echo "Expected SCALA_MAJOR_VERSION='0.0' in ${config_bzl[0]} under" \
+    "SCALA_VERSION=0.0.0, but it was absent. config.bzl:" >&2
+  cat "${config_bzl[0]}" >&2
+  exit 1
```
bor-p-s, line 42: The shared output base can keep a stale `*rules_scala_config` dir from an earlier bzlmod hash, so this glob could match more than one file. This only checks `config_bzl[0]`, while the original test grepped every match. Would it be safer to use `"${config_bzl[@]}"` in both the existence check and the grep?

test/scala_config/config_bzl_reflects_scala_version_test.sh
```diff
@@ -0,0 +1,36 @@
...
+# helper this script sources.
+
+set -euo pipefail
+
+# shellcheck source=test/expect_build_failure/nested_bazel.sh
+source "${TEST_SRCDIR:-${RUNFILES_DIR:-$0.runfiles}}/${TEST_WORKSPACE:-_main}/test/expect_build_failure/nested_bazel.sh"
+
+nested_bazel_setup "rules_scala_scala_config_output_base"
+
+nested_bazel_run build --repo_env=SCALA_VERSION=0.0.0 @rules_scala_config//:all >/dev/null 2>&1
```
bor-p-s: This nested bazel build runs unguarded under `set -e` with output sent to `/dev/null`, so if it fails (this is tagged `requires-network`) the test log gets no diagnostic. `test/macros/expect_no_ijar.sh` and `test/scaladoc/scaladoc_test.sh` guard the same kind of call with `if ! output=$(...); then echo "$output"; fi` — might be worth matching that pattern here.

test/scala_config/scalac_javac_scala_library_test.sh
```diff
@@ -0,0 +1,36 @@
...
+source "${TEST_SRCDIR:-${RUNFILES_DIR:-$0.runfiles}}/${TEST_WORKSPACE:-_main}/test/expect_build_failure/nested_bazel.sh"
+
+scala_version="${1:?usage: scalac_javac_scala_library_test.sh <scala_version> <expected_library>}"
+expected_library="${2:?usage: scalac_javac_scala_library_test.sh <scala_version> <expected_library>}"
+
+nested_bazel_setup "rules_scala_scala_config_output_base"
+
+output="$(nested_bazel_run aquery \
+  'mnemonic("Javac", //src/java/io/bazel/rulesscala/scalac:scalac)' \
+  "--repo_env=SCALA_VERSION=${scala_version}" 2>&1)"
```
bor-p-s: Similar to the build call in config_bzl_reflects_scala_version_test.sh: this aquery runs unguarded under `set -e`, so on failure the captured output never gets echoed. Could guard it the same way `expect_no_ijar.sh`/`scaladoc_test.sh` do: `if ! output=$(...); then echo "$output"; fi`.

test/scala_config/BUILD
```diff
@@ -0,0 +1,45 @@
...
+        size = "large",
+        srcs = ["scalac_javac_scala_library_test.sh"],
+        args = [
+            scala_version,
+            expected_library,
+        ],
+        data = [
+            "//:MODULE.bazel",
+            "//test/expect_build_failure:nested_bazel.sh",
+        ],
```
bor-p-s, line 21: These two targets only track `MODULE.bazel` and the nested-bazel helper, not the code that actually decides which `scala-library` version lands on the classpath. Bumping the coordinate in `third_party/repositories/scala_2_12.bzl`/`scala_2_13.bzl` wouldn't invalidate the cached PASS. Worth adding those two files to `data`.

test/scala_config/BUILD
```diff
@@ -0,0 +1,45 @@
...
+
+# A fake SCALA_VERSION must flow through into the generated @rules_scala_config
+# config.bzl.
+sh_test(
+    name = "config_bzl_reflects_scala_version_test",
+    size = "large",
+    srcs = ["config_bzl_reflects_scala_version_test.sh"],
+    data = [
+        "//:MODULE.bazel",
+        "//test/expect_build_failure:nested_bazel.sh",
```
bor-p-s, line 42: Same gap here: `scala_config.bzl` (the repo rule that writes `SCALA_MAJOR_VERSION` into the generated `config.bzl`) isn't in `data`, so editing that logic wouldn't rerun this test. Might need `exports_files` for it in the root `BUILD` first, the same way `MODULE.bazel` already is.

test/scala_config/config_bzl_reflects_scala_version_test.sh
PawelLipski: Done, saved output into a var

test/scala_config/scalac_javac_scala_library_test.sh
PawelLipski: Done, saved output into a var

test/scala_config/BUILD
PawelLipski: Done, added to `data`

test/scala_config/BUILD
PawelLipski: Done, added to `data`

test/scala_config/config_bzl_reflects_scala_version_test.sh
PawelLipski: Done, switched to `"${config_bzl[@]}"`

bor-p-s (approved):

PR #1929, opened 2026-08-18
refactor: Migrate `test_build_event_protocol.sh` to Bazel

PawelLipski:
### Description

Migrate `test/shell/test_build_event_protocol.sh` to native Bazel `sh_test` targets in `//test/build_event_protocol` (and drop its `test_rules_scala.sh` line), keeping the original's seven cases as seven targets.

Each builds a `//test` target with `--build_event_text_file` and asserts the target's output jar appears in the resulting Build Event Protocol stream -- the same check once per rule kind (`scala_binary`/`library`/`test`, `scala_junit_test`, and their `_java`-jar variants), off one parameterised script.

The signal is a generated BES file, not the build/test *outcome* the stdout-matching `expect_build_failure.bzl` macros model, so these stay bespoke `sh_test`s, mirroring `test/diagnostics_reporter` (which likewise runs bazel then inspects its output). All seven share one nested output base; being `exclusive`, they serialize rather than multiply the ~1GB of extracted Scala jars a separate base would cost.

### Motivation

### Test plan

- [x] `bazel test //test/build_event_protocol:all` -- all 7 targets pass.


test/build_event_protocol/BUILD
```diff
@@ -0,0 +1,35 @@
...
+            "//test/expect_build_failure:nested_bazel.sh",
+        ],
+        tags = [
+            "exclusive",
+            "local",
+            "requires-network",
+        ],
+    )
+    for name, target, jar in [
+        ("scala_binary", "ScalaLibBinary", "ScalaLibBinary"),
```
PawelLipski, line 27: These names all correspond to targets defined in `test/BUILD.bazel`.

test/build_event_protocol/jar_exposed_in_bep_test.sh
```diff
@@ -0,0 +1,35 @@
...
+# bespoke sh_test rather than an expect_build_failure.bzl macro. The nested
+# `bazel` (and the rationale for it) lives in the shared nested_bazel.sh helper
+# this script sources.
+
+set -euo pipefail
+
+# shellcheck source=test/expect_build_failure/nested_bazel.sh
+source "${TEST_SRCDIR:-${RUNFILES_DIR:-$0.runfiles}}/${TEST_WORKSPACE:-_main}/test/expect_build_failure/nested_bazel.sh"
+
+target="${1:?usage: jar_exposed_in_bep_test.sh <target> <expected_jar>}"
```
PawelLipski, line 22: "error if unset or empty", I didn't know this syntax previously 😅

test/build_event_protocol/jar_exposed_in_bep_test.sh
```diff
@@ -0,0 +1,35 @@
...
+expected_jar="${2:?usage: jar_exposed_in_bep_test.sh <target> <expected_jar>}"
+
+nested_bazel_setup "rules_scala_build_event_protocol_output_base"
+
+bes_file="${TEST_TMPDIR:?TEST_TMPDIR must be set}/build_events.txt"
+nested_bazel_run build "${target}" "--build_event_text_file=${bes_file}" >/dev/null 2>&1
```
bor-p-s: `>/dev/null 2>&1` here means any real build failure (compile error, flaky fetch) gets swallowed: `set -euo pipefail` exits before the `cat "${bes_file}"` block runs, so the test log shows just a bare exit code, no error text.

The original `test_build_event_protocol.sh` ran `bazel build` unredirected, and sibling scripts on the same `nested_bazel.sh` helper capture output and print it on failure instead, e.g. `test/scaladoc/scaladoc_test.sh`:

```bash
if ! BUILD_OUTPUT=$(nested_bazel_run build "${TARGET}" ... 2>&1); then
    echo "Build of ${TARGET} failed:" >&2
    echo "${BUILD_OUTPUT}" >&2
    exit 1
fi
```

Worth doing the same here.

PawelLipski: Done, fixed

bor-p-s (approved):

PR #1931, opened 2026-08-18
refactor: Migrate test_scala_specs2.sh to Bazel-native sh_test targets

bor-p-s:
### Description

Migrates the 10 shell test cases in `test_scala_specs2.sh` into 9 Bazel-native targets in `test/scala_junit_test/BUILD`, next to the existing `specs2_failing_test` fixture and its `specs2_exception_*` tests.

7 of them check `--test_filter` behavior on `test:Specs2Tests` by matching the specs2 console reporter's own "+ ..." lines in the output. They use the existing `expect_test_success_test` macro, the same one `scala_test_testfilter` already uses for `--test_filter` checks. Two of them route the filter through `bazel_arg_files` instead of `bazel_args`, since the filter regex has `(`, `|`, `[` characters that `sh_test`'s Bourne tokenization or the Windows launcher's `bash -c` would mangle.

The other 2 shell cases check the generated JUnit `test.xml` on disk, which `bazel test` never prints to stdout, so they can't reuse the stdout-matching macros. They follow the same pattern as the existing `junit_generates_xml_logs_test.sh`: a script that sources `nested_bazel.sh`, runs a nested `bazel test`, and greps the report. Since both original cases run the same fixture back to back (once without a filter, once with), they are combined into one script, `specs2_test_xml_test.sh`, giving 9 targets total instead of 10. A third XML check, `specs2_failing_test_xml_test.sh`, covers the case where the filtered example itself fails: the nested run is expected to fail, and the report must still list only that one example.

### Motivation

Deletes `test_scala_specs2.sh` and its `test_rules_scala.sh` line.

Release impact: tests only.


test/scala_junit_test/specs2_regex_unsafe.txt
```diff
@@ -0,0 +1 @@
++ 2 + 2 != 5
```
PawelLipski, line 1: Hmm this file is supposed to be _rejected_, right? pls fix file names (this and maybe some others as needed)

test/scala_junit_test/specs2_failing_test_xml_test.sh
```diff
@@ -0,0 +1,36 @@
...
+  '--test_filter=scalarules.test.junit.specs2.SuiteWithOneFailingTest#specs2 tests::fail$' \
+  //test/scala_junit_test:specs2_failing_test
+status=$?
+set -e
+if [[ "${status}" -eq 0 ]]; then
+  echo "Expected the nested \`bazel test\` to fail (the filtered example fails), but it succeeded." >&2
+  exit 1
+fi
+
+report="$(nested_bazel_run info bazel-testlogs)/test/scala_junit_test/specs2_failing_test/test.xml"
```
PawelLipski, line 30: Nit: what happens if the report is missing?

test/scala_junit_test/BUILD
```diff
@@ -94,6 +94,167 @@ expect_test_failure_test(
...
+expect_test_success_test(
+    name = "specs2_test_filter_exact_match_unsafe_characters",
+    bazel_arg_files = ["specs2_regex_unsafe_filter_flag.txt"],
+    bazel_args = [
+        "--nocache_test_results",
+        "--test_output=streamed",
+    ],
+    data = ["//test:Specs2Tests"],
+    expect = ["expected_specs2_regex_unsafe.txt"],
+    reject = ["expected_specs2_regex_escaped.txt"],
```
PawelLipski: Yeah, pls rename these rejected files

test/scala_junit_test/rejected_specs2_not_run.txt
```diff
@@ -0,0 +1 @@
++ not run
```
PawelLipski: Actually, what do these seemingly random strings correspond to? are these names of tests that should be filtered out?

test/scala_junit_test/BUILD
```diff
@@ -94,6 +94,167 @@ expect_test_failure_test(
...
+    ],
+    data = ["//test:Specs2Tests"],
+    expect = ["expected_specs2_run_from_another_test.txt"],
+    reject = ["expected_specs2_run_from_another_test_2.txt"],
+    target = "//test:Specs2Tests",
+)
+
+# `\Q...\E`-escaped filters reach the intended example even when its
+# description contains regex metacharacters. The filter's own metacharacters
+# (`+`, spaces) would be mangled by the Windows sh launcher's `bash -c` if
```
PawelLipski: Nit: the reason is likely slightly different (?)
`+` isn't a shell metachar, and spaces are handled by `_nested_bazel_test`

Maybe the extraction of separate arg files isn't even needed here (?)

test/scala_junit_test/specs2_test_xml_test.sh
```diff
@@ -0,0 +1,41 @@
...
+# expect_build_failure.bzl macros.
+
+set -euo pipefail
+
+# shellcheck source=test/expect_build_failure/nested_bazel.sh
+source "${TEST_SRCDIR:-${RUNFILES_DIR:-$0.runfiles}}/${TEST_WORKSPACE:-_main}/test/expect_build_failure/nested_bazel.sh"
+
+nested_bazel_setup "rules_scala_specs2_test_xml_output_base"
+
+report="$(nested_bazel_run info bazel-testlogs)/test/Specs2Tests/test.xml"
```
PawelLipski, line 16: Nit: same as in the other script, what if report is missing?

PawelLipski: (probably just a confusing error message instead of the right one; still, worth adding an assertion)

test/scala_junit_test/specs2_test_xml_test.sh
```diff
@@ -0,0 +1,41 @@
...
+  if [[ "${count}" -ne "${expected_count}" ]]; then
+    echo "Expected ${expected_count} testcase(s) in ${report}, found ${count}. Full report:" >&2
+    cat "${report}" >&2
+    exit 1
+  fi
+}
+
+nested_bazel_run test \
+  --nocache_test_results \
+  --test_output=streamed \
```
PawelLipski: Nit: is this `streamed` mode needed for anything here?

test/scala_junit_test/BUILD
```diff
@@ -120,7 +120,9 @@ expect_test_success_test(
     target = "//test:Specs2Tests",
 )
 
-# A method-level filter runs only that example.
+# A method-level filter runs only that example. `specs2_run_smoothly.txt` and
+# `specs2_not_run_smoothly.txt` hold one example's "+ ..." line each; which one
+# is `expect` and which is `reject` depends on the test, not the file name.
```
PawelLipski, line 125: Aww okay, didn't see that

test/scala_junit_test/BUILD
```diff
@@ -161,30 +163,32 @@ expect_test_success_test(
...
+# description contains regex metacharacters. `_nested_bazel_test` already
+# single-quotes any `bazel_args` entry containing a space, which keeps this
+# filter's spaces, `+` and `$` intact -- none of those need `bazel_arg_files`.
 expect_test_success_test(
     name = "specs2_test_filter_exact_match_unsafe_characters",
-    bazel_arg_files = ["specs2_regex_unsafe_filter_flag.txt"],
     bazel_args = [
         "--nocache_test_results",
         "--test_output=streamed",
+        "--test_filter=scalarules.test.junit.specs2.JunitSpec2RegexTest#\\Qtests with unsafe characters::2 + 2 != 5\\E$$",
```
PawelLipski, line 180: Yeah, easier to read than an arg file

PawelLipski (approved):

test/scala_junit_test/specs2_regex_unsafe.txt
bor-p-s: Renamed the fixtures that are `expect` in one test and `reject` in another to drop the direction-implying prefix (`specs2_run_smoothly.txt`, `specs2_not_run_smoothly.txt`, `specs2_run_from_another_test.txt`, `specs2_run_from_another_test_2.txt`, `specs2_regex_unsafe.txt`, `specs2_regex_escaped.txt`), matching how `first_filter_test_method1.txt` etc. are named elsewhere in this file.

test/scala_junit_test/BUILD
bor-p-s: Same rename as above.

test/scala_junit_test/rejected_specs2_not_run.txt
bor-p-s: Dropped this file. `specs2_test_filter_match_multiple_methods` now rejects `specs2_not_run_smoothly.txt` instead — that's a real example's text (from the sibling `JunitSpecs2Test` class), so the check actually guards against that example leaking in, rather than a string that never appears in the output.

test/scala_junit_test/BUILD
bor-p-s: You're right, thanks. Moved `specs2_test_filter_exact_match_unsafe_characters`'s filter to plain `bazel_args` (it only has a space and `$`, both already handled by `_nested_bazel_test`'s auto-quoting). Kept `bazel_arg_files` for `escaped_and_sanitized` (has `[`/`]`) and `match_multiple_methods` (has `(`/`|`), and reworded the comments to say why.

test/scala_junit_test/specs2_failing_test_xml_test.sh
bor-p-s: Added an explicit `[[ -f "${report}" ]]` check with a clear message before the `grep`.

test/scala_junit_test/specs2_test_xml_test.sh
bor-p-s: Same fix here too.

test/scala_junit_test/specs2_test_xml_test.sh
bor-p-s: No — this script only reads the XML file on disk, never the console output. Dropped `--test_output=streamed` from both nested runs.

PR #1932, opened 2026-08-19
ci: Let test_rules_scala.sh share the remote cache with CI's own build/test steps

bor-p-s:
### Description

Adds `--remote_cache`/`--google_default_credentials` (plus `--remote_max_connections=200`, and macOS's own GCS-bucket backend instead of RBE) to the `bazel build`/`bazel test` calls made directly in `test_rules_scala.sh`, gated on `$BUILDKITE` so a local run keeps its current behavior. Scripts it sources, like `test_scala_library.sh`, run their own bazel calls exactly as before. Excludes `last_green`: it uses a different Bazel build, so the cache does not help there -- it only slows it down (measured: 8780s vs 3476-3920s without the flag).

Also excludes `dicer`'s `assignment_generator_test` from `test/community_build:dicer_test`'s target list: its own 60s timeout leaves it running within a second of that limit even when idle, so it tipped over twice once this step started sharing the host with more concurrent cache traffic.

### Motivation

bazelci.py's own `remote_caching_flags()` is scoped to `build_targets`/`test_targets` tasks. `test_rules_scala.sh` runs under a separate task type, `shell_commands`, so every nested `bazel test test/...` it calls (4 times, under different toolchain flags) started cold every CI run, running independently of the cache the sibling `bazel test //...` step already warms.

This depends on #1939: before it, most nested `expect_*` tests carried the `local` tag, which keeps a result out of remote cache on its own, regardless of these flags. #1939 moves them to `no-sandbox` + `no-remote-exec`, the combination that actually makes them cacheable.

### Test plan

Compared build #6645 (master, before this stack) against build #6804 (this branch, stacked on #1939), same target set -- the `assignment_generator_test` exclusion above landed after both of these ran, so it played no part in the numbers below:

| Step | Before | After |
|---|---|---|
| Linux `test_rules_scala` | 3241s | 1090s |
| Linux `test_rules_scala with jdk21` | 3287s | 1073s |
| Windows `test_rules_scala` | 3240s | 1659s |
| macOS `test_rules_scala` | 5246s | 2570s |
| Linux `test_rules_scala (last_green Bazel)` | 3606s | 3754s |

`last_green` does not get these flags at all, and its time stayed almost the same (3606s vs 3754s). This shows the speedup on the other rows really comes from the cache -- if it came from something else in this branch, `last_green` would have gotten faster too.

macOS timing jumps around a lot between runs, even without this change -- from 3589s to 6358s on other, unrelated builds around the same time. So the macOS number here is rough, not exact.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -41,6 +41,20 @@ build --java_runtime_version=17
 
 build:windows --worker_quit_after_build --enable_runfiles
 
+# bazelci's own remote-cache flags (remote_caching_flags in bazelci.py) for
+# `build_targets`/`test_targets` tasks, as named configs a bazel invocation
+# must explicitly request via --config -- see test_rules_scala.sh, which
+# requests one of these for its own top-level calls while the nested `bazel`
+# calls the expect_* tests run stay on the default configuration.
+# Covers both Linux and Windows: both cache through the RBE (Remote Build
+# Execution) endpoint, only macOS (below) uses a different backend.
+build:ci-cache-rbe --remote_cache=remotebuildexecution.googleapis.com --remote_instance_name=projects/bazel-untrusted/instances/default_instance --google_default_credentials --remote_max_connections=200
+
+# macOS runs on MacService instead of GCE (Google Compute Engine), so
+# bazelci.py caches it through a GCS (Google Cloud Storage) bucket instead
+# (still with --google_default_credentials).
+build:ci-cache-gcs --remote_cache=https://storage.googleapis.com/bazel-untrusted-build-cache --google_default_credentials --remote_timeout=3600 --remote_max_connections=200
+
 # Caps each of the 3 lanes expect_build_failure_test's nested tests hash into
 # (see the "resources:laneN_lock:1" tag in expect_build_failure.bzl) at one
 # running test at a time on Linux/Windows, so two tests in the same lane
--- a/.bazelrc
+++ b/.bazelrc
@@ -48,12 +48,18 @@ build:windows --worker_quit_after_build --enable_runfiles
 # calls the expect_* tests run stay on the default configuration.
 # Covers both Linux and Windows: both cache through the RBE (Remote Build
 # Execution) endpoint, only macOS (below) uses a different backend.
-build:ci-cache-rbe --remote_cache=remotebuildexecution.googleapis.com --remote_instance_name=projects/bazel-untrusted/instances/default_instance --google_default_credentials --remote_max_connections=200
+build:ci-cache-rbe --remote_cache=remotebuildexecution.googleapis.com
+build:ci-cache-rbe --remote_instance_name=projects/bazel-untrusted/instances/default_instance
+build:ci-cache-rbe --google_default_credentials
+build:ci-cache-rbe --remote_max_connections=200
 
 # macOS runs on MacService instead of GCE (Google Compute Engine), so
 # bazelci.py caches it through a GCS (Google Cloud Storage) bucket instead
 # (still with --google_default_credentials).
-build:ci-cache-gcs --remote_cache=https://storage.googleapis.com/bazel-untrusted-build-cache --google_default_credentials --remote_timeout=3600 --remote_max_connections=200
+build:ci-cache-gcs --remote_cache=https://storage.googleapis.com/bazel-untrusted-build-cache
+build:ci-cache-gcs --google_default_credentials
+build:ci-cache-gcs --remote_timeout=3600
+build:ci-cache-gcs --remote_max_connections=200
 
 # Caps each of the 3 lanes expect_build_failure_test's nested tests hash into
 # (see the "resources:laneN_lock:1" tag in expect_build_failure.bzl) at one
--- a/test_rules_scala.sh
+++ b/test_rules_scala.sh
@@ -28,10 +28,7 @@ test_output_flag="--test_output=errors"
 # to do with what it actually tests.
 #
 # Excludes the last_green task: it runs a different (unreleased) Bazel
-# binary, whose action digests don't match what the pinned release Bazel
-# used everywhere else populates, so it pays remote-cache round trips for
-# misses instead of the hits the other tasks get -- measured 8780s against a
-# 3476-3920s baseline without this fix.
+# binary than every other task, so it has nothing to share this cache with.
 remote_cache_flags=""
 if [[ "${BUILDKITE:-}" == "true" ]] && ! is_macos && [[ "${BAZELCI_TASK:-}" != *last_green* ]]; then
   remote_cache_flags="--remote_cache=remotebuildexecution.googleapis.com --remote_instance_name=projects/bazel-untrusted/instances/default_instance --google_default_credentials"
--- a/test_rules_scala.sh
+++ b/test_rules_scala.sh
@@ -18,20 +18,26 @@ test_output_flag="--test_output=errors"
 # bazelci's shell_commands steps don't get its own --remote_cache /
 # --google_default_credentials flags (remote_caching_flags in bazelci.py), so
 # the bazel calls below share nothing with the separate "bazel test //..."
-# step. The same GCE credentials work on Linux and Windows; macOS runs on
-# MacService instead of GCE and has no ADC to find. Passed as argv, not
-# written to .bazelrc: nested_bazel.sh's nested `bazel` invocations (run by
-# the expect_* tests below) read this workspace's .bazelrc directly and would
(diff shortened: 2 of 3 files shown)


test_rules_scala.sh
```diff
@@ -13,6 +13,36 @@ test_dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/test/shell
...
+# below (whose own log turned out not to carry bazel's usual "N processes: ...
+# remote cache hit" summary line at all, cached or not -- unclear if that's
+# buildkite's shell_commands log capture dropping it, or something else), this
+# runs one small, fast, and unambiguous check first: build a jar, `bazel
+# clean` (clears the *local* cache only, not any remote one), then build it
+# again. A real remote-cache hit prints its usual summary line on the second
+# build; a credentials failure prints its own clear error either way.
+if [[ "${BUILDKITE:-}" == "true" && "$(uname)" == "Linux" ]]; then
+  echo "common --remote_cache=remotebuildexecution.googleapis.com" >>.bazelrc
+  echo "common --remote_instance_name=projects/bazel-untrusted/instances/default_instance" >>.bazelrc
```
PawelLipski: Wow Google has a public Bazel cache for use in basically any project? 😯  is it writable? is there any separation b/w projects?

This looks like sth super-prone to malicious code injection 🤔 

(This is mostly out of curiosity, not very relevant to the experiment at hand)

bor-p-s: Real auth, not a public cache: `--google_default_credentials` uses the CI machine's own GCP service account (metadata server), not anonymous access.

Isolation is by trust tier, not by project: bazelci.py puts anything from the "bazel" org (any untrusted PR, across all projects using this shared setup) into one `bazel-untrusted` project, separate from `bazel-trusted`/`bazel-public`. So no per-repo separation within that tier.

And it's not hypothetical: the cache key has a literal comment "Whenever the remote cache was known to have been poisoned increase the number below." Already happened once.

PawelLipski: Wow okay 😅 sounds... sketchy at best

test_rules_scala.sh
```diff
@@ -10,22 +10,49 @@ fi
...
+  if is_macos; then
+    remote_cache_flags="--remote_cache=https://storage.googleapis.com/bazel-untrusted-build-cache --google_default_credentials --remote_timeout=3600 --remote_max_connections=200"
+  else
+    remote_cache_flags="--remote_cache=remotebuildexecution.googleapis.com --remote_instance_name=projects/bazel-untrusted/instances/default_instance --google_default_credentials --remote_max_connections=200"
+  fi
+fi
+
 . "${test_dir}"/test_bzlmod_macros.sh
-$runner bazel build src/... test/...
+$runner bazel build $remote_cache_flags src/... test/...
```
PawelLipski: Nice... by the way, do you think entire file can realistically be gone completely in favor of `bazel test //...`? or will we need to leave something indefinitely?

test_rules_scala.sh
```diff
@@ -10,22 +10,49 @@ fi
...
+# Excludes the last_green task: it runs a different (unreleased) Bazel
+# binary than every other task, so it has nothing to share this cache with.
+#
+# Linux/Windows use RBE; macOS runs on MacService instead of GCE, so
+# bazelci.py caches it through a GCS bucket instead (still with
+# --google_default_credentials).
+remote_cache_flags=""
+if [[ "${BUILDKITE:-}" == "true" ]] && [[ "${BAZELCI_TASK:-}" != *last_green* ]]; then
+  if is_macos; then
+    remote_cache_flags="--remote_cache=https://storage.googleapis.com/bazel-untrusted-build-cache --google_default_credentials --remote_timeout=3600 --remote_max_connections=200"
```
PawelLipski: Nit: maybe extract a config (`*.bazelrc`) file and use via `--config=...`?

test_rules_scala.sh
```diff
@@ -10,22 +10,49 @@ fi
...
+# `bazel` invocations (run by the expect_* tests below) read this workspace's
+# .bazelrc directly and would inherit it otherwise, serving intermediate
+# outputs from the remote cache that never land in the nested output base --
+# a nested test that reads one such output by path then fails with a
+# NoSuchFileException that has nothing to do with what it actually tests.
+#
+# Excludes the last_green task: it runs a different (unreleased) Bazel
+# binary than every other task, so it has nothing to share this cache with.
+#
+# Linux/Windows use RBE; macOS runs on MacService instead of GCE, so
```
PawelLipski: Pls expand the 3 acronyms maybe

PawelLipski (approved):

test_rules_scala.sh
bor-p-s: Done, expanded RBE/GCE/GCS in the comment.

test_rules_scala.sh
bor-p-s: Looked into this, `--enable_platform_specific_config` (already on in `.bazelrc:34`) auto-applies bare `build:macos`/`build:linux` sections, but unconditionally on every build, including local runs, so it doesn't fit here (we only want this on CI, and not on `last_green`).

A named config does work though: I tested `build:ci-macos`/`build:ci-linux` sections with `--config=ci-macos` locally and confirmed bazel picks them up. That would get the flag strings out of the script. The `is_macos` branch in shell still has to stay, it would just pick `--config=ci-macos` vs `--config=ci-linux` instead of building the flag string directly. Want me to make that change?

PawelLipski: Yeah, pls do in a separate commit

test_rules_scala.sh
bor-p-s: Don't think so, not fully. Below this block the script runs `bazel build`/`bazel test` under several different `--extra_toolchains` overrides, plus the nested `expect_*` tests. A single `bazel test //...` only covers one toolchain config at a time, so we'd need at least this multi-invocation driver to stay.

It can keep shrinking though: each migrated fixture becomes a real `bazel test //...` target. But this script runs `bazel test .../test/...` four times total (lines 47, 52, 53, 55): once under the default toolchain, then three more times, each under a different `--extra_toolchains` override. So a migrated test runs those same four times per CI build. More real coverage from the four toolchain variants, at the cost of four times the exec time.

test_rules_scala.sh
bor-p-s: Done, pushed as a separate commit.

.bazelrc
```diff
@@ -41,6 +41,26 @@ build --java_runtime_version=17
...
 
+# bazelci's own remote-cache flags (remote_caching_flags in bazelci.py) for
+# `build_targets`/`test_targets` tasks, as named configs a bazel invocation
+# must explicitly request via --config -- see test_rules_scala.sh, which
+# requests one of these for its own top-level calls while the nested `bazel`
+# calls the expect_* tests run stay on the default configuration.
+# Covers both Linux and Windows: both cache through the RBE (Remote Build
+# Execution) endpoint, only macOS (below) uses a different backend.
+build:ci-cache-rbe --remote_cache=remotebuildexecution.googleapis.com
+build:ci-cache-rbe --remote_instance_name=projects/bazel-untrusted/instances/default_instance
```
PawelLipski, line 52: I'm still skeptical 🧐  doesn't it introduce security risks? can't just anyone put jars there that would start digging coins when loaded?

Fortunately, as we ship the sources and not any compiled artifacts, so the risk doesn't extend to the released rules_scala

PawelLipski: > start digging coins when loaded?

Or worse, leak credentials from our CI? not sure how many creds are provided to CI, and how many of them are provided to the Bazel actions (probably none via env vars as the environment is filtered, but maybe some files are visible in dot-folders)?

bor-p-s: Checked this carefully.

Coin-miner-jar angle: `--remote_cache` is content-addressed. A hit only substitutes output bytes for an action key our own build graph already computes (same target, same input hashes), it only ever satisfies an action we already declared in a BUILD file. Poisoning means writing a colliding entry under a key a real future build will request, that's an upload-time attack, and the same write surface already exists today via the sibling `bazel test //...` step. This PR reuses that surface, it was already there.

Credential angle: `--google_default_credentials` just tells bazel's client where to find the credential already sitting on the CI machine (its own GCP service account, via the metadata server). That's also the exact credential bazelci.py already wires up for the sibling `bazel test //...` step in this same CI run, for every PR, today. This PR lets `test_rules_scala.sh`'s own calls (a `shell_commands` task, sitting outside what that function covers) reach that same already-existing credential.

PawelLipski (approved):

PR #1933, opened 2026-08-20
fix: resolve rules_scala_config labels in rules_scala's own repo mapping

FrankPortman:
### Description

`setup_scala_testing_toolchain` and `setup_scalafmt_toolchain` pass `@rules_scala_config//:scala_version*` `target_settings` as plain strings, which resolve in the consumer's repo mapping — failing unless the root module does `use_repo(scala_config, "rules_scala_config")`. Wrap them in `Label()`, like `setup_scala_toolchain` already does.

### Motivation

Consumers shouldn't need to `use_repo` an internal repo to use these macros.

### Testing

Not testable in-repo — under rules_scala's own repo mapping the plain string resolves fine; the failure only exists under a consumer mapping without the `use_repo`. So the regression test is a bespoke `sh_test` (`//test/toolchain_macros_repo_mapping`, mirroring `test/scala_config` on the shared `nested_bazel.sh` helper) that synthesizes such a consumer module and asserts `setup_scala_testing_toolchain` analyzes. It reproduces the failure before the fix and passes after.


bor-p-s (approved):
LGTM

PR #1934, opened 2026-08-20
refactor: Split test/sh_tests/BUILD into topical packages

bor-p-s:
### Description

Pure refactor, no test logic changes. #1921 dumped 19 unrelated tests into one `test/sh_tests` package because the migration there was organized by rule type (`sh_test`), not by topic. This splits it into packages named after what each test checks.

`string_in_file.sh` and `string_in_filename_in_jar.sh` are shared helper scripts (several packages outside `test/` use them as a generic check for whether a string is or isn't in a file/jar). They get their own `test/string_checks` package, with explicit visibility for the other consumers, rather than living inside whichever package happens to use them for a self-test.

One naming gotcha worth flagging: `scala_test`'s resource path heuristic (`scala/private/resources.bzl`) picks the target path by finding the literal substring `"resources"` in the resource's path and keeping everything after it. A package named `test/resources_from_local_target` collided with that and silently changed the expected resource path in the test, so the package is named `test/resource_from_local_target` (singular) instead.

`test_binary_run_with_large_classpath` moved into the existing `test/src/main/scala/scalarules/test/large_classpath` package, next to the `largeClasspath` fixture it tests, instead of getting its own package. The two `--test_filter` tests that landed in `test/sh_tests` via #1921 (merged mid-review here) went into the existing `test/scala_test_testfilter` package for the same reason, instead of a new same-topic sibling.

### Motivation

Reviewer feedback on #1921: package names should describe what's tested, not how.


test/src/main/scala/scalarules/test/large_classpath/BUILD
```diff
@@ -22,3 +23,10 @@ create_dependencies(
     amount = 1000,
     length = 20,
 )
+
+sh_test(
```
PawelLipski: Actually, why under `test/src/main/scala/.../BUILD` and not under `test/<something>/BUILD`?

test/string_checks/string_in_file.sh
```diff
@@ -15,7 +15,7 @@ should_be_in_file=$1
 
 if test "$should_be_in_file" != "true" -a "$should_be_in_file" != "false" ; then
   echo "ERROR: Please use only (\"true\" or \"false\") to specify whether you need the substring to be in the file."
-  echo "Refer to test/sh_tests/string_in_file.sh for documentation."
+  echo "Refer to test/jar_signature_checks/string_in_file.sh for documentation."
```
PawelLipski: Hmmm wrong path?

test/string_checks/string_in_filename_in_jar.sh
```diff
@@ -16,7 +16,7 @@ should_be_in_file=$1
 
 if test "$should_be_in_file" != "true" -a "$should_be_in_file" != "false" ; then
   echo "ERROR: Please use only (\"true\" or \"false\") to specify whether you need the substring to be in the file."
-  echo "Refer to test/sh_tests/string_in_filename_in_jar.sh for documentation."
+  echo "Refer to test/jar_signature_checks/string_in_filename_in_jar.sh for documentation."
```
PawelLipski: Wrong path?

PawelLipski (approved):

test/src/main/scala/scalarules/test/large_classpath/BUILD
bor-p-s: Fair point, moved it to its own `test/large_classpath` package to match the rest of the split.

test/string_checks/string_in_file.sh
bor-p-s: Yes, leftover from the earlier move through `test/jar_signature_checks`. Fixed to `test/string_checks/string_in_file.sh`.

test/string_checks/string_in_filename_in_jar.sh
bor-p-s: Same issue, fixed to `test/string_checks/string_in_filename_in_jar.sh`.

PR #1935, opened 2026-08-20
refactor: Migrate `test_compiler_sources_integrity.sh` to Bazel-native sh_test targets

bor-p-s:
### Description

Migrates `test/shell/test_compiler_sources_integrity.sh` to native `sh_test` targets in `//test/compiler_sources_integrity_test`, keeping all seven original cases.

`test/compiler_sources_integrity` is a separate Bazel module excluded from this repo's package tree by `.bazelignore` (bazelbuild/bazel#22208), so it can't use the `expect_build_failure.bzl` macros. Two custom scripts drive a nested `bazel build` from inside that module directly.

Added `exports_files` for `scala/private/macros/{compiler_sources_integrity,scala_repositories}.bzl` and the remaining `third_party/repositories/scala_*.bzl` files, listed in each test's `data`, so a code change re-runs the affected tests.

### Motivation

Continues moving shell tests off the legacy `test_rules_scala.sh` runner toward a unified `bazel test //...`.


test/compiler_sources_integrity/MODULE.bazel
```diff
@@ -1,4 +1,4 @@
-"""Bazel module for test/shell/test_compiler_sources_integrity.sh"""
+"""Bazel module for compiler source integrity tests."""
```
PawelLipski: Actually... why is this specifically extracted as a separate Bazel module? does it have to be like that? if so, maybe add a comment to this file or a README (AFAICS there's nothing yet)

test/compiler_sources_integrity_test/BUILD
```diff
@@ -0,0 +1,169 @@
...
+        ("scala_2_12", scala_2_12_version),
+        ("scala_2_13", scala_2_13_version),
+        ("scala_3_1", scala_3_1_version),
+        ("scala_3_2", scala_3_2_version),
+        ("scala_3_3", scala_3_3_version),
+        ("scala_3_4", scala_3_4_version),
+        ("scala_3_5", scala_3_5_version),
+        ("scala_3_6", scala_3_6_version),
+        ("scala_3_7", scala_3_7_version),
+        ("scala_3_8", scala_3_8_version),
```
PawelLipski, line 93: There's no globbing unlike the old impl right? so with new Scala version, this list won't get updated automatically?

test/compiler_sources_integrity/MODULE.bazel
bor-p-s: Good catch, added a comment explaining it: `scala_config`/`scala_deps` extension settings only take effect from the root module's tags (see `single_tag_values` in `bzlmod.bzl`), so this test needs its own module to set bogus versions and disable version validation without touching the real root module's config. That's on top of the `.bazelignore` reason already noted in `test/compiler_sources_integrity_test/BUILD`.

test/compiler_sources_integrity_test/BUILD
bor-p-s: You're right, this list is manual now, the old shell test globbed `third_party/repositories/scala_*.bzl`. In Starlark `load()` has to be static, so it can't be built from a `glob()` result. Added a comment noting a new major.minor line needs a `load()` and a list entry added by hand.

PawelLipski (approved):

PR #1936, opened 2026-08-20
refactor: Migrate test_scala_proto_library.sh to Bazel-native sh_test targets

bor-p-s:
### Description

Migrates both cases in `test/shell/test_scala_proto_library.sh` and deletes the file (and its `test_rules_scala.sh` line). Stacked on #1941, which makes `third_party/test/proto` build again -- needed first since one of the two cases targets it.

`test_scala_proto_custom_generator` becomes an `expect_test_success_test` (`test/proto/custom_generator/BUILD.bazel`), asserting `//test/proto/custom_generator:DummyGeneratorTest` passes under the two `--extra_toolchains` the original shell function set.

`test_scala_proto_library_action_label` becomes a custom `sh_test` (`test/scala_proto_library_action_label`) built on the shared `nested_bazel.sh` helper, the same pattern as `test/scala_config`'s aquery-based checks: the signal here is `aquery` stdout, a different kind of check than the build/test outcome `expect_build_failure.bzl`'s macros model. `third_party/test/proto` is itself `.bazelignore`d, so its files can't be listed in `data` to bust the test's cache the way same-workspace fixtures do; the test is tagged `external` instead, which disables result caching so it always re-runs. The real expected label is `action 'scala @@//:proto'` -- the original's `@{0,2}` was never a valid label, and its comparison was dead anyway (`if [ "$action" != $expected ]` left `$expected` unquoted, so `[` word-split it and errored with "too many arguments", which `if` reads as false, so the failure branch never ran).

Also dropped a stale comment in `test/proto/custom_generator/BUILD.bazel` pointing at a "manual build for a failing custom generator": that scenario was already migrated to a real test in #1876 (`test/scala_proto_failing_generator`), the comment just never got cleaned up.

### Motivation

Release impact: tests only.


PawelLipski (commented):
Hmmm could you extract the non-trivial fixes into a separate PR?  and stack this (relatively) trivial PR which would then just migrate from shell script to Bazel targets

test/scala_proto_library_action_label/action_label_test.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+#
+# Asserts the Scalac action for `third_party/test/proto` (a self-contained
+# bzlmod module, walled off from this workspace via `.bazelignore`) carries the
```
PawelLipski, line 4: Crap, sorry if asked already (there're so many similar PRs) - but I can't see my comment so maybe not

Why does `third_party/test/proto` need to be kept in such an isolated module? couldn't we check for Scalac action even when this was part of the top-level module? or could the top-level module "pollute" the actions  somehow in such a setup?

test/scala_proto_library_action_label/BUILD
```diff
@@ -0,0 +1,25 @@
+load("@rules_shell//shell:sh_test.bzl", "sh_test")
+
+# `third_party/test/proto`'s Scalac action must carry this exact label.
```
PawelLipski: This comment is unclear in the context of the file (would be more clear in the script instead?)

test/scala_proto_library_action_label/action_label_test.sh
bor-p-s: Yes -- the top-level module would "pollute" the check exactly the way you're guessing. This test dates back to #1117 (2020), which fixed a bug where `Label("%s-fast" % label)` inside `scala_proto/private/scalapb_aspect.bzl` resolved relative to the repo defining that `.bzl` file (rules_scala itself), not the workspace that actually calls `scala_proto_library`. That mis-resolution only shows up when the target lives in a genuinely separate consuming workspace: from inside rules_scala's own top-level module, "the defining repo" and "the calling repo" are the same repo, so a regression here would produce the same label either way and the test couldn't catch it. `third_party/test/proto`, via `local_path_override`, simulates a real downstream consumer, which is the only way to observe the correct vs incorrect label prefix.

PawelLipski: (maybe mention that in the comment in code, unless explanation already present somewhere)

test/scala_proto_library_action_label/BUILD
bor-p-s: Good catch -- moved this: the "what it checks" line now lives only in the script's header (where it already partially duplicated it), and the BUILD comment keeps just the `.bazelignore`/`external`-tag rationale that's actually specific to this file.

PawelLipski (approved):

PR #1587, opened 2024-06-03
Add a few cross-build examples

aszady:
### Description
Adding examples illustrating some basic concepts of cross-build.

### Motivation
#1290

PR #1937, opened 2026-08-21
refactor: Migrate test_scalafmt.sh to Bazel-native sh_test targets

bor-p-s:
### Description

Migrates `test/shell/test_scalafmt.sh` (4 cases: binary, library, test, custom-conf) and the `test_scalafmt()` function from `test/shell/test_cross_build.sh` (6 cases across Scala 2.x and 3.x) to native `sh_test` targets, using a new `scalafmt_round_trip_test` macro.

Each round-trip test asserts, per fixture:
- the "formatted" target reports no source needing reformatting
- the "unformatted" target reports at least one source needing reformatting
- running the formatter over the unformatted source produces exactly the formatted fixture

`test/scalafmt/round_trip_test.bzl` runs the phase's own `.format-test`/`.format` scripts directly, matching what the original shell test did with `bazel run`. It needs `tags = ["local"]`: those scripts locate their compiled outputs by walking their own path up to the bazel-out `bin` directory, which only works against the real output tree, not a sandboxed copy. It also invokes them via `bash` explicitly, not their shebang, to keep working on Windows (the original shell test needed the same thing via `--run_under=bash`).

`test_cross_build/scalafmt/round_trip_test.bzl` can't reuse that approach: every fixture there pins its own `scala_version`, so the fixture's compiled output lives under a different build configuration than the wrapping test, and the `bin`-directory walk resolves to the wrong directory. It reads the phase's own `manifest.txt`/`.fmt.output` action outputs directly instead (recorded once by the phase, so no need to reproduce the fixture's own path logic) - this works under any sandboxing and any configuration, so `test/scalafmt` could adopt the same approach later.

`test_cross_build` didn't have `rules_shell` as a direct dependency (only inherited from `rules_scala` as a `dev_dependency`, and bzlmod doesn't propagate those to modules that depend on it) - added it, which is why `MODULE.bazel.lock` picked up a large diff (`bazel mod deps --lockfile_mode=update`, not hand-edited).

Fully retires `test/shell/test_scalafmt_helper.sh`; nothing else used it after removing its last two callers.

### Motivation

Moves shell tests off the legacy `test_rules_scala.sh`/`test_cross_build.sh` runners toward native `bazel test //...`.


test_cross_build/scalafmt/round_trip_test.bzl
```diff
@@ -0,0 +1,40 @@
+load("@rules_shell//shell:sh_test.bzl", "sh_test")
```
PawelLipski, line 1: To be considered in a separate PR: do we want to keep the structure with separate test/ and test_cross_build/? or should the latter just go under test/ like everything else?

PawelLipski (approved):

test_cross_build/scalafmt/round_trip_test.bzl
bor-p-s: One thing worth factoring into that discussion: `test_cross_build`'s `MODULE.bazel` sets a different `scala_config.settings()` default (`scala_version = "3.1.3"`, a wider `scala_versions` list) than the root `MODULE.bazel`. `scala_config`'s own implementation reads tags via `root_module_tags` -- root module only, by design (same as bzlmod's general override semantics: https://bazel.build/external/extension). So a `scala_config.settings()` call only takes effect for whichever module is root in that particular `bazel` invocation; folding `test_cross_build` under `test/` would mean losing its own default.

`test_cross_build_default_2x` specifically exercises the case where the module's own default is 2.x while individual targets pin `scala_version` to 3.x (the select-based `common_toolchain_deps` path) - that needs `test_cross_build` to actually be root with its own default, not just a directory under the main tree's root.

This isn't new with bzlmod either - `test_cross_build` predates it (#1587/#1588, issue #1290), for the same underlying reason: a second default toolchain set couldn't coexist in one `WORKSPACE` either.

Worth keeping in mind for whatever shape that follow-up takes.

PR #1938, opened 2026-08-21
test: Declare the inputs of joern_test/dicer_test

bor-p-s:
### Description

`joern_test`/`dicer_test` run a nested `bazel test` against a downstream consumer wired to this checkout via `local_path_override`, so rules_scala's own sources are code under test without being declared inputs. The `external` tag forced them to always re-run, which is why they dominate the ubuntu CI step (~1026s / ~257s).

`@rules_scala_source_fingerprint//:fingerprint.txt` (a `repository_rule` in `source_fingerprint.bzl`) hashes every file under `scala/`, `src/`, `third_party/` -- the only trees the consumer can reach -- and both tests declare it as `data`, with `external` replaced by `no-sandbox` + `no-remote-exec` (`local` also blocks the local test-result cache).

The rule uses `repository_ctx.watch_tree()` (Bazel 7.1+) instead of `local = True`: `local` repos have a known staleness bug ([bazelbuild/bazel#16217](https://github.com/bazelbuild/bazel/issues/16217)) where changes aren't reliably picked up without a `bazel shutdown`. `watch_tree` ties re-evaluation to Bazel's own file-watching, and only runs when something in the requested build actually depends on it -- not on every invocation, the way a workspace-status-command hash would.

An earlier version of this PR hand-maintained a `package_sources` filegroup per package (35 of them) instead; dropped once `watch_tree` turned out to cover the same ground with no per-package list to keep in sync.

### Motivation

Lets both tests come back `(cached) PASSED` on an unrelated rerun instead of always re-executing.

### Test plan

- [x] `bazel test` twice on Linux CI -- second run showed both `(cached) PASSED` (build [#6690](https://buildkite.com/bazel/rules-scala-scala/builds/6690)); the ubuntu `bazel test //...` step's critical path dropped from ~1042s to 62.6s.
- [x] Touched a file under `src/java/io/bazel/rulesscala/scalac/` on Linux CI, reverted after (build [#6692](https://buildkite.com/bazel/rules-scala-scala/builds/6692)) -- both re-executed (plain `PASSED`, critical path back to ~1049s).


PawelLipski (approved):

test/community_build/source_fingerprint.bzl
```diff
@@ -0,0 +1,37 @@
...
+them changes it, and an unrelated rerun is served `(cached) PASSED`.
+
+Unlike a `local = True` repository rule, `repository_ctx.watch_tree` ties
+re-evaluation to Bazel's own file-watching (added in Bazel 7.1.0 specifically
+to replace `local`'s unreliable invalidation -- see
+https://github.com/bazelbuild/bazel/issues/16217) rather than "always treat as
+stale", so this is skipped whenever nothing in the three trees changed.
+"""
+
+_ROOTS = ["scala", "src", "third_party"]
```
PawelLipski: Hmm should `scala_proto` also be included (for `dicer_test`)? or `MODULE.bazel`?

test/community_build/downstream_test.bzl
```diff
@@ -4,6 +4,23 @@ checkout, wired against *this* rules_scala checkout: first repins its Maven
 lockfile(s) (`bazel run @maven//:pin`, since the consumer's exact pin may not
 match a version this checkout's third_party repos carry), then runs a
 nested `bazel test` against `targets`.
+
+Cacheability
```
PawelLipski, line 8: Also, pls update `test/community_build/README.md`, around line 30

PawelLipski (changes requested):
As in the comments

test/community_build/source_fingerprint.bzl
bor-p-s: Good catch — the gap was bigger than just `scala_proto`. `source_fingerprint.bzl` now covers every top-level directory a `bazel_dep(rules_scala)` consumer can load from: `scala_proto`, `protoc`, `testing`, `thrift`, `twitter_scrooge`, `junit`, `specs2`, `jmh`, `scalatest`, `java_stub_template`, plus `MODULE.bazel`, `MODULE.bazel.lock`, root `BUILD`, and `scala_config.bzl`. I checked each one against actual `load()` calls in docs, `examples/`, and rules_scala's own `.bzl` files.

Also added `exposed_top_level_dirs_test`: it fails and names the directory if a new one shows up that isn't classified as public or internal, so this can't quietly go stale again.

(Would rather have used `watch_tree`'s `exclude` param and watched the whole root minus a short internal-only list — no per-directory list to maintain at all — but that needs Bazel 8+, and this repo pins 7.7.1.)

test/community_build/downstream_test.bzl
bor-p-s: Done.

MODULE.bazel
```diff
@@ -319,3 +319,17 @@ downstream_consumers.consumer(
     remote = "https://github.com/databricks/dicer.git",
 )
 use_repo(downstream_consumers, "dicer", "joern")
+
+# Hashes scala/, src/, third_party/ (what downstream_consumers' targets can
```
PawelLipski: Update the comment

test/community_build/source_fingerprint.bzl
```diff
@@ -0,0 +1,163 @@
...
+]
+
+# Test/CI/dev fixtures and repo metadata, none of it reachable from a
+# consumer's build: directories, then loose root files. `.git` and `tmp`
+# aren't source. Bazel's own convenience symlinks (`bazel-bin`, `bazel-out`,
+# `bazel-testlogs`, `bazel-<checkout-dir-name>` -- the last one varies by
+# checkout, e.g. differs in CI) are not listed here at all;
+# exposed_top_level_dirs_test.sh skips anything matching `bazel-*` by pattern
+# instead of by name.
+_INTERNAL_ENTRIES = [
```
PawelLipski, line 67: Also, maybe skip `.DS_Store` and `.idea` etc.?

test/community_build/exposed_top_level_dirs_test.sh
```diff
@@ -0,0 +1,48 @@
...
+  [[ "${found}" -eq 0 ]] && unknown+=("${entry}")
+done <<< "${actual}"
+
+if [[ "${#unknown[@]}" -eq 0 ]]; then
+  echo "OK: every top-level entry is accounted for."
+  exit 0
+fi
+
+echo "New top-level entry not in source_fingerprint.bzl's _PUBLIC_DIRS," >&2
+echo "_PUBLIC_ROOT_FILES, or _INTERNAL_DIRS -- classify it and add it there:" >&2
```
PawelLipski: `_INTERNAL_ENTRIES`? not dirs

test/community_build/source_fingerprint.bzl
```diff
@@ -0,0 +1,163 @@
+"""A repository whose one file lists every file rules_scala exposes to a
+downstream consumer's `local_path_override` (see downstream_repository.bzl),
+with its content hash.
+
+`repository_ctx.watch_tree`'s `exclude` param (which would let this watch the
+whole checkout root minus a short internal-only list) needs Bazel 8+; this
```
PawelLipski, line 6: Maybe add an issue + TODO for that

PawelLipski (approved):

test/community_build/source_fingerprint.bzl
bor-p-s: Opened #1945 and added a `TODO(#1945)` in the docstring.

MODULE.bazel
bor-p-s: Updated.

test/community_build/source_fingerprint.bzl
bor-p-s: Good idea. Added `_LOCAL_ONLY_FILES` (`.DS_Store`, `.idea`, `.vscode`) and pruned them from the hash, so browsing/opening a directory locally doesn't invalidate the fingerprint.

test/community_build/exposed_top_level_dirs_test.sh
bor-p-s: Fixed — renamed to `_INTERNAL_ENTRIES` everywhere and reworded the comments/messages to say "entry" instead of "directory".

PR #1939, opened 2026-08-24
fix: Run nested expect_* tests in 3 lanes instead of exclusive

bor-p-s:
### Description

Splits the shared nested output base `expect_build_failure_test` (and its sibling macros) use into 3 lanes (`hash(name) % 3`, picked via `--lane` in `nested_bazel_setup`), replacing the `exclusive` tag from #1894 with per-lane isolation: tests in different lanes run concurrently.

Also switches these tests' default `tags` from `local` to `no-sandbox` + `no-remote-exec`. Both keep the nested `bazel` outside the sandbox, which is what it actually needs; `local` additionally blocks the test action itself from remote cache, while `no-sandbox` leaves that path open.

How many same-lane tests can run at once is set per platform in `.bazelrc`: 1 on Linux/Windows, so they queue behind each other there; 1000 on macOS, to dodge a Bazel scheduler bug (bazelbuild/bazel#18153) where a scarce resource delays every unrelated test scheduled alongside it. On macOS a same-lane pair can genuinely run together, which matters for a test asserting on freshly-printed build output (`--clean-before-build` in `expect_build_failure.sh`): `nested_bazel.sh` adds its own lock (`_nested_bazel_acquire_lane_lock`) around exactly that pair of nested `bazel` calls to close the gap.

### Motivation

#1894 traded speed for correctness: forcing every nested test to run one at a time removed the shared-lock timeout risk (#1884, #1885), at a disclosed cost on the critical path (+8 min on the steps that run them). This restores concurrency without reintroducing full lock contention.

The `local` -> `no-sandbox` switch matters beyond lanes: #1932 adds remote cache flags to `test_rules_scala.sh`'s bazel invocations, but a `local`-tagged test stays uncached regardless of the outer invocation's flags. Stacked together, the two PRs measured a 66% drop on Linux and roughly 50% on macOS/Windows for that CI step (numbers in #1932).

### Alternatives considered

Parallel output bases at K=4 were probed before (#1890/#1891) and rejected for disk cost (~1.7GB per lane) and a shared `--disk_cache` across lanes. This reuses the same idea at K=3; lanes here only share `--repository_cache` (external-repo downloads), already shared regardless of lane count, so the added disk cost is 2 extra ~1.7GB output bases, while `--disk_cache` contention stays exactly what it already was.

Capping the macOS lane lock at 1 too, matching Linux/Windows, was tried: it removes the same-lane race outright, at the cost of the bazel#18153 stall on unrelated tests. Kept the higher macOS capacity plus the script-level lock instead -- same safety, without that throughput cost.

### Test plan

- `bazel test -- //test/strict_dependency/... //test/missing_direct_deps/...`: 22/22 pass, including the 7 targets that take the new lock (`--clean-before-build` callers).
- Verified the lock itself outside Bazel: two concurrent shell processes on the same lane, one holding the lock for a few seconds -- the second waited the whole time and acquired it right after release. A process that exits mid-lock (simulated `exit 1`) still releases it, via the `EXIT` trap in `nested_bazel_setup`.
- Full local suite (`bazel test -- //... -//test_expect_failure/...`): 344/346 pass, 2 skipped.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -40,3 +40,32 @@ build --java_runtime_version=17
 #Windows needs --worker_quit_after_build due to workers not being shut down when the compiler tools need to be rebuilt (resulting in 'file in use' errors). See Bazel Issue#10498.
 
 build:windows --worker_quit_after_build --enable_runfiles
+
+# bazelci's own remote-cache flags (remote_caching_flags in bazelci.py) for
+# `build_targets`/`test_targets` tasks, as named configs a bazel invocation
+# must explicitly request via --config -- see test_rules_scala.sh, which
+# requests one of these for its own top-level calls while the nested `bazel`
+# calls the expect_* tests run stay on the default configuration.
+# Covers both Linux and Windows: both cache through the RBE (Remote Build
+# Execution) endpoint, only macOS (below) uses a different backend.
+build:ci-cache-rbe --remote_cache=remotebuildexecution.googleapis.com
+build:ci-cache-rbe --remote_instance_name=projects/bazel-untrusted/instances/default_instance
+build:ci-cache-rbe --google_default_credentials
+build:ci-cache-rbe --remote_max_connections=200
+
+# macOS runs on MacService instead of GCE (Google Compute Engine), so
+# bazelci.py caches it through a GCS (Google Cloud Storage) bucket instead
+# (still with --google_default_credentials).
+build:ci-cache-gcs --remote_cache=https://storage.googleapis.com/bazel-untrusted-build-cache
+build:ci-cache-gcs --google_default_credentials
+build:ci-cache-gcs --remote_timeout=3600
+build:ci-cache-gcs --remote_max_connections=200
+
+# Caps each of the 3 lanes expect_build_failure_test's nested tests hash into
+# (see the "resources:laneN_lock:1" tag in expect_build_failure.bzl) at one
+# running test at a time, so two tests in the same lane access their shared
+# nested output base one at a time. A higher macOS amount was tried, to work
+# around a Bazel scheduler bug (bazelbuild/bazel#18153) where a scarce
+# resource delays every unrelated test alongside it, but its own script-level
+# lock caused more timeouts than this flat cap, so 1 stayed uniform.
+test --local_resources=lane0_lock=1 --local_resources=lane1_lock=1 --local_resources=lane2_lock=1
--- a/test/expect_build_failure/expect_build_failure.bzl
+++ b/test/expect_build_failure/expect_build_failure.bzl
@@ -242,11 +242,18 @@ def _nested_bazel_test(
 
     # Callers default `tags` to `no-sandbox` rather than `local`: both run the
     # nested `bazel` outside the sandbox, which is all it needs, but a `local`
-    # result is also kept out of the shared cache. Added here: `exclusive`,
-    # because each run takes the shared output base's lock, so running them one
-    # at a time keeps them off each other's timeout; `no-remote-exec`, because
+    # result is also kept out of the shared cache. `no-remote-exec` because
     # this reads this machine's tree.
-    execution_tags = tags + ["exclusive", "no-remote-exec"]
+    #
+    # Splits the shared nested output base into 3 lanes (hash(name) % 3) so
+    # tests in different lanes run concurrently instead of all serializing on
+    # one shared output base. The `resources:laneN_lock:1` tag asks each test
+    # for one unit of its lane's lock; .bazelrc caps every laneN_lock at 1, so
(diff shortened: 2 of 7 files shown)


test/expect_build_failure/nested_bazel.sh
```diff
@@ -84,14 +84,6 @@ _nested_bazel_real_home=""
 _nested_bazel_common_opts=()
 _nested_bazel_symlink_prefix=""
 
-# Prepares the environment for nested `bazel` invocations and `cd`s into the real
```
PawelLipski, line 87: It was a misplaced comment right? no content changes?

.bazelrc
```diff
@@ -39,4 +39,33 @@ build --java_runtime_version=17
...
+build:windows --worker_quit_after_build --enable_runfiles
+
+# bazelci's own remote-cache flags (remote_caching_flags in bazelci.py) for
+# `build_targets`/`test_targets` tasks, as named configs a bazel invocation
+# must explicitly request via --config -- see test_rules_scala.sh, which
+# requests one of these for its own top-level calls while the nested `bazel`
+# calls the expect_* tests run stay on the default configuration.
+# Covers both Linux and Windows: both cache through the RBE (Remote Build
+# Execution) endpoint, only macOS (below) uses a different backend.
+build:ci-cache-rbe --remote_cache=remotebuildexecution.googleapis.com
```
PawelLipski, line 51: Hmmm deja vu: haven't these changes been introduced already (maybe in a slightly different shape) in different PR recently?

test/expect_build_failure/nested_bazel.sh
bor-p-s: Not moved — matches master's original layout (comment directly above `nested_bazel_setup`, same as before this PR). Only real change: dropped the `Arg 2: lane` bit, since that argument's gone now.

test/expect_build_failure/expect_build_failure.bzl
```diff
@@ -242,11 +242,18 @@ def _nested_bazel_test(
 
     # Callers default `tags` to `no-sandbox` rather than `local`: both run the
     # nested `bazel` outside the sandbox, which is all it needs, but a `local`
-    # result is also kept out of the shared cache. Added here: `exclusive`,
-    # because each run takes the shared output base's lock, so running them one
-    # at a time keeps them off each other's timeout; `no-remote-exec`, because
+    # result is also kept out of the shared cache. `no-remote-exec` because
     # this reads this machine's tree.
-    execution_tags = tags + ["exclusive", "no-remote-exec"]
```
PawelLipski, line 249: By the way, there's quite a lot of `exclusive` tags on targets/macros in the project... pls make sure that none are left on the targets that use `_nested_bazel_test` (unless it's desired for some reason? then add a relevant comment)

test_rules_scala.sh
```diff
@@ -10,20 +10,46 @@ fi
 test_dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/test/shell
 # shellcheck source=./test_runner.sh
 . "${test_dir}"/test_runner.sh
+# shellcheck source=./test_helper.sh
+. "${test_dir}"/test_helper.sh
 runner=$(get_test_runner "${1:-local}")
 test_output_flag="--test_output=errors"
 
-$runner bazel build src/... test/...
+# bazelci's `build_targets`/`test_targets` tasks get their own --remote_cache
```
PawelLipski, line 38: Isn't it a content of a separate PR? has it been merged into this one (by accident?)?

.bazelrc
bor-p-s: Yeah, deja vu is right — #1932 was merged directly into this PR's own branch (`bor-p-s/nested-test-lanes-stacked`), not into master, so its changes show up in this PR's diff too. It'll stop showing up once #1932 (or this whole PR) actually lands on master.

test_rules_scala.sh
PawelLipski: (ok, #1932)

test/expect_build_failure/expect_build_failure.bzl
bor-p-s: Checked every caller of the four `_nested_bazel_test` macros — none of them pass `exclusive` in `tags`. Every `exclusive` tag left in the repo belongs to a separate, plain `sh_test` outside these macros.

One of those (`scalac_exec_group/BUILD`'s `scalac_action_exec_properties_test`) had a comment justifying its `exclusive` with "shares the nested output base of the other nested-Bazel tests, whose lock serializes them" — stale on both counts (the lock's gone, and this test never actually shared an output base with anyone). Fixed the comment.

PawelLipski (approved):

PR #1940, opened 2026-08-24
refactor: Migrate test_scala_import_source_jar.sh to Bazel-native sh_test targets

bor-p-s:
### Description

Migrates the 3 checks in `test_scala_import_source_jar.sh` into 3 `sh_test` targets in a new `test/fetch_sources/` package. All three build `//test/fetch_sources/...` and check whether `@com_google_guava_guava_21_0//:guava-21.0-src.jar` got fetched, based on the `BAZEL_JVM_FETCH_SOURCES` env var. This doesn't fit the 4 macros in `expect_build_failure.bzl` (it's a post-build file-existence check, not a build/test success-or-failure match), so it's a bespoke `sh_test` following `expect_no_ijar.sh`'s pattern, same as `scaladoc_test.sh`.

The three targets share one nested output base (matches the old shell test, which relied on the repository rule's `environ` declaration to force a re-fetch whenever the env var changes, so test order doesn't matter). Each test resolves the source jar's on-disk path itself, since the file may not exist at all in the "should not be fetched" case: resolving that label directly would fail outright rather than just report absence. Instead each test resolves the always-present main jar's path and derives the source jar's path from its directory.

Moves the `fetch_sources` fixture out of `test/src/main/scala/scalarules/test/fetch_sources/` (a nested source-fixture package, not used for tests) into a new top-level `test/fetch_sources/` package, matching every other migrated bespoke test (`test/scaladoc`, `test/diagnostics_reporter`, `test/macros`). Nothing else in the repo references the old path.

### Motivation

Deletes `test_scala_import_source_jar.sh` and its `test_rules_scala.sh` line.

Release impact: tests only.


test/shell/test_scala_import_source_jar.sh
PawelLipski, line 1: There's still a reference to this file in `scala/private/extensions/dev_deps.bzl:77`

test/fetch_sources/BUILD
```diff
@@ -0,0 +1,66 @@
...
+        "requires-network",
+    ],
+)
+
+sh_test(
+    name = "not_fetched_when_env_var_non_true_test",
+    size = "large",
+    srcs = ["fetch_sources_test.sh"],
+    args = [
+        "--env-value=false",
```
PawelLipski, line 54: Maybe also test for a wrong value (non-empty but neither `true` or `false`, like `1`)

test/fetch_sources/fetch_sources_test.sh
```diff
@@ -0,0 +1,74 @@
...
+
+# shellcheck source=test/expect_build_failure/nested_bazel.sh
+source "${TEST_SRCDIR:-${RUNFILES_DIR:-$0.runfiles}}/${TEST_WORKSPACE:-_main}/test/expect_build_failure/nested_bazel.sh"
+nested_bazel_setup "rules_scala_fetch_sources_test_output_base"
+
+if [[ -n "${ENV_VALUE}" ]]; then
+  export BAZEL_JVM_FETCH_SOURCES="${ENV_VALUE}"
+fi
+
+target="//test/fetch_sources/..."
```
PawelLipski: Nit: hmmm building all targets? just `fetch_sources` should be enough right? (as the rest is `sh_tests`)

PawelLipski (approved):

test/shell/test_scala_import_source_jar.sh
bor-p-s: Good catch, fixed — updated the comment to point at //test/fetch_sources instead.

test/fetch_sources/BUILD
bor-p-s: Good idea, added not_fetched_when_env_var_invalid_test (--env-value=1).

test/fetch_sources/fetch_sources_test.sh
bor-p-s: Right, fixed to build just //test/fetch_sources:fetch_sources.

PawelLipski (approved):

PR #1941, opened 2026-08-24
fix: Build third_party/test/proto again

bor-p-s:
### Description

`third_party/test/proto` (a self-contained bzlmod module used by `test/shell/test_scala_proto_library.sh`) does not build right now, for two independent reasons:

1. Its `MODULE.bazel` had no `bazel_dep(rules_proto)`, so `@rules_proto//proto:defs.bzl` (loaded by its own `BUILD.bazel`) could not resolve.
2. Past that, `@rules_scala//scala:unused_dependency_checker_error_toolchain`, which this module registers, needs `enable_compiler_dependency_tracking = True` in `scala_config` -- it wasn't set here.

Adds the missing `bazel_dep` (version `7.1.0`, matching `test/proto_cross_repo_boundary/repo/MODULE.bazel`, the other module with the same setup) and a `scala_config.settings()` call turning on `enable_compiler_dependency_tracking`, both `dev_dependency = True` like this module's other extensions. Regenerated the lockfile to match.

### Motivation

Release impact: tests only.


third_party/test/proto/MODULE.bazel
```diff
@@ -2,6 +2,7 @@
 
 module(name = "proto")
 
+bazel_dep(name = "rules_proto", version = "7.1.0")
```
PawelLipski, line 5: Actually, pls remind me, how come was CI passing without these fixes?

PawelLipski (approved):

third_party/test/proto/MODULE.bazel
PawelLipski: (even though this module is bazelignored - wasn't it still needed/built/analyzed in any way?)

bor-p-s: The only CI signal that ever touched this module's build was the old shell test `test_scala_proto_library_action_label`. Its assertion was dead code: `if [ "$action" != $expected ]` left `$expected` unquoted, so `[` errored with "too many arguments" and `if` read that as false, so the failure branch never ran. The test passed no matter what `bazel aquery` printed.

On the `.bazelignore` question: right, it wasn't built any other way. `bazel build //...` never discovers packages under an ignored path -- confirmed this by checking `.bazelci/presubmit.yml`, every top-level task builds/tests `//...` and none mention this module. The only lint check that touches it (`check_module_bazel`, part of `test_bzlmod_tidy.sh`) only runs `bazel mod tidy`, which resolves the `bazel_dep`/`use_extension` graph but never builds or analyzes a target. I confirmed this too: ran `bazel mod tidy` against the original, unfixed `MODULE.bazel` and it exits clean with zero changes, so this check could not have caught either bug.

PR #1946, opened 2026-08-26
refactor: Migrate the last 2 test_misc.sh cases to Bazel-native sh_test

bor-p-s:
### Description

#1921 migrated 5 of `test/shell/test_misc.sh`'s 7 cases and left `test_override_javabin` and `xmllint_test` for a follow-up. This is that follow-up, and it deletes the file (and its `test_rules_scala.sh` line) for good.

`test_override_javabin` becomes a plain `sh_test` (`test/override_javabin`): it builds `//test:ScalaBinary` as a normal `data` dependency and runs the resulting launcher directly with `JAVABIN=/etc/basdf`, asserting the run fails. `JAVABIN` handling lives entirely in the generated launcher (`phase_write_executable.bzl`'s preamble sets it from the env), so running the built launcher directly reproduces what `bazel run` does. `JAVABIN` is a bash-only mechanism; on Windows Bazel uses a native `.exe` launcher instead, so the target is `target_compatible_with`-excluded there, matching the original's `is_windows` skip.

`xmllint_test` is dropped instead of migrated. It validated that XML files under `bazel-testlogs` -- generated by every other test in the old shell suite's run -- were well-formed, which needs visibility into the whole suite's combined output; a single `sh_test` only sees its own run. Scala tests' XML comes from rules_scala's own `JUnitXmlReporter.scala` (a hand-rolled `scala.xml` serializer writing to `XML_OUTPUT_FILE`), not a Bazel-internal mechanism, so the check did have something real to catch. But checking it needs a run of the whole suite, which doesn't fit one `sh_test`'s scope -- a real regression test for the reporter's XML output belongs on the reporter itself, not as a leftover shell script.

### Motivation

Release impact: tests only.


test/override_javabin/BUILD
```diff
@@ -0,0 +1,18 @@
+load("@rules_shell//shell:sh_test.bzl", "sh_test")
+
+package(default_testonly = 1)
+
+# JAVABIN is read by the bash launcher scala_binary/scala_test generate
+# (see phase_java_wrapper.bzl); on Windows Bazel uses a native .exe launcher
```
PawelLipski: Maybe `see phase_write_executable.bzl`? as that's where the env var is captured

test/override_javabin/scala_binary_override_javabin_test.sh
```diff
@@ -0,0 +1,18 @@
...
+# (see the "Set JAVABIN to the path to the JVM launcher" preamble Bazel's
+# java_stub_template.txt generates).
+#
+# Usage: scala_binary_override_javabin_test.sh <launcher>
+
+set -euo pipefail
+
+launcher="${1:?Usage: scala_binary_override_javabin_test.sh <launcher>}"
+
+if output="$(JAVABIN=/etc/basdf "${launcher}" 2>&1)"; then
```
PawelLipski: Maybe assert on output (unless output is very brittle/not worth it)

PawelLipski (approved):

test/override_javabin/BUILD
bor-p-s: Good catch, fixed -- it's phase_write_executable.bzl that captures JAVABIN, not phase_java_wrapper.bzl.

test/override_javabin/scala_binary_override_javabin_test.sh
bor-p-s: Added an assertion that the failure output mentions the bad JAVABIN path (/etc/basdf) -- checked it's stable (bash's exec failure always names the path).

PR #1949, opened 2026-08-27
test: Migrate test_semanticdb.sh to Bazel-native sh_test targets

bor-p-s:
### Description

The other 4 functions check post-build state: provider values read from a generated script, whether `.semanticdb` output was produced, and whether a source glob picks up an added/removed file. That's outside what `expect_build_failure.bzl`'s macros model (a build/test pass/fail outcome), so these are hand-written `sh_test` scripts on `nested_bazel.sh`, the same pattern as `expect_no_ijar.sh`.

`nested_bazel.sh` gets two small fixes for the new `bazel query` call it needs: it now skips `--symlink_prefix` for `query` specifically, since that's the one subcommand that rejects the flag, and its Windows arg-conversion list now covers `labels(` too, the same label-mangling guard it already has for `--extra_toolchains=`. This second fix is reasoned from the existing guard, not verified on a real Windows run.

`scala/BUILD` and `scala/private/BUILD` each gain one `exports_files` entry (`semanticdb_provider.bzl`, `phases/phase_semanticdb.bzl`), so the Scala 2 `produces_semanticdb_test` variants can list them as `data`. Their fixture's only own action is a template expansion, which the fingerprint mechanism reads by mnemonic only, so a bug in the provider's fields wouldn't otherwise bust the test's cache.

### Motivation

Continues the move off the old shell-test runner (`test_rules_scala.sh`) toward native `bazel test //...`.

Release impact: tests only.


test/expect_build_failure/nested_bazel.sh
```diff
@@ -141,8 +142,10 @@ nested_bazel_setup() {
     _nested_bazel_common_opts+=("--repository_cache=${repository_cache}")
   fi
   # Keep the nested build's convenience symlinks out of the workspace so they do
-  # not clobber the parent invocation's `bazel-bin` etc.
-  _nested_bazel_common_opts+=("--symlink_prefix=${_nested_bazel_output_base}/convenience_symlinks/")
+  # not clobber the parent invocation's `bazel-bin` etc. Kept separate from
```
PawelLipski: Oh, so up to now there was no `bazel query` run via nested bazel?

I remember you've used `cquery` instead of `query` (probably for this reason) somewhere recently right?

test/semanticdb/BUILD
```diff
@@ -80,6 +82,126 @@ scala_library(
     ),
 )
 
+# The nested `bazel build` here reads the on-disk checkout directly, so its
+# inputs must be declared by hand -- same reasoning as the expect_*_test
```
PawelLipski, line 86: Hmmm there aren't `expect_*_test` macros above 🤔

test/semanticdb/BUILD
```diff
@@ -80,6 +82,126 @@ scala_library(
...
+        "no-remote-exec",
+        "no-sandbox",
+        "requires-network",
+    ],
+)
+
+[
+    sh_test(
+        name = "produces_semanticdb_test_scala%d_%s" % (
+            scala_majver,
```
PawelLipski: Nit: IIRC the scheme calls these segments `EPOCH.MAJOR.MINOR`, pls verify & if so consider changing to `scala_epoch` (unless this repo uses `major` for that already)

test/semanticdb/produces_semanticdb_test.sh
```diff
@@ -0,0 +1,99 @@
...
+#
+# Usage:
+#   produces_semanticdb_test.sh <toolchain-label> <is-bundle:0|1> <scala-majver:2|3> [extra-bazel-arg]...
+
+set -euo pipefail
+
+# shellcheck source=test/expect_build_failure/nested_bazel.sh
+source "${TEST_SRCDIR:-${RUNFILES_DIR:-$0.runfiles}}/${TEST_WORKSPACE:-_main}/test/expect_build_failure/nested_bazel.sh"
+
+toolchain="${1:-}"
```
PawelLipski: Nit: use named flag for these arguments, update `test/semanticdb/BUILD` accordingly

test/semanticdb/no_semanticdb_test.sh
```diff
@@ -0,0 +1,43 @@
...
+rm -rf "${target_out}"
+
+if ! build_output="$(nested_bazel_run build //test/semanticdb:all_lib 2>&1)"; then
+  echo "Expected build of //test/semanticdb:all_lib to succeed." >&2
+  echo "${build_output}" >&2
+  exit 1
+fi
+
+if [[ "$(find "${target_out}" -type f -name '*.semanticdb' | wc -l)" -gt 0 ]]; then
+  echo "Error: Semanticdb files erroneously found in target output" >&2
```
PawelLipski: Oh, so it shouldn't be a loose file but inside a jar instead? if so, pls clarify the error message

test/semanticdb/produces_semanticdb_test.sh
```diff
@@ -0,0 +1,99 @@
...
+bazel_bin="$(nested_bazel_run info "${build_args[@]}" bazel-bin 2>/dev/null)"
+execution_root="$(nested_bazel_run info "${build_args[@]}" execution_root 2>/dev/null)"
+
+# The provider values are read from this build-generated script rather than
+# from Starlark, since the assertions run in this sh_test, not in an aspect.
+# shellcheck disable=SC1090,SC1091
+source "${bazel_bin}/test/semanticdb/semantic_provider_vars_all.sh"
+
+if [[ "${semanticdb_enabled}" -ne 1 ]]; then
+  echo "Error: SemanticdbInfo.semanticdb_enabled not equal to true" >&2
```
PawelLipski: It's not clear from error message what kind of config (?) key it is and where it's read from

PawelLipski: (same for the ones below)

PawelLipski (approved):

test/expect_build_failure/nested_bazel.sh
bor-p-s: Right, switched to `cquery` — it already accepts `--symlink_prefix` and supports `labels(srcs, ...)`. Reverted the `query`-specific branching in `nested_bazel_run`, so this file's diff is now just the `labels(` addition to `MSYS2_ARG_CONV_EXCL` (still needed since the label sits mid-string here too).

test/semanticdb/BUILD
bor-p-s: You're right, my mistake — the macros are below this comment, not above. Fixed the wording.

bor-p-s: You're right, my mistake — the macros are below this comment, not above. Fixed the wording.

test/semanticdb/BUILD
bor-p-s: Confirmed — extract_major_version already returns "2.13"/"3.3" here, so scala_majver collided with that. Renamed to scala_epoch.

test/semanticdb/produces_semanticdb_test.sh
bor-p-s: Done — switched to --toolchain=, --is-bundle=, --scala-epoch=, updated BUILD to match.

test/semanticdb/no_semanticdb_test.sh
bor-p-s: Right, it's loose files, the jar check is separate below. Reworded both messages to say which and name the file.

test/semanticdb/produces_semanticdb_test.sh
bor-p-s: Reworded to show actual vs expected, plus one comment noting these are SemanticdbInfo fields from scala/semanticdb_provider.bzl.

test/expect_build_failure/nested_bazel.sh
PawelLipski: Up to you - `query` + fix to `nested_bazel.sh` can also stay here and in other places (`cquery` can arguably be confusing as we don't care about configured graph here)

test/semanticdb/produces_semanticdb_test.sh
bor-p-s: Same fix applied to all four checks.

test/expect_build_failure/nested_bazel.sh
bor-p-s: Fair point on cquery being confusing here — switched back to query + the --symlink_prefix skip, since we really just need srcs membership, not the configured graph.

PawelLipski (approved):

PR #1951, opened 2026-08-27
refactor: Migrate test_scala_library.sh to Bazel-native sh_test targets

bor-p-s:
### Description

Migrates `test/shell/test_scala_library.sh` (4 cases) to native `sh_test` targets, sharing one bespoke test script, `test/expect_build_failure/no_recompilation_test.sh`: the 4 `expect_build_failure.bzl` macros assert a single build/test outcome, while these checks need two sequential builds with a fixture mutation in between, following `expect_no_ijar.sh`'s and `disappearing_class_test.sh`'s pattern for that shape of check.

Each case builds a target, replaces a caller-supplied word inside a quoted string literal in a fixture, builds again with `--subcommands`, and asserts a specific action stays out of the second build's output. 3 cases assert a downstream Scala consumer's compiled output stays cached after an internal-only change to its dependency (the ijar/interface-jar contract): 2 in `ijar/BUILD` against the `user`/`dependency` fixture, 1 in `strict_deps/no_recompilation/BUILD` against a transitive-dependency fixture built under `--extra_toolchains=//test/toolchains:high_level_transitive_deps_strict_deps_error`. The 4th, also in `ijar/BUILD`, asserts `dependency`'s own Java sibling output stays cached after an internal-only change to its Scala sibling. All 4 restore the mutated file (from the `sed` backup) on exit either way.

Args are plain words: an embedded `(` or `"` survives Bazel's `args` tokenization but becomes a shell syntax error once Bazel's Windows test wrapper re-parses argv through its own shell.

All 4 mutate the real source checkout and share a nested output base, so they're tagged `exclusive` to prevent another test from interleaving between the initial build, fixture mutation, and verification build. Two of the `ijar` cases also mutate the same file, `B.scala`.

### Motivation

Release impact: tests only.


test/expect_build_failure/no_recompilation_test.sh
```diff
@@ -0,0 +1,92 @@
...
+# its consumer, since the consumer only ever compiles against the target's
+# interface jar.
+#
+# Builds <target> once against a nested output base, then replaces the
+# quoted string literal "<search-word>" with "<replace-word>" in
+# <changed-file> in the real source tree, and builds <target> again with
+# --subcommands. Fails if <not-expected> appears in the second build's
+# output (meaning the consumer action ran, i.e. it recompiled).
+#
+# <search-word>/<replace-word> (not a raw sed expression) so the args stay
```
PawelLipski: Pls move to right before the `sed` invocation 😅  it's irrelevant in the overview doc (pls in fact instruct your agents to avoid this pattern)

PawelLipski: On a similar note - pls tell the agents to avoid loading the PR description with implementation details, let them instead post the details as github comments close to relevant places

test/expect_build_failure/no_recompilation_test.sh
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env bash
+#
+# Regression test for the ijar/interface-jar correctness contract: a change to
```
PawelLipski, line 3: Nice (even if pre-existing)

PawelLipski: This project was (before we picked it up)/is in fact pretty well tested, just the test invocations were somewhat suboptimal 😅  and the missing downstream tests were IMO a major gap

PawelLipski (approved):

test/expect_build_failure/no_recompilation_test.sh
bor-p-s: Fixed.

PR #1952, opened 2026-08-27
refactor: Migrate test_bzlmod_macros.sh to Bazel-native sh_test targets

bor-p-s:
### Description

Migrates all 13 cases in `test/shell/test_bzlmod_macros.sh` (tests for `scala/private/macros/bzlmod.bzl`'s `root_module_tags`/`single_tag_values`/`repeated_tag_values`) into `sh_test` targets in a new `test/bzlmod_macros/` package.

Each case creates a standalone Bazel module that consumes rules_scala through `local_path_override`, matching how an external bzlmod consumer sees its module extensions, then runs a nested `bazel run --enable_bzlmod` and checks its output. `bzlmod_macros_test.sh` takes a `--case=<name>` flag and picks a target/tag-lines/expected-output from a table, one entry per case; the fake-root-module case keeps its own code path since it wraps the module in a second one. The fixture files (`BUILD.bzlmod_test`, `MODULE.bzlmod_test`, `MODULE.bzlmod_test_root_module`, `bzlmod_test_ext.bzl`) move from `scala/private/macros/test/` into the new package, next to their tests.

Scratch state for each nested module now lives under `TEST_TMPDIR` instead of a fixed `/tmp` path, so each checkout of this repo running the suite keeps its own fixture files, even at the same time as another checkout. The fake-root case also gets its own copy of `.bazelversion`/`.bazelrc`, matching every other case. Bazelisk reads these files to resolve the repo's pinned Bazel version from that scratch directory; it silently falls back to a different version when they're missing.

`deps/latest/MODULE.bazel` is added to the root `BUILD`'s `exports_files` because the tests need it as a `data` input (the fixture module declares `bazel_dep(name = "latest_dependencies")` with a `local_path_override` to it). The root `BUILD` is the nearest package that can export it, since `deps/latest/` holds only the `MODULE.bazel` file itself.

`test/expect_build_failure/nested_bazel.sh` now clears `RUNFILES_DIR`/`RUNFILES_MANIFEST_FILE`/`RUNFILES_MANIFEST_ONLY`/`JAVA_RUNFILES` before a nested `bazel` call. These env vars carry over from the outer `sh_test`'s own runfiles, so any `sh_binary` a nested `bazel run` builds and executes would read them and look in the outer test's runfiles tree instead of its own. This only broke on Windows: its `sh_binary` launchers read those env vars, while other platforms resolve runfiles through real symlinks instead. It only showed up here because this is the first nested test that actually runs a built binary, rather than just building or testing one. The same fix already exists in `test/community_build/downstream_test_driver.sh` for the same reason; this puts it in the shared helper so every nested-bazel test gets it.

Fully retires `test/shell/test_bzlmod_macros.sh`, deleted along with its `test_rules_scala.sh` line.

### Motivation

Moves shell tests off the legacy `test_rules_scala.sh` runner toward native `bazel test //...`.


test/bzlmod_macros/bzlmod_macros_test.sh
```diff
@@ -0,0 +1,239 @@
...
+run_target() {
+  nested_bazel_run run --enable_bzlmod "$1" 2>&1
+}
+
+target=""
+tag_lines=()
+expect_regex=""
+expect_fail_regex=""
+
+case "${case_name}" in
```
PawelLipski: Hmmm instead of hardcoding the cases in a `case`... would it be possible to pass the config to the script directly from test/bzlmod_macros/BUILD?

Maybe create a macro that would wrap the current `sh_test` (which would accept args like `target`, `expect_regex` etc.) - feels more idiomatic

bor-p-s: Done — moved the config into `test/bzlmod_macros/BUILD` via a new `bzlmod_macro_test` macro in `bzlmod_macros.bzl`. `bzlmod_macros_test.sh` now just parses `--target`/`--tag`/`--expect` flags instead of the `case`.

test/bzlmod_macros/BUILD
```diff
@@ -0,0 +1,110 @@
+load(":bzlmod_macros.bzl", "bzlmod_fake_root_module_test", "bzlmod_macro_test")
+
+bzlmod_macro_test(
```
PawelLipski, line 3: Nice, much better to read

PawelLipski (approved):

PawelLipski (approved):

PR #1953, opened 2026-08-27
fix: Wire the CI remote cache into the nested joern/dicer build

bor-p-s:
### Description

`downstream_test`'s nested `bazel test` builds joern/dicer cold every run. This adds the same remote cache CI's outer `bazel` already uses, gated on `BUILDKITE=true` (inert on a developer's machine), under its own `cache-silo-key`. `--cache_test_results=no` keeps the test targets (9 for joern, 163 for dicer) executing every time -- only the build-action time gets cached.

`.bazelci/presubmit.yml` adds `--test_env=BUILDKITE` to the `ubuntu2004` task: Bazel gives test actions a fixed, curated environment regardless of sandboxing, so `BUILDKITE` reaches `downstream_test_driver.sh` only with an explicit forward.

### Motivation

`joern_test` is the longest step in `bazel test //...`: ~1026-1062s, of which ~450-470s is build time for joern's own ~1030 actions, unaffected by #1938's caching.

### DIFF
--- a/test/community_build/downstream_test_driver.sh
+++ b/test/community_build/downstream_test_driver.sh
@@ -109,9 +109,13 @@ mkdir -p "${_nested_bazel_output_base}"
 cd "${NESTED_BAZEL_WORKSPACE}"
 # This driver always uses the developer's home directory. `nested_bazel_setup`
 # does so only when asked, because a `~/.bazelrc` outside the repo could then
-# decide the result of a test whose pass is cached and reused. The tests here
-# are tagged `external`, so Bazel re-runs them every time and there is no such
-# stale pass to worry about. The paths built below need a home directory anyway.
+# decide the result of a cached, reused test pass. That risk is real here too
+# now that `downstream_test` dropped `external` for `source_fingerprint`-based
+# caching (see downstream_test.bzl): a developer's own `~/.bazelrc` still
+# isn't a declared input, so a `~/.bazelrc` edit between two runs with an
+# otherwise-unchanged fingerprint can go unnoticed by the cache. Scrubbing or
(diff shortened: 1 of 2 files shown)


test/community_build/downstream_test_driver.sh
```diff
@@ -109,9 +109,13 @@ mkdir -p "${_nested_bazel_output_base}"
...
-# decide the result of a test whose pass is cached and reused. The tests here
-# are tagged `external`, so Bazel re-runs them every time and there is no such
-# stale pass to worry about. The paths built below need a home directory anyway.
+# decide the result of a cached, reused test pass. That risk is real here too
+# now that `downstream_test` dropped `external` for `source_fingerprint`-based
+# caching (see downstream_test.bzl): a developer's own `~/.bazelrc` still
+# isn't a declared input, so a `~/.bazelrc` edit between two runs with an
+# otherwise-unchanged fingerprint can go unnoticed by the cache. Scrubbing or
+# pinning the home configuration this nested invocation reads is follow-up
+# work. The paths built below need a home directory regardless.
```
PawelLipski, line 118: For clarity, does this comment change apply regardless of the script changes below? or is it only relevant now?

PawelLipski (approved):

test/community_build/downstream_test_driver.sh
bor-p-s: Independent of the script changes below. The `~/.bazelrc` risk comes from `downstream_test` switching from `external` to `source_fingerprint`-based caching (a separate earlier change) — that's what made a stale cached pass possible. The remote cache flags added later in this diff are unrelated; they just add caching for the nested build's own actions.

PR #1956, opened 2026-08-31
fix: Run only a smoke subset of jssrc2cpg's own test suite in joern_test

bor-p-s:
### Description

`joern_test`'s `jssrc2cpg` target is its single most expensive one: ~750-820s of the ~1050-1200s the whole test costs once the nested build is warm. Its ~11k lines of test sources are close to entirely Joern's own JS/TS semantic correctness (AST node shapes, CFG edges, dataflow paths, type recovery for hundreds of language-construct variations) -- an assertion about Joern's own correctness.

This adds one new file to Joern's checkout via a patch (`joern_jssrc2cpg_smoke.patch`, same mechanism as the existing `dicer_protobuf_version.patch`), keeping every one of Joern's own test files exactly as upstream ships them. The new file aggregates 6 existing test classes into one ScalaTest suite: representative AST/CFG/dataflow/type-recovery passes, plus the two closest things jssrc2cpg has to an integration smoke test (`ProjectParseTests`'s real `AstGenRunner`/file I/O, and `JsSrc2CpgHTTPServerTests`'s JVM HTTP server + full classpath).

`downstream_test.bzl` gets three new params:
- `filtered_targets` + `test_filter`: a target list run in a *second*, separate nested `bazel test` invocation with `--test_filter` set. Bazel's `--test_filter` applies uniformly across every target in one `bazel test` command line, so a ScalaTest suite name that exists only in jssrc2cpg's classpath needs its own invocation -- sharing the main one would fail every other target in it with "class not found". `downstream_test_driver.sh` now runs both invocations unconditionally (captures the first's exit status instead of letting `set -e` skip the second on a failure) and exits with the worse of the two.
- `patches`: declares the patch file as `data` on the `sh_test`, so an edit to the patch itself is a declared input that forces a real re-run -- `source_fingerprint.bzl`'s scope stops at what rules_scala exposes, leaving a patch under `test/` outside it. Applied here to both `joern_jssrc2cpg_smoke.patch` and the pre-existing `dicer_protobuf_version.patch`, which had the same gap.

jssrc2cpg's test sources still compile in full either way (real Scala 3 compile workload against the toolchain under test) -- only the smoke subset actually executes.

### Motivation

Measured on Linux, this test's own platform: jssrc2cpg 748-820s -> 165-215s. Combined with #1953's build-action cache, a rough estimate for the whole `joern_test` step: ~1026-1200s -> ~250-400s.

This saving also compounds inside `./test_rules_scala.sh`: `test/...` matches `joern_test`, and that script runs `bazel test .../test/...` under several different `--extra_toolchains`, each one discarding Bazel's analysis cache (confirmed directly: `WARNING: Build option --extra_toolchains has changed, discarding analysis cache`) and re-running `joern_test` in full. This cost lands up to 4 times per `./test_rules_scala.sh` run, on a step that is on the real CI critical path.

### Trade-off

`downstream_test` only cares about the toolchain signal here. There's a small chance this cuts something useful -- going ahead anyway.

### DIFF
--- a/test/community_build/BUILD
+++ b/test/community_build/BUILD
@@ -5,7 +5,11 @@
 load(":downstream_test.bzl", "downstream_test")
 load(":source_fingerprint.bzl", "exposed_top_level_dirs_test")
 
-exports_files(["downstream_test_driver.sh", "dicer_protobuf_version.patch"])
+exports_files([
+    "downstream_test_driver.sh",
+    "dicer_protobuf_version.patch",
+    "joern_jssrc2cpg_smoke.patch",
+])
 
 exposed_top_level_dirs_test(name = "exposed_top_level_dirs_test")
 
@@ -28,19 +32,37 @@ exposed_top_level_dirs_test(name = "exposed_top_level_dirs_test")
 #
 # extra_bazel_flags's --test_timeout bounds each nested test, not size
 # (which bounds this whole sh_test action).
+#
+# test_filter selects jssrc2cpg's own SmokeSuite (see
+# joern_jssrc2cpg_smoke.patch): jssrc2cpg is this test's single most
+# expensive target (~750-820s of the ~1050-1200s this whole test costs once
+# the nested build itself is warm), and its ~11k lines of test sources are
+# ~99% Joern's own JS/TS semantic correctness -- an assertion about Joern's
+# own correctness. The patch adds one new
+# file aggregating a handful of existing test classes (AST, CFG, dataflow,
+# type-recovery, plus the closest things to an integration smoke test: a
+# real AstGenRunner file-parse and a JVM HTTP server + full classpath) into
+# one suite; test_filter selects just that suite, so jssrc2cpg's test
+# sources still compile in full (real Scala 3 compile workload against the
+# toolchain under test) but only the smoke subset executes: measured
+# 748-820s -> 165s.
 downstream_test(
     name = "joern_test",
+    size = "enormous",
+    extra_bazel_flags = "--test_timeout=60,1800,1800,3600",
+    filtered_targets = ["//joern-cli/frontends/jssrc2cpg:tests"],
+    patches = ["//test/community_build:joern_jssrc2cpg_smoke.patch"],
     repo_name = "joern",
     scala_version = "3.7.4",
-    extra_bazel_flags = "--test_timeout=60,1800,1800,3600",
-    size = "enormous",
     target_compatible_with = ["@platforms//os:linux"],
     targets = [
         "//...",
         "-//joern-cli/frontends/php2cpg/...",
         "-//joern-cli/frontends/rust2cpg/...",
         "-//joern-cli/frontends/swiftsrc2cpg/...",
+        "-//joern-cli/frontends/jssrc2cpg/...",
     ],
+    test_filter = "io.joern.jssrc2cpg.SmokeSuite",
 )
 
 # dicer covers rules_scala's scala_proto/ScalaPB (gRPC) codegen path and
@@ -60,9 +82,10 @@ downstream_test(
 # "large" (900s) default in total.
 downstream_test(
     name = "dicer_test",
+    size = "enormous",
+    patches = ["//test/community_build:dicer_protobuf_version.patch"],
     repo_name = "dicer",
     scala_version = "2.12.21",
-    size = "enormous",
     target_compatible_with = ["@platforms//os:linux"],
     targets = [
         "//...",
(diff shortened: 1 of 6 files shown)


test/community_build/downstream_test_driver.sh
PawelLipski, line 143: Nit: maybe extract to separate PR?

PawelLipski: Oh, it actually looks like sth that came from the base PR 🤔

test/community_build/downstream_test_driver.sh
```diff
@@ -6,7 +6,8 @@
 # repository rule already wrote into the consumer's MODULE.bazel.
 #
 # Usage: downstream_test_driver.sh --marker-rootpath <path> --scala-version <v> \
-#   --output-base-name <name> [--extra-bazel-flags <flags>] -- <target-pattern>...
+#   --output-base-name <name> [--extra-bazel-flags <flags>] [--test-filter <value>] \
```
PawelLipski, line 9: Nit: not sure if the comment is meant to lists all flags, but `--filtered-targets` are missing

test/community_build/BUILD
```diff
@@ -28,19 +32,37 @@ exposed_top_level_dirs_test(name = "exposed_top_level_dirs_test")
...
+# real AstGenRunner file-parse and a JVM HTTP server + full classpath) into
+# one suite; test_filter selects just that suite, so jssrc2cpg's test
+# sources still compile in full (real Scala 3 compile workload against the
+# toolchain under test) but only the smoke subset executes: measured
+# 748-820s -> 165-215s.
 downstream_test(
     name = "joern_test",
+    size = "enormous",
+    extra_bazel_flags = "--test_timeout=60,1800,1800,3600",
+    filtered_targets = ["//joern-cli/frontends/jssrc2cpg:tests"],
```
PawelLipski, line 53: Hmmm why both `filtered_targets` and `test_filter` are needed?

PawelLipski: Oh okay I see, more tests in the target 🤦🏻

test/community_build/downstream_test_driver.sh
```diff
@@ -237,8 +251,32 @@ done
...
+# Captured rather than let `set -e` exit here: both invocations always run
+# regardless of the first's outcome, and the combined exit status (see
+# below) reflects either one failing.
+status=0
 # shellcheck disable=SC2086 # intentional word-splitting: extra_bazel_flags
 # and targets are each meant to expand to multiple words/patterns.
 nested_bazel_run test --test_output=errors --cache_test_results=no \
   --repo_env=SCALA_VERSION="${scala_version}" \
-  ${extra_bazel_flags} -- "${targets[@]}"
+  ${extra_bazel_flags} -- "${targets[@]}" || status="$?"
```
PawelLipski, line 267: Wait, doesn't it also run the filtered-out targets? 🤔

test/community_build/downstream_test.bzl
```diff
@@ -58,24 +92,29 @@ def downstream_test(
...
-            "--output-base-name",
-            name,
-        ] + (["--extra-bazel-flags", extra_bazel_flags] if extra_bazel_flags else []) + [
+                   "--marker-rootpath",
+                   "$(rootpath @{}//_bazel_native_marker:marker.txt)".format(repo_name),
+                   "--scala-version",
+                   scala_version,
+                   "--output-base-name",
+                   name,
+               ] + (["--extra-bazel-flags", extra_bazel_flags] if extra_bazel_flags else []) +
```
PawelLipski: Extract separate vars, this is confusing in the current shape

test/community_build/downstream_test_driver.sh
bor-p-s: Good catch, added --filtered-targets to the usage line.

test/community_build/downstream_test_driver.sh
bor-p-s: No double run — `targets` in BUILD already has a negative pattern (`-//joern-cli/frontends/jssrc2cpg/...`) excluding the filtered target, so it only runs through the second (filtered) invocation. Verified with `bazel query`: a `-//pkg/...` pattern removes targets declared directly in `pkg`, not just subpackages. Added a comment here pointing at that contract.

test/community_build/downstream_test.bzl
bor-p-s: Extracted `args` into a separate list built up with `if`s instead of a chained conditional-concat expression — same generated args, verified via `bazel query --output=build`.

PawelLipski (approved):

PawelLipski (approved):

PR #1960, opened 2026-09-03
refactor: Replace nested-bazel build-success tests with transitions

bor-p-s:
### Description

Converts Category B nested-bazel positive-build fixtures (`expect_build_success_test`) to a native `--extra_toolchains` transition + `bazel_skylib`'s `build_test`, so a plain in-graph build answers "does this build" instead of a nested `bazel` process. The shared transition and macro live in `test/expect_build_failure/toolchain_transition.bzl`; `collect_actions.bzl`'s own copy of the same transition now reuses it.

13 of 16 audited candidates convert this way. The other 3 (`sdeps`, `dummy_generator_test`, the jacocorunner test) assert a test/coverage run succeeds, not just a build, which this design doesn't cover -- they stay nested.

`test_rules_scala.sh` also skips the converted targets in its extra `--extra_toolchains` sweeps, since their own transition already fixes the toolchain and their result can't change between sweeps.

### Motivation

Each converted test no longer pays the cost of a nested `bazel` invocation (its own startup, re-analysis, no shared cache with the outer build). Measured locally, touching a fixture's source and rerunning: the 13 targets go from about 99s combined to about 4s.

Release impact: tests only.


PawelLipski (approved):
Nice!

PR #1962, opened 2026-09-03
ci: Skip the redundant default-toolchain build+test in test_rules_scala.sh on Linux and macOS

bor-p-s:
### Description

`test_rules_scala.sh`'s first lines (`bazel build/test src/... test/...` and `bazel test third_party/...`) go back to the script's original 2016 version. Since #28fcc676/#3a746bc4 (2026-07-02) broadened the `ubuntu2004`/`macos` tasks' `build_targets`/`test_targets` from `//test/...` to `//... -//test_expect_failure/...`, these two tasks already cover the same scope.

Adds a `SKIP_DEFAULT_TOOLCHAIN_RUN` env var to `test_rules_scala.sh` that skips those lines, and sets it only on the `test_rules_scala_linux` and `test_rules_scala_macos` shell_commands tasks. `test_rules_scala_win`, `test_rules_scala_jdk21`, and `test_rules_scala_linux_last_green` keep running them unchanged -- none of the three has an equivalent `build_targets`/`test_targets` task, so for them this is the only place the default toolchain gets exercised.

Confirmed the removed scope is fully covered: `bazel query 'tests(src/... + test/...) except tests(//... - //test_expect_failure/...)'` returns empty (same for `third_party/...`).

### Motivation

`ubuntu2004`/`macos` are native `build_targets`/`test_targets` tasks that get their own remote cache wired in by `bazelci.py`. `test_rules_scala_linux`/`test_rules_scala_macos` run as `shell_commands` -- since #1932/#1939 landed, they share that same cache too, so this isn't about eliminating a full from-scratch rebuild anymore. It's removing a second, redundant run of a scope the sibling task already built and tested in the same CI run.


test_rules_scala.sh
PawelLipski, line 77: If it comes to the remaining commands here - I suppose they must be listed explicitly due to the non-standard toolchain config right?

PawelLipski (approved):

test_rules_scala.sh
bor-p-s: Right -- those run under `--extra_toolchains=...` (strict-deps toolchain variants, `minimal_direct_source_deps`, `plus_one_deps`, stats file toolchain, etc.), and none of them has an equivalent native CI task anywhere else in `.bazelci/presubmit.yml`. So this script stays the only place they get exercised, and they're left unconditional.

PR #1965, opened 2026-09-04
fix: Skip executing dicer's own business-logic tests in dicer_test

bor-p-s:
### Description

Adds a `smoke_test_targets` mechanism to `downstream_test`: the nested build still compiles everything in `targets`, but only the listed targets actually execute as tests. Applied to `dicer_test`: its other ~150 test targets assert dicer's own business logic, unrelated to what `dicer_test` checks (scala_proto/ScalaPB codegen, Scala 2.12 support). Two targets that round-trip real generated messages (`toProto`/`fromProto`) stay as the actual runtime check for that.

Also drops `dicer_test`'s 9 target exclusions (etcd-dependent targets, a flaky statistical test, a timeout-sensitive test): they mattered only for test execution, and the main invocation now only builds them.

### Motivation

`dicer_test`'s test-execution phase was spent almost entirely on tests unrelated to rules_scala's own correctness. Measured once locally: build + these two targets ~46.5s, vs ~64s to also run all ~150 -- roughly 18s removed from every run.

### DIFF
--- a/test/community_build/BUILD
+++ b/test/community_build/BUILD
@@ -81,12 +81,22 @@ downstream_test(
 # size = "enormous" for the same reason as joern_test above: dicer's nested
 # test suite is large enough, with some tests minutes-long, to exceed this
 # sh_test's own "large" (900s) default in total.
+#
+# smoke_test_targets: dicer's other ~150 test targets assert dicer's own
+# business logic; these two are the actual runtime check for
+# scala_proto/ScalaPB codegen, round-tripping real generated messages
+# (`toProto`/`fromProto`). gRPC service/stub code keeps compile-only
+# coverage, same as everything else in `targets`.
 downstream_test(
     name = "dicer_test",
     size = "enormous",
     patches = ["//test/community_build:dicer_protobuf_version.patch"],
     repo_name = "dicer",
     scala_version = "2.12.21",
+    smoke_test_targets = [
+        "//dicer/common:assignment_test",
+        "//dicer/common:diff_assignment_test",
+    ],
     target_compatible_with = ["@platforms//os:linux"],
     targets = [
         "//...",
--- a/test/community_build/BUILD
+++ b/test/community_build/BUILD
@@ -68,12 +68,6 @@ downstream_test(
 # dicer_test's job: exercise rules_scala's scala_proto/ScalaPB (gRPC)
 # codegen and Scala 2.12 support.
 #
-# The 8 targets excluded below start a real etcd binary. That binary is
-# what repeatedly times out in CI.
-#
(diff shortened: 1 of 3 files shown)


test/community_build/BUILD
PawelLipski: Hmmm should these tests still be excluded?
Were they excluded since their **build** was too heavy, or rather their **test runs**?
If the latter, maybe they don't need to be excluded from `targets` anymore?

bor-p-s: You're right — verified by actually building all 9 excluded targets (etcd ones + algorithm_chaos_test) plus assignment_generator_test directly: all 10 compile cleanly. They were excluded only because of test-execution behavior (etcd binary startup, a flaky statistical assertion, a tight test timeout under load), which doesn't apply once `targets` only builds. Dropped the exclusions in the latest commit -- `targets` is now just `["//..."]`.

test/community_build/BUILD
```diff
@@ -68,40 +68,34 @@ downstream_test(
...
+# (assignment_generator_test) were all test-execution problems, moot once
+# a target only builds.
 downstream_test(
     name = "dicer_test",
     size = "enormous",
     patches = ["//test/community_build:dicer_protobuf_version.patch"],
     repo_name = "dicer",
     scala_version = "2.12.21",
-    target_compatible_with = ["@platforms//os:linux"],
-    targets = [
```
PawelLipski, line 91: Maybe a similar simplification (build all, test only selected) can be done for `joern_test`?
And then probably some attrs can be removed from macro (`targets`? `filtered_targets`? `test_filter`?)

I don't like that there's now `filtered_targets` and `smoke_test_targets`, it's unclear which does what and whether/how they differ

bor-p-s: Measured it: joern's other 8 targets (everything but jssrc2cpg, already smoke-tested) cost 242s of test execution locally (x2cpg 105s, rubysrc2cpg 53s, kotlin2cpg 38s, javasrc2cpg 23s, rest under 11s each), the same category jssrc2cpg was in before #1956.

I'd rather keep that out of this PR: it needs a smoke patch per big frontend, each requiring real understanding of that frontend's test suite, roughly 4x this PR's effort plus 4x the future maintenance surface. Happy to track it as a follow-up.

On `filtered_targets` vs `smoke_test_targets`: different granularity. `filtered_targets` selects one ScalaTest class inside a target (needed for jssrc2cpg's aggregated smoke suite). `smoke_test_targets` selects whole targets to execute; the rest only build. Can tighten the docstrings if they still read as interchangeable.

PawelLipski (approved):

PR #1966, opened 2026-09-05
[fix] Release job adjustments

WojciechMazur:
- Mark downstream_repository as dev_dependency.
- Skip `test/scalac_exec_group:scalac_action_exec_properties_test` on the release job: it runs a fresh, dedicated nested-bazel output base near the very end of the wildcard test run, when the GitHub-hosted runner's disk is already tight from everything before it -- likely why it failed on both v7.3.0 attempts. Still covered on regular CI, just not on this release job.

(community_build itself is back in the release test command -- #1965 cut `dicer_test` down to 2 executed targets with the rest build-only, which should remove the timing-sensitive surface that was failing it before.)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,7 +29,7 @@ jobs:
   release:
     uses: bazel-contrib/.github/.github/workflows/release_ruleset.yaml@v7.4.0
     with:
-      bazel_test_command: "bazel test //src/... //test/... //third_party/..."
+      bazel_test_command: "bazel test -- //... -//test/community_build/..."
       prerelease: false
       draft: true
       release_files: rules_scala-*.tar.gz
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,7 +29,7 @@ jobs:
   release:
     uses: bazel-contrib/.github/.github/workflows/release_ruleset.yaml@v7.4.0
     with:
-      bazel_test_command: "bazel test -- //..."
+      bazel_test_command: "bazel test -- //... -//test/scalac_exec_group:scalac_action_exec_properties_test"
       prerelease: false
       draft: true
       release_files: rules_scala-*.tar.gz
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -29,7 +29,7 @@ jobs:
   release:
     uses: bazel-contrib/.github/.github/workflows/release_ruleset.yaml@v7.4.0
     with:
-      bazel_test_command: "bazel test //src/... //test/... //third_party/..."
+      bazel_test_command: "bazel test -- //... -//test/community_build/..."
```
PawelLipski: Out of curiosity - what's the failure cause?

@bor-p-s maybe add some tag to these community builds? so that they're not selected by default (maybe just `manual`)

WojciechMazur: I'm not quite sure, so useful information from logs https://github.com/bazel-contrib/rules_scala/actions/runs/33959859120/job/101294857746#step:4:408 

Still, maybe we should take the community_build from the wildcard tests, and have them executed in their own job? 15minutes is execution is really long

bor-p-s: Took a look — both release runs (Sept 3, Sept 5) failed only on `dicer_test`, `joern_test` passed both times. Couldn't get the actual failure text though, this job doesn't pass `--test_output=errors` and doesn't upload test logs.

One thought: `dicer_test` still runs ~150 of dicer's own tests here, and this job's nested build has no remote cache (gated on `BUILDKITE=true`) on a plain GitHub-hosted runner, so it's colder and tighter on resources than our usual CI. A couple of dicer's tests are already known to be timing-sensitive under load.

#1965 cuts that down to 2 executed targets, rest build-only. Might be worth revisiting this exclusion once that lands, in case it's enough on its own.

bor-p-s: It's merged now.

bor-p-s (approved):

PR #1967, opened 2026-09-06
fix: map source paths explicitly in -paths-for-coverage.txt

### LINKED ISSUE #1101: Code coverage only working with deep directory structure
If I have a file `c.scala` in package `a.b` then code coverage only seems to work if its path relative to the `BUILD` file ends with `/a/b/`. So it works if the `BUILD` file is in the root and I have `/a/b/c.scala` or `/src/a/b/c.scala` but not `/a/b/src/c.scala`. It also doesn't work if I have `/a/b/c.scala` but the BUILD file is in `/a` or in `/a/b`. This is only an issue with code coverage, it works fine for building, running and testing.

By "not working" I mean the file doesn't show up in the coverage results, there are no errors.

I found this Jacoco issue that seems related: https://github.com/jacoco/jacoco/issues/939

avbhandaru:
### Description

Emit explicit `<real source path>///<class-derived path>` mappings in `-paths-for-coverage.txt` so coverage is attributed to sources whose path on disk does not end with their package path.

`JacocoInstrumenter` writes `-paths-for-coverage.txt` as a bare newline-joined list of source paths. Bazel's `JacocoLCOVFormatter` keys coverage data by the path it derives from the *class* — `<package>/<SourceFileName>` — and resolves it against that list with a **suffix match**. A source file therefore only gets coverage when its path happens to end with its package path: `src/main/scala/com/example/util/Time.scala` works, while `util/time/Time.scala` declaring `package com.example.util` is silently left out of the report.

[bazelbuild/bazel#12627](https://github.com/bazelbuild/bazel/pull/12627) added an explicit mapping form to that file — `<real source path>///<class-derived path>` — which bypasses the suffix match entirely (`parts[1].equals("/" + sourceFile)`). This PR adopts it.

The `package` clause isn't readable from the instrumenter (sources are passed as paths, not declared as inputs to the action), but the jar being walked names every package directory it contains, so the instrumenter emits one mapping per (source, package directory) pair. Pairs that don't describe a real class are simply never looked up; pairs whose package holds a class named after the source file are emitted first so two sources sharing a basename in one jar still resolve. The bare source paths are kept at the end, so layouts where the suffix match already works behave exactly as before.

Fixes #1101

### Motivation

`bazel coverage` currently produces no data at all for any repository that does not mirror its package hierarchy in its directory layout — the case in #1101, open since 2020. Nothing fails and nothing warns; the affected records are just absent from `coverage.dat`, which makes the whole feature quietly useless outside a Maven-style tree. This is exactly what the mapping format in bazelbuild/bazel#12627 was added for, and rules_scala had never adopted it.

### Testing

- New `test/coverage_source_path_mapping`, wired into `./test_coverage.sh` alongside the existing coverage fixtures. Its sources sit directly in the package directory and declare deep packages (`com.example.coverage.mapping`, `com.example.coverage.other`), so their paths do not end with their package paths — the #1101 case, which no existing fixture covers. Two assertions: a golden `expected-coverage.dat` diff, and an explicit check that each instrumented source has `DA:` lines with non-zero hit counts and a non-zero `LH:`.
- **Both new assertions fail without the `JacocoInstrumenter.java` change.** Verified by reverting just that file: the produced `coverage.dat` is *empty* — the two sources are dropped from the report entirely — so the golden diff fails and the hit-line check reports `no coverage record for test/coverage_source_path_mapping/AlsoMapped.scala`.
- `./test_coverage.sh` passes in full (9 test cases). The four pre-existing goldens — `coverage_scalatest`, `coverage_specs2_with_junit`, `coverage_filename_encoding`, `coverage_scalatest_resources` — are byte-identical with the change applied; I also re-ran `//test/coverage_scalatest:test-scalatest` with `--nocache_test_results` and diffed it against the checked-in golden to confirm.
- `bazel build //...` and `bazel run //tools:lint_check` are clean.

Tested on macOS (arm64, Bazel 8) only; I have not run the suite on Linux.

### DIFF
--- a/src/java/io/bazel/rulesscala/coverage/instrumenter/JacocoInstrumenter.java
+++ b/src/java/io/bazel/rulesscala/coverage/instrumenter/JacocoInstrumenter.java
@@ -16,12 +16,24 @@ import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.StandardOpenOption;
 import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 import org.jacoco.core.instr.Instrumenter;
 import org.jacoco.core.runtime.OfflineInstrumentationAccessGenerator;
 
 public final class JacocoInstrumenter implements Worker.Interface {
+
+  /**
+   * Separates the real source path from the class-derived path in a {@code
+   * -paths-for-coverage.txt} entry. Must stay in sync with {@code
+   * JacocoLCOVFormatter.EXEC_PATH_DELIMITER}.
+   */
+  private static final String EXEC_PATH_DELIMITER = "///";
 
   public static void main(String[] args) throws Exception {
     Worker.workerMain(args, new JacocoInstrumenter());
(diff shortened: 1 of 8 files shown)


bor-p-s:
Superseded by and merged as #1987 (same fix, plus a fix for a basename-collision bug found during review, and unit tests).

PR #1968, opened 2026-09-06
feat: Add coverage_skip_oversized_methods to scala_toolchain

avbhandaru:
### Description

Adds an opt-in `coverage_skip_oversized_methods` attribute to `scala_toolchain`. When set, a class that JaCoCo cannot instrument because a method would exceed the JVM's 64KB limit is passed through to the output jar uninstrumented instead of failing the build.

JaCoCo's probes add bytecode, so instrumenting a method that is already close to the 64KB per-method limit pushes it over. ASM throws `MethodTooLargeException`, `JacocoInstrumenter` wraps it in a `RuntimeException`, and the whole action fails:

```
java.io.IOException: Error while instrumenting /com/example/generated/AttributeExtractor$.class with JaCoCo 0.8.11...
Caused by: com.google.testing.coverage.jarjar.org.objectweb.asm.MethodTooLargeException:
  Method too large: com/example/generated/AttributeExtractor$.apply (...)Lscala/collection/immutable/Seq;
```

With the attribute set, such a class is instead:

- written to the output jar uninstrumented, so the runtime classpath stays complete and tests still run;
- omitted from the coverage report rather than reported as 0%, by skipping the `.class.uninstrumented` twin entry that `JacocoCoverageRunner` uses to build its analysis;
- reported with a warning on stderr naming the class and the reason.

`--skip_oversized_methods` is threaded to the instrumenter worker through the param file that `phase_coverage.bzl` already builds. `MethodTooLargeException` is repackaged by jarjar into the JaCoCo runner jar, so it isn't on the instrumenter's compile classpath and cannot be caught by type — it is detected by walking the cause chain and matching the simple class name.

Refs #1969

### Motivation

In a large Scala repository, one generated `apply` method exceeded the limit under instrumentation. Its library's `JacocoInstrumenter` action failed, and **652 test targets that transitively depended on that library could not build** — targets that compile and test fine without coverage. A single oversized method therefore makes `bazel coverage` unusable across a large part of the build, with an error that gives no hint that coverage instrumentation is the cause. Splitting the offending generated method is not always possible, and it is not worth blocking coverage for the other 651 targets.

The default is unchanged: `coverage_skip_oversized_methods` defaults to `False` and the action still fails exactly as before, so this is strictly additive.

Tested with a new `test/shell/test_coverage_oversized_method.sh` (wired into `./test_coverage.sh`) over a fixture whose single method compiles to ~44KB and genuinely trips `MethodTooLargeException` once instrumented. It asserts that the target passes `bazel test` without coverage, that `bazel coverage` still fails with `MethodTooLargeException` by default, and that with the attribute set the build succeeds, the warning is emitted, a sibling class is still measured, and the skipped class is absent from `coverage.dat`. Also ran the full `./test_coverage.sh` (all 10 cases pass), `bazel build //...`, and `./lint.sh` (clean).


bor-p-s:
@avbhandaru thanks for this, ported it forward in #1979. Closing as superseded.

PR #1970, opened 2026-09-07
ci: Tag the remaining disk-heavy nested-bazel tests no-release

bor-p-s:
### Description

Extends the `no-release` tag (introduced for `scalac_action_exec_properties_test` and `community_build`) to the rest of the nested-bazel test family: every `sh_test` whose script sets up its own dedicated nested Bazel output base (~1GB each, extracted fresh into `/tmp`). 68 targets across 16 BUILD files.

Left out: `expect_build_failure`/`expect_analysis_failure` fixtures (already share 3 lanes, not disk-heavy individually) and `no_recompilation_test` (already `exclusive`, separate concern).

### Motivation

The last two release runs both hit "No space left on device" on GitHub's 14GB-disk runner. Excluding one target at a time just moved the failure to whichever test happened to run last. These are already covered by the regular BuildKite CI (much bigger machines) on every PR and on master before a release is cut, so the release job doesn't need to re-run them.

### DIFF
--- a/test/argument_file_in_runner/BUILD
+++ b/test/argument_file_in_runner/BUILD
@@ -19,6 +19,7 @@ sh_test(
         # after that toolchain changes.
         "external",
         "local",
+        "no-release",
         "requires-network",
     ],
     # rules_scala doesn't support argfiles on Windows yet.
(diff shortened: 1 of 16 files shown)


PawelLipski (commented):
~AFAICS this PR covers `sh_test`s directly declared in BUILD files... should the macros like in `test/expect_build_failure/expect_build_failure.bzl` also be covered?~  nvm, I see the description now (curiosuly, I can't delete this comment?)

PawelLipski (approved):

PR #1971, opened 2026-09-07
fix: Smoke-filter joern_test's remaining frontend targets

bor-p-s:
### Description

Extends #1956's smoke-suite mechanism to `joern_test`'s other 4 expensive frontend targets: x2cpg, rubysrc2cpg, kotlin2cpg, javasrc2cpg.

All 5 filtered targets now share one `test_filter` value in a single nested `bazel test` invocation, so each one's smoke suite needs the identical class name: `io.joern.smoke.SmokeSuite` (jssrc2cpg's existing patch moves to that package too.

### Motivation

Measured in a Linux container, one run: these 4 targets went from ~225s to ~19s combined test-execution time.
EOF
)


test/community_build/joern_jssrc2cpg_smoke.patch
```diff
@@ -1,33 +1,17 @@
...
-# lines of test sources still compile in full (real Scala 3 compile
-# workload against the toolchain under test), but only this smoke subset
-# executes: measured 748-820s -> 165-215s.
+# Wraps 6 existing classes (AST, CFG, dataflow, type-recovery passes, plus
+# a file parse and an HTTP round trip) as io.joern.smoke.SmokeSuite -- see
+# BUILD's test_filter comment for why the package is io.joern.smoke, not
+# io.joern.jssrc2cpg. ~800s -> ~165s.
 --- /dev/null
-+++ b/joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/SmokeSuite.scala
++++ b/joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/smoke/SmokeSuite.scala
```
PawelLipski: Oh! so the same base file is patched in multiple ways right? depending on what package it's viewed from?

test/community_build/downstream_test.bzl
```diff
@@ -56,15 +56,20 @@ def downstream_test(
         filtered_targets: target patterns run only through a chosen
             ScalaTest suite (`test_filter`), instead of their whole test
             source tree -- e.g. joern_test uses this to run just a smoke
-            suite for jssrc2cpg, its most expensive target, while every
-            other target in `targets` keeps running unfiltered. Needs a
-            *second*, separate nested `bazel test` invocation, with
+            suite for each of its 5 most expensive targets, while every
+            other target in `targets` keeps running unfiltered. `test_filter`
+            is one exact class name (Bazel's `--test_filter` becomes
```
PawelLipski: Would it make sense to instead introduce a dict with `<target>: <class name>` entries? to avoid this hackish patches?

PawelLipski: So that you select a (potentially different) class to run from each package separately?

test/community_build/joern_jssrc2cpg_smoke.patch
bor-p-s: Each patch creates a new file (`--- /dev/null`), it doesn't modify an existing one. These are 5 separate new files, one per frontend module (`joern-cli/frontends/<name>/...`). They share the same relative suffix (`src/test/scala/io/joern/smoke/SmokeSuite.scala`) because every filtered target needs a class with the identical fully-qualified name (`io.joern.smoke.SmokeSuite`) for `test_filter` to resolve it -- see the comment above `joern_test` in BUILD.

test/community_build/downstream_test.bzl
bor-p-s: Implemented this in 1e737093 -- `filtered_targets` is now a dict, and the driver does one nested `bazel test` per entry. Measured in a Linux container: running them as separate invocations added about 10% overhead on a warm cache (166.8s vs 183s combined), but was actually faster on a cold one (183s vs 229.9s) -- so no real regression either way. 4 of 5 targets now point straight at their own existing class, no patch needed. Only jssrc2cpg still needs its own patch, since no single existing class combines what its smoke suite needs. This also makes the earlier question about the shared file path moot -- there's no more shared package now.

PawelLipski (approved):

test/community_build/downstream_test.bzl
bor-p-s: Good idea, BTW.

PR #1972, opened 2026-09-08
ci: Skip the toolchain sweep on macOS/windows in test_rules_scala.sh

bor-p-s:
### Description

The toolchain sweep in `test_rules_scala.sh` (repeats `test/...` under `high_level_transitive_deps_strict_deps_error`/`_warn`, `minimal_direct_source_deps`, and the `plus_one_deps` toolchain) ran unconditionally on ubuntu, macOS, and windows CI. Wraps it in `if ! is_macos && ! is_windows; then ... fi` so it only runs on ubuntu.

The toolchains involved configure `scala_toolchain`'s `dependency_mode`/`strict_deps_mode`/`dependency_tracking_method`, pure Bazel-graph/JVM dependency-checking logic. This runs identically on every OS, with one exception: `DependencyAnalyzer.scala` normalizes jar paths differently on Windows (`isWindows` branch, backslash to forward slash). That branch only runs when strict-deps or unused-deps checking is on, and those same toolchains are already exercised independently, per target, by nested tests in `test/missing_direct_deps`, `test/strict_dependency/*`, and `test/plus_one_deps/*`, each supplying its own `--extra_toolchains` via `build_args` and running unconditionally on every platform. So this Windows-specific branch keeps its coverage; what macOS/windows lose here is the tree-wide rebuild itself, which ubuntu already exercises identically.

Ran the full script locally on macOS: default-toolchain part still builds/tests green, and the sweep commands do not appear in the run log, confirming the guard actually skips them.

Also updates `test_rules_scala_macos`'s comment in `.bazelci/presubmit.yml`: that task already skips the default-toolchain run (`SKIP_DEFAULT_TOOLCHAIN_RUN=1`, since #1962), so after this change it only builds the script's three trailing manual-tagged targets (`//test/binary_in_genrule`, the two `//test_statsfile` targets).

### Motivation

Measured cost on a recent build (#6911, same script shape): macOS ~1133s (19min), windows ~553s (9min), ubuntu ~454s, kept as the baseline. A repeated tree-wide rebuild with no coverage gain on those two platforms.


test_rules_scala.sh
```diff
@@ -65,13 +65,26 @@ if [[ -z "${SKIP_DEFAULT_TOOLCHAIN_RUN:-}" ]]; then
...
-$runner bazel test $remote_cache_config "${test_output_flag}" "${toolchain_sweep_test_flag}" --extra_toolchains=//test/toolchains:high_level_transitive_deps_strict_deps_error -- test/...
-$runner bazel test $remote_cache_config "${test_output_flag}" "${toolchain_sweep_test_flag}" --extra_toolchains=//scala:minimal_direct_source_deps -- test/...
-$runner bazel build $remote_cache_config test/missing_direct_deps/internal_deps/... --strict_java_deps=warn --extra_toolchains=//test/toolchains:high_level_transitive_deps_strict_deps_warn
-$runner bazel test $remote_cache_config "${test_output_flag}" "${toolchain_sweep_test_flag}" //test/... --extra_toolchains="//test_expect_failure/plus_one_deps:plus_one_deps"
+# The toolchain sweep below rebuilds the whole test/... tree under 2-3
+# alternative --extra_toolchains values, to catch a tree-wide regression under
+# those configs. Skipped on macOS/windows: ubuntu already covers it, saving
+# ~19min/~9min per run respectively.
+#
+# Windows coverage of DependencyAnalyzer.scala's isWindows path-normalization
```
PawelLipski, line 73: Actually... _how_ does it stay intact? as in, where's that logic for keeping this part intact?

bor-p-s: The chain, step by step:

1. `DependencyAnalyzer.scala`'s `isWindows` branch only runs when the `dependency_analyzer` scalac plugin is actually loaded.
2. The plugin is loaded only when `dependency_info.use_analyzer` is true (`scala/private/rule_impls.bzl:65`).
3. `use_analyzer` is `is_strict_deps_on or is_unused_deps_on` (`scala/private/dependency.bzl:22`) -- true only when strict-deps or unused-deps checking is turned on for that build.
4. The default toolchain (`dependency_mode = "direct"`) computes `strict_deps_mode = "off"` and ships `unused_dependency_checker_mode = "off"`, so a plain default-toolchain build never loads the plugin at all.

So the branch only fires under the toolchains that do turn strict/unused checking on: `high_level_transitive_deps_strict_deps_error`/`_warn`, `minimal_direct_source_deps`, `plus_one_deps`. Those are the ones this sweep repeats.

The point of my comment: those same toolchains already get exercised elsewhere, per target, regardless of this sweep -- via `build_args`/`extra_toolchains` on plain (non-`manual`) `sh_test` targets in `test/missing_direct_deps/BUILD`, `test/strict_dependency/BUILD` (+ `scala_import`, `scala_proto_deps`), `test/scala_binary_missing_direct_deps/BUILD`, and `test/plus_one_deps/*/BUILD`. Each of those is an ordinary test that `bazel test test/...` runs on every platform, sweep or no sweep -- so the `isWindows` branch keeps getting covered there.

PawelLipski: Wow that's complicated 🤯

PawelLipski (approved):

test_rules_scala.sh
bor-p-s: Fair, that reads dense written out like that. I'd say it's about the same amount of complexity as before, just spread across a pile of separate scripts. This moves it into Bazel's own graph, where it's explicit and checked instead of hidden in a shell script.

PR #1974, opened 2026-09-09
test: Migrate test_thirdparty_version.sh to native Bazel targets

bor-p-s:
### Description

Migrates `test_thirdparty_version.sh` to native `expect_test_success_test` targets in `test_thirdparty_version/BUILD`, one per Scala version, each driving the same nested `bazel test //third_party/... --repo_env=SCALA_VERSION=X` the script ran.

The package sits at the repo root, not under `test/`: `test_rules_scala.sh` runs a `test/...` wildcard for several jobs, including one that sets `--java_language_version=21`, and Scala 3.1.3 fails to build under JDK 21. Tagged the targets `no-release`, matching the rest of the nested-bazel test family (#1970): each one rebuilds the whole `//third_party/...` tree per version, which the release job never had to do before since this was a shell script, not a Bazel target.

### Motivation

Continues the shell-to-native test migration (#1867). Simplest of the remaining scripts: no dynamic behavior, no source mutation, each target is a single flat nested build (unlike e.g. `disappearing_class`'s two sequential nested builds with a source swap between them).

### Test plan

- Broke one target on purpose (an unknown Scala version) to confirm the check goes red.
- Appended the jdk21 job's `.bazelrc` flags locally and confirmed Scala 3.1.3 fails to build there.
- Full `bazel test //...` shows no new failures versus master.


.bazelci/presubmit.yml
```diff
@@ -125,10 +125,12 @@ tasks:
     shell_commands:
       - "./test_version.sh"
   thirdparty_version_linux:
-    name: "./test_thirdparty_version.sh"
+    name: "test_thirdparty_version"
     platform: ubuntu2204_java17
-    shell_commands:
-      - "./test_thirdparty_version.sh"
+    build_targets:
+      - "//test_thirdparty_version/..."
```
PawelLipski, line 133: Are these targets also included when `//...` is tested (like in `ubuntu2004`->`test_targets`)? is it expected?

bor-p-s: Yes, expected. These are plain (non-`manual`) targets, so they also get picked up by the `ubuntu2004`/`macos` `test_targets: //...` steps, same as the rest of the nested-bazel test family (`test/semanticdb`, `test/scalacopts_invalid`, etc.) which also run both under their own dedicated task and under the general wildcard. Deliberately excluded from test_rules_scala.sh's toolchain sweeps and the jdk21 job (those only match `test/...`/`third_party/...`, not this root-level package) and from the release workflow (tagged `no-release`).

PawelLipski (approved):

PawelLipski (approved):

PR #1975, opened 2026-09-09
refactor: Migrate test_bzlmod_tidy.sh to Bazel-native sh_test targets

bor-p-s:
### Description

Migrates `test/shell/test_bzlmod_tidy.sh` into native `sh_test` targets under a new `test/bzlmod_tidy/` package: one per MODULE.bazel directory, kept in sync with `git ls-files '**MODULE.bazel'` by a small coverage test (a `glob()` only sees its own package, so this needed a hardcoded list instead).

Restores the checkout on exit even on failure, a deliberate change from the original script's behavior, to match how the other checkout-mutating tests in this migration behave. Tag reasoning is in `bzlmod_tidy.bzl`'s docstring.

One-line fix in `test/expect_build_failure/nested_bazel.sh`: `bazel mod` needs the same `--symlink_prefix` exclusion `query` already has.

Fully retires `test/shell/test_bzlmod_tidy.sh`.

### Motivation

Moves shell tests off the legacy `test_rules_scala.sh`/`test_lint.sh` runners toward native `bazel test //...`.

### Verification

Broke a fixture's MODULE.bazel on purpose to confirm the check goes red, then restored it and confirmed green. Ran the full `test/bzlmod_tidy/...` suite three times in a row, checking `git status` after each run.


test/bzlmod_tidy/BUILD
```diff
@@ -0,0 +1,53 @@
+load("@rules_shell//shell:sh_test.bzl", "sh_test")
+load(":bzlmod_tidy.bzl", "bzlmod_tidy_test")
+
+# Every repo-tracked MODULE.bazel file's directory ("." for the root), kept in
+# sync with `git ls-files '**MODULE.bazel'` by :all_module_bazel_files_covered_test.
+MODULE_DIRS = [
+    ".",
```
PawelLipski, line 7: Hmmm why is the explicit list needed now, even though it (apparently) wasn't previously?

PawelLipski (approved):

test/bzlmod_tidy/BUILD
bor-p-s: Good question. The shell script's `git ls-files '**MODULE.bazel'` is plain shell — it walks the whole repo using git's own file list. A Bazel `glob()` only looks inside the package it's written in, and most of these directories are their own package (each has a BUILD file), so a glob only sees the MODULE.bazel right next to it, not the others. That's why we need a plain list here instead — the coverage test right below checks it against git, so it fails if the list ever gets out of sync.

PR #1976, opened 2026-09-09
refactor: Migrate test_cross_build.sh to Bazel-native sh_test targets

bor-p-s:
### Description

Migrates `test/shell/test_cross_build.sh` (2 checks: `bazel test //...` and a cross-major-version build with `--repo_env=SCALA_VERSION=2.13.18` on `since_3_1`/`scala_2_13`/`scala_2_12`) to native `sh_test` targets in `test/cross_build_test/BUILD`.

`test_cross_build` is a separate Bazel module (its own `MODULE.bazel`/`WORKSPACE`, `local_path_override`'d onto this checkout), excluded from this repo's own package tree by `.bazelignore` (bazelbuild/bazel#22208). The `expect_build_failure.bzl` macros only build labels inside this same workspace, so this drives a nested `bazel` directly instead, the same pattern `test/compiler_sources_integrity_test` (#1935) uses, sharing its `nested_bazel.sh` helper.

The cross-major-version check guards a select-based fix for cross-version default dep providers (#1825): it makes a `scala_library` whose `scala_version` picks the non-default major version get that version's own compiler JARs, where it previously got the workspace default's. That fix lives in `scala/toolchains_repo.bzl`, an existing dependency outside `test_cross_build`, so it's listed in the new `sh_test`'s `data` (with a new `exports_files` entry in `scala/BUILD`) to re-key the test's cached result when it changes, the same way `compiler_sources_integrity_test` lists its own mechanism's files.

The `cross_build_linux` task in `.bazelci/presubmit.yml` is removed: the new targets live under `test/...`, already covered by the `ubuntu2004`/`macos` "bazel test //..." tasks.

Also removes the now-dead root `test_cross_build.sh` wrapper, which only sourced the now-deleted `test/shell/test_cross_build.sh` and was called only from `test_all.sh` and the removed presubmit task, plus the matching now-stale references in `test_all.sh` and `test/community_build/source_fingerprint.bzl`'s `_INTERNAL_ENTRIES` list.

### Motivation

Continues the `test/shell/*.sh` to Bazel-native test migration.


test/cross_build_test/BUILD
```diff
@@ -0,0 +1,59 @@
...
+
+_NESTED_BAZEL_TAGS = [
+    "local",
+    "no-release",
+    "requires-network",
+    "exclusive",
+]
+
+sh_test(
+    name = "test_test",
```
PawelLipski: This name is somewhat opaque 😅 `full_suite_test`?

test/cross_build_test/BUILD
```diff
@@ -0,0 +1,59 @@
...
+    # verifies; listing it re-keys that test when the fix changes.
+    "//scala:toolchains_repo.bzl",
+]
+
+_NESTED_BAZEL_TAGS = [
+    "local",
+    "no-release",
+    "requires-network",
+    "exclusive",
+]
```
PawelLipski, line 28: Should these tests (or some of them) receive `skip-toolchain-sweep`?

PawelLipski (approved):

test/cross_build_test/BUILD
bor-p-s: Renamed to `full_suite_test`, thanks.

test/cross_build_test/BUILD
bor-p-s: Added it to both — neither test's nested `bazel` call uses the outer `--extra_toolchains` at all, so a toolchain sweep here always builds the same thing. Checked the code path (`nested_bazel_run` builds its own command line, doesn't forward it) and confirmed with `--test_tag_filters=-skip-toolchain-sweep`.

PR #1977, opened 2026-09-10
refactor: Migrate test_coverage.sh to Bazel-native sh_test targets

bor-p-s:
### Description

Migrates test_coverage.sh and its 4 sub-scripts (test/shell/test_coverage_*.sh) into native sh_test targets in their fixture packages (test/coverage_specs2_with_junit, test/coverage_scalatest, test/coverage_filename_encoding, test/coverage_scalatest_resources/consumer).

All 7 cases follow one of two patterns: run bazel coverage and diff the resulting coverage.dat against a checked-in expectation, or run it with --instrument_test_targets=True and grep the .dat for a source file. Added one parametrized macro, coverage_test (new test/coverage_test/ package), instead of 8 near-identical copies, the same way expect_build_failure.bzl became a shared family instead of copy-paste. bazel coverage can't run from inside a plain sh_test action (the sandbox can't reach the Bazel that started it), so like the rest of this migration it runs a nested bazel coverage via nested_bazel.sh.

Tagged exclusive: every coverage_test shares one nested output base, and two running concurrently on the same fixture target (one with --instrument_test_targets=True, one without) can race on the same coverage.dat.

Removes the now-redundant test_coverage_linux/test_coverage_macos CI tasks: the new targets already run under the existing ubuntu2004/macos CI tasks that build/test the whole repo.

### Motivation

Moves shell tests off the legacy test_rules_scala.sh/test_coverage.sh runners toward native bazel test, part of #1867.

### Verification

Broke expected-coverage.dat on purpose to confirm the diff test goes red, restored it and confirmed green. Ran the full test suite for the whole repo (excluding the negative test_expect_failure targets); the 7 new tests pass, and the remaining failures are pre-existing (missing Java runtime under the nested build's scrubbed HOME), unrelated to this change.


test/coverage_filename_encoding/BUILD.bazel
```diff
@@ -1,4 +1,13 @@
 load("//scala:scala.bzl", "scala_library", "scala_specs2_junit_test")
+load("//test/coverage_test:coverage_test.bzl", "coverage_test")
+
+coverage_test(
+    name = "coverage_matches_expected_test",
+    # Default to 2.12.21 because other Scala versions change the output.
```
PawelLipski: Hmm this sounds like _2.12.21 doesn't change the output but the other version do change the output_.

Is this intended? or is it more about _each Scala version has its own output format, so let's stick to a specific one to avoid drift caused by using whatever the default Scala happens to be_?

PawelLipski: (same remark to other places where this comment occurs)

test/coverage_scalatest_resources/consumer/BUILD.bazel
```diff
@@ -1,4 +1,23 @@
...
+load("//test/coverage_test:coverage_test.bzl", "coverage_test")
+
+coverage_test(
+    name = "coverage_matches_expected_test",
+    bazel_args = [
+        # Default to 2.12.21 because other Scala versions change the output.
+        "--repo_env=SCALA_VERSION=2.12.21",
+        "--instrumentation_filter=^//test/coverage_scalatest_resources[:/]",
+    ],
+    expected = "test/coverage_scalatest_resources/expected-coverage.dat",
```
PawelLipski: Hmm so `expected` is in the output of the command itself, and `grep` in the output file? (I haven't read the implementation itself yet)

Maybe make the attribute naming more clear

test/coverage_test/coverage_test.bzl
```diff
@@ -0,0 +1,94 @@
...
+    another run without it expected to read its own).
+
+    Args:
+        name: test target name.
+        target: label whose nested `bazel coverage` must succeed. A
+            package-relative label (`":foo"` or `"foo"`) is resolved against
+            this package.
+        expected: workspace-relative path to a checked-in coverage.dat that the
+            real run's coverage.dat must match exactly.
+        grep: pattern that must appear in the real run's coverage.dat.
```
PawelLipski: Okay, maybe `expected_file_path` and `expected_....something_dunno_what`?

test/coverage_filename_encoding/BUILD.bazel
PawelLipski: (oh, pre-existing comment - still, good to clarify)

test/coverage_test/coverage_test.bzl
```diff
@@ -0,0 +1,94 @@
...
+        bazel_args: extra flags forwarded verbatim to the nested `bazel
+            coverage` (e.g. `"--instrument_test_targets=True"`).
+        size: test size; defaults to `"large"` (the nested Bazel invocation is
+            slow and, on a cold cache, serializes on the shared output base).
+        tags: test tags; defaults to
+            `["no-sandbox", "external", "exclusive", "requires-network"]` --
```
PawelLipski: Should `skip-toolchain-sweep` be included here?

PawelLipski: Or maybe not, we want to check how coverage behaves under different toolchains right?

PawelLipski: But how about `no-release`? I don't remember how much is supposed to be excluded from that runs

test_coverage.sh
```diff
@@ -1,15 +0,0 @@
-#!/usr/bin/env bash
```
PawelLipski, line 1: File still listed in `source_fingerprint.bzl`

test/coverage_test/coverage_test.sh
```diff
@@ -0,0 +1,82 @@
...
+coverage_dat="${testlogs_dir}/${target_relpath}/coverage.dat"
+
+if [[ -n "${expected}" ]]; then
+  if ! diff "${expected}" "${coverage_dat}"; then
+    echo "coverage.dat for ${target} does not match ${expected}" >&2
+    exit 1
+  fi
+else
+  if ! grep -q "${grep_pattern}" "${coverage_dat}"; then
+    echo "coverage.dat for ${target} does not contain expected line: ${grep_pattern}" >&2
```
PawelLipski: Super-nit: _line_ - apparently just substrings are matched, not whole lines (which is implied by the message)

test/coverage_test/coverage_test.bzl
```diff
@@ -0,0 +1,94 @@
...
+        name = name,
+        size = size,
+        srcs = [_COVERAGE_TEST_SH],
+        args = args,
+        data = [
+            "//:MODULE.bazel",
+            _NESTED_BAZEL_LIB,
+        ],
+        tags = tags,
+        # `bazel coverage` doesn't produce a coverage.dat on Windows yet.
```
PawelLipski, line 97: Wow how does it work on Windows? a different file format? or unsupported at all?

test/coverage_test/coverage_test.bzl
```diff
@@ -0,0 +1,94 @@
...
+coverage` (see nested_bazel.sh) and then either diffs the resulting
+coverage.dat against a checked-in expectation or greps it for a line.
+"""
+
+load("@rules_shell//shell:sh_test.bzl", "sh_test")
+
+_COVERAGE_TEST_SH = "//test/coverage_test:coverage_test.sh"
+_NESTED_BAZEL_LIB = "//test/expect_build_failure:nested_bazel.sh"
+
+def _absolutize(target):
```
PawelLipski, line 14: Does it belong in some shared utils? or is this particular way of absolutizing coverage-specific?

test/coverage_scalatest_resources/consumer/BUILD.bazel
```diff
@@ -1,4 +1,23 @@
...
+    target = ":tests",
+)
+
+coverage_test(
+    name = "coverage_includes_resource_test_targets_test",
+    bazel_args = ["--instrument_test_targets=True"],
```
PawelLipski, line 19: Wow a rare flag 😯  this is to make sure that the test code itself is all executed/covered? sounds paranoid to me 😅
(no action required ofc, given that it's pre-existing/doesn't harm)

test/coverage_scalatest_resources/consumer/BUILD.bazel
```diff
@@ -1,4 +1,23 @@
 load("//scala:scala.bzl", "scala_library", "scala_test")
+load("//test/coverage_test:coverage_test.bzl", "coverage_test")
+
+coverage_test(
+    name = "coverage_matches_expected_test",
+    bazel_args = [
+        # Default to 2.12.21 because other Scala versions change the output.
+        "--repo_env=SCALA_VERSION=2.12.21",
+        "--instrumentation_filter=^//test/coverage_scalatest_resources[:/]",
```
PawelLipski, line 11: Nit: are there some other commonly used coverage-related flags that we might want to test here? not that I can think of anything specific, just checking - maybe a different coverage format (xml?) is typically used in some cases? but ofc fine if not

test_coverage.sh
bor-p-s: Good catch, fixed — removed test_coverage.sh from _INTERNAL_ENTRIES in source_fingerprint.bzl.

test/coverage_test/coverage_test.bzl
bor-p-s: bazel coverage on Windows doesn't produce a coverage.dat at all here (confirmed from the Windows CI job log after this PR's targets first ran there: output 'coverage.dat' was not created). Not something new this PR introduces -- the old test_coverage.sh was never invoked on Windows either (no test_coverage_windows task existed). Added target_compatible_with to skip these on Windows, formalizing what was already true before.

test/coverage_filename_encoding/BUILD.bazel
bor-p-s: Clarified: it's the latter -- each Scala version formats coverage.dat differently, so diffing against a fixed expectation needs a fixed Scala version. Updated the comment in all 4 places to say that explicitly.

test/coverage_scalatest_resources/consumer/BUILD.bazel
bor-p-s: Renamed to expected_file/expected_line -- both check the real run's coverage.dat, not the command's own output (expected_file diffs it whole against a checked-in file, expected_line greps it for a substring).

test/coverage_test/coverage_test.bzl
bor-p-s: Went with expected_file / expected_line.

test/coverage_test/coverage_test.sh
bor-p-s: Fixed, changed the message to 'expected text' since it's a substring match.

test/coverage_test/coverage_test.bzl
bor-p-s: Added both. skip-toolchain-sweep: the nested bazel coverage runs under its own output base with its own flags, so the outer sweep's --extra_toolchains never reaches it -- every sweep would just repeat what the default sweep already checked. no-release: matches #1970, which tagged every other disk-heavy nested-bazel test the same way for the release workflow's disk budget.

test/coverage_test/coverage_test.bzl
bor-p-s: Left it duplicated for now -- it's private (leading underscore) in expect_build_failure.bzl, so pulling it out means either exporting it there or creating a new shared-utils package just for this one ~10-line function. There's no existing shared-bzl-utils convention in this repo (the one shared package here, expect_build_failure.bzl, groups a much bigger mechanism used by 4 macros, not a single helper). Would reconsider if a third nested-bazel macro family needs the same absolutizing logic.

test/coverage_scalatest_resources/consumer/BUILD.bazel
bor-p-s: Yeah, exactly -- it forces coverage instrumentation onto the test's own code too, not just the code under test. Niche flag, agreed, but it's what test_coverage.sh already checked before this migration, so kept as-is.

test/coverage_scalatest_resources/consumer/BUILD.bazel
bor-p-s: Nothing else specific to this migration's scope -- bazel coverage only ever produces the lcov-format coverage.dat here, that's the only format rules_scala's instrumenter emits. Didn't look for other flags beyond what the original test_coverage.sh already checked.

PawelLipski (approved):

PR #1978, opened 2026-09-10
refactor: Migrate test_reproducibility.sh to Bazel-native sh_test targets

bor-p-s:
### Description

`test_reproducibility.sh` diffed md5 hashes of `test/...`/`test/coverage_*` jars across two from-scratch builds -- a two-build comparison. `expect_build_failure.bzl`'s macros model a single build/test pass/fail outcome, so this stays a hand-written `sh_test` on `nested_bazel.sh`, the same pattern as `expect_no_ijar.sh`.

This test gets its own dedicated nested output base rather than joining the shared 3-lane pool: it's an order of magnitude heavier (~270-275s per run locally, 502-839s in earlier CI) and its whole point is a from-scratch rebuild every time, which would stall a lane's other tests. Tagged `exclusive` (same reasoning as `diagnostics_reporter`), `size = "enormous"`, and `external` -- `external` forces a fresh run every time instead of reusing a cached PASS, since re-executing the two builds is the only way this test can catch a build turning non-deterministic.

Also tagged `skip-toolchain-sweep`, since this test always builds with the default toolchain and `test_rules_scala.sh`'s sweep re-running it under alternate `--extra_toolchains` would just repeat the identical build. `target_compatible_with` limits it to linux/macos, matching where the original shell test actually ran -- Windows is new ground for this script's MSYS2 behavior.

Coverage packages are still collected dynamically at test time, so this stays correct regardless of what #1977 (migrating `test_coverage.sh`, not yet merged) does to `test/coverage_*`.

The standalone linux/macos presubmit.yml tasks are gone; the new target gets picked up by `//...` in the existing `ubuntu2004`/`macos` "bazel test //..." tasks, which drops the legacy `ubuntu1804` pin that was only tied to the shell task.

### Motivation

Continues the shell-to-Bazel-native test migration (#1867).


test/reproducibility/BUILD
```diff
@@ -0,0 +1,36 @@
...
+#
+# manual: kept off any //... or test/... wildcard (including
+# test_rules_scala.sh's toolchain sweep) since ~14 min per run on macOS CI
+# would otherwise serialize into whichever job picked it up. Instead it's
+# its own dedicated presubmit.yml task, run by naming this target directly.
+# size=enormous: matches that ~14 min per-run cost.
+# external: forces a fresh run every time instead of reusing a cached PASS,
+# since re-executing the two builds is the only way this test can catch a
+# build turning non-deterministic.
+# target_compatible_with: limited to platforms where the original shell test
```
PawelLipski, line 15: Don't mention old code

test/reproducibility/build_is_identical_test.sh
```diff
@@ -0,0 +1,53 @@
...
+md5_util() {
+  if [[ "${OSTYPE}" == "darwin"* ]]; then
+    echo "md5"
+  else
+    echo "md5sum"
+  fi
+}
+
+# Collected dynamically so new/renamed/removed test/coverage_* packages
+# need no change here.
```
PawelLipski, line 24: Somewhat unclear what this comment refers to 🤔

test/reproducibility/build_is_identical_test.sh
```diff
@@ -0,0 +1,53 @@
...
+coverage_packages=()
+for package_dir in test/coverage_*; do
+  coverage_packages+=("//${package_dir}/...")
+done
+
+non_deploy_jar_md5_sum() {
+  local bazel_bin
+  bazel_bin="$(nested_bazel_run info bazel-bin)"
+  find "${bazel_bin}/test" -name '*.jar' ! -name '*_deploy.jar' ! -path "${bazel_bin}/test/jmh/*" \
```
PawelLipski, line 33: Actually, why excluding deploy jars?

PawelLipski: Is it due to timestamps in jar's zip entries? maybe add a comment in code

test/reproducibility/build_is_identical_test.sh
```diff
@@ -0,0 +1,53 @@
...
+
+nested_bazel_run clean
+sleep 10 # make sure timestamp-based nondeterminism, if any, would show up as different timestamps
+
+disk_cache_dir="$(mktemp -d)"
+nested_bazel_run build "--disk_cache=${disk_cache_dir}" test/...
+nested_bazel_run build "--disk_cache=${disk_cache_dir}" --collect_code_coverage -- "${coverage_packages[@]}"
+non_deploy_jar_md5_sum > "${hash2}"
+
+diff "${hash1}" "${hash2}"
```
PawelLipski, line 53: Maybe add some more pronounced assertion (like, what files/directories to check if the hashes differ)

test/reproducibility/BUILD
bor-p-s: Fixed.

test/reproducibility/build_is_identical_test.sh
bor-p-s: Reworded.

test/reproducibility/build_is_identical_test.sh
bor-p-s: Not timestamps -- singlejar normalizes those. Verified deploy jars are reproducible too, removed the exclusion.

test/reproducibility/build_is_identical_test.sh
bor-p-s: Added an explicit failure message; diff already shows the exact jar path.

PawelLipski (approved):

PR #1987, opened 2026-09-17
fix: map source paths explicitly in -paths-for-coverage.txt

### LINKED ISSUE #1101: Code coverage only working with deep directory structure

bor-p-s:
### Description
Ports #1967 by @avbhandaru onto current master. That PR adds an explicit source-to-class mapping in `-paths-for-coverage.txt` so coverage attribution follows a source's declared package instead of relying only on its directory path (see the class doc on `JacocoInstrumenter.pathsForCoverageContent` for the mechanism and its limits).

Two changes on top of #1967's original heuristic, both about the same underlying issue: a mapping can only be trusted when the pairing is unambiguous. It's skipped when another source shares its basename, and skipped when its basename matches a class in more than one package (a class name alone doesn't say which source declared it).

Also drops the original heuristic's "rest" bucket: those mappings could collide with an unrelated target's real class once every target's file is merged into one set for the coverage run.

Also splits `JacocoInstrumenter.java` into its own `java_library` (`instrumenter_lib`), with the existing `java_binary` depending on it via `runtime_deps`, so `pathsForCoverageContent` and its helpers can get a direct `java_test` instead of only going through a full nested `bazel coverage` run. Same shape as this repo's `worker` and `scalac/compileoptions` packages.

Fixes #1101

### Motivation
For a source whose disk path diverges from its package path, `bazel coverage` silently drops it from the report.

### Test plan
- Added `coverage_matches_expected_same_basename_test` for the basename-collision case above: fails on the pre-fix heuristic (one source's coverage lines end up duplicated into the other's record), passes with the uniqueness guard.
- Added `JacocoInstrumenterTest`, a direct unit test on `pathsForCoverageContent` and its helpers: both ambiguity guards above, the dropped "rest" bucket, plus two cases the coverage fixtures don't exercise (a doubly-nested class, and a class in the default package).


test/coverage_source_path_mapping/TestSourcePathMapping.scala
```diff
@@ -0,0 +1,15 @@
...
+import com.example.coverage.other.AlsoMapped
+import org.scalatest.flatspec._
+
+class TestSourcePathMapping extends AnyFlatSpec {
+
+  "Mapped.covered" should "work" in {
+    assert(Mapped.covered("ab") == "ba")
+  }
+
+  "AlsoMapped.covered" should "work" in {
```
PawelLipski: Pls refine the test naming (here and the other file) to make it clear what case (path==package, path!=package) is tested

src/java/io/bazel/rulesscala/coverage/instrumenter/JacocoInstrumenter.java
```diff
@@ -122,10 +141,81 @@ private static boolean isMethodTooLarge(Throwable t) {
     return false;
   }
 
+  /**
+   * Builds the contents of `-paths-for-coverage.txt`. `JacocoLCOVFormatter`'s suffix match
```
PawelLipski, line 149: Could you give an example (just in github comment) how this file looks like?

PawelLipski: Also, how/where is this file consumed? I might've missed that

src/java/io/bazel/rulesscala/coverage/instrumenter/BUILD
```diff
@@ -19,6 +17,20 @@ java_binary(
...
 
+java_binary(
+    name = "instrumenter",
+    main_class = "io.bazel.rulesscala.coverage.instrumenter.JacocoInstrumenter",
+    visibility = ["//visibility:public"],
+    runtime_deps = [":instrumenter_lib"],
+)
+
+java_test(
+    name = "JacocoInstrumenterTest",
```
PawelLipski, line 28: Wow, instrumenter was untested previously 😯

test/coverage_source_path_mapping/TestSourcePathMapping.scala
bor-p-s: Renamed both to name the case directly.

src/java/io/bazel/rulesscala/coverage/instrumenter/JacocoInstrumenter.java
bor-p-s: For the simple case (`Mapped.scala`, deeper package than directory), the generated file looks like:
```
test/coverage_source_path_mapping/Mapped.scala////com/example/coverage/mapping/Mapped.scala
test/coverage_source_path_mapping/Mapped.scala
```
First line is the explicit mapping, second is the bare fallback path.

bor-p-s: It's consumed by Bazel's own `JacocoCoverageRunner`/`JacocoLCOVFormatter` (see the comment a bit above, around line 97, with links to their source) — they scan every jar for a file ending in `-paths-for-coverage.txt`, and `JacocoLCOVFormatter.getExecPath` reads each line to resolve which real source file a class's coverage data belongs to.

src/java/io/bazel/rulesscala/coverage/instrumenter/JacocoInstrumenter.java
```diff
@@ -29,6 +36,13 @@ public final class JacocoInstrumenter implements Worker.Interface {
    */
   private static final String SKIP_OVERSIZED_METHODS_FLAG = "--skip_oversized_methods";
 
+  /**
+   * Separates the real source path from the class-derived path in a {@code
+   * -paths-for-coverage.txt} entry. Must stay in sync with {@code
+   * JacocoLCOVFormatter.EXEC_PATH_DELIMITER}.
```
PawelLipski, line 42: I'm assuming it can't be just imported from that class right?

src/java/io/bazel/rulesscala/coverage/instrumenter/JacocoInstrumenter.java
```diff
@@ -62,10 +76,14 @@ private void processArg(Instrumenter jacoco, String[] args) throws Exception {
 
     JarCreator jarCreator = new JarCreator(outPath);
 
+    // Jar-relative paths of every class we instrument, e.g. `/com/example/util/Time$.class`.
+    // Collected during the walk below and used to map sources onto their packages.
+    Set<String> classEntries = new LinkedHashSet<>();
```
PawelLipski: Maybe instead of generic "entries" use a more specific name - to make it clear it's a *path*, not e.g. a FQCN

test/coverage_source_path_mapping/TestSameBasenameMapping.scala
```diff
@@ -0,0 +1,14 @@
+package com.example.coverage.aliasing
+
+import org.scalatest.flatspec._
+
+class TestSameBasenameMapping extends AnyFlatSpec {
+
+  "one.Same, path mirrors package" should "keep its own coverage" in {
```
PawelLipski, line 7: 👍🏻

PawelLipski (approved):
Minor nits left

src/java/io/bazel/rulesscala/coverage/instrumenter/JacocoInstrumenter.java
bor-p-s: Checked the real `JacocoLCOVFormatter` source — it's `private` there, so it can't be imported. Kept it as a literal here, with the doc comment pointing at it so the two stay in sync.

src/java/io/bazel/rulesscala/coverage/instrumenter/JacocoInstrumenter.java
bor-p-s: Renamed to `instrumentedClassPaths`.

PR #1989, opened 2026-09-18
Add standalone-root build test for three nested modules

bor-p-s:
### Description

Adds nested-bazel build tests for `test/proto_cross_repo_boundary/repo`, `third_party/test/example_external_workspace`, and `third_party/test/new_local_repo`, as their own build root. See `run_test.sh` and `BUILD` for what and why.

### Motivation

Found while landing #1786, whose CI failure led to #1988's fix; this covers the same gap for the other two `dev_dependency`-wired modules and adds a permanent check.

Verified this catches a regression: reverting #1988's fix locally reproduces the same failure here, red; restoring it goes green.

Building `test/proto_cross_repo_boundary/repo` standalone surfaced two Windows-only failures, fixed in its `.bazelrc` and in `run_test.sh`'s output_base name (see the comments there for why). Confirmed fixed by a green Windows run in this PR's own CI.


test/proto_cross_repo_boundary/repo/.bazelrc
```diff
@@ -0,0 +1,7 @@
+# protobuf builds its own protoc/upb toolchain from C++ source by default
+# (this module doesn't opt into protobuf's prebuilt-protoc flag), which
+# needs C++17.
+common --enable_platform_specific_config
+build:linux --cxxopt=-std=c++17 --host_cxxopt=-std=c++17
+build:macos --cxxopt=-std=c++17 --host_cxxopt=-std=c++17
+build:windows --cxxopt=/std:c++17 --host_cxxopt=/std:c++17
```
PawelLipski, line 7: Windows always needs to be different 😅

PawelLipski (approved):

test/proto_cross_repo_boundary/repo/.bazelrc
bor-p-s: 💯

PR #1990, opened 2026-09-21
Fail loudly when the compile classpath carries two versions of the Scala standard library

### LINKED ISSUE #1781: `scala3-library_3` defined for `scala_version` might be not taken into account for evicition
Under `dependency_mode=transitive` `scala3-library_3` (and possible `scala-library`) artifacts might not be a subject to evicition rules resulting in chossing version different to the version of the Scala compiler. Such case might lead to compilation crashes observed in https://github.com/scala/scala3/issues/22890  

We should either enforce to always use matching stdlib artifact for the verison of the compiler or to ensure we that version selected by `scala_version` would be compared with version defined in transitive dependenices.  

Based on the Slack chat:  

> I’m now getting these messages
package scala contains object and package with same name: caps.
This indicates that there are several versions of the Scala standard library on the classpath.
The build should be reconfigured so that only one version of the standard library is on the classpath.
and an error pointing to a usage of Expr.summonIgnoring, which only exists in Scala 3.7.x
Based on [this issue](https://github.com/scala/scala3/issues/22890) it looks like this is happening because I have multiple versions of scala3-library_3 on my classpath — one from my build and one that’s pulled in transitively from each of my Scala library dependencies. Is there a way to suppress only those transitive deps? 
... 
Thanks, just confirmed that adding an explicit dependency on scala3-library_3 fixes it

bor-p-s:
### Motivation

Under `dependency_mode = "transitive"` or `"plus-one"`, a `scala_library`/`scala_binary` target's compile classpath can carry two versions of `scala3-library_3` (or `scala-library`): the one `scala_version` resolves, plus a different one pulled in transitively by some other dependency. The compiler crashes with an unhelpful error ("package scala contains object and package with same name: caps", or a TASTy version mismatch) that doesn't point at the cause. See #1781, which traces back to https://github.com/scala/scala3/issues/22890.

`dependency_mode` also decides which jars reach the compiler's classpath (`scala/private/rule_impls.bzl`): `"direct"` exposes only each dep's own compile jar, `"transitive"`/`"plus-one"` pull in a dep's full transitive closure, stdlib jars included.

### Description

`compile_scala` (`scala/private/rule_impls.bzl`) now checks the classpath before compiling: groups jars by stdlib artifact (basename match on `scala3-library_3-`/`scala-library-`, stripping wrapper prefixes/suffixes like `scala_import`'s `-stamped` or rules_jvm_external's `header_`), fails with the mismatched dependency and both versions named.

`scala-library` (Scala 2.x) tolerates a patch-only difference (compared by major.minor): real downstream builds carry mismatched patch versions of it via separate maven resolution, and that's harmless since Scala 2.x keeps binary compatibility within a major.minor line. `scala3-library_3` still needs an exact match, since a version mismatch there is exactly what crashes the compiler in issue #1781.

`test/scala3_library_eviction/` adds a fixture with a stdlib jar two hops below the target (through a `scala_import`, so it never compiles at the wrong version itself), plus tests: fails under `transitive`/`plus-one`, succeeds under `direct`, covers the rules_jvm_external jar naming, matching versions (must not trip the check), a `scala-library`-only mismatch across a major.minor line, and a `scala-library` patch-only difference (must not trip the check).

Fixes #1781


scala/private/stdlib_version_check.bzl
```diff
@@ -0,0 +1,83 @@
+"""Fails a scala compile if its classpath carries two incompatible versions of
+the Scala standard library, naming the mismatched dependency and both
+versions in the error.
+
+Left unguarded, this crashes the compiler with an unreadable error; see
+https://github.com/bazel-contrib/rules_scala/issues/1781.
+"""
+
+# Basename prefixes of the Scala standard library's own jars.
+_STDLIB_JAR_PREFIXES = ["scala3-library_3-", "scala-library-"]
```
PawelLipski: Does it also check against mixing scala3 and scala(2) libs in the same target? or is it for some reason unlikely (as compared to mixing major versions within one Scala 2/3 epoch)?

scala/private/stdlib_version_check.bzl
```diff
@@ -0,0 +1,83 @@
...
+
+# Prefixes a compile-jar-producing rule puts in front of those basenames
+# (rules_jvm_external's jvm_import: "header_"; a processed/re-signed jar: "processed_").
+_JAR_WRAPPER_PREFIXES = ["header_", "processed_"]
+_DIGITS = "0123456789"
+
+# Maven classifier suffixes (sources/javadoc jars); matching one would parse
+# a bogus "version" out of e.g. scala3-library_3-3.3.1-sources.jar.
```
PawelLipski: A bit unclear what bogus version means here... sth like `3.3.1-sources`?

scala/private/stdlib_version_check.bzl
```diff
@@ -0,0 +1,83 @@
...
+def compat_key(artifact, version):
+    if artifact == "scala-library-":
+        return ".".join(version.split(".")[:2])
+    return version
+
+# Fails if classpath_jars carries two incompatible versions of the same
+# stdlib artifact.
+def fail_on_mismatched_stdlib_versions(target_label, classpath_jars):
+    versions_by_artifact = {}
+    for jar in classpath_jars.to_list():
```
PawelLipski: Whooops this is an antipattern, it can easily make Bazel analysis quadratic on the projects that use rules_scala :/

PawelLipski: Hmmm albeit... `compiler_classpath_jars` are probably already expanded into CLI args anyway, right? so the complexity should not increase (?)

If safe, pls add a comment why so

scala/private/stdlib_version_check.bzl
bor-p-s: No, this check doesn't cover that, and I think that's fine: scala3-library_3 and scala-library are different artifacts, not two versions of the same one, so they're expected to coexist (scala3-library_3 itself depends on a matching scala-library). This check only catches two copies of the same artifact at different versions. Mixing a Scala 2-compiled library into a Scala 3 target (or the other way around) is a different kind of problem, outside what #1781 is about.

scala/private/stdlib_version_check.bzl
bor-p-s: Good catch, thanks — verified it, Args.add_all() defers expansion to execution phase, so this is a real new cost, not something already paid elsewhere. Real risk on deep graphs under transitive/plus-one. Let me think through a proper fix (likely an aspect-propagated provider instead of flattening the full classpath) rather than rush it — will follow up.

PawelLipski: Yeah, judging by how similar linters worked in my commercial client's codebase - it was typically an aspect; `to_list` is almost religiously avoided

scala/private/stdlib_version_check.bzl
PawelLipski: Okay, maybe drop a comment (unless it's already clear but I've missed that) that this case isn't covered

Scala 2.x vs 2.y major version libs are notoriously binary incompatible - doesn't Scala 3 have better guarantees tho? for compatibility b/w 3.x and 3.y

scala/private/stdlib_version_check.bzl
bor-p-s: That aspect precedent makes sense. But the check needs to match the compiler's exact classpath -- including dependency modes, exports, and macro runtime jars, five separate exposure paths in collect_jars. I'd rather check the assembled classpath in the worker, which sidesteps duplicating that logic, than risk a subtly wrong aspect. Trade-off: it fails at execution time, not analysis time. If analysis-time reporting turns out to matter, an aspect built around shared exposure logic is the natural follow-up.

PawelLipski: I think execution-time failure is fine, and strictly better than what we have now
In fact it might allow for more powerful checks and/or better error messages than analysis-time check

scala/private/stdlib_version_check.bzl
bor-p-s: Moved to StdlibVersionCheck.java (Java), same fix there.

scala/private/stdlib_version_check.bzl
bor-p-s: Scala 3 does guarantee more than 2.x -- backward across minors, both ways across patches. But that's excluded for experimental APIs, and the scala.caps break was exactly that (RC against a much older minor, not a patch pair -- checked the actual versions). So scala3-library_3 stays exact-match; added an RC/prerelease guard too since those carry no guarantee either. Scope note's in the javadoc.

src/java/io/bazel/rulesscala/scalac/StdlibVersionCheck.java
```diff
@@ -0,0 +1,174 @@
...
+        return new ArtifactVersion(prefix, version);
+      }
+    }
+    return null;
+  }
+
+  // The granularity two versions of the same artifact must agree on to coexist
+  // on one classpath. scala-library (Scala 2.x) keeps binary compatibility
+  // across patch releases within one minor line, so only a major.minor
+  // difference matters there (mixing 2.12 and 2.13 breaks; 2.12.20 vs 2.12.21
```
PawelLipski: Pls check what's the official naming & what's used in rules_scala, I remember it used to be called `EPOCH.MAJOR.MINOR` rather than `MAJOR.MINOR.PATCH`

src/java/io/bazel/rulesscala/scalac/StdlibVersionCheckTest.java
```diff
@@ -0,0 +1,188 @@
...
+        assertThrows(
+            StdlibVersionCheck.MismatchedStdlibVersions.class,
+            () ->
+                StdlibVersionCheck.check(
+                    "//test:app",
+                    new String[] {
+                      "bazel-out/bin/scala3-library_3-3.7.3.jar",
+                      "bazel-out/bin/scala3-library_3-3.7.4.jar",
+                    }));
+    assertTrue(exception.getMessage().contains("multiple versions of scala3-library_3"));
```
PawelLipski, line 140: Maybe check for the exact full message? here and/or in other tests

src/java/io/bazel/rulesscala/scalac/StdlibVersionCheckTest.java
```diff
@@ -0,0 +1,188 @@
...
+            StdlibVersionCheck.MismatchedStdlibVersions.class,
+            () ->
+                StdlibVersionCheck.check(
+                    "//test:app",
+                    new String[] {
+                      "bazel-out/bin/scala3-library_3-3.6.4.jar",
+                      "bazel-out/bin/header_scala3-library_3-3.7.4.jar",
+                    }));
+    assertTrue(exception.getMessage().contains("multiple versions of scala3-library_3"));
+  }
```
PawelLipski, line 222: Also check that a lib with `-sources` suffix doesn't cause a failure

test/scala3_library_eviction/BUILD
```diff
@@ -0,0 +1,283 @@
...
+    build_args = ["--extra_toolchains=//test/scala3_library_eviction:transitive"],
+    target = ":app_patch_only_scala_library_mismatch",
+)
+
+expect_build_failure_test(
+    name = "transitive_deps_mode_rejects_scala3_library_patch_mismatch_test",
+    build_args = ["--extra_toolchains=//test/scala3_library_eviction:transitive"],
+    expect = ["expected_mismatch_error.txt"],
+    target = ":app_patch_only_scala3_library_mismatch",
+)
```
PawelLipski, line 301: Wow so many targets 😅

test/scala3_library_eviction/expected_mismatch_error.txt
```diff
@@ -0,0 +1 @@
+multiple versions of scala3-library_3 on the compile classpath
```
PawelLipski, line 1: Consider checking for full message here as well

src/java/io/bazel/rulesscala/scalac/StdlibVersionCheck.java
```diff
@@ -0,0 +1,174 @@
...
+          if (versionsList.length() > 0) {
+            versionsList.append(", ");
+          }
+          versionsList.append(versionAndPath[0]).append(" (").append(versionAndPath[1]).append(")");
+        }
+        throw new MismatchedStdlibVersions(
+            String.format(
+                "%s: multiple versions of %s on the compile classpath: %s. A dependency was"
+                    + " built against a different version of %s than the toolchain's scala_version"
+                    + " resolves; make sure every dependency that carries %s was built against the"
```
PawelLipski: Would it be possible to state this toolchain's scala_version explicitly here in the message?

src/java/io/bazel/rulesscala/scalac/StdlibVersionCheckTest.java
bor-p-s: Done -- assertEquals with the full message where I control every input.

src/java/io/bazel/rulesscala/scalac/StdlibVersionCheckTest.java
bor-p-s: Good catch -- added check_sourcesJarAlongsideMatchingClasspath_passes.

test/scala3_library_eviction/BUILD
bor-p-s: Yeah -- each variant needs its own dep chain to actually exercise a different code path, not just a different assertion 😅

test/scala3_library_eviction/expected_mismatch_error.txt
bor-p-s: The message embeds bazel-out paths that vary by execroot, so a full exact match only works for the fixed parts. Added a second expect file with the static tail (everything but the paths) -- the whole non-path text is now checked.

PawelLipski (approved):

src/java/io/bazel/rulesscala/scalac/StdlibVersionCheck.java
bor-p-s: Done -- threaded toolchain.scala_version through as a new --ScalaVersion arg, now stated explicitly in the message.

src/java/io/bazel/rulesscala/scalac/StdlibVersionCheck.java
bor-p-s: Checked both. Scala's own docs name these per-series, no "epoch" term: for Scala 2, "13" in 2.13.18 is the major version and "18" is minor; for Scala 3, "7" in 3.7.4 is minor and "4" is patch. Matched what rules_scala already uses instead of chasing Scala's own inconsistent vocab: extract_major_version/extract_minor_version in scala_cross_version.bzl call the first two numbers (2.12, 3.7, ...) the major version and the third the minor version. Updated all the comments to that.

Issue #1793, opened 2025-11-23
`scala_repl` is broken for Scala 3


WojciechMazur:
REPL in Scala 3 uses different main class then the one used for Scala 2, the main class should be selected dynamically 
for Scala 2: `scala.tools.nsc.MainGenericRunner` 
for Scala 3: `dotty.tools.repl.Main` 

Also in upcoming Scala 3.8 REPL was seperated from compiler artifact and introduces dependenices on [fansi](https://index.scala-lang.org/com-lihaoyi/fansi) for syntax highliting / rendering

PR #1991, opened 2026-09-21
fix: Use dotty.tools.repl.Main for scala_repl under Scala 3

bor-p-s:
### Description
`phase_write_executable_repl` now picks `dotty.tools.repl.Main` for Scala 3 based on the resolved toolchain's `scala_version`, the same pattern (`toolchain.scala_version.startswith(...)`) already used elsewhere in this codebase to branch on major Scala version (e.g. `phase_compile.bzl`, `phase_semanticdb.bzl`). Scala 2's path is unchanged.

Adds a `Scala3Repl` target (`scala_version = "3.7.4"`) and a check in the existing `test/repl:repl_test` that feeds it `1 + 1` and checks for `res0: Int = 2`.

Pinned to 3.7.4: from Scala 3.8 the REPL moves into its own `scala3-repl_3` artifact with new dependencies (fansi, pprint). That's addressed in a follow-up PR stacked on this one.

### Motivation
Addresses the entry-point part of #1793 (the fansi/scala3-repl split for 3.8+ is handled separately in the stacked follow-up): `scala_repl` hardcoded `scala.tools.nsc.MainGenericRunner`, the Scala 2 REPL entry point, as its main class. That class doesn't exist on a Scala 3 toolchain's classpath, so `scala_repl` failed at startup with `ClassNotFoundException` for every Scala 3 project, making the rule completely unusable there.

### Test plan
Reverted the main_class change, confirmed the new test fails with `ClassNotFoundException: scala.tools.nsc.MainGenericRunner`. Reapplied the fix, confirmed it passes.


test/BUILD
```diff
@@ -320,6 +320,14 @@ scala_repl(
     srcs = ["A.scala"],
 )
 
+# Pinned to 3.7.4: from 3.8 the REPL moves into its own scala3-repl artifact
+# (fansi/pprint deps).
+# TODO: wire up scala3-repl support for 3.8+.
```
PawelLipski, line 325: Pls create an issue + link it here (`TODO (#number):`)

PawelLipski (approved):

test/BUILD
bor-p-s: The follow-up PR (#1992) already removes this exact TODO comment when it wires up 3.8/3.9 support -- it links back to #1793, which already describes this limitation, so there's no separate issue to create.

PR #1992, opened 2026-09-21
feat: Add scala3-repl deps for scala_repl under Scala 3.8 and 3.9

### LINKED ISSUE #1793: `scala_repl` is broken for Scala 3

bor-p-s:
### Description
Follow-up to the `scala_repl` main-class fix (#1991): adds Scala 3.8 and 3.9 REPL support. From 3.8, `dotty.tools.repl.Main` moves out of `scala3-compiler_3` into its own `scala3-repl_3` maven artifact, and its extra runtime deps differ by version:
- 3.8: fansi_3, pprint_3, sourcecode_3, plus coursier-interface/using_directives/slf4j-api for the REPL's `:dep` command.
- 3.9: its own jline 4.0.14 artifacts (added alongside the repo's existing 3.30.6 set, used elsewhere), a new `scala3-directives-parser_3` artifact, and coursier-interface/slf4j-api again. fansi/pprint are gone in 3.9; `:dep` already existed in 3.8, only its own jline/directives-parser deps are new to 3.9.

Adds a new toolchain dep provider, `scala_repl_classpath`, layering these version-specific extras on top of the existing `scala_compile_classpath`. `setup_scala_toolchain()` also takes an optional `scala_repl_classpath` override, matching the other classpath params it already has. `phase_scalac_provider.bzl` falls back to the plain compile classpath for a toolchain predating this provider, preserving a custom `scala_toolchain()` caller's explicit `dep_providers` list.

A Scala 3.x version whose minor version needs REPL-specific deps must have a matching entry in this mapping, or `scala_repl` analysis fails loudly with a message pointing at the gap, instead of building a REPL that breaks at runtime. That check runs inside `phase_write_executable_repl`, so it only fires when a `scala_repl` target is actually analyzed, keeping other builds under that version green. One tradeoff: an explicit `scala_repl_classpath` override for an unmapped version now also hits this check, even when the override is already complete. This only matters for a hypothetical future consumer overriding an unmapped version; today's supported versions are unaffected.

### Motivation
Fixes #1793: this PR closes the fansi/scala3-repl split part of it, and it's stacked on #1991, which fixes the main_class part. Both need to land for the issue to actually be resolved.

### Test plan
Manually confirmed the 3.8 `:dep` command resolves a real dependency (`:dep com.lihaoyi::upickle:3.1.0` → "Resolved 1 dependencies (8 JARs)"). The automated `1 + 1` check added for both versions verifies REPL startup and evaluation; `:dep` and the legacy-provider fallback are covered only by this manual check.


scala/private/macros/scala_repositories.bzl
```diff
@@ -266,4 +267,9 @@ def scala_version_artifact_ids(scala_version):
             "org_scala_sbt_util_interface",
         ])
 
+    # From 3.8, dotty.tools.repl.Main moves out of scala3-compiler into its
+    # own artifact with its own, version-specific extra deps -- see
+    # repl_extra_deps in repl_deps.bzl for the full story.
+    result.extend([dep.lstrip("@") for dep in repl_extra_deps(scala_version)])
```
PawelLipski: Pls also explain why the `@` needs to be stripped specifically, as it's not obvious (`@` suggests an external repo... but why keep the external repo name then?)

scala/private/phases/phase_write_executable.bzl
```diff
@@ -34,6 +38,12 @@ def phase_write_executable_scalatest(ctx, p):
 
 def phase_write_executable_repl(ctx, p):
     toolchain = ctx.toolchains["//scala:toolchain_type"]
+    if not repl_is_known_supported(toolchain.scala_version):
+        fail(
```
PawelLipski: For clarity, this error is supposed to be shown to the maintainers of rules_scala, not end users, right?

scala/private/macros/repl_deps.bzl
```diff
@@ -0,0 +1,56 @@
+# scala_repl's extra deps key off the exact minor version, finer-grained
```
PawelLipski: Maybe mention also in `docs/scala_repl.md`? not sure if very relevant to end users (but maybe mention `scala_version` argument? AFAICS it's not mentioned there)

scala/private/macros/repl_deps.bzl
```diff
@@ -0,0 +1,56 @@
...
+
+    3.8 and 3.9 each changed scala_repl's own deps once already, so treat a
+    future 3.x minor this mapping has yet to cover as unknown territory,
+    requiring an update here, rather than assuming it needs nothing:
+    silently assuming that would build a scala_repl that fails at runtime
+    with ClassNotFoundException. Called only when a scala_repl target is
+    actually analyzed, scoping the check to the targets it actually concerns
+    rather than every general toolchain/repository setup.
+    """
+    if not scala_version.startswith("3.") or int(scala_version.split(".")[1]) < 8:
```
PawelLipski: Is it guaranteed that `scala_version` will always be of `3.x` and not `3.x.y` shape?

scala/private/macros/scala_repositories.bzl
bor-p-s: repl_extra_deps's own two other callers (setup_scala_toolchain.bzl, toolchains_repo.bzl) build classpath labels, so its list is "@repo_name" form. This function's own result is bare repo names throughout (see every entry above) -- lstrip("@") just bridges the two conventions. Added a comment explaining it, thanks.

scala/private/phases/phase_write_executable.bzl
bor-p-s: Not exactly -- scala_versions is a public setting (scala_config.settings(scala_versions=[...])), so an end user can add a Scala version rules_scala hasn't caught up on yet, and hit this too. Reworded the message to cover both cases.

scala/private/macros/repl_deps.bzl
bor-p-s: Added scala_version to docs/scala_repl.md -- that one was a real gap. Left the fansi/3.8-split internals out of it though: that's implementation detail an end user doesn't act on, repl_deps.bzl's own comment already covers it for contributors.

scala/private/macros/repl_deps.bzl
bor-p-s: Yes -- scala_version is backed by a Bazel string_setting with allowed_values pinned to SCALA_VERSIONS (all X.Y.Z). I actually hit this directly while testing: passing scala_version = "3.10.0" on a target that isn't in that list gets rejected by Bazel itself with "invalid value... Allowed values are [...]" before this code ever runs. extract_minor_version elsewhere in this file already relies on the same 3-part shape, so this isn't a new assumption. And even a weird suffix would only land in .split(".")[2], not [1], which is all this line reads.

scala/private/phases/phase_write_executable.bzl
```diff
@@ -5,6 +5,10 @@ load("//scala/private:common.bzl", "rlocationpath_from_file")
 #
 # DOCUMENT THIS
```
PawelLipski: Looks like missing docs 😅  pls open an issue for filling these up while you're here

PawelLipski (approved):

scala/private/phases/phase_write_executable.bzl
bor-p-s: Opened #2002 for it.

PR #1994, opened 2026-09-21
Use rules_jvm_external for internal dependencies

bor-p-s:
### Description

Carries forward #1786 (by @jadenPete), rebased onto current master, with the adjustments below.

This is the second in a series to refactor the ruleset to use `rules_jvm_external` for fetching all Maven artifacts. It deletes `scala/private/extensions/dev_deps.bzl` and refactors the ruleset to pull all internal dependencies (previously declared in that module) using `rules_jvm_external`. It also makes `rules_jvm_external` a required dependency of the ruleset.

### Motivation

See #1775 for the initiative and motivation behind it.

### Changes from #1786's own diff

- Fixed a regression: JFFI lost its native classifier when switching from `com.github.jnr:jffi:jar:native:1.2.17` to a plain `com.github.jnr:jffi:1.3.13`, dropping test coverage for importing a classified artifact in `//test/src/main/scala/scalarules/test/scala_import:jffi_native_external`. Restored via `maven.artifact(classifier = "native", force_version = True)`.
- Dropped the file-group ("with_file") test variant, since rules_jvm_external generates a `jvm_import` target per artifact rather than the raw-file import the old macro produced for that case.
- Restored the four `test/fetch_sources` cases on `io_bazel_rules_scala_guava`, replacing the fixture from the deleted `dev_deps.bzl`. They exercise `BAZEL_JVM_FETCH_SOURCES`, an env var `scala_maven_import_external.bzl` reads on every call to `scala_toolchains()` -- the macro every rules_scala consumer uses, untouched by this migration. Confirmed with fresh-output-base nested Bazel builds: the source jar is present for unset and `TruE`, and absent for `false` and `1`.


bor-p-s:
It's a bit big, but no easy way to shrink I'm afraid.

test/expect_build_failure/expect_build_failure.sh
```diff
@@ -153,7 +153,7 @@ _resolve_message_file() {
   printf '%s' "${file}"
 }
 
-nested_bazel_setup "rules_scala_expect_build_failure_output_base_lane${lane}"
+nested_bazel_setup "ebf_lane${lane}"
```
PawelLipski: Hmmm why this changed? 🤔

MODULE.bazel
```diff
@@ -181,26 +183,44 @@ use_repo(
...
+        "org.typelevel:cats-core_2.12:2.13.0",
+        "org.typelevel:kind-projector_2.12.20:0.13.4",
+    ],
+    fetch_sources = True,
+    lock_file = "//:rules_scala_test_maven.json",
+    repositories = [
+        "https://repo.maven.apache.org/maven2",
+        "https://maven-central.storage-download.googleapis.com/maven2",
+        "https://mirror.bazel.build/repo1.maven.org/maven2",
+        "https://jcenter.bintray.com",
```
PawelLipski: JCenter is dead; you can clean up on a separate PR (coz it's present on master as well)

test/src/main/scala/scalarules/test/sources_jars_in_deps/BUILD
PawelLipski: Hmm why remove this test?

test/src/main/scala/scalarules/test/scala_import/BUILD
```diff
@@ -51,21 +51,12 @@ scala_specs2_junit_test(
     deps = [":imported_genericlib_jar"],
 )
 
-scala_library(
```
PawelLipski: Why removed?

test/src/main/scala/scalarules/test/scala_import/BUILD
```diff
@@ -51,21 +51,12 @@ scala_specs2_junit_test(
...
-    # jvm_maven_import_external doesn't fetch source jars automatically
-    deps = ["@org_typelevel__cats_core//jar"],
-)
-
 ##Runtime deps
 scala_import(
     name = "indirection_for_transitive_runtime_deps",
     testonly = True,
     jars = [],
-    # jvm_maven_import_external doesn't fetch source jars automatically
```
PawelLipski: Oh, b/c this is no longer a problem?

test/missing_direct_deps/external_deps/BUILD
```diff
@@ -24,26 +24,7 @@ scala_library(
     srcs = [
         "B.scala",
     ],
-    deps = ["@com_google_guava_guava_21_0//jar"],
-)
-
-scala_library(
```
PawelLipski: Why removed?

test/fetch_sources/BUILD
PawelLipski: Removed coz rules_jvm_external handles it now easily, right?

WORKSPACE
```diff
@@ -138,10 +138,6 @@ rbe_preconfig(
     toolchain = "ubuntu2004-bazel-java11",
```
PawelLipski: Btw is there a plan/issue to remove WORKSPACE in favor of bzlmod? or does the repo still maintain the former for the sake of compatibility (?)

test/expect_build_failure/expect_build_failure.sh
bor-p-s: Shortened the nested output base name here (and in `no_recompilation_test.sh`) because it broke Windows CI: with the old ~50-char name, the runfiles tree path for one of rules_jvm_external's own internal tools (`AddJarManifestEntry`) came out to 262 characters, 2 over Windows' 260-character path limit. A shorter name gives enough margin (see the comment I added in `nested_bazel.sh` right above `nested_bazel_setup`).

MODULE.bazel
bor-p-s: Agreed, will clean it up in a separate PR.

test/src/main/scala/scalarules/test/sources_jars_in_deps/BUILD
bor-p-s: This test checked that cats-core's source jar does NOT end up on the classpath (see the comment in third_party/repositories/scala_2_12.bzl: "testing that we don't include sources jars"). Now `rules_scala_test_maven.install()` always fetches source jars for everything in that install (`fetch_sources = True`, one flag for the whole list). So the thing this test checked isn't true anymore, and there's no per-artifact way to turn it off without a second install().

test/src/main/scala/scalarules/test/scala_import/BUILD
bor-p-s: Same target, same reason as the copy in `sources_jars_in_deps/BUILD`: `rules_scala_test_maven.install()` now fetches source jars for everything in it, so the "source jar isn't on the classpath" check this test did isn't true anymore.

test/src/main/scala/scalarules/test/scala_import/BUILD
bor-p-s: Yes, that's it exactly -- source jars are now always fetched (`fetch_sources = True`), so that comment was no longer true.

test/missing_direct_deps/external_deps/BUILD
bor-p-s: These backed `fails_on_missing_direct_external_file_group_dep`, which checked a buildozer hint naming the right repo among two identically-wrapped repos for the same guava jar. That needs a second, artificial `maven.install()` under rules_jvm_external, since a coordinate dedupes to one repo per install -- see the PR description for details.

test/fetch_sources/BUILD
bor-p-s: Not quite -- it's deleted here only because it depended on a fixture (`com_google_guava_guava_21_0`) that this PR removes along with `dev_deps.bzl`. #1995 (open) repoints it at `io_bazel_rules_scala_guava`, an unrelated, already-existing fixture, before this PR lands. Once #1995 merges and this branch rebases, this file won't need to change at all.

WORKSPACE
bor-p-s: Not that I know of -- I didn't find an open issue or plan for dropping WORKSPACE entirely. This PR just keeps maintaining both, matching what the repo already does everywhere else (every dependency declared twice). Might be worth asking the maintainers directly if it matters for planning.

PawelLipski: > Might be worth asking the maintainers directly if it matters for planning.

We're the maintainers now 😅 

image

PawelLipski (approved):

